PR #251, opened 2015-03-04
cforest can handle missings in features via surrogate splits

zmjones:
the documentation for `ctree_control` says that it is restricted to ordered predictors but there doesn't appear to be any error checking or problems in doing so. I have only given a cursory look through `SurrogateSplits.c`, however, so I am not sure what is up there yet. If nominal missings are a problem should this be checked in `mlr`?

The below examples are from the `cforest` example section.

```{r}
## classification, where i insert missingness into an unordered factor
data("mammoexp", package = "TH.data")
mammoexp$HIST[sample(1:nrow(mammoexp), 25)] <- NA
mamofit <- cforest(ME ~ ., mammoexp)

## survival
data("GBSG2", package = "TH.data")
GBSG2$tsize[sample(1:nrow(GBSG2), 25)] <- NA
bst <- cforest(Surv(time, cens) ~ ., data = GBSG2, 
               control = cforest_unbiased(ntree = 50))

data(iris)
iris$Petal.Width[sample(1:nrow(iris), 10)] <- NA
iris.cf <- cforest(Species ~ ., data = iris, 
                          control = cforest_unbiased(mtry = 2))


```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,7 +5,9 @@ mlr_2.4:
 - option fix.factors in makeLearner was renamed to fix.factors.prediction for clarity
 - getTaskData now has recodeY = "drop.levels" which drops empty factor levels
 - new measures
--- brier
+- brier
+- add missings properties to all ctree and cforest methods:
+  regr/classif for ctree, regr/classif/surv for cforest, and regr/classif for blackboost
 
 mlr_2.3:
 - resample now returns an object of class ResampleResult (downward compatible)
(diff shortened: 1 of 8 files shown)


berndbischl:
Hi. What I would do in such a "borderline case"

a) just set the missing prop like you did, but add one explaining sentence in the "note" where you reference the cforest doc. 

b) In this case you could / should also mention this change in NEWS.

Can you adapt please, then I will merge.

Thx!

berndbischl:
a) What I also meant here: I would add an extra check in mlr just for that case (if cforest would really break on missings in nominal attributes)

zmjones:
ok sounds good

the note i am adding is "see ?ctree_control for possible breakage for nominal features with missingness"

is that alright?

i also thought it should be added to the `ctree` methods as well, since this is the case there also. it may also apply to some of the methods for `mboost` methods which use `ctree`. I will look through those this afternoon and test them if i can.

for NEWS should this be added under 2.4?

berndbischl:
ok to all what you said.

mllg:
Thanks for the PR!

Issue #254, opened 2015-03-07
Setting rpart options for classif.ada


phASL:
Setting rpart options for ```classif.ada``` via ```par.vals``` or directly e.g. as ```maxdepth``` argument is not possible:
```R
library("mlr")

param = list(maxdepth = 5)
lrn.ada = makeLearner("classif.ada", par.vals = param)
train(lrn.ada, sonar.task)
# Error in ada.default(x, y, ..., na.action = na.action) : 
#   Error:  Argument maxdepth not matched

lrn.ada = makeLearner("classif.ada", maxdepth = 5)
train(lrn.ada, sonar.task)
# Error in ada.default(x, y, ..., na.action = na.action) : 
#   Error:  Argument maxdepth not matched
```
You can use a trick: Set the mlr option ```on.par.without.desc``` to "warn" and use the ```...``` argument:

```R
param = list(maxdepth = 5)
# also possible: param = rpart.control(maxdepth = 5)

configureMlr(on.par.without.desc = "warn")
lrn.ada = makeLearner("classif.ada", ... = param)
# Warning message:
# In setHyperPars2.Learner(learner, insert(par.vals, args)) :
# classif.ada: Setting parameter ... without available description object!
# You can switch off this check by using configureMlr!

train(lrn.ada, sonar.task)
# Model for id = classif.ada class = classif.ada
# Trained on obs: 208
# Used features: 60
# Hyperparameters: ...=<list>
```

The problem is that ```ada()``` requires a named list for the rpart options, so I think it's a better solution to insert some code in ```trainLearner.classif.ada```, just like the [ctrl argument] (https://github.com/berndbischl/mlr/blob/0eec258c8b5e39691aeaf229e647ca0796e64435/R/RLearner_classif_boosting.R#L34) in ```trainLearner.classif.boosting```.

jakob-r:
@berndbischl See if you like this (https://github.com/berndbischl/mlr/commit/bc27e5122c8c6eee35818eec8f0a5ffcfe95163f) solution and I will merge it into master branch

Issue #244, opened 2015-02-17
predict.threshold argument of makeLearner


schiffner:
Below I specify a threshold via `predict.threshold` in `makeLearner` without setting `predict.type = "prob"`. Shouldn't we return some warning/error or maybe, for full service, automatically set the predict.type to "prob" internally if a threshold is given?

```
th = c(setosa = 1, virginica = 4, versicolor = 8)
lrn = makeLearner("classif.lda", predict.threshold = th)
mod = train(lrn, task = iris.task)
pred = predict(mod, task = iris.task)
pred

Prediction: 150 observations
predict.type: response
threshold: 
time: 0.00
  id  truth response
1  1 setosa   setosa
2  2 setosa   setosa
3  3 setosa   setosa
4  4 setosa   setosa
5  5 setosa   setosa
6  6 setosa   setosa
```

berndbischl:
You are right. But I rather have the more verbose option where one is forced to set the type, so your case will be an error.

schiffner:
Ok. Thanks.

PR #255, opened 2015-03-07
Issues #244 #254

erenturunc:
I changed default predict.type value in learner if threshold is set.

I added rpart parameter control to classif.ada


larskotthoff:
@all -- can this be merged?

Issue #247, opened 2015-02-27
Improvements in plotROCRCurves


schiffner:
Some things I noticed when writing the tutorial. 

  * Defaults:
  
    1. The default is `avg = "threshold"`. That does not make much sense if `obj` is a (`list` of)
       `Prediction` objects. The default should depend on `obj`. In my opinion we should set `avg = "none"` for (`list`s of) `Prediction` objects and `avg = "threshold"` for (`list`s of) `ResamplePrediction` objects and `BenchmarkResult`s.

    2. The colored boxes in the legend are not needed. Therefore keep `fill = NULL` (the default in
       `legend`).

    3. The default is `add.diag = TRUE`: The diagonal does only make sense for ROC curves
       and it's tedious to set is to `FALSE` when plotting other performance measures.
       Maybe the default should be `FALSE` or depend on `meas1` and `meas2`.

  * At the moment standard `plot` arguments like `xlab` or `ylim` cannot be used.
    Many graphical options of ROCR's `plot` function also do not work.
    Either give `plotROCRCurves` a `...` argument or something like `plot.args` to pass
    graphical parameters.

  * Determine `ylim` internally to prevent the following:

    ```
    lrn1 = makeLearner("classif.lda", predict.type = "prob")
    mod1 = train(lrn1, sonar.task)
    pred1 = predict(mod1, task = sonar.task)
    
    lrn2 = makeLearner("classif.ksvm", predict.type = "prob")
    mod2 = train(lrn2, sonar.task)
    pred2 = predict(mod2, task = sonar.task)

    plotROCRCurves(list(lda = pred1, ksvm = pred2), meas1 = "acc", meas2 = "cutoff", avg = "none",
      add.diag = FALSE, legend.args = list(x = "bottom"))
    ````

zmjones:
about ii. do you mean the boxes around the colors inside the legend, or the box around the legend itself? in the former case this can be controlled using `fill` and `border` but in the latter case `bty = "n"` removes the box.

schiffner:
@zmjones: I mean the colored boxes inside the legend. There is no need for them, as there are only colored lines in the plot.
Thanks very much!

schiffner:
I updated i. (default `average` setting) in the first post:
In my opinion we should set `average = "none"` for (`list`s of) `Prediction` objects and `average = "threshold"` for (`list`s of) `ResamplePrediction` objects and `BenchmarkResult`s.
[...]

PR #259, opened 2015-03-14
improvements to plotROCRCurves from #247

zmjones:
things that were added

 - a (by default empty) list to the arguments: `plot.args`
 - `ylim` is set by finding the range of the `y.values` slot of elements of the list `rocr.perfs` across all elements
- `add.diag` is now `NULL` by default, and is set to `FALSE` unless `meas1 = "tpr"` and `meas2 = "fpr"`, appropriate assertions were added in the generic method
 - two defaults were added to the legend, `fill = FALSE` and `bty = "n"`. the former removes the box drawn around the colors corresponding to each line and the latter removes the box around the entire legend

i haven't implemented i from #247 as it is not immediately obvious to me what the defaults should be.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,12 @@ mlr_2.4:
 - getTaskData now has recodeY = "drop.levels" which drops empty factor levels
 - new measures
 -- brier
+- new defaults for plotROCRCurves
+-- avg and add.diag set depending on inputs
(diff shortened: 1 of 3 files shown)


zmjones:
I ran `devtools::check` on my copy of this repo's `master` branch and the check also failed.

PR #260, opened 2015-03-16
Add neural network related learners

hetong007:
I add `neuralnet::neuralnet` as a new classification learner.

The interface of `neuralnet::neuralnet` is not following the common R packages:
- Not supporting `formula` with `Y ~ .` grammar
- It does binary classification, but the label can only be `numeric` binary values
- It does multiclassification, but requires the `formula` has multiple names on the left of `~`, e.g. `A+B~C+D+E`. 

Therefore I only created the binary classification learner for this function. Includes
- The learner script
- The test script
- Modified DESCRIPTION
- Update NAMESPACE with `roxygen2`

P.S. I am opening this pull request as an applicant of the GSoC project. And it failed to past Travis test mainly because `xgboost` is not on CRAN currently.


hetong007:
For `avNNet`, I can only find it in `caret`. It is basically averaging a bunch of `nnet` models with the same parameters.

Besides those params in `nnet`, there are two more parameters: `repeats` and `bag`:
- `repeats` means the number of `nnet` models.
- `bag` means whether to shuffle data or not.

I name these two new params following `caret`.

PR #262, opened 2015-03-18
Adding learners from sparseLDA package

kerschke:
Just added two learners from the sparseLDA package (sda and smda).


larskotthoff:
Hmm, the build failed, but that seems to be unrelated. Could you add something to NEWS about this though please?

kerschke:
About the added learners or the fact that the build fails?! ;-)

PR #265, opened 2015-03-19
fixes #247 plotROCRCurves

### LINKED ISSUE #247: Improvements in plotROCRCurves

zmjones:
- can pass arguments to plot via plot.args
 - ylim is set by finding the range of the y.values slot of elements of the list rocr.perfs across all elements
 - add.diag is now NULL by default
   + set to FALSE unless meas1 = "tpr" and meas2 = "fpr", appropriate
 - assertions were added in the generic method
 - fill = FALSE and bty = "n"
 - news updated
 - docs updated

cleaned version of previous PR #259


zmjones:
should I rebase this and fix the merge conflicts myself?

berndbischl:
would be thankful for any dilligent fixing, i am traveling

zmjones:
ok, should be good to go now, it was just a conflict with NEWS.

zmjones:
I was wondering what, if anything to do with failed tests due to things other than the code I am working on. I see that `predict.kohonen’ is no longer being exported and the associated error in `predict_classif_xyz.R` and I get a note about the `thirdparty` directory. Is there a reason (or a discussion I can read) about why `predict.kohonen` is referenced explicitly?

PR #290, opened 2015-04-30
plotROCRCurves with plotROC

zmjones:
rewrite plotROCRCurves to use plotROC
 - use ggplot2 instead of base graphics
 - optionally use ROCR::performance, otherwise just plotROC
 - optional produce interactive plot

I also added a test that does some basic things. If people are happy with how this looks (can change defaults and so on, or add things to the ggplot object before it is printed) or how it is structured then I will work on adding this to the tutorial.

```dt check``` passed with only the existing travis errors and a warning I am not sure about.

```
Warning in Sys.setlocale("LC_CTYPE", "en_US.utf8") :
  OS reports request to set locale to "en_US.utf8" cannot be honored
```


larskotthoff:
Ok, seems to be fine (apart from the unrelated errors), merging.

PR #296, opened 2015-05-07
Fixed threshold-prefix-issue, cf. 242

kerschke:


berndbischl:
merged on cmd line and improved

PR #307, opened 2015-05-25
plotROCRCurves with ggplot2

zmjones:
- use `ROCR::performance` but not `ROCR::plot`
 - handle averaging in mlr rather than `ROCR::plot`
- `ggplot2` instead of base graphics
 - test for basic input to all methods

Also I removed the global assignment I mentioned to Lars/Julia.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -16,6 +16,7 @@ mlr_2.4:
 - tuneIrace: allows to set the precision / digits within irace (using the argument 'digits'
   in makeTuneControlIrace); default is maximum precision
 - generateLearningCurve and plotLearningCurve helps to plot the learning curve demonstrating how the performance increases with more observations.
+- plotROCRCurves has been rewritten to use ggplot2
 
 - new learners:
 -- classif.bst
--- a/R/plotROCRCurves.R
+++ b/R/plotROCRCurves.R
@@ -14,33 +14,34 @@
 #' @param meas2 [\code{character(1)}]\cr
 #'   Measure on y-axis. Note that this is a measure name from *ROCR* and not from mlr!
 #'   Default is \dQuote{fpr}.
-#' @param avg [\code{character(1)}]\cr
-#'   How to average results from resampling.
+#' @param avg [\code{chracter(1)}]\cr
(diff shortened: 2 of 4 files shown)


zmjones:
the build error appears unrelated but i am not sure.

larskotthoff:
Ok, the master mlr build succeeds, but I can't tell the difference between them because travis cuts the output at 10000 lines. I'm not sure what the best way to proceed is there -- it may be beneficial for you to create a new travis job that uses your copy of the repository (also so you can test without creating pull requests) and runs only the relevant subset of the tests, but IIRC travis doesn't make running a subset of the tests easy unfortunately.

larskotthoff:
I ran the test locally and it worked fine. I'll merge this for now, we can always revert later.

schiffner:
Sorry for not commenting earlier, was visiting relatives and had to drive home first. I just saw we have now two files with nearly identical tests that should be merged:
https://github.com/berndbischl/mlr/blob/master/tests/testthat/test_base_rocr.R and
https://github.com/berndbischl/mlr/blob/master/tests/testthat/test_base_plotROCRCurves.R

zmjones:
ah ok, just did a poor job searching for existing tests. thanks!

schiffner:
Not a problem, I am merging the files now. (And this way, I just found a typo in the old tests that I missed before. So thank you :) )

zmjones:
[here](https://api.travis-ci.org/jobs/64010699/log.txt?deansi=true) is the full log. 

there is this

```
Client failed to connect to the D-BUS daemon:
  Unable to autolaunch a dbus-daemon without a $DISPLAY for X11
  GConf Error: No D-BUS daemon running
```

and then this

```
R: malloc.c:3795: _int_malloc: Assertion `victim->fd_nextsize->bk_nextsize == victim' failed.
```

and then check fails. i thought maybe this was my fault since the test was outputting to the display device  but this was when i was under the impression that there was no older test for `plotROCRCurves`.

larskotthoff:
Ah, useful to know that you can get the full output!

The error seems to be unrelated to your changes though, apparently it's caused by `tests/run-classif.R`.

PR #310, opened 2015-05-28
bugfix for plotROCRCurves

zmjones:
alpha should be NULL when horizontal or vertical averaging is used
tests added for all averaging param values


larskotthoff:
Thanks!

PR #314, opened 2015-06-02
separate plotROCRCurves into plot/generate

zmjones:
separates plotting and generation of ROCR curves


larskotthoff:
Thanks for the PR, but could you please check in `plotROCRCurves` if the right class of data is being passed in (as discussed)? Also, I would rename the data generation function to `generateROCRCurvesData` to make its purpose clearer -- at the moment it sounds like this may be generating a plot as well.

schiffner:
Thanks also from my side.
I agree with Lars.
Apart from that I also have really minor stuff concerning the documentation: Could you please add `@family roc` and `@family predict` for cross-references and make sure that each `@title`, `@description` and `@param` end with a period?

zmjones:
ok will do. about to eat dinner now but will do this afterwords!

@larskotthoff as far as the name I agree. I was imitating generateLearningCurve and plotLearningCurve. Perhaps that should be changed to match?

larskotthoff:
Yes, I would make it clear what each function produces, i.e. have `data` in the name of a function that produces data.

zmjones:
ok i made both of those changes. if you want i can modify generateLearningCurve accordingly or i can do that separately

larskotthoff:
I would do that separately. Thanks.

zmjones:
ah sorry, accidentally pushed to the same branch. let me know and if needed i can fix it. what would be the best way to do that with git?

larskotthoff:
Thanks, that's fine for now. I'll merge.

zmjones:
Sweet thanks!

schiffner:
Thanks from my side, too!

PR #315, opened 2015-06-02
update tutorial with generateROCRCurves

zmjones:
mostly I composed the new functions except for one instance at the beginning where I printed the output from `generateROCRCurves`. I included a bit about the (very simple) object created by `generateROCRCurves`.


larskotthoff:
Is this ready to go or does it need to be updated wrt to the latest changes?

zmjones:
Yes this is fully updated. This reflects changes for both `generateROCRCurves` and `generateLearningCurve`, the latter of which I meant to issue in a separate PR (sorry!).

PR #318, opened 2015-06-05
multiple filter methods for plot/getFilterValues

zmjones:
- getFilterValues can now take a vector of methods
 - plotFilterValues can use facetting
 - plotFilterValuesGGVIS can use a Shiny app to interactively
   show different filter methods
 - minor modifications to class 'FilterValues'
 - minor modifications to function 'FilterFeatures'

This passed the feature selection tests and check. I can't write tests for the Shiny app though.

### DIFF
--- a/R/filterFeatures.R
+++ b/R/filterFeatures.R
@@ -7,7 +7,8 @@
 #' @template arg_task
 #' @param method [\code{character(1)}]\cr
 #'   See \code{\link{listFilterMethods}}.
-#'   Default is \dQuote{rf.importance}.
+#'   Default is \code{NULL}, however, if \code{method} and \code{fval} are \code{NULL},
+#'   the default filter \dQuote{rf.importance} is used.
 #' @param fval [\code{\link{FilterValues}}]\cr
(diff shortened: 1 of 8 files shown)


zmjones:
So I implemented the latter suggestion. Now in `filterFeatures` the default method is `NULL`. If `fval` is passed and the `data` element has length 1 then I just extract the name of this element (i.e. that becomes the default method`. If there are multiple filters in `fval` then the user has to set one as `method` which is then extracted. I avoided the side effect of not having a default filter when calling `filterFeatures` by substituting the default filter (by removing the method argument) when no `fval` is passed and `method` is still `NULL`. I am not sure I documented it correctly though.

zmjones:
I talked with Bernd a bit and am now working on returning a `data.frame` instead of a named list from `getFilterValues`.

larskotthoff:
Ok, could you also please check the default behaviour (i.e. no method selected explicitly)? It looks to me like you'd get an error message at the moment, which is different to the behaviour before.

zmjones:
ok i will check that again.

PR #319, opened 2015-06-08
remove unneeded arguments from plot functions

zmjones:
- plotThresVsPerf loses linesize.
 - for plotFilterValues and plotFilterValuesGGVIS feat.type.cols is
   changed to a flag for whether to color by feature type (no extra
   color defaults). Users can change the colors by using
   scale_fill_manual in ggplot2 or scale_nominal in ggvis.


larskotthoff:
Merged, thanks.

PR #323, opened 2015-06-09
interface consistency

zmjones:
had to merge the branches to finish up the interface consistency changes. i guess i should have instead branched off of the branch that i wrote `generateThreshVsPerfData` in? that would be the best way to have done this?


zmjones:
Looks like the build failed because it exceeded 50 minutes. Any ideas for how I can fix this? I guess that it was barely under 50 minutes before and my addition of some tests pushed it over.

larskotthoff:
I would have made the two changes completely separately, but don't worry about it now. Again the interface changes to make them more consistent is something that should be mentioned in NEWS, in particular what functions are affected.

Regarding the build, you could use cheaper classifiers in the tests as they're not testing anything directly related to the classifiers. You could even create some very cheap mock classifiers (although this might be something that should be done separately).

zmjones:
The problem was that in the second bit I wanted to make changes to `generateThreshVsThreshData` which hadn't been merged into master yet.

I'll add some more info to news.

I think I can just create some `BenchmarkResult`, `Prediction`, `ResampleResult` (lists thereof) objects and then use them to test the generation/plotting functions. Where should I do that though?

larskotthoff:
You can put them in `helper_objects.R`. And I just remembered that we do actually have mock learners already in the appropriately named `helper_mock_learners.R`. You could use those to speed things up.

larskotthoff:
Merged, thanks.

PR #324, opened 2015-06-10
control over color/facetting for some plots

zmjones:
- plotThreshVsPerf
 - plotThreshVsPerfGGVIS
 - plotLearningCurve
 - plotLearningCurveGGVIS

smarter defaults
arguments for color/facet for ggplot
arguments for color/interaction for ggvis
added additional args to news

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,7 +11,7 @@ mlr_2.4:
   of the plotting first, into well-defined objects; these can then be plotted with mlr or custom code;
   the naming scheme is always generate<Foo>Data and plot<Foo>
 - getFilterValues is deprecated in favor of generateFilterValuesData
-- plotFilterValues and  can now plot multiple filter methods using facetting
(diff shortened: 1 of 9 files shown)


larskotthoff:
Thanks -- just out of curiosity, why did you comment the tests for the ggvis versions?

zmjones:
I run them interactively but they all call Shiny, which you have to kill manually, so they don't seem to work with this sort of test. I am not sure if there is a better way to test them.

larskotthoff:
Ok, this is something to think about for the future, but fine for now.

Another thing that I've noticed is that facet and colour are mutually exclusive and binary, so you could have just a single argument and infer the other from that. The advantage would be less code for argument checking. Would it make sense to only specify facet and infer colour from that? Again something to think about.

Merging the PR now.

zmjones:
Yes that probably does make sense. I can just change that now. I also think I missed the case where the user sets both to the same thing, which this would also solve.

PR #334, opened 2015-06-16
introduced budget parameter to tuning

kerschke:
once again adding the budget parameter


kerschke:
I just tried to stay consistent with the rest of the code.. And as you can see, you guys already used `convertToShortString` within the same function.
Or am I misinterpreting it?

berndbischl:
Yes, that functon is for complex objects like lists and so on.

berndbischl:
You just have an integer.

berndbischl:
Do you have unit tests that try to check whether the budget is properly respected?

berndbischl:
I just merged and will iterate a bit

berndbischl:
If you read this go on hangout so we can review 10 min?

PR #335, opened 2015-06-16
introduce tunable-flag

kerschke:
iterated over the parameters of the integrated learners and eventually defined tunable = FALSE (as the default is tunable = TRUE)


berndbischl:
I sometimes see "tunable = TRUE"
This looks like a bug. Please dont merge without PROPERLY checking.

jakob-r:
I have to admit I oversaw these three instances
@kerschke any specific reason?

berndbischl:
They look like typos, as the params seem to be stuff like "verbose".

And usually we want to rely on the default TRUE. And not note it explicitly.

kerschke:
You're right. They were just typos and I've fixed them.
Sorry for that..

PR #345, opened 2015-06-17
generatePartialPrediction

zmjones:
+ adds exported functions
  - generatePartialPredictionData
  - plotPartialPrediction
  - plotPartialPredictionGGVIS
+ adds non-exported functions
  - generateFeatureGrid
  - doPartialPredictionIteration
+ tests

curious about what people think about using facetting/interaction rather than background color as done in `plotLearnerPrediction`. I could probably add that functionality as well, or replace what I've done with that.

Also the tests are somewhat limited because execution is slow, so if anyone can think of likely problems I would appreciate that as well.


larskotthoff:
Thanks, could you update NEWS as well please?

zmjones:
yep always seem to forget that

larskotthoff:
No worries. Re the facetting, having more options (e.g. background colour) is always good, so I'd go for it unless it turns out to be too difficult to implement in a general fashion.

zmjones:
Ok would you rather I do that before a merge or is doing it later ok too?

larskotthoff:
Fine in a separate PR.

Regarding the tests, it would be good if you could (in the future) wrap the calls to the functions to check that no warnings are produced (see https://stackoverflow.com/questions/22003306/is-there-something-in-testthat-like-expect-no-warnings). It would also be great if you could think about better ways of testing in general, in particular wrt "verifying" the generated plot. For example, you could save as SVG, parse the XML and verify simple things like that the number of circles corresponds to the number of rows in the data.

zmjones:
Yea I agree that the testing is unsatisfying. I will start looking into that.

zmjones:
An over 50 minute error. I timed my new test and it is < 1min. I think maybe the package installation took too long. Maybe retry the build a bit later?

larskotthoff:
Just restarted it.

larskotthoff:
Looks good, merging.

schiffner:
Thanks, that looks really great!

Regarding visualization for classification: I don't know why that didn't occur to me earlier. Have you considered spinograms/spine plots? They might give a clearer representation for a larger number of classes and better reflects the fact that the sum of probabilities/proportions is 1.
I am also ok with the current solution. Just thought you might want to consider it.

zmjones:
Oh that is interesting. I gave them a quick look and it is essentially a stacked bar chart with the width mapped to a feature? I am not at all opposed to adding more functionality. I am already going to try out some of the things in `plotLearnerPrediction`.

schiffner:
> Oh that is interesting. I gave them a quick look and it is essentially a stacked bar chart with the width mapped to a feature?

Exactly. Spine plots are stacked bar charts (for categorical x) with the width representing frequencies of x values, as you already said, and spinograms are stacked histograms (for numeric x).
Just for inspiration, if you haven't already found these R functions: https://stat.ethz.ch/R-manual/R-devel/library/graphics/html/spineplot.html, http://www.inside-r.org/packages/cran/vcd/docs/spine

PR #349, opened 2015-06-17
fix doc error in generateLearningCurveData

zmjones:
return type was wrong and needed a link to the object documentation


schiffner:
Thanks again!

PR #352, opened 2015-06-18
bugfixes for generatePartialPredictions

zmjones:
one is related to factor levels in `generateFeatureGrid` and the other the group aesthetic for probability plots in `plotPartialPrediction`.

I am working on making tests for plots and adding more error checking for the plotting functions (here specifically) as well. Neither are adequate now.

### DIFF
--- a/R/generatePartialPrediction.R
+++ b/R/generatePartialPrediction.R
@@ -154,7 +154,7 @@ plotPartialPrediction = function(obj, facet = NULL) {
   if (all(target %in% obj$task.desc$class.levels)) {
     out = reshape2::melt(obj$data, id.vars = obj$features, variable = "Class", value.name = "Probability")
     out$Class = gsub("^prob\\.", "", out$Class)
-    plt = ggplot2::ggplot(out, ggplot2::aes_string(feature, "Probability", color = "Class"))
+    plt = ggplot2::ggplot(out, ggplot2::aes_string(feature, "Probability", group = "Class", color = "Class"))
   } else {
     plt = ggplot2::ggplot(obj$data, ggplot2::aes_string(feature, target))
   }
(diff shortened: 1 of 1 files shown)


larskotthoff:
Thanks. Is it possible to have a unit test for the bug you've fixed?

zmjones:
Not for the plot (yet) but for the factor levels yes. I will amend this in a moment.

berndbischl:
Will pull it in now, zach made a note for the unit test

larskotthoff:
I would have at least waited until the travis build finished, but ok :)

berndbischl:
Sorry, wanted to try it out :)

berndbischl:
(will not do this often though, busy and also was talking to zach)

zmjones:
I have a question about `generateFeatureGrid`. When an integer feature is passed I am not sure what to do. `seq.int` returns a numeric even if `length.out` is less than the range of `abs(from-to`. e.g, `is.integer(seq.int(1:3, length.out = 3))` is `FALSE`. If I am correct in thinking that since integers are also numerics (and are treated the same by all learning algorithms) then this won't be a problem. If you wanted you could subsample or bootstrap the observed values of the feature (an option to the function).

zmjones:
I guess I could just do `fmin:fmax` and throw an error if that isn't equal to `cutoff` but that seems like a pain.

berndbischl:
Ints are not treated always the same. I would be careful here. If it was an int before, it should be later.

berndbischl:
I mean, there is a certain stepsize right for every feature? Cant you round this to the next reasobale value?

berndbischl:
Or are we simply talking about the data type that seq.int returns?

zmjones:
yes I suppose I could do that

berndbischl:
But try to make sure that it stays an int, to be sure...

zmjones:
I think I can just coerce the sequence to an `int` and that will be fine.

zmjones:
actually `sort(rep(min:max, length.out = cutoff))` is probably the best. thanks for the help

PR #353, opened 2015-06-18
error checking and tests for generatePartialPrediction

zmjones:
separate test for generateFeatureGrid
more tests for generatePartialPredictionData (including expecting errors)
more error checking in generatePartialPredictionData

still working on tests for the actual plots. when i get that working i'll go back and write tests for all of them


larskotthoff:
Thanks.

PR #356, opened 2015-06-22
compute partial predictions for features w/o conditioning

zmjones:
- adds interaction flag to generatePartialPredictionData,
   when FALSE and length(features) > 1, partial predictions
   for each feature are computed
 - plotPartialPrediction and plotPartialPredictionGGVIS can
   use facetting/shiny to visualize the above type of output

### DIFF
--- a/R/generatePartialPrediction.R
+++ b/R/generatePartialPrediction.R
@@ -10,6 +10,14 @@
 #' @param data a \code{data.frame} with the same columns as are present in the training data.
 #' @param features \code{character}\cr
 #'   A vector of feature names matching the training data.
+#' @param interaction \code{logical(1)}\cr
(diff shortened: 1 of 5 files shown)


larskotthoff:
Thanks. Could you update NEWS as well please?

zmjones:
I didn't think it made sense to since this function wasn't in 2.4, so it will appear at the next CRAN release that it is just a new function with this feature.

berndbischl:
yup, if a new function appears, only mention it under "new functions". We do not need to track changes before it ever went to cran. only in extreme case explain in the text above why this function was added. in most cases it should be inferrable from the function name. and people can simply read the doc file.

larskotthoff:
Ok, thanks.

PR #360, opened 2015-06-23
fixes for plot functions

zmjones:
sorry this is in the same PR, branching mistake on my part

 - `plotPartialPredicton` and `plotPartialPredictionGGVIS` only plot the positive class for binary classification, though they retain the legend to make this clear
 - `plotFilterValuesGGVIS` drops the least important features first when `n_show` is less than the number of features

### DIFF
--- a/R/generateFilterValues.R
+++ b/R/generateFilterValues.R
@@ -217,12 +217,12 @@ plotFilterValuesGGVIS = function(fvalues, feat.type.cols = FALSE) {
   }
 
   gen_plot_data = function(data, sort_type, value_column, factor_column, n_show) {
-    if (sort_type != "none")
+    if (sort_type != "none") {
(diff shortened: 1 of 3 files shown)


larskotthoff:
Thanks, merging.

larskotthoff:
Could you also please add a test that verifies that the least important features are dropped (in a future PR)?

zmjones:
That change was only in the ggvis function, so I'll have to have a way to test shiny apps to do that.

zmjones:
I could factor out that little function and test that though.

larskotthoff:
Yes to both :) See http://cran.r-project.org/web/packages/RSelenium/vignettes/shinytesting.html for testing Shiny. I'm particularly keen on this because it's easy to break things there without anybody but a poor user noticing.

zmjones:
Yea definitely true. Thanks for the link I will check this out.

PR #362, opened 2015-06-30
plotPartialPredictionGGVIS title error

zmjones:
for classification tasks the y-axis is always a probability


larskotthoff:
Thanks -- could you check `obj$task.desc$type` instead of the length of `target` though please? That will make the intention of the check much clearer.

zmjones:
yep that makes more sense. thanks

PR #364, opened 2015-07-01
substitute measure names for ids in plots

zmjones:
e.g., instead of "tpr" we'd have "True positive rate". 

I thought about doing this for filters too, but the filter descriptions are often somewhat long. It would be nice to have a more proper name though.

I think this needs to done for the predictions from the survival learners. As @schiffner noted they are all measures of risk but what precisely they are differs amongst some of them and is a pretty important distinction I would think.

### DIFF
--- a/R/generateLearningCurve.R
+++ b/R/generateLearningCurve.R
@@ -123,6 +123,12 @@ plotLearningCurve = function(obj, facet = "measure") {
   assertChoice(facet, mappings)
   color = mappings[mappings != facet]
 
+  for (i in 1:length(obj$measures)) {
+    measure.name = obj$measures[[i]]$name
+    colnames(obj$data)[colnames(obj$data) == obj$measures[[i]]$id] = measure.name
+    obj$measures[[i]] = measure.name
+  }
+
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, I'm merging this for now, but it looks to me like you should be replacing those identical bits of code in different places with a function :)

Maybe there's a way of shortening the filter descriptions (or even introducing a new field for a short description)?

zmjones:
Yes I initially wrote one. There is a bit of inconsistency in how measures are stored in the objects which is why I didn't include it. `generateThreshVsPerfData` has an element which is the id element of the measures, and `generateLearningCurveData` has an element which is a list of the actual measure objects. I think the latter case is a better choice, as then the operation is more simple. But I didn't think the code repetition between the ggplot2/ggvis versions was a problem since they will (presumably) be combined at some future date. Does that sound reasonable?

larskotthoff:
Ok, that sounds good. I would keep an eye on not having too much repeated (or very similar code) in the future though.

zmjones:
Yep, fair enough.

PR #365, opened 2015-07-01
adds generic permutation.importance filter

zmjones:
- user specified aggregation, contrast, measure, and learner

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -23,6 +23,9 @@ mlr_2.5:
 - new learners
 -- multilabel.rFerns
 
+- new filters
+-- permutation.importance
+
 mlr_2.4:
 - WrappedModel printer was slightly improved
 - ReampleResult now stores the runtime it took to resample in a slot
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -357,3 +357,46 @@ makeFilter(
     })
   }
 )
+
+makeFilter(
+  name = "permutation.importance",
+  desc = "the aggregate difference between feature permuted and unpermuted predictions",
+  pkg = character(0L),
+  supported.tasks = c("classif", "regr", "surv"),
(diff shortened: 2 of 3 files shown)


larskotthoff:
Thanks. Could you document it as well please?

zmjones:
where should I document it?

larskotthoff:
Ah, good point -- I thought that the other ones were documented in the source file. This is probably something that should go in the tutorial (my main point here is that it's not entirely obvious what `contrast` should be and an example and some explanation would help). I'll merge this for now.

zmjones:
Yea no that occurred to me too. Maybe we could add some more information to the filter objects?

larskotthoff:
Yep, sounds good. It might even be worth making this a separate section of the tutorial and then only referencing it in the feature selection section. @berndbischl @schiffner

zmjones:
Yea that makes sense. 

It occurred to me also, just now, that `generateFilterValuesData`, which can take a vector of filter methods, might be a little weird with this filter. Shouldn't it be possible to calculate and compare permutation importance using many different learners and/or measures, contrast functions, aggregation functions, etc.?  Maybe this could be made more general by writing a `generatePermutationImportanceData` function which is not set up to only support feature selection. 

However this is maybe a bit messy. Thoughts?

larskotthoff:
That makes sense, but you could get that by simply having a loop around the filter function, calling it with different arguments every time. Unless I'm missing something this doesn't warrant a new function for me.

zmjones:
Now that I think about it more this is consistent with the others in that you can pass different parameters to other functions and get different output. This just *seemed* different because the user specifies more or less the whole thing.

zmjones:
It would be nice to have a section on this in the tutorial though.

PR #367, opened 2015-07-02
fix for "get" trouble

zmjones:
- getThreshVsPerfData now returns measure objects
 - simplifies name/id replacement code in learningcurve and threshvsperf
   plot functions

### DIFF
--- a/R/generateLearningCurve.R
+++ b/R/generateLearningCurve.R
@@ -123,11 +123,8 @@ plotLearningCurve = function(obj, facet = "measure") {
   assertChoice(facet, mappings)
   color = mappings[mappings != facet]
 
-  for (i in 1:length(obj$measures)) {
-    measure.name = obj$measures[[i]]$name
-    colnames(obj$data)[colnames(obj$data) == obj$measures[[i]]$id] = measure.name
-    obj$measures[[i]] = measure.name
-  }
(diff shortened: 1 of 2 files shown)


schiffner:
Looks good to me. Thanks!
@berndbischl : Do you want to check, too?

zmjones:
Yea that is right. I wasn't aware of mapValues. I can use that though. Will change momentarily.

berndbischl:
Actually we wrote that for when we use ggplot. You so often want to "rename" stuff for plots.

zmjones:
Ah ok.

zmjones:
mapValues is in use now. that is a nice solution

larskotthoff:
Could you add a unit test that fails with the old code please?

zmjones:
I don't know how it fails. Bernd said that he didn't really either.

schiffner:
Yes. I didn't manage to reproduce the error when running the code interactively. It only failed when building the tutorial (on my computer and on Travis).

What we could do is to replace one of the mlr built-in measures in the tests for e.g. plotThreshVsPerf, plotLearningCurve and others by a user-defined one. This wouldn't have helped with the get-issue today, but might be useful in the future.

larskotthoff:
Hrm, it seems to be very specific to knitr -- no idea what exactly is going on. Merging with the hope that I don't break everything again :)

zmjones:
Ok, I'll issue a new PR for the test.

PR #368, opened 2015-07-02
Naming

zmjones:
I added a test for the custom measure case that broke the tutorial and expanded the original idea a little bit so that it is more consistent across all the plots.


larskotthoff:
Wait, this test won't actually break with the previous implementation that used `get()`, or am I missing something?

zmjones:
No. I was just doing what @schiffner suggested in #367.

larskotthoff:
Ok thanks, just checking.

zmjones:
It would be good if we could regenerate the tutorial after this. Partially to be sure everything worked correctly, but also to have the better plot labels in there.

larskotthoff:
Sure, I'll kick off another build after this is merged.

Issue #366, opened 2015-07-02
how to label stuff in plots


berndbischl:
1) How do we "name" / label objects in plots.
Answer: This is why we have "id"s in objects. So if you show something related to a task, you use its id. Same for measures, there is also this function measureAggrName, that concats the aggregation method's name to the measure name. But that gets kinda long, and seems to be overspecific.  Just mentioning it for completeness sake.
For learners it is a bit complicated, because we have: id, name, and shortname. "shortname" is, well, short, and was exactly intended to be used in plots.
So please use it there, always a default.

berndbischl:
Zach says:

I should add that although the measure ids have been the default labels for measures. I don't get why that is the case. Wouldn't the measure names be better if their length isn't a problem? Why would we prefer tpr to true positive rate. It would be much harder to misread the latter. The default in plotROCRCurves is to use this longer measure name (though it is a measure from ROCR::performance there). I was doing this for consistency and because I thought it made sense for the above reason.

berndbischl:
@zmjones Lets move the discussion from emails here

zmjones:
Hello I am here!

zmjones:
So what do you want me to do on this?

zmjones:
And I want to know precisely how I broke things so I can avoid it again. I guess I kind of knew `get` was bad but not any specific reasons, tests passed, etc.

berndbischl:
So as always it seems to be a think one might want to configure. In cases where length is not a problem, take the longer name. Maybe also as a default. In cases where it is likely that long names are problematic, take the shorter ones. But make that choice changeable for the user I guess. Try to have something consistent across the plots.

berndbischl:
For the get stuff, lets do this over emails. Please tell me what you want to achieve in the 3 lines of code there

zmjones:
Ok well I put that in the email. You should just revert it though. That commit stands on its own, nothing else builds on it.

zmjones:
It is in my first reply, the third little section.

berndbischl:
I answered. Can you please simply repair as I suggested?

berndbischl:
And tell me here whether you agree to what I said above about the plots. I mean if you already use a longer name where it is feasible, maybe you do not need to make this configurable. Just think about this for a second, I do not want to complicate things too much.

berndbischl:
Maybe the best might be to have an argument: 
labels.learners, labels.measures, lables.task.
[...]

PR #376, opened 2015-07-07
solution to labelling problem in #366

zmjones:
- aggregations now have long names
 - if measure ids/names match in generation functions, aggregation ids
   are appended to column names
 - plotThreshVsPerf/GGVIS and plotLearningCurve/GGVIS have argument
   'pretty.names' which optionally uses measure and aggregation names
   instead of ids for plotting (default is TRUE)
 - new helper functions measureAggrPrettyName and replaceDupeMeasureNames

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -13,6 +13,15 @@ mlr_2.5:
 - there was a nasty bug in measure "mcc". fixed and unit tested. and apologies.
 - removed getTaskFormulaAsString and improved getTaskFormula so the former is
   not needed anymore
+- aggregations now have a 'name' property, which is a long name
+- generateLearningCurveData and generateThreshVsPerfData now append the aggregation id
+  to the output column name if the measure ids are the same
+- plotLearningCurve, plotLearningCurveGGVIS, plotThreshVsPerf, plotThreshVsPerfGGVIS now have an argument
+  'pretty.names' which plots the 'name' element of the measures instead of the 'id'.
+- makeCustomResampledMeasure now has arguments 'measure.id' and 'aggregation.id'
+  instead of only 'id' which corresponded to the measure. Also, 'name' and note (corresponding to the measure)
+  as well as 'aggregation.name' have been added.
+- makeCostMeasure now has arguments 'name' and 'id'.
 
 - new functions
 -- getDefaultMeasure
--- a/R/Aggregation.R
+++ b/R/Aggregation.R
@@ -28,6 +28,8 @@ NULL
 #'
 #' @param id [\code{character(1)}]\cr
 #'   Name of the aggregation method. (Preferably the same name as the generated function)
+#' @param name [\code{character(1)}]\cr
+#'   Long name of the aggregation method.
 #' @param fun [\code{function}]\cr
 #'   A function with following signature: \code{function(task, perf.test, perf.train, measure, group, pred)}
 #'   \itemize{
@@ -42,12 +44,13 @@ NULL
 #' @return \link{Aggregation} object
 #' @examples
 #' # computes the interquartile range on all performance values
-#' test.iqr = makeAggregation(id = "test.iqr",
+#' test.iqr = makeAggregation(id = "test.iqr", name = "Test set interquartile range",
 #'   fun = function (task, perf.test, perf.train, measure, group, pred) IQR(perf.test))
 #' @export
-makeAggregation = function(id, fun) {
+makeAggregation = function(id, name, fun) {
   assertString(id)
-  setClasses(list(id = id, fun = fun), "Aggregation")
+  assertString(name)
+  setClasses(list(id = id, name = name, fun = fun), "Aggregation")
 }
 
 #' @export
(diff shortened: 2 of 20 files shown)


zmjones:
do not just merge this please! this is my own interpretation of what is going on in #366 and should be discussed.

berndbischl:
Does this simply mean we shall review it carefully? Or are you not done?

zmjones:
The former. Looks like travis failed because of the binary installation.

larskotthoff:
What is `replaceDupeMeasureNames` supposed to do? It looks like it replaces names that are duplicate, but doesn't remove the duplicates.

zmjones:
Yea that is right. Maybe it should have a more intelligent name. `substituteDupeMeasureNames`?

larskotthoff:
Why do you replace the duplicates with their aggregate names?

zmjones:
I'm not. For measures with duplicate ids I append the aggregation measure to the id, which then goes into the column name in the container object.

Then if you want pretty names in the plot there is a flag that can do the same thing with the long names.

larskotthoff:
Ok, I guess what I'm wondering is what the function is for. My initial impression was that you want to make things that have the same name distinguishable, but that doesn't seem to be the case from what you're saying. Is the purpose to identify the measures that have been aggregated and label them accordingly?

zmjones:
Your initial impression was my intention. I am confused about what you are confused about.

The function takes measures, extracts their ids or names (depending on whether `x` is `name` or `id`, finds the indices of those with counts > 1, pastes the aggregation id or name (using `measureAggrName` or `measureAggrPrettyName` accordingly) to those measure ids or names, and then returns a vector of modified ids/names.

larskotthoff:
Ah right, those duplicated measures would have different aggregations. That was the part I was missing. Makes perfect sense.

zmjones:
Right, otherwise they would be the same measure.

larskotthoff:
Exactly, hence my confusion. I'll leave this open so Bernd can check, but as far as I'm concerned it can be merged.

zmjones:
Ok cool. Can you restart the build?

larskotthoff:
Yep, did that just now.

larskotthoff:
BTW you should be able to do that too.

zmjones:
Ah, didn't know that. Thanks!

zmjones:
Going to bed but it occurred to me that I didn't document the changes to makeCustomResampleMeasure and makeCostMeasure in NEWS. The former case is one area I'd like someone to look at, as I thought it was a bit messy (esp. using the argument inheritance).

zmjones:
This is fixed now. I am curious why `makeCostMeasure` sets up the aggregation scheme differently than `makeMeasure`.

schiffner:
Thanks.

About your question about aggregation for `makeCostMeasure` and `makeMeasure`: I'm not sure what you mean. Normally, `makeCostMeasure` generates a `Measure` object that can be aggregated like every standard measure, e.g. `mmce`.

zmjones:
Bernd just explained it to me. I was just confused about the level that was being aggregated at. I think I get it now. 

I am making a few minor changes Bernd suggested (just checking them now) and will push them momentarily.

schiffner:
Ah ok, thanks.

zmjones:
Ok it should be good to go now after Travis finishes.

berndbischl:
I am trying to fix my stuff, I broke travis

berndbischl:
My stuff should be fixed now

zmjones:
Ok, so how should I do this. Should I rebase my branch off of master?

berndbischl:
Can you do this? And rerun travis? That would be great

zmjones:
Ok rebased now, Travis is rerunning now.

zmjones:
It appears that the build failed due to something related to #377.

zmjones:
Rebased again and restarted.

zmjones:
issue with tuneThreshold now

berndbischl:
seen and fixed

zmjones:
ok rebased and restarted

larskotthoff:
Thanks, will merge.

schiffner:
I know the PR is already merged and I'm a bit late with this...

Can we please set a default `name = id` in `makeAggregation` (just like in `makeMeasure` and `makeCostMeasure`) to prevent that the changes in `makeAggregation` break older mlr code (and most importantly the tutorial :) )?

larskotthoff:
Ah, so that's what it was. I'm on it.

schiffner:
Thanks.

I have hopefully fixed the tutorial. It was only in one section where it makes sense to explain that `makeAggregation` has a `name` argument.

larskotthoff:
And I've found an easy way to speed up building the tutorial, so we should know whether this fixed it soon :)

schiffner:
Cool. Again, thanks. :+1:

zmjones:
Ah thanks! Just when you think you've fixed everything... I really should just build the tutorial along with all the other normal checks/tests.

larskotthoff:
No worries, that's what we have Travis for!

schiffner:
Exactly! :)

Issue #123, opened 2014-07-10
LiblineaRLogReg / weights


berndbischl:
Hi Eric,

why do you do this:

```
-  LiblineaR(data = d$data, labels = d$target, ...)
 +  if (!is.null(.weights)) {
 +    .weights = .weights[unique(names(.weights))]
 +  }
 +  LiblineaR(data = d$data, labels = d$target, wi = .weights, ...)
  }
  
```
The part:

= .weights[unique(names(.weights))]

berndbischl:
@kerschke 
Can you please close this soon?

kerschke:
After having read your exhaustive dialogue, I'm not quite sure whether I get the task correct. So let me try to summarize it and correct me, if I'm wrong.
1) mlr allows the usage of observation weights but not class weights (at least in general).
2) The former (observation weights) is precisely, what is defined in the properties of a learner (and also highlighted in the learner summary table within the tutorial)
3) Some learners (which @studerus already listed), are able to handle class weights on their own.

And based on that, we now need a property (e.g. classweights) for those learners, so we can easily select them (e.g. using `listLearners("classif", properties = "classweights")`? However, in that case, one would still need to look into the original learner and try to find the correct name of the class weight argument. So, instead of (or in addition to) the property, we actually might need an additional list element for the learner, which states the name of the class weight parameter. So in case of ksvm, it might look like the following:

```{r}
      ...
      makeNumericVectorLearnerParam(id = "class.weights", len = NA_integer_, lower = 0),
      makeLogicalLearnerParam(id = "fit", default = TRUE)
    ),
    par.vals = list(fit = FALSE),
    ## add classweights property:
    properties = c("twoclass", "multiclass", "numerics", "factors", "prob", "classweights"),
    ## add id of the classweights-parameter, here "class.weights":
    classweights = "class.weights",
    ...
```
Or am I misunderstanding the task?

berndbischl:
> 1) mlr allows the usage of observation weights but not class weights (at least in general).

Yes, in principle. But note that you can use obs weights to generate class weights with the WeightedClassesWrapper. And that this wrapper also allows to directly use a class.weighting parameter for this purpose. But you have to enter in manually. Read that Wrapper now completely if you havent.

> 2) 

Yes. We might have called the property "obs.weights" to be precise. 

> Some learners (which @studerus already listed), are able to handle class weights on their own.

[...]

PR #378, opened 2015-07-07
including class weights (cf. issue #123)

kerschke:
Included the class weights property for classification learner and also added some man-files (which apparently were not committed so far).


berndbischl:
For now it is OK, but pls dont add man add NAMESAPCE to the PR. this just makes it harder to merge I guess.

berndbischl:
Don use \code\link for stuff like character. Pls fix

berndbischl:
Also pls do NOT put unrelated stuff into one PR. You are also changing params in other learners and so on.

berndbischl:
And the unit test you wrote belongs to group "base", not "classif". 
"classif" is (as one can really see) just a group for all indivdual classif learners. 
Please really look at this with open eyes, so I dont have to check all of this....

berndbischl:
AND: getClassWeightsParams looks not good. 
- You are using assertion wrong
- Why do you return a named list of values? 
- Isnt the arg just a string and not a char vec?

kerschke:
> And the unit test you wrote belongs to group "base", not "classif".
"classif" is (as one can really see) just a group for all indivdual classif learners. 

I thought about that, but (as you mentioned in previous conversations) *base* should contain only the minimum stuff, so I decided to put it into *classif*.

> Also pls do NOT put unrelated stuff into one PR. You are also changing params in other learners and so on. 

In that case, it would be great if everybody could run all checks and all builds before committing. Otherwise, I'll always have to check in each commit which files I need to ignore (again)..

> Don use \code\link for stuff like character. Pls fix

Why not? I mean, I definitely can change it, but do not see the point.

> AND: getClassWeightsParams looks not good. 
> - You are using assertion wrong

Could you actually be more precise? What am I doing wrong / how should I use it instead?

> - Why do you return a named list of values? 

Because I thought it's useful to have the entire information of the class weight parameter of the learner - not just the ID. If you prefer the latter, I could do that instead..

> - Isnt the arg just a string and not a char vec?

Actually you're right. I forgot that there exists `assertString` as well.

berndbischl:
> base should contain only the minimum stuff,

No, base should contain as MUCH as possible, if it creates no problems on CRAN. Eg timing problems. 
If you can create a FAST test, put in in base. Stuff only goes into other groups if it is a big loop over like all filters or does something technically annoying like parallel stuff. check the basic API in base

berndbischl:
> Don use \code\link for stuff like character. Pls fix Why not? I 

Because I say so (= it is consistent with the rest of package)

berndbischl:
> You are using assertion wrong

You cannot put an arbitrary logical clause into assert

berndbischl:
Lets try to talk about the rest in 20 mins, some stuff I dont get

kerschke:
Alright.

PR #379, opened 2015-07-08
use prediction get methods in partialPrediction functions

zmjones:
uses 3bfd3faa9ec4687747979faa66aac6d8bc2c4ef

some small adjustments since `getPredictionProbabilities` only returns the positive class for binary classification, which I pulled out in `generatePartialPredictionData` before.

### DIFF
--- a/R/generatePartialPrediction.R
+++ b/R/generatePartialPrediction.R
@@ -102,9 +102,12 @@ generatePartialPredictionData = function(obj, data, features, interaction = FALS
 
   if (td$type == "regr")
     target = td$target
-  else if (td$type == "classif")
-    target = td$class.levels
-  else
+  else if (td$type == "classif") {
+    if (length(td$class.levels) > 2L)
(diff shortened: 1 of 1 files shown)


larskotthoff:
Thanks. Could you add a unit test for the different cases you handle in the change please?

zmjones:
I am not sure what you mean. There are already tests for binary and multiclass classification and regression in the tests for this, so if any of these changes broke that then they would fail.

larskotthoff:
Ok, I'm probably misunderstanding what you did here -- is this only a cleanup commit, i.e. the behaviour of the function for different numbers of classes hasn't changed?

zmjones:
Correct, this is just using the helper "get" functions just added for `Prediction` objects, instead of pulling things out manually, and this resulted in me having to modify the internal behavior slightly.

larskotthoff:
Ok, thanks for the clarification. Merging.

PR #383, opened 2015-07-08
introducing class weights param

kerschke:
I integrated `class.weights.param` in the learner structure (cf. issue #123). Note, that I also removed `wcw.param` from the `makeWeightedClassesWrapper`, because the learner now contains that information.


kerschke:
Could someone of you check, if you get the same errors? I mean, on my mac I get the same errors as Travis does right now - but I can't explain them at all. As Travis shows, the errors occur in test_base - for measures, plotViperCharts, etc. However, when I run those tests separately (by calling test_base_measure, test_base_plotViperCharts, etc.), I don't get an error..

berndbischl:
> Note, that I also removed wcw.param from the makeWeightedClassesWrapper, 
> because the learner now contains that information.

Pls, if you do API changes like this: ASK!

If you remove it, OK, but then document how one would proceed if we forgot to tag a param as the class.weights.param.

Actually: I would like to still have that arg in there. Just use a default setting. This is also what we agreed upon...

kerschke:
I did ask.. and since I didn't know when I'll get a response, I decided to do it this way. In the worst case, I'll just have to revert those changes..

kerschke:
And yes, you suggested that. But you also mentioned that I should use my head - which I did ;)

berndbischl:
Isnt it more flexible if we keep it with the arg? I would suggest that.

kerschke:
Well.. why would you still need the arg? Only if the learner is incomplete, right?!

berndbischl:
Hmm. Ok then at least doc in the wrapper how the mechnism would work, what one would need to do if we forgot to add that info to the learner. Well. I mean it will also break existing code. Lets not change API if we dont have to. Here we dont have to. lets keep it.

berndbischl:
Well its your fault. You broke it.

Here

  learners = listLearners(binaryclass.task, "class.weights")

You cannot do that in base I guess.

berndbischl:
listLearners loads all packages of all learners. We now have too many.

berndbischl:
(thats a line from one of your new unit tests)

kerschke:
Why not? I want to test all learners that allow class weights.. if I run that file on its own, it works perfectly fine.

berndbischl:
Just read what I said. listLearners loads a crapload of packages. If you run this with the other test files, even more packages get loaded. At a certain point, you reach the magic number 100. Crash.

berndbischl:
Try to move that test to the file "learners_all". There exactly that is done already.

kerschke:
Ok. Thanks :)
So I'll fix that part and also re-initiate the param. Right?

berndbischl:
Yes. And dont commit man/

PR #386, opened 2015-07-09
added reference, family and yeast example to multilabelBinaryRelevanceWrapper.R

IcedragonP:


berndbischl:
1) Please adhere to style guide. You have stull like "Labels" in there. Which should be lower case.

2) it was nice that you added yeast.task! But then use it directly in the example! Why do you extract the data, then create a task again?

berndbischl:
3) Dont use qassert! Use assertCharacter. Look the options like min.len etc.

Pls fix.

IcedragonP:
I think it is more illustrative if you begin with the data and not with the task. fixed the issues, added yeast.RData and deleted yeast.task.RData

berndbischl:
Argh. No. please talk about stuff. 

1) I want to have the multilabel task. As a talk. As an accessible object in mlr.
2) If you want to show how to create a task, do it in an example for make task. You can IMHO show this by using "stupid random dummy data". The example of the BRWrapper is the wrong place for this. This uses the yeast.task.

Also: Please dont use commit messages like "fixed". Make them precise.

IcedragonP:
Ok, sometimes it is very difficult to "guess" what you really want and this is a lot of timewasting for me.

berndbischl:
well i hinted at the simpe solution. just ask and commuicate. we other guys have to do the same thing :)

IcedragonP:
Ok. is it ok now? Because then I can push some tests...

berndbischl:
I'll merge on the cmd line now, please do PRs later with one commmit with precise message

berndbischl:
merged

PR #393, opened 2015-07-14
name bug in plotThreshVsPerfGGVIS

zmjones:
"perf" should have been replaced by "performance" in the plot labels


larskotthoff:
Thanks.

PR #396, opened 2015-07-15
tests for ggplot functions

zmjones:
- writes to svg, reads in xml, and parses xml to count points, lines,
   bars, and facets


larskotthoff:
Very nice, thanks! I'll merge once Travis has finished.

PR #398, opened 2015-07-16
allow listMeasures("multilabel")

IcedragonP:
did not work before (I need this for the tutorial)


berndbischl:
Plase always try in these cases to extend the unit tests, too.
That is the most important aspect. Because this here indicates that we dont test this.

Will merge now and do this.

PR #400, opened 2015-07-17
H20

giuseppec:
Can't push anymore, therefore pull request


giuseppec:
parameter boundaries should be checked from h2o guys, I skipped the ones where I was not sure...

PR #404, opened 2015-07-18
adding a getter for converting caret param set defaults to mlr

kerschke:
added a getter for the default param sets of caret, cf. issue #207 and PR #403

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -47,6 +47,7 @@ mlr_2.5:
 -- plotBenchmarkResult, convertBMRToRankMatrix, generateRankMatrixAsBarData, plotRankMatrixAsBar,
    generateBenchmarkSummaryData, plotBenchmarkSummary,
    friedmanTestBMR, friedmanPostHocTestBMR, generateCritDifferencesData, plotCritDifferences
+-- getCaretParamSet
 
 - new measures
 -- hamloss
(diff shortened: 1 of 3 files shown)


larskotthoff:
So wait, why is this a separate PR? Shouldn't this be a part of #403?

kerschke:
Just wanted to make sure that all changes are within the same commit - @berndbischl once mentioned that he prefers it this way, because it's easier to read. And if I would have changed something now afterwards and added it to the PR than there would be 2 commits, right?
Now, we can just close #403 and hopefully merge #404.

berndbischl:
Ok, but then in general close the older PR when you make a new one. So we dont get confused.

berndbischl:
I would like to give some feedback on stuff in the code, I can better do this on hangout

kerschke:
So you would like to hangout now?

berndbischl:
yup so i dont forget

berndbischl:
and NEVER forget to update NEWS!

kerschke:
Arghh.. that rebase and squash stuff is horrible.. apparently it's still considering the several commits. So I don't see the advantage. But probably I did something wrong..
Do you want a clean, single commit (then I would just put everything into a new PR) or can you live with it for this time?

PR #407, opened 2015-07-20
Add new learners from SwarmSVM

hetong007:
I figured out the previous error in travis is because I haven't set values to some parameters, and taht triggered a warning.

Is it true that the parameter `default` in `makeIntegerLearnerParam` is only for documentation, other than setting a default value and do pass to the training function?


larskotthoff:
Ok great. Not sure what you mean with your question -- the default parameter tells mlr what the default value of that particular parameter is.

hetong007:
If I set 
```r
makeIntegerLearnerParam(id = "m", default = 3, lower = 2)
```
I don't have `m=3` passing to my training function if I don't pass it explicitly, do I?

larskotthoff:
Yes, the default parameters aren't set automatically.

hetong007:
Although there's actually an ERROR from other functions, now the learners i add have passed the travis check.

larskotthoff:
But travis passed?

hetong007:
Yes. The existing ERROR is introduced in the master branch of mlr: https://travis-ci.org/mlr-org/mlr#L3100

larskotthoff:
Hmm ok, this works for me locally and travis passes despite the error. I'll have a look.

larskotthoff:
Ok, I've just fixed the error in travis. Could you rebase please? Sorry about the hassle.

hetong007:
It seems good now. Do I still need to squash all the commits? I am afraid the commits from other authors will also be squashed.

larskotthoff:
Thanks, I'll merge this.

PR #410, opened 2015-07-22
Add new learner RSNNS::mlp 

hetong007:
Add `RSNNS::mlp` as a new classification learner.

I have just learnt to use `par.vals`, so I also editted `gaterSVM` and `clusterSVM`.

One side effect is generating `man/getCaretParamSet.Rd` by `roxygen2` automatically, it seems forgetten in a previous commit on the master branch.


larskotthoff:
Thanks. Please make a separate PR for changes to other learners in the future though.

PR #418, opened 2015-07-29
Add hill climbing algorithm to learn the weight to do average in StackedLearner.R

hetong007:
As stated in https://github.com/mlr-org/mlr/issues/288#issuecomment-124255728, I have implemented the hill.climb algorithm to learn the weights of base learners, then do the average. However as in the issue, this idea may need more discussion. My next step is to add examples and test files, but before that I want to know exactly where to put this algorithm in. From the current status, I think it fits in the other functions well. So this code is not at a mergable stage, but it might need someone's review and suggestion. Thank you very much.

I have also fix one of the `FIXME` in `StackedLearner.R`: https://github.com/mlr-org/mlr/blob/master/R/StackedLearner.R#L201. Instead of calling `table` in each row, I count the table manually for each class. This speeds up the calculation dramatically as in the following example:

```{r}
oldfun = function(probs) t(apply(probs, 1L, function(x) (table(factor(x, 1:5) )/length(x))))
newfun = function(probs) {
  mat = matrix(0,nrow(probs),5)
  for (i in 1:5) 
    mat[,i] = rowSums(probs==i) 
  mat = mat/ncol(mat)
  return(mat)
}

probs = matrix(sample(1:5,100000,replace = TRUE), 20000,5)

system.time({res1 = oldfun(probs)})
system.time({res2 = newfun(probs)})
sum((res1-res2)^2)
```

The running time is 5 secs versus <0.01 sec, and the difference is zero. Another question is do you think I should put this fix in a new pull request? Thank you.


larskotthoff:
Thanks Tong. Could you please update NEWS and add tests for the new method? It would also be great if you could clean up the commits -- there are a lot of unrelated ones (it might be cleaner to close this and  do a separate PR).

hetong007:
Thanks Lars for the comment.  I will rebase the commits and reopen the PR.

Issue #73, opened 2014-05-01
resample(..., extract = [defaultValue]) with Wrappers


jakob-r:
The usual case when using a wrapped learner like `makeFeatSelWrapper` is that you also want to access the used Features at the end. So why doesn't these wrappers tell `resample` directly what should be extracted in the default case? It's not that costly to  extract with `getFeatSelResult` by default, isn't it?

Suggestions:
* makeTuneWrapper : getTuneResult
* makeFeatSelWrapper : getFeatSelResult
* makeFilterWrapper : getFilteredFeatures
* Did I miss something?

When preforming the benchmark (Issue #40) for example it would be also more handy when the user doesn't have to take care of it. Or should it be implemented in the benchmark locally?

jakob-r:
Pull request here: https://github.com/mlr-org/mlr/pull/419
Will auto merge in 3 days

berndbischl:
Apparently we have this now

PR #419, opened 2015-07-29
Solve Issue #73: automatic resample extraction

jakob-r:


R/getResampleExtract.R
```diff
@@ -0,0 +1,35 @@
+getResampleExtract2 = function(learner) {
+  UseMethod("getResampleExtract2")
+}
+
+getResampleExtract2.NULL = function(learner) {
```
mllg: This should be `getResampleExtract2.default`.

R/getResampleExtract.R
```diff
@@ -0,0 +1,35 @@
...
+
+getResampleExtract2.FeatSelWrapper = function(learner) {
+  c(list(FeatSelResult = getFeatSelResult), getResampleExtract2(learner$next.learner))
+}
+
+getResampleExtract2.FilterWrapper = function(learner) {
+  c(list(FilteredFeatures = getFilteredFeatures), getResampleExtract2(learner$next.learner))
+}
+
+getResampleExtract = function(learner){
```
mllg: What are you trying to do here? This seems pretty odd.

tests/testthat/test_base_TuneWrapper.R
```diff
@@ -36,8 +36,9 @@ test_that("TuneWrapper", {
   lrn1 = makeLearner("classif.ksvm", predict.type = "prob")
   lrn2 = makeTuneWrapper(lrn1, resampling = makeResampleDesc("Holdout"), par.set = ps1, control = makeTuneControlGrid())
   expect_equal(lrn2$predict.type, "prob")
-  r = resample(lrn2, binaryclass.task, makeResampleDesc("Holdout"), measures = mlr::auc)
+  r = resample(lrn2, binaryclass.task, makeResampleDesc("CV"), measures = mlr::auc)
```
mllg: Try to keep the unit tests as inexpensive as possible. Better use Holdout or CV with iters = 2.

jakob-r: My mistake - I changed it just to see what happens. forgot to change it back

R/getResampleExtract.R
jakob-r: will change

R/getResampleExtract.R
jakob-r: We find out what we can extract by default for each wrapper. For multiple wrapped learners we can extract more than just one information so we collect all usable functions and then build one function which returns a list. Compare with [this tutorial part](http://mlr-org.github.io/mlr-tutorial/devel/html/nested_resampling/index.html#filter-methods-with-tuning).

mllg: Well okay, then throw in some comments please. The `force` is not required btw.

jakob-r:
I am not sure what's the best way to test this. 
a) Test if the extract works in any function wich have `extract` as an argument
b) Collect all possible functions in one file and test if extracts works as expected
c) Do not write additional test to save execution time

jakob-r:
d) put the extract tests only there were we use the things (Wrappers) we want to extract things from

jakob-r:
Additionally:
When we use multiple Wrappers, by default, a list with all different extractions is returned. However if we only have one Wrapper which offers an extraction at the moment we don't put the result in an extra list (with only one item)
Somehow I don't like this inconsistency: Not defined if result is list or single item. What is the consensus here?

R/getResampleExtract.R
jakob-r: added comments on top of file

PR #420, opened 2015-07-29
add learners from rknn

florianfendt:
Hi Guys,
I added the regr. and classif. learners from rknn.
As this is my first contribution you might want to check thoroughly.
Cheers,
Flo


jakob-r:
I checked not so thoroughly but it looks pretty good! 👍 Thanks.
Don't worry about the "All checks have failed" from travis. That problem is on our side (or even beyond).
One tiny remark. The few occurrences in the tests of `<-` could bother someone 😉

florianfendt:
@jakob-r 
ok thanks, the <- should be gone now ;)
So should I undo my commit on travis.yml or leave it like this now?

jakob-r:
travis.yml should be fine. The error is caused by another dependency.

It surprises me that `rknn` doesn't offers any possibility to access the single predictions of the random knns. That would be pretty nice to use them to obtain some kind of probability for the classification and an uncertainty estimator for the regression. If you feel eager it would be nice to further investigate / contact the author.
Also we could use the package as a variable selection method if it's promising.

florianfendt:
ok, so I'll leave the PR as is.
Yeah, I thought so, too. Will contact the author about that!
Thank's heaps for your help.

larskotthoff:
@florianfendt The travis build for master is fixed now -- could you rebase please so we get passing checks here?

florianfendt:
travis can't locate the package? What's the problem here?

larskotthoff:
Looks like it's not available as a binary package. Just remove it from the list of binary packages in `.travis.yml`.

larskotthoff:
Ok, I have no idea why Travis says that the build succeeded. Looks like libgmp isn't installed. Could you add 

     - sudo apt-get install libgmp-dev

to the `before_script:` section of `.travis.yml`?

florianfendt:
still the same problem unfortunately.
Travis says: "No repository set, so cyclic dependency check skipped"
How can I fix this? All solutions I found on the net suggest to set the repository in the .Rprofile,
but how can this be achieved with travis?
Sorry, I've never worked with travis before.

larskotthoff:
The cyclic dependency check message is harmless and unrelated. It looks like before_script installs the package too late. Could you put the apt-get line in its own `before_install:` section please? That should fix it.

florianfendt:
Ok great, this worked like a charm!
Thanks a lot Lars

PR #423, opened 2015-07-31
Add hill climbing algorithm to learn the weight to do average in StackedLearner.R

hetong007:
This is a cleaner version of https://github.com/mlr-org/mlr/pull/418 . 

As stated in https://github.com/mlr-org/mlr/issues/288#issuecomment-124255728, The hill climbing algorithm can be used to learn the weights of the base models.

I have also fix one of the `FIXME` in `StackedLearner.R`: https://github.com/mlr-org/mlr/blob/master/R/StackedLearner.R#L201. Instead of calling `table` in each row, I count the table manually for each class. This speeds up the calculation dramatically as in the following example:

```{r}
oldfun = function(probs) t(apply(probs, 1L, function(x) (table(factor(x, 1:5) )/length(x))))
newfun = function(probs) {
  mat = matrix(0,nrow(probs),5)
  for (i in 1:5) 
    mat[,i] = rowSums(probs==i) 
  mat = mat/ncol(mat)
  return(mat)
}

probs = matrix(sample(1:5,100000,replace = TRUE), 20000,5)

system.time({res1 = oldfun(probs)})
system.time({res2 = newfun(probs)})
sum((res1-res2)^2)
```

The running time is 5 secs versus <0.01 sec, and the difference is zero.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,6 +38,7 @@ mlr_2.5:
 - Learner Propoerties are now implemented object oriented as a state of a Learner.
   Only RLearners have the properties stored in a slot.
   For each class the getter can be overwritten.
+- The hill climbing algorithm for stacking (Caruana 04) is implemented as method 'hill.climb' in 'makeStackedLearner' to select models from base learners, which is equivalent to weighted average.
 
 - new functions
 -- getDefaultMeasure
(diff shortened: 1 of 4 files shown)


larskotthoff:
Thanks Tong!

PR #428, opened 2015-08-05
Fix broken travis by xgboost

hetong007:
Fix the broken Travis caused by `xgboost`.

### DIFF
--- a/R/RLearner_classif_xgboost.R
+++ b/R/RLearner_classif_xgboost.R
@@ -29,10 +29,11 @@ makeRLearner.classif.xgboost = function() {
       makeIntegerLearnerParam(id = "early.stop.round", default = 1, lower = 1),
       makeLogicalLearnerParam(id = "maximize", default = TRUE)
     ),
+    par.val = list(nrounds = 1),
     properties = c("twoclass", "multiclass", "numerics", "factors", "prob", "weights"),
     name = "eXtreme Gradient Boosting",
     short.name = "xgboost",
-    note = ""
+    note = "`nrounds` set to 1 by default"
   )
 }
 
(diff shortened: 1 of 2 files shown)


hetong007:
It seems that the `xgboost` issue has been resolved. But the error in the example remains.

larskotthoff:
Hmm, all I see are lots and lots of warnings about mtry. I'll run the build again.

larskotthoff:
I'm also seeing lots of warnings like

    makeRLearner.classif.xgboost: warning in makeRLearnerClassif(cl =
    "classif.xgboost", package = "xgboost", par.set =
    makeParamSet(makeUntypedLearnerParam(id = "params", default =
    list()), makeDiscreteLearnerParam(id = "booster", default = "gbtree",
    values = c("gbtree", "gblinear")), makeIntegerLearnerParam(id =
    "silent", default = 0), makeNumericLearnerParam(id = "eta", default =
    0.3, lower = 0), makeNumericLearnerParam(id = "gamma", default = 0,
    lower = 0), makeIntegerLearnerParam(id = "max_depth", default = 6,
    lower = 0), makeNumericLearnerParam(id = "min_child_weight", :
    partial argument match of 'par.val' to 'par.vals'

larskotthoff:
Seems to be fine apart from that typo.

hetong007:
Sorry for the typo! I have fix it now. In the log file the error from the example remains, and there's no warning/error on `xgboost`'s learners now. It seems the `mtry` messages are caused by the error.

larskotthoff:
Thanks. I'll merge this and investigate what's going on with the other stuff.

larskotthoff:
Ok, it seems that the source of the messages are your hill climbing tests. I've fixed it now I hope. Two things for the future:

- Never *ever* use `listLeaners()` with `create = TRUE` in a test. This will load *all* the packages and break everything because the maximum number of DLLs will be reached.
- List the learners you use for testing explicitly and make sure that they are cheap to run. If you use `listLearners()` the learners that are being tested will change as the package develops and lead to unexpected behaviour. Keep the unit tests cheap and as predictable as possible.

I should have caught this when checking your PR for this though, so it's really my fault :)

hetong007:
Oh I see. Thank you very much for pointing that out! I am still on the way learning about every piece of `mlr`, and I will keep it in mind.

PR #430, opened 2015-08-06
Add compress model algorithm in Stackedlearner

hetong007:
As stated in this thread:https://github.com/mlr-org/mlr/issues/288#issuecomment-125407467, I have  implemented the model compression algorithm as a succeeding algorithm:

1. Run the hill climbing algorithm on a relatively small size of data
2. Generate some psuedo training data and label the data by the ensemble model from the hill climbing algorithm
3. Train a super learner (neural network by default) to mimic the behaviour of the ensemble. The result is this model, thus the size and the prediction efficiency of the model is reduced.

Since it relies on the hill climbing algorithm, I have implemented this algorithm in the same file. To make examples work, I also removed the `listLearners` command.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -39,6 +39,7 @@ mlr_2.5:
   Only RLearners have the properties stored in a slot.
   For each class the getter can be overwritten.
 - The hill climbing algorithm for stacking (Caruana 04) is implemented as method 'hill.climb' in 'makeStackedLearner' to select models from base learners, which is equivalent to weighted average.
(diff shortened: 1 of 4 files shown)


larskotthoff:
Thanks, merging. It would be good to be able to pass the super learner as an argument.

larskotthoff:
Wait, you've already implemented that. Ignore my comment :)

hetong007:
Thank you Lars for reviewing my code :)

Issue #373, opened 2015-07-06
LiblineaR


giuseppec:
Currently, we have three learners `classif.LiblineaRMultiClass`, `classif.LiblineaRLogReg` and `classif.LiblineaRBinary`. However, all of them support multiclassification since one-vs-all is implemented in LiblineaR for LogReg and Binary SVMs. We should use one learner for each classification method that is implemented in LiblineaR:

type
0 – L2-regularized logistic regression (primal)
1 – L2-regularized L2-loss support vector classification (dual)
2 – L2-regularized L2-loss support vector classification (primal)
3 – L2-regularized L1-loss support vector classification (dual)
4 – support vector classification by Crammer and Singer
5 – L1-regularized L2-loss support vector classification
6 – L1-regularized logistic regression
7 – L2-regularized logistic regression (dual)

For example:
`classif.LiblineaRL2LogReg` (allowing primal and dual option)
`classif.LiblineaRL1LogReg`
`classif.LiblineaRL2SVC` (allowing primal and dual option)
`classif.LiblineaRL1L2SVC`
`classif.LiblineaRL2L1SVC`
`classif.LiblineaRMultiClassSVC`

florianfendt:
I'll start working on this today!

florianfendt:
While I'm on it, why not implement type 11, 12 ,13 (for regression) as well, right?

berndbischl:
@florianfendt 
What is the status here?

You do the other "measure" units first or this? Version 2.6?

florianfendt:
This can be closed. We already merged that about a week ago.

berndbischl:
thx. you can close in these cases yourself. But asking also does not hurt

florianfendt:
ok will keep that in mind thx

PR #432, opened 2015-08-08
Issue #373 LiblineaR 

florianfendt:
So, I added the LiblineaR learners according to @giuseppec 's suggestion. I also added the Regression learners from LiblineaR.

I wanted to delete the old learners we already have, as these are unnecessary now, but then I would have to delete a line in 
test_base_checkTaskLearner.R
and wanted to check back with you first if this is alright?


larskotthoff:
Looks ok to me. Please go ahead with removing the old learners.

jakob-r:
I don't know about all these `makeDiscreteLearnerParam(id = "type", values = 3L)` etc.
If we say that one type is one specific learner then this is not a parameter anymore. Especially if it has only one value. 
IMHO we need something like
```r
trainLearner.classif.LiblineaRL2L1SVC = function(.learner, .task, .subset, .weights = NULL, ...) {
  d = getTaskData(.task, .subset, target.extra = TRUE)
  LiblineaR::LiblineaR(data = d$data, target = d$target, type = 3L, ...)
}
```

florianfendt:
That seems more appropriate. I'll change that as soon as I'm returning from my holidays

florianfendt:
ok, changed everything where there was only one type per learner!

jakob-r:
In the comments it says `#default epsilon dependent on type`. Now that we have a different learner for each type we should/could add the suitable default epsilon for each type, no?

florianfendt:
@jakob-r 
thx, fixed it now

jakob-r:
No, thank you for all the effort :wink: ! Looks very fine to me. Maybe @mllg or @berndbischl want to have a look and merge.
If they don't say anything I will merge in some days.

giuseppec:
I also reviewed the code. Here my suggestions:
- do we need the `cross` parameter from `LiblineaR`? I think it should either be removed or have at least the `tunable = FALSE`.
- For `classif.LiblineaRL2LogReg`, for example, the user can choose between `type = 0` and `type = 7`. In `note = "type 0 is primal and type 7 is dual problem"` one can look up what the type means. I think it would be more user-friendly if the user could directly choose  `type = "primal"` or `type = "dual"`. Same holds for `LiblineaRL2SVC`
- `makeRLearner.classif.LiblineaRL2SVC` still has the `type` parameter. This should be changed as suggested by jacob (post of 10 Aug). 
- I know it is currently not possible with paramHelpers to use a default for `epsilon` when it depends on other parameters (see also https://github.com/mlr-org/mlr/issues/438 where according to @berndbischl this should be possible soon). You added a comment in the line above, which is good, however, you should also add a `#FIXME' in the beginning of your comment (so this issue can be found much easier).
- Another point, which I think is rather general because this also refers to other learners (such as `glmnet`, for example): Do we want to add support for sparse matrices in mlr, e.g. by additionally store them in the task?

berndbischl:
> do we need the cross parameter from LiblineaR? I think it should either be removed
> or have at least the tunable = FALSE.

set tunable = FALSE

> For classif.LiblineaRL2LogReg, for example, the user can choose between type = 0 and type = 7.
> In note = "type 0 is primal and type 7 is dual problem" one can look up what the type means. 
> I think it would be more user-friendly if the user could directly choose type = "primal" or type = "dual". > Same holds for LiblineaRL2SVC

Lets not change the algorithms too much. 

> makeRLearner.classif.LiblineaRL2SVC still has the type parameter. 
>This should be changed as suggested by jacob (post of 10 Aug).

OK

> I know it is currently not possible with paramHelpers to use a default for epsilon when it depends on 
> other parameters (see also #438 where according to @berndbischl this should be possible soon).
> You added a comment in the line above, which is good, however, you should also 
> add a `#FIXME' in the beginning of your comment (so this issue can be found much easier).

OK lets do this.

> Another point, which I think is rather general because this also refers to other learners
> (such as glmnet, for example): Do we want to add support for sparse matrices in mlr, e.g. by 
> additionally store them in the task? We can discuss this here: #471

We have a different issue for this, lets not discuss this here.

@flo:
Please do this and squash all commits into one for the pull request.

florianfendt:
ok, did all that!

berndbischl:
thx for being dilligent and picking up stuff quickly!

larskotthoff:
Thanks, merging. For this like this, could you please in the future mention in NEWS that a renaming has taken place? At the moment there's no information on that, it looks just like a bunch of new learners were added.

Issue #431, opened 2015-08-07
Bug in performance calculation for a ResamplePrediction?


schiffner:
Some functions like `tuneThreshold` or `generateThreshVsPerfData` call `performance` on a `ResamplePrediction`.
For example `tuneThreshold` and `generateThreshVsPerfData` both contain something like

```
performance(setThreshold(pred, ...))
```

where `pred` and thus `setThreshold(pred, ...)` are `Prediction`s (and therefore can also be `ResamplePrediction`s) to calculate performance values for different threshold settings.

`performance` seems to treat the `ResamplePrediction` as a `Prediction`, the aggregation functions in the `Measure`s are not taken into account, which can lead to totally wrong performance values, for example:

```
set.seed(123)
m = mmce
m = setAggregation(m, b632)

rdesc = makeResampleDesc("Bootstrap", iters = 5, predict = "both")
r1 = resample(makeLearner("classif.lda", predict.type = "prob"), sonar.task, rdesc, list(mmce, m))
r1$aggr
# mmce.test.mean      mmce.b632 
#     0.2695533      0.1898192 

d1 = generateThreshVsPerfData(r1, list(mmce, m))
plotThreshVsPerf(d1)

performance(r1$pred, list(mmce, m))
#     mmce     mmce 
#0.109375 0.109375
performance(setThreshold(r1$pred, 0.5), list(mmce, m))
    # mmce     mmce 
#0.109375 0.109375

set.seed(123)
rdesc = makeResampleDesc("Bootstrap", iters = 5)
r2 = resample(makeLearner("classif.lda", predict.type = "prob"), sonar.task, rdesc)
r2$aggr
# mmce.test.mean 
#      0.2695533 

d2 = generateThreshVsPerfData(r2)
plotThreshVsPerf(d2)

performance(r2$pred)
#      mmce 
#0.2690217 
performance(setThreshold(r2$pred, 0.5))
#      mmce 
#0.2690217
```

I hope I'm not doing something wrong.
In case of `tuneThreshold` I'm not sure if it was originally intended for `ResamplePrediction`s.

jakob-r:
The problems are two:
* `predict = "both"` has not been taken into account in `performance.R`
* `performance.R` is not made for `ResamplePrediction` and it just works by chance

jakob-r:
Well. Theoretically we have all the information we need in `r1$pred$data` to split the double occurrences back to train and test  as this information is in `pred$data$set`.
[...]

PR #434, opened 2015-08-10
adds plotROCCurves using to generateThreshVsPerf

zmjones:
- adds aggregate argument to generateThreshVsPerfData
   which gives performance per iteration or avg performance
   (temporary until #431 fixed)
 - tests/news update


larskotthoff:
Thanks, merged.

PR #435, opened 2015-08-13
Add Feature-subspace Aggregating Ensemble Algorithm

hetong007:
As mentioned here: https://github.com/mlr-org/mlr/issues/288#issuecomment-125206990, I have implemented the "Feating" ensemble algorithm to do the classification. The basic idea is the following:

1. Build a tree which uses the same feature on each level to split the data. Each feature only appear at most once in each tree.
2. On each leaf either train a model on enough data points, or simply use the majority otherwise.
3. Loop through all the possible trees, or a random subset of it.

Currently I created two individual files `tests/run-feating.R` and `tests/testthat/test_feating.R` to test, because I don't know other more appropriate location. Maybe you have better idea. If so I would like to modify them.


larskotthoff:
Thanks, looks good. Just two things. Could you please squash the irrelevant commits and add a reference to the paper you took the algorithm from in the documentation?

PR #436, opened 2015-08-13
Add Feature-subspace Aggregating Ensemble Algorithm

hetong007:
This is a clearner version of https://github.com/mlr-org/mlr/pull/435. The git log of my repository became a bit messy after merging back from mlr-org/mlr, so I didn't figure out how to squash every commit comveniently. I reforked mlr and copied the changed files.

As mentioned here: https://github.com/mlr-org/mlr/issues/288#issuecomment-125206990, I have implemented the "Feating" ensemble algorithm to do the classification. The basic idea is the following:

1. Build a tree which uses the same feature on each level to split the data. Each feature only appear at most once in each tree.
2. On each leaf either train a model on enough data points, or simply use the majority otherwise.
3. Loop through all the possible trees, or a random subset of it.

Currently I created two individual files `tests/run-feating.R` and `tests/testthat/test_feating.R` to test, because I don't know other more appropriate location. I have also added the reference of the paper.


larskotthoff:
Thanks, merging.

PR #437, opened 2015-08-14
adds individual conditional expectation curves

zmjones:
- refactoring and expansion of generatePartialPredictionData
 - interact insteat of interaction argument for
   plotPartialPredictionGGVIS
 - bugfixes and additional tests


larskotthoff:
Thanks. You're very diligent about updating NEWS, but in this case I don't think that's even needed as there's no change to a released version. I'll merge and fix that.

zmjones:
Ah ok thanks. Forgot about that. I will update the tutorial soon as well.

PR #439, opened 2015-08-16
allow getCaretParamSet to use tasks

kerschke:
This PR is just a minor fix of `getCaretParamSet`, allowing to use `task` instead of `x` and `y`.

### DIFF
--- a/R/getCaretParamSet.R
+++ b/R/getCaretParamSet.R
@@ -18,10 +18,8 @@
 #'   generating the grid of tuning parameters. \code{caret} generates either as
 #'   many values per tuning parameter / dimension as defined by \code{length}
 #'   or only a single value (in case of non-tunable \code{par.vals}).
-#' @param x [\code{data.frame}]\cr
-#'   A data frame with the input data.
-#' @param y [\code{factor} | \code{numeric}]\cr
-#'   A response vector.
+#' @param task [\code{\link{Task}}]\cr
+#'   Learning task, which might be requested for creating the tuning grid.
 #' @param discretize [\code{logical(1)}]\cr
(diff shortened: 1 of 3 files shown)


larskotthoff:
Thanks. Looks like you forgot to update the examples -- could you fix that please?

kerschke:
The R-file contains the updated examples. I just didn't add the updated Rd-file (on purpose, because I was told I should not always update the Rd-files)..
Or do you see any other missing updates?

larskotthoff:
If you change the documentation, you should regenerate it and commit the changes to the Rd files :) The Travis build is failing because of this.

kerschke:
Ok, I can do that. However, that was exactly what I was told in the past ;-)

berndbischl:
Why does travis break if RDs are not regenerated?

kerschke:
Why doesn't travis actually generate the RDs on its own?

berndbischl:
I told Pascal not to commit them as diffs are otherwise harder to read. I never saw an error on travis due to this

berndbischl:
Am I wrong? If so sorry, but I though this is auto-generated on travis anyway?

larskotthoff:
No, Travis doesn't regenerate the documentation. This could be set up, but then we'd need Travis to commit that back to the repo as well (again can be set up, but a little bit more work).

Not regenerating the documentation won't break Travis unless the old examples don't work anymore like in this case.

kerschke:
Ah ok.. So this means it used to work in the past, because I didn't break the previous examples with my previous commits?

larskotthoff:
Yep.

larskotthoff:
Thanks, merging.

PR #444, opened 2015-08-17
make center, fmin, and fmax a named list in generatePartialPredictionData

zmjones:
use a named list instead of a vector to avoid errors due to features that are not numeric


larskotthoff:
Thanks.

PR #445, opened 2015-08-17
parallelmap drops lists of length 1

zmjones:
- don't call rbind on result of one call to
   doPartialPredictionIteration
 - adds tests for bug


larskotthoff:
Are you saying that this is a `parallelMap` bug?

zmjones:
no i don't think so. i am looking at the docs now.

PR #446, opened 2015-08-17
output mismatch when using center argument

zmjones:
- don't call rbind on result of one call to
   doPartialPredictionIteration
 - adds tests


larskotthoff:
Thanks, merging.

Issue #298, opened 2015-05-08
gsoc-hyperparameter


kerschke:
I'll try to track all the changes within the wiki (https://github.com/berndbischl/mlr/wiki/GSOC2015-Tracker:-Improving-mlr%27s-hyperparameter-and-tuning-system.)

Nevertheless, feel free to commit in here as well.

berndbischl:
Your PH / irace fix is not good. Please always try to thing about what the real problem is. 
What you implemented still truncates (small) values. So 1e-5 still becomes 0 with digits = 4.
Right?

I mean, the code was not perfect before and irace maybe does not allow scientific notation, but if we look at such stuff in case of problems we should try to strive for the best solution.

berndbischl:
Maybe what we should do:

1) Write garbage range in string
2) Overwrite x$boundary$<paramname> in the resulting irace object with the proper values
3) then remove our digit argument
4) add unit test to show what we specify gets into irace object exactly

Problem: We kinda access irace internal structure. But I dont know a better alternative now?

kerschke:
Would you prefer to have that fix within ```ParamHelpers::convertParamSetToIrace```? Or should I leave that function untouched and just fix it within ```mlr::tuneIrace```? I would go with the former one, because I don't see a reason for NOT using the correct boundaries.
Given the case that we discard those rounding issues, do we actually still need the ```digits``` argument within ```ParamHelpers::convertParamSetToIrace```? If we just keep the correct boundaries, I don't see a reason for keeping this argument.

kerschke:
I've played a little bit with that issue and here's what I came up with:
1) I agree to keep the correct boundaries when given the output of ```convertParamSetToIrace```. This way, we still have the correct values and get rid of a redundant argument (```digits```) within ParamHelpers.
2) We still need the ```digits``` argument within mlr's ```tuneIrace```. If we remove it there, ```irace``` will automatically use 4 digits - which will lead to a lot of errors, e.g. when having the boundaries ```1e-5``` to ```2e-5``` (both will be rounded to ```0``` and thus, will be "out of bounds").
We could obviously try to set a "smart" default for the digits, but then the question would be: how many digits are smart? Let's assume we have the interval from ```1.3e-7``` to ```2.72456252e-5```. Should we take 8 digits (needed to represent all values of the lower bound), 13 digits (needed to represent all values of the upper bound and of the distance between the boundaries) or any amount of digits in between?!
[...]

Issue #223, opened 2015-01-30
Use sub-model trick for hyperparameter tuning


zachmayer:
Many models actually fit multiple models in one "pass".  For example, a `gbm` with 1,000 trees can be used to make predictions for 1-999 trees and a `glmnet` model can be used to predict for many different lambda values along the regularization path.

The various methods for parameter turning (e.g. grid search and irace) should be aware of this.

For example, with a gbm, you only need to fit each combination of distribution, interaction.depth, n.minobsinnode, shrinkage, bag.fraction and train.fraction with the maximum number of trees, and you get predictions for each lesser value of `n.trees` for free.

berndbischl:
Yes, I thought about this many time, you are perfectly right.

I am unsure whether it is "trivial" to make that work. But we can of course try.
Before, we have more a less favored "generality" instead of exploiting such specific extra tricks.

But I guess we are now very general and can start with the fiddling around for the special cases ;-)

I need to warn you though, that we have some other issues before that, that are more important than this. So I am unsure how long solving this here will take.

Can I also ask - just out of curiosity - what you are using the package for?

zachmayer:
Using the "sub model trick" with irace is definitely non-trivial.  I think it's easier for grid search.

I do a lot of predictive modeling, mostly using the caret package to tune gbms.  I noticed that mlr supports the xgboost package, so I decided to check it out!

caret only does grid search at the moment, and utilizes the sub-model trick.  Basically, you remove the lower values of `n.trees` from the model fitting grid and use those models to evaluate predictions on the full tuning grid.

berndbischl:
1) We are always happy to cooperate. If you want to help with this, I am happy to support.

2) Can we make a list of (important) models and params where this would be doable.

3) Am I right to assume, that if it works for param x, I would never actually "vary" x in tuning, I would always set it to the max value, and then look at all submodels "below" / "included"?

4) If 3) is true, I guess we might simply change mlr's tunerFitnessFunction, which resamples a learner on a task, with given settings.
(Actually a better point to do this, might be evalOptimizationState?)
We use this for basically every optimizer so we have smart code reuse.
So for "submodels" we could simply check all submodels, and return the best value we found? 
We would then simply also need to store the "sub params"?

Does this make sense? I am just improvising here, without looking at stuff in detail....

zachmayer:
1) If I have time to take a crack at this, I will!
2) Yes, I can do this
[...]

PR #453, opened 2015-08-19
initiate submodel approach

kerschke:
Initialized usage of submodel approach (cf. issue #223 and #298).
So far, I only implemented it for `gbm` and `randomForest`. I will try to add `glmnet` next, but I wanted to confirm, whether the current approach is ok.


larskotthoff:
@kerschke Is this still relevant or can it be closed?

ja-thomas:
Moved the diff to the (closed) issue if we will need this in the future for some reason

Issue #456, opened 2015-08-27
plotPartialPrediction and other new plotters by zach


berndbischl:
Hi Zach,

could you please look a few minor things I noticed while trying out the plot?

1) You dont use the "standard" in mlr to document parameters. 
Eg the curly brackets around data types and the line breaks and so on.
Can you please look and an older mlr function to compare and adapt your roxygen code.

2) For some stuff we should use roxygen templates 
Eg we already have ret_ggplot2. We should also have that for ggvis.
Copypaste is always bad.

3) Does it make sense to doc all plot/generate functions for a specific plot on 1 page?
I would say "yes". Can we do that? 
With "@rdname".

Example code for now:

```
m = train("regr.randomForest", bh.task)
d = generatePartialPredictionData(m, data = getTaskData(bh.task), features = getTaskFeatureNames(bh.task))
plotPartialPredictionGGVIS(d)
```

4) Can we please pass the task the model was trained on into the function instead of the data.frame?
That seems much better. Then we can also set a default for the features. Eg.
all features if there are <= 10, or the first 10?

5) If you look at the x axis above for some features, numbers overlap and the axis looks bad. 
I know we said we do "default settings", but shall we not at least turn the labels by 45 degrees and doc this?

berndbischl:
@zmjones

berndbischl:
Shall we also talk about whether you would be up to contribute further to the project?

zmjones:
I definitely plan on continuing, though of course now I have a bunch of other things to do as well; that is, if you all are still interested.

I will work on the documentation today.

Do you think it always makes sense to only use the training data as the data for computing partial predictions? If that is not always the case then relying on the task only would not work. That would be nice in producing a reasonable default. And changing the default orientation of the axes is a good idea as well.

I've had some trouble with it but a partial derivative estimator will make it into this function soon as well. Then I'd like to implement the functional anova decomposition as well.

schiffner:
About the plot docs:

Re 3) I'm not sure, since there is not that much overlap in the function arguments. It's just my personal preference, but I usually hate it to scroll through a long list of mostly irrelevant arguments to find the ones that apply to the function I want to use.

Re 5) Would it make sense to cut decimal places instead/additionally?

6) I noticed that sometimes (e.g. `generateThreshVsPerfData`) the returned object is not mentioned. As it's an S3 object with a special class I think it should be.
In case of learning curves `LearningCurveData` has its own page with only little information. I guess that's not really needed.

[...]

PR #458, opened 2015-08-27
task or data for partial predictions

zmjones:
another fix related to #456

 - generatePartialPredictionData now takes a task or a data.frame.
   if a task is passed and no features are specified, the 5 most
   important features are used.

### DIFF
--- a/R/generatePartialPrediction.R
+++ b/R/generatePartialPrediction.R
@@ -11,10 +11,24 @@
 #'
 #' @param obj [\code{\link{WrappedModel}}]\cr
 #'   Result of \code{\link{train}}.
+#' @param task [\code{\link{Task}}]\cr
+#'   The task. If this is passed, data from this task is predicted.
 #' @param data [\code{data.frame}]\cr
-#'  With columns as are present in the training data.
+#'  Data to be used to compute partial predictions.
+#'  Pass this alternatively instead of \code{task}.
 #' @param features [\code{character}]\cr
 #'   A vector of feature names contained in the training data.
+#'   If not specified the features will be extracted from the \code{task}.
+#'   If \code{n} is unspecified, half of the available features will be used, with the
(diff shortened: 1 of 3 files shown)


larskotthoff:
I would check if `features` is passed when a task is given and if so, use those features. Could also require this (as for data) to remove the somewhat arbitrary 5.

zmjones:
I am already doing the first part. I am not what you mean by the second part.

larskotthoff:
Require `features` to be passed in and error if not. That way you'd never need to make the decision which features to choose.

zmjones:
So before this commit it was required. @berndbischl suggested that it might be nice to have a default.

larskotthoff:
That makes sense, but I don't think it's 5 :) I would make this number relative to the total number of features, setting the default to maybe 50%, and allow the user to specify it.

zmjones:
Hmmm. The reason I chose 5 was to make visualization easier. It seems like 50% would be a lot if the problem is even somewhat sparse though. I agree that just an arbitrary number is not the best though. 

I was thinking that I could expand this a bit and give the user control over `generateFilterValuesData`. Then the user could say `nselect = x`, and have another argument for filter method selection, with defaults that are small and fast respectively.

larskotthoff:
I think choosing the `n` most important features is the most reasonable thing to do here. If the user wants to select them by other means, they could pass `features`. I'd also be happy if `n` was a user-specifiable parameter (defaulting to 5) and not a percentage.

zmjones:
Yea ok I will do that and push it shortly.

larskotthoff:
Cool thanks.

zmjones:
Should be good to go now.

larskotthoff:
Sorry to nitpick, but the documentation isn't correct anymore now -- 50% of the features are picked instead of 5 if I read it correctly?

zmjones:
Quite right. Should be fixed now.

larskotthoff:
Thanks, merging.

PR #463, opened 2015-09-06
new learners randomUniformForest

florianfendt:


larskotthoff:
It looks like you've added a test that takes a very long time. We can't have those with Travis unfortunately as it will abort the build if no output has been produced for 10 minutes. Could you reduce the time required for the test please?

berndbischl:
if problems occur on travis, note that i removed randomIUniformForest from mlr before, so we might wanna search the tracker

florianfendt:
yea, the test is only taking 3.5 seconds on my machine now and it wasn't that much more before either.
Couldn't find anything related in the Issue tracker, though.

berndbischl:
these tests are technical tests, just reduce ntree to 5.
3.5 secs is already "a lot", we have to run many models

PR #466, opened 2015-09-09
new learner rotationForest

florianfendt:


jakob-r:
There were some tiny issues which i cleant here: b0449e8f4d67f0c31b63549d328d92b56ff8263d
I checked and the learner seems to support factors out of the box.
If you really have to use dummy variables for factors the learner does not support factors in the technical terms and you should not set the flag.

florianfendt:
hey @jakob-r,
thanks for fixing this. You could have just dropped me a message and I would have done it myself,
was my mistake after all.
But anyway, thanks again!

PR #467, opened 2015-09-10
New Learner obliqueRF

florianfendt:


larskotthoff:
Looks like the test for the new learner is failing. Could you have a look please?

florianfendt:
````obliqueRF::obliqueRF```` can't find the predict method.
Looks to me like I have to add obliqueRF to imports in DESCRIPTION, or is there another workaround?
Btw: Why don't I get this error with ````R CMD check````?

larskotthoff:
Did you run roxygen and then commit that? Sounds like you forgot one of the two :)

florianfendt:
No, I ran roxygen, all my S3Methods are in the NAMESPACE.
````obliqueRF::obliqueRF```` is training and predicting and therefore needs
````predict.obliqueRF```` in my trainLearner method.
Only reasonable suggestion I found on the web is putting the package to imports.

larskotthoff:
Ok yes, you'll have to add the package in DESCRIPTION where the other learner packages are. That should have failed when you ran it locally though.

berndbischl:
it is possible, that obqliqeRF has a problem in their NAMESPACE. I have seen this before.
Try addind a "!" before the package in the "package" slot. Then we load the package into the global namespace. But only do that if necessary

berndbischl:
But it MUST be in the suggests in DESCRIPTION. Do we not have this documented? We should

florianfendt:
putting the "!" in there didn't help. Has anybody got some other ideas or should we put this in the todo-files for now?

berndbischl:
put it into todo-files. not important now, then close here.

florianfendt:
put it in the todo files, could you merge pls, don't want to fall into the habit of pushing to master branch directly ;)
cheers

PR #473, opened 2015-09-13
add functions to merge BenchmarkResult - need this for OpenML soon

giuseppec:
This two functions are the basis to merge two BenchmarkResult. Since a BenchmarkResult needs all learner-task combinations, there are two simple usecases:

1) In my experiments I forgot to train a learner on all my task -> use `mergeBenchmarkResultLearner`
2) In my experiments I forgot to train all my learners on a task -> use `mergeBenchmarkResultTask`

If you are really forgetful and forgot both, a learner and a task, then, well, then you have to write your own function that might be based on my two functions...
I would need this two functions for `OpenML` as soon as possible.


larskotthoff:
Thanks, merging.

PR #474, opened 2015-09-13
new learners randomForestSRCSyn

florianfendt:


pfistfl:
One thing i saw while using it:

Measures `timetrain`and `timepredict` on this learner are highly skewed,
 as seperate training and prediction is not possible.
Not sure how to deal with that.

berndbischl:
@pfistfl 
Then we have to ignore that problem

berndbischl:
due to the 2 commits i have merged this on the CLI and also updated news

PR #477, opened 2015-09-14
closes #464, predict.type = "se" added to partial predictions

### LINKED ISSUE #464: plotPartialPrediction should support predict.type = "se"
...and then plot uncertainty regions. We should have kind of similar code (maybe not the best) in mlrMBO.

Something like that should work
```r
library(mlr)
x = rnorm(20, sd = 2)
y = 2 + 0.5*x + rnorm(20)
t = makeRegrTask("test", data.frame(x,y), target = "y")
l = makeLearner("regr.lm", predict.type = "se")
m = train(l, t)
pd = generatePartialPredictionData(m, getTaskData(t), "x")
plotPartialPrediction(pd)
```

zmjones:
- new extractor method getPredictionSE
 - new argument 'bounds'
 - new unit test


larskotthoff:
Thanks, looks good. Just a tiny thing about NEWS: As all of this visualisation stuff hasn't been released yet, it's not really necessary to say that it "now supports" something. It should be a "diff" to the previous release, so the best way to go may be to edit the NEWS item that mentions the function in the first place.

zmjones:
Ok I just wasn't keeping track of the releases. As you say none of this is in 2.4 so I just removed everything but the "new function" line.

larskotthoff:
Thanks, merging.

PR #480, opened 2015-09-16
ksvm: add cache parameter

jakob-r:


berndbischl:
Please add to NEWS and squash. 
(Actually the cache can affect the runtime measures, but lets keep tuneable = FALSE)

jakob-r:
done

berndbischl:
Well spotted for "regr." Merging.

berndbischl:
1) The names are much too long now. Note that they will appear in objects and prints. 
test.rmseaggr? 
That was my best output after 10 secs of thought? 

2) You also need to add the NAMESPACE, for travis. See break.

PR #481, opened 2015-09-16
svm: scale param also for regression

jakob-r:


jakob-r:
I merged the PR qith the up to date master - in fear of something strange i prefer not to squash

berndbischl:
NEWS: Please refer to the learners with their real name, so classif.ksvm, otherwise it is a bit confusing.

Just for the fun of it, try to rebase your real changes on top. You are not in tne master right?
So this should be no problem?
Then I also dont have to pull in the merge.

berndbischl:
we experimented here. and fucked it up. will delete.

Issue #468, opened 2015-09-11
why rmse.test.sqrt.of.mean?


iaolier:
Hi, I just have noticed MLR aggregates `rmse` on `rmse.test.sqrt.of.mean` (sqrt of the mean RMSE) when doing resampling. Is there a reason for that? Shouldn't it be just the mean RMSE (`test.mean`) which is the one we usually use?  

Below, I would expect `rmse.test.mean=0.310`, instead.

Example:
```
> resample(learner = lrn, task = tsk, resampling = rin, measures = rmse)
[Resample] cross-validation iter: 1
[Resample] cross-validation iter: 2
[Resample] cross-validation iter: 3
[Resample] cross-validation iter: 4
[Resample] cross-validation iter: 5
[Resample] cross-validation iter: 6
[Resample] cross-validation iter: 7
[Resample] cross-validation iter: 8
[Resample] cross-validation iter: 9
[Resample] cross-validation iter: 10
[Resample] Result: rmse.test.sqrt.of.mean=0.557
```

berndbischl:
This has been a very unfortunate bug, introduced by refactoring. It is fixed in 2.5 on Github now. 
Thanks a lot for the good report.


Please check on your side the adapted rmse now. 

@iaolier:
Shall we talk about the coop anyhow, for what you do in meta qsar? Maybe with joaquin?

berndbischl:
I have also opened this #487

PR #483, opened 2015-09-16
fix RMSE aggregation

jakob-r:
Fix Issue #468


berndbischl:
1) The names are much too long now. Note that they will appear in objects and prints. 
test.rmseaggr? 
That was my best output after 10 secs of thought?

2) You also need to add the NAMESPACE, for travis. See break.

jakob-r:
like this okay?

berndbischl:
Thx looks good

PR #486, opened 2015-09-16
work on randomForest se estimators

jakob-r:
Maybe we can think of a more general way to handle ensemble learners in general than a special implementation for the randomForest.
However this code seems very unreviewed which is dangerous as it is a crucial part of mlrMBO.
I would like @mllg to join me on reviewing this code based on [Standard errors for bagged and random forest estimators, Sexton and Laake, 2009](http://dl.acm.org/citation.cfm?id=1465325).

### DIFF
--- a/R/RLearner_regr_randomForest.R
+++ b/R/RLearner_regr_randomForest.R
@@ -54,7 +54,7 @@ trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = N
 
       # fit models on the bootstrap samples
       models = apply(samplesIdx, 2, function(bootstrapIdx) {
-        randomForest::randomForest(f, data = train[bootstrapIdx,, drop = FALSE],...)
+        randomForest::randomForest(f, data = train[bootstrapIdx,, drop = FALSE], ntree = ntree, ...)
       })
 
       # save models in attrribute
--- a/R/RLearner_regr_randomForest.R
+++ b/R/RLearner_regr_randomForest.R
@@ -26,7 +26,8 @@ makeRLearner.regr.randomForest = function() {
     ),
     properties = c("numerics", "factors", "ordered", "se"),
     name = "Random Forest",
-    short.name = "rf"
+    short.name = "rf",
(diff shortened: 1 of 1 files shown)


zmjones:
I am unaware of variance estimators for ensemble methods generally but maybe that is possible.

The bias-corrected infintesimal jackknife has a [canonical implementation](https://github.com/swager/randomForestCI) by the lead author of [this paper](http://arxiv.org/pdf/1311.4555.pdf) (in JMLR). It needs the bootstrap matrix of observations, trees where the entries are the number of times the ith observation was used in the jth tree (randomForest w/ keep.inbag = TRUE only returns an indicator matrix). Anyways when that gets patched this would be both easier to use and better (in terms of statistical/computational properties).

I am a bit short on time now but I'll look at the estimators in there now when I get a chance.

jakob-r:
That idea sounds very promising! For every mlr-learner we could implement a method which extracts that *bootstrap matrix* from the base learners and of course the BaggingWrapper so we could implement the JAB and the IJK in a general way.

berndbischl:
Hi. Can we please move this discussion to another place?
I am quite interested in the topic and also agree with Jakob that we need to figure this out properly.

Now for the PR:

- What is done here? summarize the changes. "work on" is not enough. 
- Travis apparently breaks. Why?

jakob-r:
1. we missed to actually set the ntrees to a lower value for the noisy bootstrap (fixed that with do.call)
 the variables now follow more strictly the mentioned paper
2. it will still fail because the bias can be higher then the estimated variance. hence variance-bias turns to <0 which we cannot take the root of
   * either we got the formulas wrong or the approximation does not work for such low values of R and M

berndbischl:
ok, I'll leave this open for some time. 
What I see now:
a) some variable names (uppercase letters) dont follow the style guide. pls change.
b) Thx for the note. In general we should cite work in the R docs, with roxygen, but I guess for the learners we cannot do that. (more a general remark from my side)

berndbischl:
What I dont get: I would like to pull in the code changes soon, they all seem to be reasonable.
Why does the test now fail?

jakob-r:
I think #595 deals better with it

PR #495, opened 2015-09-23
refactor Learner constructer(s)

berndbischl:
refactor Learner constructer(s): 
a) makeLearner.R has its own file and was simply moved

b) new makeLearnerBaseConstructor
this should always be called as endpoint during learner construction

c) removed 'short.name' and 'name' slots from some of the ensembles. 
this was never really used, makes stuff more complex, and I dont see the need. we need it for RLearners. but here we have the id. we can introduce it, but then it must be consistent in all wrappers.


berndbischl:
to clarifiy: this should be a PURELY internal change, without exported API consequences

berndbischl:
playing around with git here, i am stupid

Issue #130, opened 2014-07-17
DBScan clustering algorithm doesn't work on all machines


larskotthoff:
Getting class not found exceptions even though Weka package is installed apparently.

mllg:
The error message returned. Is the test re-activated?

larskotthoff:
No, must be coming from something else.

PR #503, opened 2015-09-25
add cluster.DBScan learner

larskotthoff:
Issue #130 is hopefully fixed with this.


larskotthoff:
I think the error is related to the class names parameter arguments. I cannot reproduce this on my machine and will check this later.

berndbischl:
can reproduce on my uni laptop but dont know how to help

berndbischl:
travis fails, pls repair.
in such cases, run a full local rtest with all groups locally. 
always at least "base" (but here it might be in the featsel test group)

larskotthoff:
Ok, apparently the latest 3 versions of the DBScan package (1.0.5, 1.0.4, 1.0.3) don't work. 1.0.1 works fine.

berndbischl:
Ok, so you mainly exhanged the DBscan package here? (downgrade?).

Great that you also added a note. 

Can you pls squash, so we can then merge for 2.6?

larskotthoff:
Well anybody who installs the DBScan package the normal way and doesn't read the note (i.e. everybody) will get this unhelpful error message. I'm wondering whether it's a good idea to include something that by default will be broken for most people. We could have a more prominent warning about the package version.

The `fpc` package implements DBScan as well, so that's probably the better thing to include?

berndbischl:
Ok, I didnt think about this. 

> The fpc package implements DBScan as well, so that's probably the better thing to include?

Yes, agreed.

berndbischl:
One way to deal with this here, might be to internally try to catch that error msg and then inform people about the downgrade / version problem. But that might be a lot of effort for something we are not really responsible... OTOH it might finally close this here.

larskotthoff:
Yep, will do that. I'll have a look at integrating the other DBScan later on.

berndbischl:
Sure you want to close here?

larskotthoff:
Yep. Not worth it, especially given that we have an alternative implementation we can use.

PR #518, opened 2015-10-07
getLearnerProperties.Learner should be exportet

jakob-r:
Mostly we only use it internally but not always and we also want to use it from mlrMBO


berndbischl:
i restarted travis

larskotthoff:
Do we want to include this in 2.5 (i.e. merge now) or wait?

berndbischl:
check looks clean and this seems trivial, merging now

PR #521, opened 2015-10-07
fixes parallelization bug in generatePartialPredictionData

zmjones:
- adds test


berndbischl:
Thx.
parallelMap is in mlr IMPORTS. do not state parallelMap:: in these cases, pls. Just makes code longer without necessity.

berndbischl:
This is always a matter of confusion.
For only SUGGESTED packages, on must use ::.

Should we note this in the contributor guidelines?

zmjones:
Yes I will do that.

larskotthoff:
Thanks, merging.

PR #522, opened 2015-10-08
speeds up partial prediction tests

zmjones:
subset the input data frame when estimating derivatives (the very slow part). this can (will be when i have some more time) be sped up a bit more with some more work.


larskotthoff:
Thanks, merging.

Issue #511, opened 2015-09-29
Integrate DBScan learner


larskotthoff:
The [fpc package](https://cran.r-project.org/web/packages/fpc/fpc.pdf) provides one.

berndbischl:
we already have a PR and a list in the wiki. Only open this when unexpected problems occur.

PR #526, opened 2015-10-10
add dbscan clusterer

larskotthoff:


larskotthoff:
This fixes #511.

jakob-r:
looks good. maybe add another param setting to the test or is it too slow?

berndbischl:
Lars, here you potentially return NA logicals and not integers. Can you please fix that?

Also do we really allow this in the API? 
What happens with all subsequent stuff if some predictions are NA?
Is this anywhere defined?

larskotthoff:
Will have a look at this later.

larskotthoff:
Don't see where this is potentially returning something other than integers -- the return value from the clusterer is wrapped in `as.integer()`. It's not defined what happens when the predictions are NA.

berndbischl:
To be really type safe you should really write NA_integer_.

That returning the NAs is kinda undefined is the problem, isnt it? Maybe we should discuss this orally.

larskotthoff:
I don't know what you mean when you say that I should write NA_integer_. I'm not generating these predictions.

What happens when any of the other learners return NA?

larskotthoff:
Argh, I was looking at the wrong old dbscan cluster branch. Fixed the logical NA thing now.

berndbischl:
status here?

larskotthoff:
Ready to be merged. NEWS may have to be fixed manually after the merge.

berndbischl:
I had to fix NEWS manually, done.

Still, we should define better what happens with 'NAs in the predictions

PR #528, opened 2015-10-10
fix required argument handling in cluster.cmeans

larskotthoff:


berndbischl:
travial change, merging now

PR #529, opened 2015-10-11
add generic clustering as classification wrapper

larskotthoff:
Not sure about the name, it's a bit clunky. Suggestions appreciated.

### DIFF
--- a/R/ClassificationViaClusteringWrapper.R
+++ b/R/ClassificationViaClusteringWrapper.R
@@ -1,0 +1,86 @@
+#' @title Classification via clustering wrapper.
+#'
+#' @description
+#' Clusters the training data and determines the distribution of class labels for each cluster. New data is assigned to a cluster and the class distribution (or label) for this cluster is returned as the prediction.
+#'
+#' Fuzzy cluster memberships are currently not used.
+#'
+#' @template arg_learner
+#' @param predict.type [\code{character(1)}]\cr
+#'   \dQuote{response} (= labels) or \dQuote{prob} (= probabilities and labels by selecting the ones with maximal probability).
+#' @param ... [any]\cr
+#'   Additional parameters passed down to the filter.
+#' @template ret_learner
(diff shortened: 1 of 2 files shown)


jakob-r:
I think this is more some kind of preprocessing and it should be completely possible to do it with `makePreprocWrapper`. But I have to admit that I only hat a quick glimpse at your code.

larskotthoff:
Hmm, not sure that this could be done entirely in preprocessing -- during the prediction phase I also need to determine the class distribution for the assigned cluster and then determine the label based on that for response.

This can also be extended to support fuzzy clustering and then the determination of the label during prediction becomes even more complex.

berndbischl:
@jakob-r 
You misunderstand. Lars simply wants to allow the usage of a clustering algorithm as a classifier.
(He also documented this kinda nicely?)
I am not a big fan of this approach, but some people do this and having this in mlr would allow easy comparisons.

berndbischl:
NB: I did NOT review this properly yet

larskotthoff:
This could probably also be extended to allow stacking on top of it, e.g. build a separate classification model for each cluster.

zmjones:
I have nothing to add but this is cool.

larskotthoff:
I've renamed this to `ClassificationViaClustering` to be less confusing.

berndbischl:
Lars, can you pls convince me again we should put that into the package?

larskotthoff:
Talked about this, I'll try to find some concrete examples where this helps.

Also Weka has it, see http://weka.sourceforge.net/doc.stable/weka/classifiers/meta/ClassificationViaClustering.html and also http://weka.sourceforge.net/doc.dev/weka/classifiers/meta/ClassificationViaRegression.html

berndbischl:
the docs say you dont support probs, but the code does.
also apparently you do at least some kind of normalization.

please precisely say what happens here in docs.

Issue #433, opened 2015-08-09
Implement way to estimate overfitting of models


larskotthoff:
There's an interesting way to estimate the relative overfitting of a model described in Improvements on Cross-Validation: The .632+ Bootstrap Method by Bradley Efron and Robert Tibshirani (page 552, equation 28). This would be useful to have in mlr.

I'm opening this issue primarily to discuss whether the way I've implemented it makes sense. The way it is described in the paper requires O(n^2) performance function evaluations where n is the size of the dataset -- this is quite slow in practice.

larskotthoff:
Summary from call with Bernd about this:

- change implementation to allow ResampleDescription, ResamplePrediction, separate train/test predictions
- change implementation of b632+ aggregation to call this method
- clean up checks
- rename to something imperative

larskotthoff:
#236 is related in some way (or so I'm told by Bernd).

berndbischl:
it is a very different procedure, but tries to roughly reach the same goal. we could, but dont have to, discuss it jointly.

pat-s:
Merged in #530.

PR #530, opened 2015-10-12
implement estimation of relative overfitting for more preds-ish (#433)

larskotthoff:
It would be good if somebody could check this, especially wrt extracting train/test data -- not sure at all about this part.


larskotthoff:
No idea why Travis fails -- seems like it's something completely unrelated and I can't reproduce it locally.

jakob-r:
I think you spotted why it failed - because caret changed their defaults.

larskotthoff:
@berndbischl and I will talk about this.

ja-thomas:
Please describe what exactly is the purpose of this. And it needs a better title (Any issues connected to this?

berndbischl:
@larskotthoff 
please answer the question here. this is soon getting extremely old. we already have conflicts. 
if we dont hurry up, this will simply be lost...

can you please 
a) explain what goes on here
b) resolve the conflicts

larskotthoff:
This is a cleanup of the relative overfitting code to make it work better with the objects mlr provides. No issue related to this.

mllg:
1. No need to document `NULL` here. This way you don't see the function definition in the docs.
2. The object is document as `obj`, but `predish` is used in the code.
3. You don't need `assertClass` on the first object in S3 methods.
4. See line comments.

R/relativeOverfitting.R
```diff
@@ -19,37 +23,64 @@
...
-    df = (perf.test - perf.train) / (perf.permuted - perf.train)
-    names(df) = stri_paste("relative.overfit", mids, sep = ".")
-    cbind(data.frame(iter = i), df)
+  iterations = unique(predish$data$iter)
+  rbindlist(lapply(iterations, function(i) {
+    data = predish$data[predish$data$iter == i & predish$data$set == "test",]
+    pred.test = makePrediction(task$task.desc, row.names(data), data$id, data$truth, predish$predict.type, predish$predict.threshold, data$response, predish$time[i])
+
+    data = predish$data[predish$data$iter != i & predish$data$set == "test",]
+    pred.train = makePrediction(task$task.desc, row.names(data), data$id, data$truth, predish$predict.type, predish$predict.threshold, data$response, predish$time[i])
```
mllg: Sure about `predish$time[i]`?

R/relativeOverfitting.R
```diff
@@ -19,37 +23,64 @@
...
+  measures = checkMeasures(measures, task)
+  mids = vcapply(measures, function(m) m$id)
+
+  perf.test = performance(predish, measures = measures)
+  perf.train = performance(pred.train, measures = measures)
+
+  nrows = nrow(predish$data) + nrow(pred.train$data)
+  pred.permuted = predish
+  pred.permuted$data = data.frame(truth = rep(c(predish$data$truth, pred.train$data$truth), each = nrows),
+    response = rep(c(predish$data$response, pred.train$data$response), times = nrows))
```
mllg: This seems odd. Are you sure you don't want to call `sample()` somewhere?

mllg: (Without having a deeper look into the objects): I suspect `predish$data` and `pred.train$data` have the same number of rows -- why do you need to call `rep` once with `each` and once with `times`? Maybe a short comment would clarify this line.

R/relativeOverfitting.R
larskotthoff: Yes. Is that not the way you're supposed to access it?

R/relativeOverfitting.R
larskotthoff: It's doing all permutations, so no need to sample. I'll add a comment.

larskotthoff:
Do we want this or not?

Issue #408, opened 2015-07-21
makePreprocWrapperCaret


schiffner:
I encountered some problems with `makePreprocWrapperCaret`.
These are not difficult to solve. I'm opening this issue because I wanted your opinion and because I didn't
want to create any conflicts if @jakob-r is working on the wrapper (#220).
- `preProcess` works only for numeric features.
  At the moment this is not handled consistently:
  
  When I generate a wrapped learner it still has property `"factors"` (if it had this originally).
  
  When trying to `train` the wrapped learner on `bh.task` (which has numeric and factor columns)
  `preProcess`, of course, throws an error.
  
  My favorite solution would be to use only numeric columns as `work.cols` in `makePreprocWrapperCaret`.
  Any objections?
  
  Concrete example:

```
lrn = makePreprocWrapperCaret("regr.lm", ppc.pca = TRUE, ppc.thresh = 0.9)
lrn
# Learner regr.lm.preproc from package stats
# Type: regr
# Name: ; Short name: 
# Class: PreprocWrapper
# Properties: numerics,factors,missings,weights,se
# Predict-Type: response
# Hyperparameters: ppc.BoxCox=FALSE,ppc.YeoJohnson=FALSE,ppc.expoTrans=FALSE,ppc.center=FALSE,ppc.scale=FALSE,ppc.range=FALSE,ppc.knnImpute=FALSE,ppc.bagImpute=FALSE,ppc.medianImpute=FALSE,ppc.pca=TRUE,ppc.ica=FALSE,ppc.spatialSign=FALSE,ppc.thresh=0.9,ppc.na.remove=TRUE,ppc.k=5,ppc.fudge=0.2,ppc.numUnique=3

train(lrn, bh.task)
traceback()
#12: stop("all columns of x must be numeric")
#11: preProcess.default(x = list(crim = c(0.00632, 0.02731, 0.02729, 
#     0.03237, 0.06905, 0.02985, 0.08829, 0.14455, 0.21124, 0.17004, 
#10: (function (x, ...) 
#     UseMethod("preProcess"))(x = list(crim = c(0.00632, 0.02731, 
#     0.02729, 0.03237, 0.06905, 0.02985, 0.08829, 0.14455, 0.21124, 
#9: do.call(caret::preProcess, c(list(x = x), cargs))
#8: .learner$train(data = getTaskData(.task, .subset), target = getTaskTargetNames(.task), 
#        args = pvs)
[...]

PR #531, opened 2015-10-12
fix compatibility for caret_6.0-57

larskotthoff:


berndbischl:
if you do stuff like this you need to increase the required version in DESCRIPTION

berndbischl:
i will do this after the merge

schiffner:
Are you really sure that caret changed the defaults of `preProcess` in version 6.0-57? As far as I can see centering and scaling have been the default for at least a year. So we don't need to change the required version because of the defaults.

Anyhow, increasing the required version might be a good idea because the improved handling of non-numeric variables in `preProcess`. (See https://github.com/mlr-org/mlr/issues/408)

berndbischl:
Ok, I am not sure and did not double check this. Maybe Lars should comment on this and I will wait

larskotthoff:
So caret doesn't have any kind of useful changelog or NEWS or even commit messages. With the current implementation, no methods get selected by default and that fails with an error message. I don't know what they actually changed or why, but adding the defaults listed in the man page fixes the problem.

berndbischl:
after talking to lars I will merge this now

Issue #447, opened 2015-08-17
add a few standard ResampleDesc objects, to save some typing


berndbischl:
maybe cv3, cv5, cv10

jakob-r:
Is that really necessary as we have `crossval(learner, task, iters = 10L)` ?

berndbischl:
yes, because we also want to pass them to tuneParams and other stuff.

jakob-r:
That on the other hand would mean that we can drop `crossval()`, no? Or am I to keen on removing things we have to maintain?

berndbischl:
no, i dont want to drop that. it is just 2 very similar, but still slightly different things.
also: people already use crossval

berndbischl:
done. thx

PR #533, opened 2015-10-12
#447 resampleDescs

florianfendt:
@berndbischl 
Put the objects in the zzz.R file like you suggested.
Should I change the docType (roxygen set docType to data)?
What would be a reasonable type here though? 
Class of the Objects or simply (helper)-object?


larskotthoff:
Thanks. Could you mention this in NEWS as well please? And you'll have to rebase on master once #531 has been merged unfortunately because caret broke mlr.

berndbischl:
1)
please fix the spelling, and watch this yourself.
Object --> object
Cross-Validation --> Crossvalidation.

2) We dont use "seealso" too much, use family, probably with "family resampling"

3) In this case I would really doc this on one page, in this case the ResampleDesc page.
Do you now how to do this?

4) cv2 is also useful, and "holdout"
What about the nameclash then? is there a way around this?

florianfendt:
So, I put a new section to the ResampleDesc page. Or should this rather go to details? I wasn't so sure.
And sorry for the spelling mistakes. I left the dash for cross-validation though, since it is used this way throughout the help page.

florianfendt:
ok @berndbischl, added everything we just discussed

berndbischl:
Looks good will merge now

PR #537, opened 2015-10-16
h2o learners

florianfendt:
@giuseppec
So, I added the missing tests, and did the documentation stuff. I also changed the learner params in test_classif_h2ogbm since I found the tests were taking quite some time on my machine.


jakob-r:
One problem I see here is that we need `h2o.init()`. This is not part of the standard mlr workflow and thus tests as `test_learners_classiflabelswitch.R` fail because they assume that a constructed learner also works. Can you put `if (#h2o.not.initialized#) h2o.init()` into the train function? Do you also have to uninitialize h2o?

berndbischl:
I talked to Matt (from h2o) about this during useR.

It should really be possible to check in the train fun whether h2o has been initialized, and otherwise call h2o.init. Like Jakob suggests above. A teardown call should not be necessary, from what I know.
Look into the source of h20 init or the package in general. There is some kind of an "init" object that one can check for existence or whatever. It took us <= 5 min during useR to see this and figure this out.

florianfendt:
@berndbischl @giuseppec 
So I had a quick look at this again today, the code should do what we want and I get passing checks locally. 
I can't reproduce the error travis is throwing either, e.g: This is working:
````
lrn = makeLearner("classif.h2oglm", predict.type = "prob")
pred = predict(mod, binaryclass.task)
pred
````
Do you see anything that might be wrong?

**Comment by giuseppec** [comment]
@florianfendt: Do you maybe have an old h2o package version locally? If yes, then update the package first. I reported a bug for the h2oglm to h2o and I think they fixed it. In mlr I did a workaround to fix the bug in mlr, but now it might be fixed in the h2o package and you could replace the current predict function for the h2oglm with:

```r
predictLearner.classif.h2oglm = function(.learner, .model, .newdata, ...) {
  m = .model$learner.model
  h2of = h2o::as.h2o(.newdata)
  p = h2o::h2o.predict(m, newdata = h2of, ...)
  p.df = as.data.frame(p)
  
  pCol = grepl("p[[:digit:]]", colnames(p.df))
  if(any(pCol)) colnames(p.df)[pCol] = gsub("p", "", colnames(p.df)[pCol])
  
  if (.learner$predict.type == "response") {
    return(p.df$predict)
  } else {
    p.df$predict = NULL
    ret = as.matrix(p.df)
    return(ret)
  }
}
```

can you do this and look if travis is still complaining?

florianfendt:
Thx @Giuseppe that might actually be possible, haven't updated the package for a couple of weeks.
I'll fix that and report back here

florianfendt:
?! Why is travis complaining about the flare package now??
If I read the log correctly, the package is installed, no?

Apart from that it should work now.
Thx again!

berndbischl:
because our tests and the setup for it is so complex you need a phd to understand them. 
fortunately i am already prof so this is a piece of cake for me, just give me 5 min :)

berndbischl:
so one side info: if we get the dreaded error "too many DLLs loaded", the next learners that cannot load their packages due to this error, will display this misleading error messages.

so that COULD be the reason we see it here. although i hope it is not. i searched the travis logs for "DLL" but could not find it.

florianfendt:
Ok thx.
And I appreciate the info, so I might be able to figure such stuff out myself in the future!

berndbischl:
i have i have to look at this locally, the learners_regr test seems to load too many packages now.
maybe we should rewrite everything in python.

giuseppec:
Remarks:
- We need a learner "regr.h2odeeplearning", which is currently missing
- Because `hidden` can be a vector, also "hidden_dropout_ratios" should be a vector, e.g. something like `makeNumericVectorLearnerParam("hidden_dropout_ratios", default = c(0.5, 0.5), len = NA_real_)`

@florianfendt could you also check the default values and ranges etc. for each parameter? I have the feeling that h2o is changing them from time to time (you can look this up in the h2o package description or http://h2o.ai/docs/master/model/deep-learning/).

berndbischl:
dealt with in #875

PR #538, opened 2015-10-17
update tests for soon-to-be-released ggplot2 version

larskotthoff:
This will fail at the moment, should be merged after the new ggplot2 version has been released.


zmjones:
What is the "ns" bit?

larskotthoff:
I was getting warnings about no namespace being specified for the query. This is what that does (i.e. map "svg" to the URL of the namespace if I understand correctly).

berndbischl:
does anybody know the release date of ggplot2?

larskotthoff:
Hadley and God, although I'm not sure about God.

larskotthoff:
@zmjones Apparently ggplot doesn't support `vjust` in `annotation` anymore. I've removed this from `generateCritDifferences` to make the build pass again, but could you please check whether everything is still being shown correctly (and what to use instead of `vjust`)?

berndbischl:
Yo, I saw this on travis too. I guess we should check it out at least briefly and document the results here.

berndbischl:
As this is merged I will put this info in the ggplot2 issue in the tracker. We also really need to be able to run on rdevel.

PR #539, opened 2015-10-19
refactor and add tile plots to plotPartialPrediction

zmjones:
also adds some new tests, removes some unnecessary ones, as well as comments for the old tests.


larskotthoff:
Thanks, will merge for 2.6.

zmjones:
can we go ahead and merge this now? it doesn't really affect anything else so the chance of breakage is very low and i'd like to use it in a blog post.

berndbischl:
let me bump the version pls

PR #542, opened 2015-10-22
default nodesize for regr.randomForest is 5

jakob-r:
Quote from the RF Help
> Minimum size of terminal nodes. Setting this number larger causes smaller trees to be grown (and thus take less time). Note that the default values are different for classification (1) and regression (5).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -48,6 +48,7 @@ mlr_2.5:
   For the new ones, see below.
 - Added some commonly used ResampleDesc description objects, to save typing in resample experiments:
   hout, cv2, cv3, cv5, cv10.
+- regr.randomForest: changed default nodesize to 5 (according to randomForest defaults)
 
 - new functions
 -- getDefaultMeasure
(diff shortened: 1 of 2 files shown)


berndbischl:
I dont see NEWS update

jakob-r:
What is the relevance for the end user? The default values are not used so far anyway. Or are they?

berndbischl:
who says that the user does not query and use them? now you change them behind their back

berndbischl:
(i tried to make the same argument as you before for something else, then @larskotthoff  corrected me...)

jakob-r:
changed

berndbischl:
merging now thx

larskotthoff:
Can we please finish 2.5 *before* any other PRs are merged?

berndbischl:
I am only merging stuff which takes less then 30 secs tp review.

I currently cannot work magic, everybody is free to help with 2.5....

(I have scheduled time again for tomorrow)

PR #548, opened 2015-10-25
fix performance for ResamplePrediction (somewhat) (#431)

larskotthoff:


larskotthoff:
Merging, as discussed with Bernd.

PR #549, opened 2015-10-25
add generic classification via regression wrapper

larskotthoff:
Will update NEWS after 2.5 release.

At the moment this simply truncates predictions < 0 to 0 and > 1 to 1 for the class probabilities. Not sure if this is the best way though, maybe normalization based on the range of the values?


catastrophic-failure:
Cool. Perhaps an option for logit transformations could be added?

larskotthoff:
That's something to think about once this has been merged, yes.

berndbischl:
> Cool. Perhaps an option for logit transformations could be added?

well: 
a) i am still unsure about the usefulness of the method. a bit. and would like to see more arguments why this is needed.

b) i do agree with @catastrophic-failure that a logit-trafo / sigmoid should be used before merge. what is currently in the code, truncation at [0,1] for probs is NEVER done. and the logit trafo is the normal and reasonable way to go. it also just adds 1-2 line of code.

catastrophic-failure:
@berndbischl the most common LS-SVM classifier is formulated as a regression over Fisher targets {-1,1}. It's a valid strategy.

berndbischl:
> @berndbischl the most common LS-SVM classifier is formulated as a regression over Fisher targets {-1,1}. It's a valid strategy.

i am not sure i understand? are you answering to my question whether lars' code is useful? 
of course i know that very many (binary) classifiers are constructed via regression approaches. 
what i meant was: should we support this via mlr? as i would guess in every package for every method where this is possible, this already exists?

ok, i might be wrong with that guess. also, MAYBE it does not hurt us too much to add this here.

what about b) and the estimation of probabilities?

larskotthoff:
Ok, I've added a logit transform -- is this what you had in mind? This will return probabilities of 0.5 for predicted probabilities of 0 though, which isn't particularly nice...

catastrophic-failure:
@larskotthoff I think not. Something along the lines of:

```splus
x = seq(-1,2, length.out = 201L)
l = 1/(1+exp(-5*(x-0.5)))

plot(x, l, type = "l"); abline(a = 0, b = 1, col = "red"); abline(v = 0.5, h = 0.5)
```

Under this scenario, 0.5 still maps to 0.5, but infinity maps to 1 and -infinity maps to 0.

That's still too drastic though, perhaps material for other PRs.

larskotthoff:
Hrm, I was trying to avoid putting any magic constants in there. @berndbischl what's your opinion?

catastrophic-failure:
Well, I had in mind it could be a tunable parameter aswell, just like other wrappers have their parameters.
Also, a comparison to another sigmoidal transformation, way less drastic this time and with a nice rounding pi multiplication instead of an arbitrary parameter.

```splus
x = seq(-1,2, length.out = 201L)
l = 1/(1+exp(-pi*(x-0.5)))
s = 0.5+0.5*tanh(pi*(x-0.5))

par(mfrow = c(1,2))
plot(x, l, type = "l", main = "logistic"); abline(a = 0, b = 1, col = "red"); abline(v = 0.5, h = 0.5)
plot(x, s, type = "l", main = "hyperbolic tangent"); abline(a = 0, b = 1, col = "red"); abline(v = 0.5, h = 0.5)
```

berndbischl:
> Hrm, I was trying to avoid putting any magic constants in there. @berndbischl what's your opinion?

use this: 1 / (1 + exp(-x)).

this is the logistic transformation from logistic regression. 

yes, one could parametrize this. then I guess we would need to do Platt scaling IMHO

https://en.wikipedia.org/wiki/Platt_scaling

Platt scaling is probably preferable here, but harder to implement.

catastrophic-failure:
1 / (1 + exp(-x)) is incorrect.

1 / (1 + exp(-x+0.5)) at least maps 0.5 to 0.5, but it changes too much other reasonable numbers.

berndbischl:
ok, like i thought this is getting more complicated now

1) remove the "oneclass" property. thats novelty detection you dont support this here, we dont support this even really in mlr

2) how do you even support multiclass? you dont. and this is also non-trvial to do this. remove the property. 

3) now whether you need to take my sigmoid-formula or the one from @catastrophic-failure depends on how you code the target labels in training. my formula is for -1,+1! his for 0,1. 
you have coded with 0,1 and then taken my formula, thats wrong.

please code the labels with -1,+1 thats more common.

4) please doc a little bit better how exactly your method works for the user.

larskotthoff:
1) Ah, I think I was confused by this before. Why is it there if mlr doesn't support it?

2) Multiclass is supported in the same way as binary classes, by training a model for each class. This is tested as well.

3) I've transformed the 0,1 interval to -1,+1.

4) Done.

berndbischl:
> 1) Ah, I think I was confused by this before. Why is it there if mlr doesn't support it?

it was added a very long time ago, for completeness sake and because we want to support it. i still do, but if you wanna remove the prop for now, also fine with me.

> 2) Multiclass is supported in the same way as binary classes, by training a model for each class. This is tested as well.

i completely missed you lapply there, sorry, was too tired. so effectively you are using the one-vs-rest principle then. and also somewhat doced this.

> 3) I've transformed the 0,1 interval to -1,+1.

well, i meant this: duing the FIT model the label as -1,+1! that learns a score for the positive class. then use my formula in predict to transform to probs. pls do that.

5) and the prob trafo does not work for multiclass. you need to ensure that the rows sum to 1, e.g.
please use the standard softmax trafo in this case and also doc this.
https://en.wikipedia.org/wiki/Multinomial_logistic_regression

larskotthoff:
Changed as discussed earlier.

larskotthoff:
ping @berndbischl

larskotthoff:
This is ready to be merged. Please let me know if we want this.

jakob-r:
Hmm. Why do this Wrapper offers multiclass if it does not come naturally? We already have the `MultiClassWrapper` which implements this functionality. Personally I would strip the multiclass thing from the wrapper here.

R/ClassificationViaRegressionWrapper.R
```diff
@@ -0,0 +1,82 @@
...
+#' lrn = makeLearner("regr.rpart")
+#' lrn = makeClassificationViaRegressionWrapper(lrn)
+#' mod = train(lrn, iris.task, subset = 1:100)
+#' predictions = predict(mod, newdata = iris[101:150, 1:4])
+makeClassificationViaRegressionWrapper = function(learner, predict.type = "response", ...) {
+  learner = checkLearner(learner, "regr")
+  ddd = list(...)
+  assertList(ddd, names = "named")
+
+  lrn = makeLearnerBaseConstructor(classes = c("ClassificationViaRegressionWrapper", "Wrapper"),
```
jakob-r: One final question. Why don't we use `makeBaseWrapper`?

jakob-r:
i reworked it a bit in branch: classificationViaRegression_jr

larskotthoff:
Do you want to merge that into this branch or merge the new one into master?

jakob-r:
I merged it in here.

berndbischl:
lets finalize this then

1)
> We already have the MultiClassWrapper which implements this functionality. Personally I would strip the multiclass thing from the wrapper here.

agreeing with @jakob-r 

2) travis breaks.

3) if this in WEKA (i think you told me?) could you link to it in docs please?

larskotthoff:
Ok, removed the multiclass functionality, added a ref to WEKA and fixed everything else. Ready to merge.

Issue #225, opened 2015-02-03
default params in param description: when we deviate from the defaults


berndbischl:
We must have a consensus (best formally written down) how one can deviate from the defaults of an implemented underlying learner.

Questions: 
* in ParamDesc: old default value or new under entry "deafault"
* should we set the value in par.vals or do it "manually" in training? the former is more transparent, but "clutters" objects more, was we see something which is a default.

kerschke:
Ok, I'll add it to the wiki.

berndbischl:
Should this not go into tutorial / extend / own learner?

berndbischl:
We also have to define it first....

mllg:
I don't get it. Do you want something like `resetParam()`?

berndbischl:
No. The problem is that we sometimes do this: 
A learner cannot be called with an empty argument list. Due to 2 reasons usually: 
a) a certain param must be set , but it has no default (eg nnet has 'size')
b) the defaults really do not work. because they are for classification but we have a regr.foo variant.

Now we need to formally write down what we do here. 
Usually I set the extra needed values in 'par.vals'. And make a 'note' in the learner. So what happens in transparent to the user. And the PH Param object has the default slot. There is usually leave the default which is used in the underlying package. 

Do you see how much text I wrote? And I am NOT sure whether this is done exactly the same for every learner. Fortunately we did not have to do this very often. But we MUST document it. 
It does not really need new functions, but a documented standard, and a tedious check whether we follow it.

Got it now?

jakob-r:
Related: #344
The problem is, if we don't have a good specification it is nearly impossible to have a reliable and working feasibility check. *Our* defaults in `par.vals` can clash with manually set parameters when creating the learner or calling `setHyperPars`.

schiffner:
The consensus so far was:

* The `ParamSet` contains possible values, bounds, defaults **exactly** as in the docs/signature of the underlying learner.
* In `trainLearner`, `predictLearner`: Don't hardcode learner params.
Exception: Functions that can deal with several types of learning problems like `glm`: For classification the `family` argument is hardcoded as "binomial".

Deviating from defaults for example because the defaults are inconvenient, unstable, temporarily buggy, ...
* Put the parameter values in `par.vals` and add a `note`. (Then the changed param values are visible for example if the `Learner` is printed or by `getHyperPars`.)
* We **never** change the defaults in the `ParamSet`.
[...]

PR #558, opened 2015-11-04
allow to override all parameters for cforest (fixes #556)

### LINKED ISSUE #556: Error when setting argument "replace" with classif.cforest
**Error:**
When I try to set the argument `replace` of learner `classif.cforest`, I get the following error when the tuning starts:

    Error in cforest_control(teststat = "quad", testtype = "Univ", mincriterion = 0,  : 
    formal argument "replace" matched by multiple actual arguments

**Minimal example:**

    library("mlr")
       
    lrn <- makeLearner("classif.cforest", predict.type = "prob",
                       par.vals = list(ntree = 1000, mtry = 5, minbucket = 10, maxdepth = 0, replace = FALSE))

    ps <- makeParamSet(
        makeDiscreteParam(id = "ntree", values = c(100, 200)),
        makeDiscreteParam(id = "mtry", values = c(5, 10)))
    ctrl <- makeTuneControlGrid()
    rdesc.inner <- makeResampleDesc("Subsample", iters = 5)
    rdesc.outer <- makeResampleDesc("CV", iters = 5)
    lrn.tuned <- makeTuneWrapper(learner = lrn, resampling = rdesc.inner, par.set = ps, control = ctrl)
   
    ## learning
    res <- resample(learner = lrn.tuned, task = pid.task, resampling = rdesc.outer)

However `getParamSet("classif.cforest")` tells me that `replace` is an argument of `classif.cforest`.

**Settings:**
R version: 3.2.2
mlr version: 2.5 (Github version)

larskotthoff:
The code to set defaults is a bit ugly -- I guess what we really need is a function that merges parameter sets, or replaces missing things in parameter sets. I didn't see anything that does it in ParamHelpers, but I could be mistaken.

### DIFF
--- a/R/RLearner_classif_cforest.R
+++ b/R/RLearner_classif_cforest.R
@@ -40,7 +40,13 @@ trainLearner.classif.cforest = function(.learner, .task, .subset,
   nresample, maxsurrogate, maxdepth, savesplitstats, ...) {
   f = getTaskFormula(.task)
   d = getTaskData(.task, .subset)
-  ctrl = learnerArgsToControl(party::cforest_unbiased, ntree, mtry, replace,
+  defaults = getDefaults(getParamSet(.learner))
+  if (missing(teststat)) teststat = defaults$teststat
+  if (missing(testtype)) testtype = defaults$testtype
+  if (missing(mincriterion)) mincriterion = defaults$mincriterion
+  if (missing(replace)) replace = defaults$replace
+  if (missing(fraction)) fraction = defaults$fraction
+  ctrl = learnerArgsToControl(party::cforest_control, ntree, mtry, replace,
     fraction, trace, pvalue, teststat, testtype, mincriterion, minprob,
     minsplit, minbucket, stump, randomsplits, nresample, maxsurrogate,
     maxdepth, savesplitstats)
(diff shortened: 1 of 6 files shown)


schiffner:
Just a question, because that would have been my first approach:
Wouldn't it be easier to put the defaults (at least if not data-dependent) in `par.vals` in `makeRLearner` (like for example in https://github.com/mlr-org/mlr/blob/master/R/RLearner_classif_mda.R)? Then you wouldn't need all the `if(missing(...))`s?

larskotthoff:
Not sure what you mean. Specifying a default value in a parameter set doesn't mean that this parameter gets automatically set to this value if it's not specified. Without the `if(missing(...))`s you'll get errors about missing arguments later on.

schiffner:
I mean `par.vals`, not the `par.set`. In most learner definitions `par.vas` is an empty `list`, but we often use it to set defaults (as e.g. in mda). These settings can be of course overwritten by the user. And all `par.vals` are passed on to training, so there should not be problems with missing arguments.

larskotthoff:
Ah you're right, I hadn't even thought of that. I guess that would require having the defaults in two places though, in `par.set` and `par.vals`. Not sure if this is a better solution here.

schiffner:
Yep. That's the downside. :)

larskotthoff:
Travis failure seems to be unrelated (probably because I didn't regenerate all the documentation?). This is ready to be merged.

berndbischl:
Another comment, for completeness sake:
One option to drastically shorten Lars' code is to simply put the default values into the signature of the trainLearner method. But that has again the downside of having the values in 2 places. 

Actually, we might take this a perfect option to deal with this here now:
#225 
@schiffner 

Right?

larskotthoff:
This is ready to be merged.

berndbischl:
thx merging now

Issue #487, opened 2015-09-17
add unit tests for measures


berndbischl:
we need to be careful here.
we need this:
- a very tiny data set, 
- then "mock" predictions for them
- the measures then need to be compared with manual calculations.

PR #561, opened 2015-11-09
#487 tests for measures

florianfendt:
@berndbischl 
So I wrote the tests as discussed:
1) with manual calculations, where I tried to write the code as "raw" as possible, like you suggested. The corresponding objects are called ``` nameofthemeasure.test  ```
2) check if ```performance()```calculates the measures correctly, too. These objects are called ```nameofthemeasure.perf```

As you can see I put both tests in one place to safe some typing, or should I rather put part 2) in a separate file, like ```test_base_performance_measures.R``` ?


berndbischl:
Hi, thanks a lot.

IMHO the tests can stay where you put them. Are we done here?
Does anybody else have comments?

larskotthoff:
I would put the helper objects into `helper_objects.R`.

berndbischl:
Which helper objects exactly? Stuff should only go into helper_objects, if it is of general use and used in multiple places. If it is used only locally, it should only be defined locally.

larskotthoff:
Well at least the smaller datasets can be used in other contexts as well, especially for more in-depth testing of the learners where we check the individual predictions.

berndbischl:
ok no problem.

florianfendt:
Moved them to helper_objects.
Thanks for the input.

florianfendt:
@berndbischl 
Are we done here, so we can merge?

berndbischl:
Only merging this now, as it is only tests and no API or code change

PR #569, opened 2015-11-20
fix: iter for CV must be at least 2 to prevent nonesense

jakob-r:
Because:
```r
> ri = makeResampleInstance(makeResampleDesc("CV", iter = 1), task = iris.task)
> ri$train.inds
[[1]]
integer(0)
```

### DIFF
--- a/R/ResampleDesc.R
+++ b/R/ResampleDesc.R
@@ -123,7 +123,7 @@ makeResampleDescHoldout = function(iters, split = 2/3) {
 }
 
 makeResampleDescCV = function(iters = 10L) {
-  iters = asCount(iters, positive = TRUE)
+  iters = asInt(iters, lower = 2L)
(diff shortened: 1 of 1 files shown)


berndbischl:
yupp, but pls not merge now, I am exactly now managing the CRAN upload now

berndbischl:
also, if that happens, i know it is tedious for arg checks, but pls add a test.
and what about repcv? must it be checked there too?

berndbischl:
actually, maybe disregard what i said about the unit test.
it is just an (suboptimal) arg check. It can stay like this without a test. sorry.
If we begin unit-testing these, we really overdo this.

jakob-r:
> and what about repcv? must it be checked there too?

There the checks are already good

berndbischl:
ok then i guess we are done here, and i will merge when 2.5 is on cran and I have inced the version nr

Issue #326, opened 2015-06-11
impute should really work on task


berndbischl:
i am constantly getting annoyed that this does not work.
it also makes no sense in the spirit of the package

berndbischl:
@mllg 
FYI

florianfendt:
Was done in #576 
Closing here

PR #576, opened 2015-11-26
#326 impute and reimpute for tasks

florianfendt:
I extended the function ````impute()```` to work for tasks, too now.
Also, I did that for ````reimpute()````. I realize not many people will use
````reimpute()```` on tasks, but I I think in some cases this might be convenient.
What do you guys think about that? Should I keep the method for ````reimpute()````?


berndbischl:
Please dont add an argument to the function but use S3.
Please closely look at eg capLargeValues or other preproc operations that already have this structure.

And, if this is in the future, only unclear in the slightest, ask in the thread how to design such stuff, so you dont waste time.

florianfendt:
Sorry was unclear here, I did that as S3 method.
Whats that Error in Travis: Github repo contains submodules... 
Couldn't reproduce locally.

berndbischl:
> Whats that Error in Travis: Github repo contains submodules... 

Ignore that for now.


> Sorry was unclear here, I did that as S3 method.

Does not matter what you wrote I can see the diffs anyway and read them :)
impute is not s3? and now has a new arg "task"?
like i said, look at capLargeValues, impute should look like that, in its signature.

florianfendt:
Ah ok, sorry I thought you were referring to ````reimpute()````.
Will change that, of course!

berndbischl:
what is the status here? that I should check it?

berndbischl:
what i just noticed: capLargeValues might be missing a "target" arg for the data.frame case
currently we seem to cap everything, including possibly the target.

berndbischl:
Do we really need the extra check on "target" with the warning?

florianfendt:
@berndbischl 
I was searching the repo a bit and couldn't find any other functions than capLargeValues and impute, where this could be an issue.
So you can decide what you want and I'll
1) change here and rebase
2) do the same for capLargeValues when I fix the bug I reported

florianfendt:
@berndbischl 
So this now consistent with PR #640
I think we are done here

florianfendt:
Possible update for NEWS:
````
- the functions impute and reimpute now also support task as an argument.
````

berndbischl:
i will merge this now and fix some stuff on master

Issue #443, opened 2015-08-17
benchmark should add an option keep.models


berndbischl:
this is very mem hungry, but some users might want this for convenience

giuseppec:
In the `BenchmarkResult` Object, we have redundant info for the learner, which increases the object size:

```
task = makeClassifTask(data = iris, target = "Species")
rdesc = makeResampleDesc("CV", iters = 2)
b = benchmark(makeLearner("classif.qda"), task, rdesc)

all.equal(b$learners$classif.qda, b$results$iris$classif.qda$learner)
```

jakob-r:
I think this "problem" we have multiple times. But what you mention is just the definition of the learner and and thus independent of the task size and usually pretty small as it's more or less just a basic list.
```r
object.size(b$learners$classif.qda)
# 11392 bytes
```

berndbischl:
We dont care about the redundant learners at the moment.

But will add the models with an option.

I assigned @florianfendt

berndbischl:
maybe also add a getter for the models in bmr_operators

berndbischl:
this is done now

PR #578, opened 2015-11-30
fix memory leaks for benchmark()

mllg:
I ran into a problem where I could not perform a very basic benchmark study of different SVMs w/o mlr eating >30GB of memory. This should fix it:

* `keep.preds` is now correctly passed down to `resample()`
* Added argument `models` to `benchmark()`, also passed down to `resample()`
* Some minor cleanups and simplifications in `benchmark()`

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+mlr_2.7:
+- New argument "models" for function benchmark
+- fixed a bug where 'keep.pred' was ignored in the benchmark function
+
 mlr_2.6:
 - cluster.kmeans: added support for fuzzy clustering (property "prob")
 - regr.lm: removed some erroneous param settings
--- a/R/benchmark.R
+++ b/R/benchmark.R
@@ -19,11 +19,14 @@
 #'   Performance measures for all tasks.
 #'   If missing, the default measure of the first task is used.
 #' @template arg_keep_pred
+#' @param models [\code{logical(1)}]\cr
+#'   Should all fitted models be returned?
+#'   Default is \code{TRUE}.
 #' @template arg_showinfo
 #' @return [\code{\link{BenchmarkResult}}].
 #' @family benchmark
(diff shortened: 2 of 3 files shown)


larskotthoff:
Thanks, looks good.

berndbischl:
Hi, please dont merge too quickly. 
First of all, we are currently not really on CRAN, because we still have a bug. I already invested quite a lot of time into this, and every further API change  makes it harder.

Second, we even have an open issue about this? At least this should be referenced?

berndbischl:
Also, please:

1) There is no documentation about the effect of this option in the BMR object, as far as I can see?
2) Apparently this is a bug fix and and API change. For both I dont see unit tests.

larskotthoff:
Wait, what's the bug? Do we have an issue for that?

berndbischl:
The bug was that keep.preds did not work correctly. Apparently @mllg discovered this and immediatly fixed this. Which is good, but we still have a rule that we never do that without a reproducing unit test.

The issue about the "models" option is here #443

berndbischl:
Summary of what needs to be done IMHO:

1) unit tests for keep.pred / the detected error
2) unit test for option "models"
3) doc effect of "models" in BMR object doc
4) either implement a BMR getter for the models or open a new issue for this. as this is in the thread for #443.

"models" is indeed a better name than "keep.models", because it is more consistent with the name in resample. I already agreed about this with @florianfendt

mllg:
> 1) unit tests for keep.pred / the detected error

In #581 

> 2) unit test for option "models"

In #581 

> 3) doc effect of "models" in BMR object doc

Part of ResampleResult, already documented there.

> 4) either implement a BMR getter for the models or open a new issue for this. as this is in the thread for #443

See #582.

berndbischl:
thx we seem to be done here

PR #584, opened 2015-11-30
refactor benchplots

berndbischl:
Hi I need some urgent help with this, please review this. I had to refactor this so we can get 2.6 as 2.7 on CRAN. 2.6 is currently broken, due to a bug in the Benchmark Summary Plot by Florian P. 

Therefore I refactored this quite a lot. 

I will now outline my changes below, and update NEWS when we are done. But we really have to solve this in about 24h.

Please do NOT change code now in the master or this branch, but complain about missing or bad stuff ASAP.


berndbischl:
@larskotthoff @mllg @zmjones @pfistfl

berndbischl:
Now the description is coming, please wait a bit

berndbischl:
1) I mainly touched these old functions

plot- and generateBenchmarkSummary.R
plotBenchmarkResult

2))
First of all, some of these had data generators, which was just code bloat. The plotters simply operate on the "BMR" objects. So I removed the generators. Then, the names seemed suboptimal. 
"plotBenchMarkResult" had a very general name and was a "box plotter". So I renamed it. I also tried to use the "BMR" in the names here, which we do in other functions names as well.

3)
Then the plotBenchMarkSummary function actually produced a bug an CRAN, a very annoying one. The problem was that we were displaying "relative scores" when compared with the best and worst. And sometimes we divided by 0, whatever. The point is: This was really hard to fix, in general. And while I thought about the plot, I disliked it more and more. My new plotBMRSummary does not have that problem anymore and wastes a lot less space. 

4) I also did some internal code clean up. Not as much as I would have liked, but as much as I could now

5) Many examples were kinda duplicated code. So I wrote 1 example under "benchmark" that showcases all of these methods. And referenced this from the other example places.

Now: help?





        modified:   R/generateRankMatrixAsBar.R
        new file:   R/plotBMRBoxplots.R
        new file:   R/plotBMRSummary.R
        deleted:    R/plotBenchmarkResult.R
        renamed:    tests/testthat/test_base_plotBenchmarkResult.R -> tests/testthat/test_base_plotBMRBoxplots.R
        new file:   tests/testthat/test_base_plotBMRSummary.R
        deleted:    tests/testthat/test_base_plotBenchmarkSummary.R

berndbischl:
I will use this place to list some TODOS:

* I am unsure about the plotRankMatrixasBar Plot, I need to look at it
* I need to update NEWS with the info from here
* Some functions use the arg "pretty.names". But not all of the benchmark plotters do this.

berndbischl:
@schiffner 
I guess we have to update minor things in the tutorial too, when this is merged.

The bottom line is that the code by Florian was really not reviewed enough. Which is really not his fault. And I am like 80% sure that I was the one who merged it without good enough review. We cannot do this anymore.

larskotthoff:
If it this this urgent then my vote is for removing this entirely and adding it back properly in a later release rather than trying to fix a huge chunk of functionality in a few hours. If I understand correctly, then a quick fix would also be to not do the normalisation.

My main point is really that you're complaining that we're merging things too quickly without proper review while at the same time asking us to review a massive refactoring that you're not sure about yourself within 24 hours.

larskotthoff:
Also, why did this generate an error on CRAN and not in our testing?

R/plotBMRBoxplots.R
```diff
@@ -0,0 +1,48 @@
...
+#' @family benchmark
+#' @export
+#' @examples
+#' # see benchmark
+plotBMRBoxplots = function(bmr, measure = NULL, style = "box", order.lrns = NULL, order.tsks = NULL,
+  pretty.names = TRUE) {
+
+  choices = c("learner.id", "task.id")
+  assertClass(bmr, "BenchmarkResult")
+  assertClass(style, "character")
```
mllg: Wrong assertion.

mllg:
I like the generateXData functions. You can write unit tests for the generated object and otherwise just can ensure that no exception is raised, but the plot may be complete garbage.

R/convertBMRToRankMatrix.R
```diff
@@ -55,12 +44,12 @@ convertBMRToRankMatrix = function(bmr, measure = NULL, ties.method = "average",
     "df = ddply(df, .(task.id), transform, alg.rank = rank(x, ties.method = '",
```
mllg: There has to be another way around? Is this fixed?

berndbischl:
@larskotthoff 
1) WRT the bug, and why it only appeared on CRAN: The bug is stochastic, and also appears in a test, that used runtime measurements. So it is not even seedable. Actually the unit test was kinda "weird" for it. Yes, the easiest fix would be to change the test or even remove it. 
So this works in your favor of pushing the larger change back, that I committed here.

2) The real problem is that this bug forced me to review the code, and I found too many problems. Problems in naming and behavior. Now they are on CRAN only a very brief time, so I would be OK with changing them quickly. If we keep them there longer, we need to maintain or deprecate this stuff what I really dont want.
I really would like a quick clean fix. And IMHO it is not a massive change.
And disregard the 24h. If we need a bit more time, so be it.

berndbischl:
And yes, I think we should really slow down on merging and changing stuff, doing this even more dilligently. But here the damage is already done and we need to react now.

berndbischl:
> I like the generateXData functions. You can write unit tests for the generated object and 
> otherwise just can ensure that no exception is raised, but the plot may be complete garbage.

100% agreed! I was really happy when this was introduced by @zmjones, exactly for the reasons you state. Here. the situation was a bit different. Because the BMR object was *already* the perfect container object. So the generateXData did nothing, it was like an "identity" map. I dont think we should do that, just for consistency. You cannot test more, you gain nothing. NB: that this also resulted from some of the simplification I did to the code. Before, in FPs version, there was maybe a reason to have the genXData function.

berndbischl:
"There has to be another way around? Is this fixed?" 

That is a string of R code, that is evaled right? Yes I saw that too, but forgot to fix it, thanks.
This is 100% not necessary.

berndbischl:
eval,parse,crap: "There has to be another way around? Is this fixed?"

Yes, now.

berndbischl:
WRT: Michel's note for bad arg checks above: fixed

berndbischl:
I also refactored the barchart. That leaves pretty.names and NEWS

berndbischl:
I added a issues for the pretty.names now, #592

pfistfl:
I will leave my 2 cents in regards to the refactored plots here.

``plotBMRBoxplots:``
- Nothing to add, pretty much the same just more efficient I guess.

``plotBMRSummary`:`
What I like:
- Saves a lot of Ink in comparison to old plot.
- Works on true scale, not scaled.
What I don't like:
- Overplotting: If two or more learners perform equally, dots are stacked and thus we can not see them. So for some learners I get 0 info. 
- Small dots are pretty underwhelming when comparing on larger BMR's.
- I can currently not see the point in the trafo = "rank" option.

``plotBMRRanksAsBarChart:``
Good:
-  Solves shared rank problem

I can currently not test / add more as I am running out of battery.

berndbischl:
> Overplotting: If two or more learners perform equally, dots are stacked and thus we can not see them. 

Correct. I have added a small vertical jitter and an arg for that.

> Small dots are pretty underwhelming when comparing on larger BMR's.

I added an arg for pointsize and set default to 4.

berndbischl:
> I can currently not see the point in the trafo = "rank" option.

It easily allows the comparions of ranks? Why is that not useful?

berndbischl:
merging now

Issue #582, opened 2015-11-30
Missing BMR getter for models


mllg:
See #443 and #578 .

berndbischl:
Hi @florianfendt 
Michel aready solve the issue #443 for you, you were too slow :)
Just do the getter quickly, then we can move on.

berndbischl:
we are done PR was merged

PR #585, opened 2015-11-30
#582 BMR getter for models

florianfendt:


berndbischl:
Seems good to me.
Lets merge this when the other PR is merged for the bench plot.

Other comments?

berndbischl:
Please document here what happens if "models = FALSE" in benchmark.
This is also unit tested correct?

You could check the same questions for the other getters!
Then we can merge.

berndbischl:
ping @florianfendt

florianfendt:
@berndbischl 
Ok, I documented that for   ````getBMRModels()````and ````getBMRPredictions()````.

I also put a note in ````getBMRPerformances()````and ````getBMRAggrPerformances()````,
the return value can't be ````NULL```` there but the data frames could consist of ````NA```` values
if the measures were calculated on the training sets.

berndbischl:
Ok I will merge this now, it looks good. I will then have one final look at it in master branch

berndbischl:
final remark. I checked it briefly some more but cannot see anything for now which is bad

Issue #235, opened 2015-02-05
control which class is over- or undersampled


schiffner:
Currently, we can undersample the larger or oversample the smaller class.
This is fine as long as we aim to make the class frequencies more equal.
Sometimes, e.g., in cost-sensitive classification, one may want to make the classification problem more imbalanced.

berndbischl:
@schiffner 
Julia, this is in master now.
Would you please check whether you think it is OK now? 
Then you can close please

schiffner:
@florianfendt: Thank you very much for doing this.

I had a short look at code and docs. And I extended the examples on the tutorial page on cost-sensitive classification. Everything looks good and works fine, so I will close here.

berndbischl:
perfect! thx

PR #588, opened 2015-12-01
#235 Control which class is over- or undersampled

florianfendt:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -20,6 +20,9 @@ mlr_new:
   predict were not passed down correctly
 - Fixed a bug where the function capLargeValues wasn't working if you passed a task.
 - capLargeValues now has a new argument "target", to prevent from capping response values.
+- added functionality to the functions oversample and undersample. It is now possible
+  to control which class to use in the corresponding function via the parameter 'cl'.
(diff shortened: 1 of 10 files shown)


berndbischl:
we agreed: 
- cl must be a class name, not "min" or "max", default is NULL, that implies the smaller or larger class, depending on the function
- first assert on cl is: either NULL or string
- later, when we have the data we check wether it truely is a class name

florianfendt:
@berndbischl 
changed what we discussed. Open for review

berndbischl:
I will merge this on the CLI now, that way it is easier for me to review

berndbischl:
1) Please dont use testNull when is.null exists

berndbischl:
2) Please have a look at how I am shortening the code of oversample() and undersample() now.

berndbischl:
3) one thing one can notice here, is that the "finer" more subtle thing of how the sampling works are actually underdocumented. I will fix this now.

berndbischl:
4) Please note that assertString exists

berndbischl:
branch is on my laptop, must be uploaded NOW
imbalancy_add_arg_to_select_sampled_class

berndbischl:
merged on console. we should be done here.

PR #595, opened 2015-12-04
adds fixed rf regr se estimators

zmjones:
- no infjackknife since randomForestCI not on CRAN yet


berndbischl:
Ok, lets maybe review this for 2.8

zmjones:
Ok sounds good. I forgot to delete one of the old tests that is duplicated, should pass now.

zmjones:
Maybe for 2.9 I will have all of this available for bagging wrappers.

berndbischl:
that would have indeed been my next question. how urgently do we need this. shall we do one unifying package update, that handles RF and BW both at the same time?

zmjones:
i would prefer this to be merged first as at least the jackknife estimator is broken in the current release. i will rewrite it to make it as general as possible a bit later. there is also another estimator i can add and, if randomForestCI is  on CRAN soon, that as well.

berndbischl:
ok, then lets review this soonish and keep it in 2.8

larskotthoff:
Thanks Zach, looks good. Two small niggles though -- please document the changes in NEWS and also add a reference to the paper where the methods for SE estimation are described.

zmjones:
yes thank you for reminding me. do you mean in the docs for `predict.WrappedModel`?

berndbischl:
you mean where to put the refs for the RF se estimators?

zmjones:
yes

berndbischl:
hmm. it should be on the page for the RF. but we dont have one.... 
and its too long for the note

berndbischl:
you could actual create a rox page ?regr.randomforest, doc it there and refer to it in the note? is that good?

zmjones:
sure that is possible but what else would we document there? and also would people know to look at the docs there? it also seems a bit ad-hoc without doing it for the other learners.

berndbischl:
well

1) like i said we could refer to it in the note of the learner, so its visible

2) you do know that you have to to make an alternative suggestion then? ;-)

zmjones:
haha i was afraid of that. 

i guess i would be ok documenting it in the functions themselves. we could then get more specific (e.g, eqn: 5, pg 13) and we could mention somewhere else in the docs that we do this.

berndbischl:
sorry, i dont get this maybe i am stupid. where do you want to doc it?

zmjones:
in the se methods themselves. after the function header we could just have a series of comments that explain what the implementation is based on. you'd only see this if you looked at the source, but we could mention that we are doing this elsewhere in the docs.

berndbischl:
i dont like this. it should be visible on a regular doc page.....

zmjones:
but the only doc page we can make for it is one that probably nobody will see right (because it is only for 1 or a few learners)? or do you not think that is the case because this is mentioned in the note?

berndbischl:
well, our convention was: if we do something specific for the learner we mention it in the note.
also: we just need an extra page to put the refs in here. so this seems clear. in the sources nearly noboby will find it. and if we ref it in the note, people have a chance to find it. 

also if it is mentioned in the note and a doc page, which are both in turn rendered to html and online, it is googleable right?

we could also display the note when a learner is printed. do you want that, in general i mean?

IMHO the only decision is how to call the extra doc page (if nobody can name a better existing one). I would suggest to call it like the learner, so ?regr.randomForest.

zmjones:
Ok just pushed these changes. Let me know what you all think.

zmjones:
ok looks all good now, fixed all the check errors

larskotthoff:
Thanks, looks good. Now there are conflicts though -- could you rebase on master please?

PS: If you create new branches in the mlr repo and PRs from those, I'd be happy to fix small things like this myself.

berndbischl:
WRT the branches: Yes I guess this is the better approach for us developers.

I would like to look at this at least for a few mins before it is merged. Tell me when you think you are done, Lars

zmjones:
it is just a news conflict which can be easily fixed in the merge, but can rebase if preferred

larskotthoff:
@berndbischl I'm done looking over this.

@zmjones Could you rebase please? Then we can merge the PR directly.

zmjones:
ok rebased and re-running on travis

zmjones:
ugh another conflict now. someone please merge this before another news conflict occurs!

larskotthoff:
3...2...1...

larskotthoff:
Merged!

Issue #587, opened 2015-12-01
Multilabel-classification: sparse labels 


QuayAu:
I have some data sets, where labels are very sparse (e.g. only one observation has that label). 
Using the binary relevance method, i get following error during resampling:

![medicaltask](https://cloud.githubusercontent.com/assets/15329438/11495586/aea09896-980c-11e5-8aab-ec05766a0845.png)

It would be nice, if the binary classifier would always predict "FALSE" for very sparse labels.

berndbischl:
we need to discuss how to handle this. another option for now is to remove these cols from the data, as you already "know" what would happen to them during prediction. 

but we need a consistent way to handle this technically. either an option in the learner or a wrapper.

berndbischl:
we now have makeConstantClassWrapper which should be able to handle this.

PR #599, opened 2015-12-06
wrapper that adds support for one-class problems to learners (#587)

larskotthoff:


berndbischl:
Hi, first of all thx.

But like I said in the issue, we should discuss this....

1) The name: "OneClass Problems"  and so on is actually a terminus technicus for something else.
So we cannot use this.

2) I am REALLY unsure whether this is should be a wrapper. It is more or less a really "technical" problem. And I guess it should be handled seamlessly by mlr by default. So IMHO it should be somehow an option. 

Can we please define first what we do, textually?

berndbischl:
And the correct place might be in the issue.

larskotthoff:
The issue seems broader -- in particular it's about *sparse* labels.

Any suggestions for a better name? I've modelled this on the "oneclass" property of learners.

Regarding the wrapper approach. The alternative would be to modify train and predict to do this transparently. The problem I see with that approach is that the model returned by train would be different depending on the data. In particular we can't really make the model behave the same in every case.

We've taken the same wrapper approach in other cases, e.g. imputing missing values for learners that don't support missing values. This isn't done transparently either but you have to explicitly wrap the learner.

berndbischl:
> The issue seems broader -- in particular it's about sparse labels.

Not really. It is the same thing. It is purely technical, if a learner sees to few different labels, it breaks. It shouldnt, at least if all labels are the same, IMHO, also without a wrapper. We also have other options for
problems with factor levels.

> Any suggestions for a better name? I've modelled this on the "oneclass" property of learners.

The oneclass property of a learner means whether a learner supports "true" oneclass classification
(which we then did not cover in mlr...., still an open todo)
https://en.wikipedia.org/wiki/One-class_classification

Like I said, I am also unsure where to put this code, so lets discuss this.

berndbischl:
We hangouted and agreed:

1) the wrapper is probably a good idea, also it should only be seldomly needed by users.

2) it is more flexible if we allow the class to be nearly constant. we need an arg for this. the behavior should be similar as in removeNearlyConstantFeatures.

3) A good name might be NearlyConstantClassWrapper.

4) Remove the "oneclass" property stuff

5) Inherit from basewrapper, like in filterwrapper and so on

larskotthoff:
Done all of that, pushed.

We seriously need some documentation for the internals though, figuring this out was extremely frustrating and in the end I just copied the code that does the same thing from the cost-sensitive wrapper. I have no idea why it works now; it certainly breaks if I change even minor details that as far as I can see shouldn't have any effect.

berndbischl:
1) The sentence in NEWS needs to be adapted

2) I can try to extends the docs a bit if I merge

3) Can we try to get closer with the names to removeConstantFeatures? 
There the "nearly" is not in the name, and we have "perc" as an argument.

I know this is nitpicking, but we really have to reduce inconsistencies.

Regarding the API docs: I 100% agree, but we have to plan this.
We can add a issue to the tracker, but it will rot there. 
I mean you opened one just for the names of the test files, it is still there.

Can you also ask already 2-3 questions which came up when you tried to solve this?

larskotthoff:
Regarding the name of the argument: I think `perc` is actually misleading in removeConstantFeatures as it is not a percentage but a fraction.

larskotthoff:
I'm going to call this `frac`.

berndbischl:
its too late here to get worked up about this 
http://cdn.meme.am/instances/60981391.jpg

larskotthoff:
I think we should change this in removeConstantFeatures actually. But that would be an API change.

Ceterum censeo `=` is ad reponi cum `<-`.

larskotthoff:
Travis failure seems to be unrelated, cannot reproduce locally.

berndbischl:
Logs indicate that there seems to be a dim drop going on 

```

4: apply(mod$predicted, MARGIN = 1L, which.max) at test_classif_randomForestSRCSyn.R:21
5: stop("dim(X) must have a positive length")

```

Has the RFSRCSyn package been updated?

larskotthoff:
Yep. Master build fails as well now. I'm just opening another issue for this.

larskotthoff:
This is ready to be merged now I think.

catastrophic-failure:
I see this was resolved way back, but I saw you were in trouble regarding the choice of name for this wrapper function. Elsewhere I saw this kind of functionality being refered to as _novelty detection_. NoveltyDetectionWrapper perhaps?

berndbischl:
> I see this was resolved way back, but I saw you were in trouble regarding the choice of name for this wrapper function. Elsewhere I saw this kind of functionality being refered to as novelty detection. NoveltyDetectionWrapper perhaps?

hmm, this here is not really novelty detection. although mlr should soon support this.
and when we do i find "OneClassClassification" a good title. 
I will try to merge this at one point

berndbischl:
@larskotthoff 
i guess i could merge this. a few comments:

- can you use prop.table to directly compute proportions / relative frequencies?

- instead of the complex S3 model can you simply create a list containing the most frequent class?

larskotthoff:
- prop.table: Done.
- list instead of S3: This wouldn't work as nicely with the chained model though, would it? I'm checking the class of the model and it feels like changing that check to look for a list would be more brittle.

berndbischl:
@larskotthoff 
ok lets merge this. please specify that the default of "frac" is 0, in the Param.
and add a sentence in the docs how probs are predicted here, pls

larskotthoff:
Done.

PR #601, opened 2015-12-07
fix assertion for upcoming checkmate update

mllg:


berndbischl:
I guess you need to update the DESCRIPTION >=  then too, maybe later.

mllg:
> I guess you need to update the DESCRIPTION >= then too, maybe later.

Nope. The check was just nonsense and the updated check works for the current and upcoming version of  checkmate. Should wait for Travis to be fixed though.

berndbischl:
Ok, so basically we can merge this anytime then?

Issue #628, opened 2015-12-19
there is no doc page for the mini resample objects like cv10


berndbischl:
?cv10 does not work

florianfendt:
@berndbischl 
I put up a PR on this.

berndbischl:
@florianfendt 
thx for being so quick. merged + done + closing.

PR #630, opened 2015-12-19
#628 resample objects help page

florianfendt:
Since we want to dock the help page to ResampleDesc I couldn't find another way than
doing this:
````
#' @rdname makeResampleDesc
#' @export
#' @usage NULL
#' @docType NULL
#' @format NULL
#' @keywords NULL
````
again for every object, so we really only add the ````@section```` part to the ResampleDesc help page.
Is there really no other way to achieve this?


berndbischl:
I guess we can draw this in, looks good

PR #633, opened 2015-12-20
fix bug in irace with unnamed discrete values, fixes #627

### LINKED ISSUE #627: tuneParams using IRace fails with discrete parameters that are not character
`tuneIrace()` calls `ParamHelpers::convertParamSetToIrace()` which converts discrete parameters into discrete *character* parameters using the *names()* of the parameter `$values`. However, the fitness function (`hookRun` in tuneIrace.R) makes no effort translating from names back to parameter values (it *does* translate logical parameters though).

This works fine, as long as the discrete parameter `values` vector is of mode character for which names equal values (default when created using `makeDiscreteParam()`); in other cases it causes errors:
```R
> lrn = makeLearner("classif.randomForest")
> ctrl = makeTuneControlIrace(maxExperiments = 200L)
> ps = makeParamSet(makeDiscreteParam("ntree", c(100, 200))) # !! create discrete numeric param
> tuneParams(lrn, iris.task, rdesc, par.set = ps, control = ctrl)
[Tune] Started tuning learner classif.randomForest for parameter set:
          Type len Def  Constr Req Tunable Trafo
ntree discrete   -   - 100,200   -    TRUE     -
With control class: TuneControlIrace
Imputation value: 1
[Tune-x] Setting hyperpars failed: Error in setHyperPars2.Learner(learner, insert(par.vals, args)) :
  100 is not feasible for parameter 'ntree'!

Error in getIndex(par$values, x) : Value not found!
```
(The error message is a bit misleading: the parameter value that IRace tries to set for ntree is the **character** "100", it should be the **numeric** 100. Consider:
```R
> lrn = makeLearner("classif.randomForest")
> ctrl = makeTuneControlIrace(maxExperiments = 200L)
> ps = makeParamSet(makeDiscreteParam("ntree", c(name100=100, name200=200)))
> tuneParams(lrn, iris.task, rdesc, par.set = ps, control = ctrl)
[Tune] Started tuning learner classif.randomForest for parameter set:
          Type len Def          Constr Req Tunable Trafo
ntree discrete   -   - name100,name200   -    TRUE     -
With control class: TuneControlIrace
Imputation value: 1
[Tune-x] Setting hyperpars failed: Error in setHyperPars2.Learner(learner, insert(par.vals, args)) :
  name100 is not feasible for parameter 'ntree'!

Error in getIndex(par$values, x) : Value not found!
```
)

berndbischl:
I did the following:

- I improved the conversion code for irace. 
- I added a new option in ParamHelpers so, we now rely on version 1.7 for this
- I removed a convertX function, which was only there for irace and is not needed now

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,7 @@ Depends:
     R (>= 3.0.2),
     BBmisc (>= 1.9),
     ggplot2,
-    ParamHelpers (>= 1.6),
+    ParamHelpers (>= 1.7),
     stats
 Imports:
     checkmate (>= 1.6.2),
--- a/NEWS
+++ b/NEWS
@@ -13,6 +13,7 @@ mlr_new:
 - Fixed a bug where the resampling objects hout, cv2, cv3, cv5, cv10 were not documented
   in the ResampleDesc help page
 - regr.xgboost, classif.xgboost: add feval param
+- fixed a bug in irace tuning interface with unamed discrete values
 
 - API changes
 -- listLearners now returns a data frame with properties of the learners if
--- a/R/convertX.R
+++ b/R/convertX.R
@@ -21,15 +21,6 @@ convertXMatrixCols = function(xs, par.set) {
   })
 }
 
-# convert logical param values from chars to true logicals, eg irace produces strings in tuning
-convertXLogicalsNotAsStrings = function(x, par.set) {
-  types = getParamTypes(par.set, use.names = TRUE)
-  j = types %in% c("logical", "logicalvector")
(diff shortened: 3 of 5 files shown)


berndbischl:
no idea why we have a conflict now. but please review first. maybe we have to rebase on master

larskotthoff:
Looks good, happy to merge when the conflicts are fixed.

berndbischl:
Hey, to Zach you said if it would be in a branch you would do it....!

larskotthoff:
But he doesn't bombard us with 10 PRs at once ;)

PR #635, opened 2015-12-20
set some crossval params to tuneable = FALSE and clean up mini param stuff

### LINKED ISSUE #605: params which trigger crossvals not tunable
Is that a conscious decision, that i.e. `cross` for *.svm and `xval` for *.rpart are not set to `tunable = FALSE` ?
Both trigger internal cross validations which I gues should not be counted as a tunable param?

berndbischl:
fixes #605


larskotthoff:
Looks good, thanks.

PR #638, opened 2015-12-21
clean up learner and measure notes

schiffner:


berndbischl:
I have to admit  I did not check every single char changed, but you are also much more precise than I am in these thing. So I will simply say "thx" and pull this in now.

schiffner:
Maybe I should have put something in the comment field. So just for the record:
I did some reformatting such that in the tutorial learners table everything looks ok.
The only "real" changes are that I added notes to 
https://github.com/mlr-org/mlr/blob/master/R/RLearner_regr_LiblineaRL2L1SVR.R and https://github.com/mlr-org/mlr/blob/master/R/RLearner_regr_LiblineaRL2L2SVR.R about the defaults of `svr_eps`.

Issue #590, opened 2015-12-01
Bug in capLargeValues.Task 


florianfendt:
````capLargeValues()````is not working if you pass a task.
See for example:
````r
> capLargeValues(binaryclass.task)

Error in capLargeValues(binaryclass.task) : 
Assertion on 'positive' failed: Must be element of set {''}

> capLargeValues(multiclass.task)

Supervised task: multiclass
Type: classif
Target: Species

Features:
numerics  factors  ordered 
       0        0        0 
Missings: FALSE
Has weights: FALSE
Has blocking: FALSE
Classes: 0
````
It's just a simple typo, where we then pass ```NULL```, to the data.frame method.
We're also missing a unit test for ````capLargeValues.Task()````.
I'll assign myself here and fix both of those things over this week.

florianfendt:
Closing here. Has been fixed in PR #640

PR #640, opened 2015-12-22
#590 capLargeValues

florianfendt:
This will fix the bug I reported in #590 
I also put the check for target consistency in a little function, I can use it then for impute and createDummyFeatures.


berndbischl:
Thx. I can merge this in now.
@florianfendt Is it OK if I iterate over this quickly. Or will this create conflicts on your side?

berndbischl:
Thx, I will merge this now, then quickly look at this in master branch

PR #641, opened 2015-12-23
ROC-plots for multilabel classification

PhilippPro:
First prototype. Would appreciate your comments before continuing. 

"generateMultilabelThreshVsPerfData" works for prediction, resample and benchmark. "PlotMultilabelROCCurves" only for prediction and resample. 

At the moment I am creating a list for each benchmark result instead of one list with all the data in the data argument. But before changing this please tell me if you are happy with the Prediction/Resampling solution. 

(General) Problem in plotROCCurves: It can happen that only "FALSE"-data in the MultilabelThreshVsPerfData, then only NAs are returned (in the specific column). That's why I used function "try" here, because otherwise plotROCCurves crashes. 

Other comment: I prefer ROCCurves of different benchmarks in one graphic, because they are easier to evaluate. 

Still Missing: 
- plotMultilabelThreshVsPerf
- plotMultilabelThreshVsPerfGGVIS

### DIFF
--- a/R/generateMultilabelThreshVsPerfData.R
+++ b/R/generateMultilabelThreshVsPerfData.R
@@ -1,0 +1,138 @@
+#' @title Generate threshold vs. performance(s) for multilabel classification.
+#'
+#' @description
+#' Generates data on threshold vs. performance(s) for multilabel classification that can be used for plotting.
+#' @title Generate threshold vs. performance(s) for Multilabel-classification.
+#'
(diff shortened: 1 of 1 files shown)


larskotthoff:
Is this still being worked on or can it be closed?

PhilippPro:
Can be closed, I am not motivated on working on the complicated structure of the plot functions. ;)

Issue #647, opened 2016-01-06
makeModelMultiplexer for regr.glmnet produces error


ecrutchfield:
After creating a model multiplexer for regr.glmnet, any attempt to call predict with that learner will generate an error. For example...
```
  library(mlr)
  base.learners = list(
    makeLearner("regr.glmnet"),
    makeLearner("regr.randomForest")
  )
  learner = makeModelMultiplexer(base.learners)
  expect_equal(learner$predict.type, "response")
  # now lets see that the next code runs and does not complain about matrix output for
  # base learner predict output
  r = holdout(learner, regr.task)
 
Error in checkPredictLearnerOutput(.learner, .model, p) : 
  predictLearner for ModelMultiplexer has returned a class matrix instead of a numeric! 
```

This same error is also generated if you call tuneParams().

To fix this you should be able to call glmnet's predict method with the 's' param and it will generate a numeric vector, instead of a matrix. It appears the simple fix is to modify the predictLearner.regr.glmnet method in RLearner_regr_glmnet.R as follows:
```
drop(predict(.model$learner.model, newx = .newdata, s=getHyperPars(.learner)$s, ...))
```

Here's a suggested test case for test_tune_ModelMultiplexer.R
```
test_that("ModelMultiplexer works for regression tasks", {
  base.learners = list(
    makeLearner("regr.glmnet"),
    makeLearner("regr.randomForest")
  )
  learner = makeModelMultiplexer(base.learners)
  expect_equal(learner$predict.type, "response")
  # now lets see that the next code runs and does not complain about matrix output for
  # base learner predict output
  r = holdout(learner, regr.task)
  
  # now check that we can tune the threshold
  ps = makeModelMultiplexerParamSet(learner,
                                     makeDiscreteParam("alpha", c(0.5,0.9)),
                                    makeDiscreteParam("mtry", c(2, 3))
  )
  rdesc = makeResampleDesc("Holdout")
  ctrl = makeTuneControlGrid()
  res = tuneParams(learner, regr.task, resampling = rdesc, par.set = ps, control = ctrl)
})
```

I tried forking master and fixing this myself, but I had lot's of problems running the others tests i.e. I kept getting seg. faults and the R session would crash. Anyway, if you'd prefer I submit this fix via a pull request, I can give it a try.

Cheers,
Eric

larskotthoff:
[...]

PR #649, opened 2016-01-06
fix bug in ModelMultiplexer predict, where hyperpars were not passed on

berndbischl:
Fixes issue #647 

Please review


schiffner:
Looks good to me.
* Maybe mention in NEWS that the bug was fixed.
* In the tests (l. 129) there is a copy-paste remnant (comment about threshold tuning that does not apply here)

larskotthoff:
Agreed with Julia.

berndbischl:
ok, in can do both. NEWS i left out because it is really particular, but i can do it

berndbischl:
done.
I decided to remove the later part of the test completely. It has nothing to do with bug, and just wastes time.

review and then merge pls

larskotthoff:
Thanks, merged.

PR #655, opened 2016-01-11
Add multilabel classifier chains wrapper

QuayAu:
This is one of the four multi label wrappers I want to implement. I start with this one first and if everything is ok, I will work on the last three.

Classifier Chains ( from http://www.cs.waikato.ac.nz/ml/publications/2009/chains.pdf ):
"Classifiers are linked along a chain where each classifier deals with the binary relevance
problem associated with label l_j∈L. The feature space of each link in the chain is extended with the 0/1 label associations of all previous links."


larskotthoff:
Thanks! Could you check in the construction method whether the passed learners support binary classification please? The "random" order parameter also isn't documented.

I don't understand how you're adding the predictions as features in the train method. Why did you not use the same way to do it as in the predict method?

QuayAu:
I added the check for binary classification and edited the documentation for the "random" parameter.

In the train method, the additional features are true labels, which are not available at prediction time.
At prediction time, when true label information is not available, they are replaced by the predicted labels. This is why a different method in the training method is used. 
(The similiar method "Nested Stacking", will also use predicted labels in the train method. I will also add this method, when this is done.)

The lengthier code block (line 68-70) replaces the boolean values of the variables with 1 or 0, since I got an error, when I tried to use boolean values as features.
```
+    if (length(data2[, which(names(data2)%in%setdiff(targets, tn))]) != 0) {
+      data2[, which(names(data2) %in% setdiff(targets, tn))] = ifelse(data2[, which(names(data2) %in% setdiff(targets, tn))] == TRUE, 1, 0)
+    }
```

Thank you for checking my code!

larskotthoff:
Thanks for the update. To convert a Boolean to numeric, you can simply do `as.numeric(data2[, which(names(data2) %in% setdiff(targets, tn))])`. I would pull out `which(names(data2) %in% setdiff(targets, tn))` into a variable as you're using it several times.

Could you add the description of how the labels are taken into account during prediction to the documentation please? It doesn't mention at the moment that during prediction the previously predicted labels are added as features.

Also I think your check can be simplified to

    if (length(names(data2) %in% setdiff(targets, tn)) != 0) {

and where possible, you should use the accessor methods that mlr provides, i.e. `length(.task$task.desc$target)` should be `length(getTaskTargetNames(.task))`; same thing for `$data$response` -- use `getPredictionResponse()` instead (and `getPredictionProbabilities()` later).

Finally, could you add something to NEWS please to document the change?

Sorry, I'm finding more things looking at it again :) They're all small things though.

mllg:
> if (length(names(data2) %in% setdiff(targets, tn)) != 0) {

This is always true for `length(names(data2)) > 0`. I think you are looking for `any()`? :wink:

larskotthoff:
Ah yes you're right, thanks!

QuayAu:
When I make a PR for the next wrapper, should the unit tests of the Classifier Chains also be in there?
Will it work with Travis or should I wait until this wrapper is merged?

berndbischl:
Can you maybe do a PR for the measures first?

berndbischl:
This can be reviewed now, but should be merged later when mlr 2.8 is on CRAN.

larskotthoff:
Travis will automatically check the PR. So yes, please always include unit tests.

PR #656, opened 2016-01-12
add multilabel measures (subset01, f1mult, jaccard)

QuayAu:
measures taken from http://www.sciencedirect.com/science/article/pii/S0031320313004019

subset01: prediction is right only if all labels are predicted right
f1mult: multilabel version of f1 score https://en.wikipedia.org/wiki/F1_score
jaccard index: https://en.wikipedia.org/wiki/Jaccard_index

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,5 @@
 mlr_new:
+- Add multilabel performance measures (subset01, f1, jaccard)
 - Feature filter "univariate" had a bad name, was deprecated and is now called
  "univariate.model.score". The new one also has better defaults.
 - (generate/plot)PartialPrediction: added new arg "geom" for tile plots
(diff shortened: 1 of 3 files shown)


berndbischl:
This can be reviewed now, but should be merged later when mlr 2.8 is on CRAN.

larskotthoff:
Could you please include some unit tests for the new measures? And mention in NEWS.

QuayAu:
I just realized I made a mistake with the F1 and Jaccard measures. I misinterpreted the formulas, I will update this later.

QuayAu:
I corrected the formulas for F1 and Jaccard measures.
0/0 is defined as 1 in this context (see page 2 in http://jmlr.org/proceedings/papers/v28/dembczynski13.pdf)

PR #657, opened 2016-01-12
Remove mlrMBO multifid support (Closes #629)

### LINKED ISSUE #629: remove multid special code from evalOptState
why is that even in there and on cran?

```

  if (inherits(control, "TuneControl") || inherits(control, "TuneMultiCritControl")) {
    if (inherits(control, "TuneControlMBO") && control$mbo.control$multifid) {
      #transform multifid lvl to param setting
      multifid.param = control$mbo.control$multifid.param #string
      multifid.values = control$mbo.control$multifid.lvls #factor over discrete values
      state[[multifid.param]] = multifid.values[state$.multifid.lvl]
      state = dropNamed(state, ".multifid.lvl")
      par.set = c(par.set, makeParamSet(
        makeDiscreteParam(multifid.param, values = multifid.values)))
    }

```

@jakob-r 

We should really not do something like this. Please talk about this before "hacking" something like this in there.

jakob-r:


berndbischl:
Thx, will merge now

Issue #579, opened 2015-11-30
Implementation of Rsquared/adjusted Rsquared as an performance measure


competulix:
It would be really cool to see Rsquared/adjusted Rsquared implemented as a possible choice for regression tasks.

larskotthoff:
I'm guessing you want the [coefficient of determination](https://en.wikipedia.org/wiki/Coefficient_of_determination) as a performance measure?

competulix:
Yes, precisely. :-)

berndbischl:
@competulix is perfectly right, this is missing for far too long now.

berndbischl:
@florianfendt Would you do this and add this to your todo list?

florianfendt:
I implemented those measures in #659, which has been merged, therefore closing here.

PR #659, opened 2016-01-12
#579 new regression measures

florianfendt:
I added the two varieties of the R^2 measure named rsq and expvar, and adjusted R^2 arsq.
NEWS then needs an update:
````
new measures
-- expvar
-- arsq
-- rsq
```

Also, I made the code in the tests for all the measures a bit more readable, imo.


larskotthoff:
Thanks, looks good. Will hold off merging until 2.8 is released though.

berndbischl:
I dont see a reason not to merge this now. I will add some more notes to the measures to make stuff even clearer and add NEWS

PR #653, opened 2016-01-09
various doc fixes

schiffner:
I know this PR is quite large, but most changes are pretty straightforward.

I fixed some typos, formatting and roxygen, ..., added some missing info with regard to multilabel and survival and doc'ed the `CalibrationData` object.

Here is a list of files where you might want to check that I didn't miss anything. 

Aggregation:
* extended and reformatted the docs of arg `fun` of `makeAggregation`

BenchmarkResult_operators:
In the docs of `getBMRPerformances`, `getBMRAggrPerformances` there was a
note saying that "Measure values will be NA for ResampleDescs with predict is “train”"
I find this more confusing than helpful. Of course, when you are doing things
wrong (i.e. don't set a suitable aggregation) you will get meaningless results.
Therefore I removed the notes, but can add a reminder about measure aggregation
to the benchmark page in the tutorial.

generateCalibration: added missing return field and doc'ed the CalibrationData object.
@zmjones: Would you mind having a look?

Measure:
* extended docs of arg `fun` of `makeMeasure`
* in the docs for `setAggregation` the arg_measure template is not used anymore because the info about the default does not apply here (see below)

Measure_custom_resampled:
* extended docs of arg `fun` of `makeCustomResampledMeasure`

RLearner:
* added multilabel to the learner class template (arg_lrncl) and use this template in RLearner
* added missing learner properties

Task:
* added info about length of the `target` arg with regard to multilabel.

TaskDesc:
* extended docs with regard to multilabel and survival

generateFilterValues, generateLearningCurveData, generatePartialPrediction:
* Because of a small error in roxygen the info about the generated `*Data` objects was not
shown on the *Data help pages. I moved the `*Data` docs to the
`generate*Data` help page (which is just my personal taste) and made some small changes.
* @family filter was twice in the docs, therefore removed one
* also removed the alias FilterValues from the docs of deprecated function `getFilterValues`

makeLearner:
* add multilabel info

plotBMRSummary:
* The last sentence in the description field was incomplete.
  @berndbischl: I guess your intention was to point to the transformations available in ggplot2?

predictLearner:
* The @details were copied from `trainLearner`, I added some info for predict instead.
* @return: added survival, cluster, and multilabel. Please check if everything is correct.
I wasn't totally sure whether to doc the case `pred.type = "prob"` for survival.
This case is also doc'ed everywhere else (e.g. ?makeLearner) but `predictLearner` says that "prob"
for survival is not yet supported.

arg_measure:
[...]

PR #664, opened 2016-01-15
Update manual after PR #653

florianfendt:
The .Rd files weren't updated after the huge manual fix in #653.
I just wanted to open a PR myself, so I thought it might be better to have this separated.

Also, there was a little bug in the RLearner.R documentation which I fixed here, too.


berndbischl:
I really dont get it why travis sometimes autogenerates this, and sometimes does not.
thx.

schiffner:
As far as I can see the last builds on the master branch errored. I guess that's the reason for the not-updated docs.

berndbischl:
ah yes. that was travis, the system, yesterday. i guess it work now. I am looking at it currently

Issue #612, opened 2015-12-15
Problems with regr.penalized.ridge and regr.penalized.lasso 


berndbischl:
First of all, the NAMESPACE is again broken of that package.
The learners only work if we load the package into the global NAMESPACE.
We only dont see this, because in our unit tests we already load the package into globalenv....

So we need to add 
    package = "!penalized"
to the learner.

Second, it seems that there is a bug as both learners seem to produce the same output.

berndbischl:
We should also by default set trace=FALSE, it really sucks

berndbischl:
And the reason that the weka team reported similar outputs is because we not lambda1 and lambda2 correctlty in the defaults!

berndbischl:
Actually I am stupid. The reason both learners return the same model, because the default is lamda = 0, lambda is the penalty control constant. so with lam = 0, we fit a linear model for lasso and ridge. so everything is ok.

But I do see that some other params are missing description objects. 

Can these be added. Flo would you do that?

So TODOS

1) Change trace = FALSE (and add note)
2) load package with !
3) add other params and look at docs for this

florianfendt:
will do!

berndbischl:
Sorry, to simply paste unfinished stuff here, this is what i could do before time ran out

```
     par.set = makeParamSet(
       makeNumericLearnerParam(id = "lambda1", default = 0, lower = 0),
+      makeUntypedLearnerParam(id = "unpenalized"),
+      makeLogicalVectorParam(id = "positive"),
+      makeNumericVectorLearnerParam(id = "startbeta"),
+      makeNumericVectorLearnerParam(id = "startgamma"),
+      makeLogicalLearnerParam(id = "standardize", default = FALSE),
+      makeIntegerLearnerParam(id = "steps", lower = 1L, default = 1L),
+      makeNumericLearnerParam(id = "epsilon", lower = 0, default = 1e-10),
+      makeIntegerLearnerParam(id = "maxiter", lower = 1L),
       makeLogicalLearnerParam(id = "trace", default = FALSE, tunable = FALSE)
+
-    note = ""
+    note = "trace=FALSE was set by default to disable logging output"
   )
 }
 
diff --git a/R/RLearner_regr_penalized_ridge.R b/R/RLearner_regr_penalized_ridge.R
index f04d511..da602db 100644
--- a/R/RLearner_regr_penalized_ridge.R
+++ b/R/RLearner_regr_penalized_ridge.R
[...]

PR #665, opened 2016-01-15
Enhancement of learners from penalized package (Issue #612)

florianfendt:
Ok, got a bit of a bigger project than I expected.
What I did:
1) got the parameters right (hopefully ;) )
2) added the classification learners
3) To have the survival learner consistent with the above I split up the learner into the 3 learners ridge, lasso, fusedlasso

Update for NEWS:
````
- Added the missing learner parameters for regr.penalized.lasso and regr.penalized.ridge

new learners
-- regr.penalized.fusedlasso
-- classif.penalized.ridge
-- classif.penalized.lasso
-- classif.penalized.fusedlasso
-- surv.penalized.ridge
-- surv.penalized.lasso
-- surv.penalized.fusedlasso

removed learners
-- surv.penalized
````


larskotthoff:
Thanks! I'm not sure how I feel about the hardcoded 0.5 threshold for converting the probabilities to class labels though. I would make this another parameter for the learner (with a default of 0.5) so that the user can adjust this easily.

Opinions?

berndbischl:
> Thanks! I'm not sure how I feel about the hardcoded 0.5 threshold for converting the probabilities to class labels though. I would make this another parameter for the learner (with a default of 0.5) so that the user can adjust this easily.

No, that is standard for all learners. mlr allow predict.type = "prob" to handle that specfically, then you have setThreshold and tuneThreshold and so on

larskotthoff:
Ok, fair enough.

Looks like some tests are failing though. Could you please have a look @florianfendt ?

florianfendt:
Fixed now.
I had a typo in the survival learners,
and the class labels for the classification learners were switched, as penalized predicts probabilities on
second level of target apparently.

berndbischl:
So the label switching unit test actually did something good and detected that?

florianfendt:
yep.
Btw:
While trying to figure this out I switched the predict type in the label switching test to "response".
Thereby, I found some other learners that might cause problems. I will look into this in more detail,
so we can talk about it on Tuesday and open a new issue for that.

berndbischl:
new PR is #850

PR #681, opened 2016-01-29
Added classif_cvglmnet learner

mllg:


berndbischl:
Looks good. I dont see any further problem and travis runs. 
I will update NEWS now.

PR #684, opened 2016-01-29
Multilabel RandomForestSRC

PhilippPro:
This is the implementation of the Multilabel-RandomForestSRC algorithm.

I also included some tests.

### DIFF
--- a/R/RLearner_multilabel_randomForestSRC.R
+++ b/R/RLearner_multilabel_randomForestSRC.R
@@ -1,0 +1,50 @@
+#' @export
+makeRLearner.multilabel.randomForestSRC = function() {
+  makeRLearnerMultilabel(
+    cl = "multilabel.randomForestSRC",
+    package = "randomForestSRC",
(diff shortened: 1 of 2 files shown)


berndbischl:
I would like to merge this, but we cannot have so much copy-past for the testing. 
This is not maintainable. 

I will iterate over this now. But please DONT copy-paste so much any more

schiffner:
@berndbischl : Did you already work on this PR?

I just wanted to add that also the par.set, par.vals, and hardcoded args should be updated before this gets merged.

berndbischl:
hi  @PhilippPro 

please note that this PR #959 already contains generic code to test all basic multilabel RLearners 
(learners_all style). so please

- remove the whole stuff in test_base_multilabel
- create a normal individual learner test file

and look at what julia said.

berndbischl:
this seems to be handled in PR https://github.com/mlr-org/mlr/pull/991

Issue #654, opened 2016-01-09
Incorrect plots by `plotThreshVsPerf` and `plotROCRCurves`


schiffner:
`plotThreshVsPerf` and `plotROCRCurves` show incorrect plots for `ResampleResult`s (and thus also for `list`s of `ResampleResult`s and `BenchmarkResult`s) if the curves are not aggregated. Instead of separate curves for the resampling iterations one very wiggly line is drawn.

```
library(mlr)

lrn = makeLearner("classif.lda", predict.type = "prob")
r = resample(lrn, sonar.task, cv3, measures = list(fpr, tpr, mmce))

d = generateThreshVsPerfData(r, measures = list(fpr, tpr, mmce), aggregate = FALSE)
plotThreshVsPerf(d) # not ok
plotROCCurves(d) # ok

d = generateROCRCurvesData(r, avg = "none")
plotROCRCurves(d) # not ok
```

zmjones:
I will look into this on Monday.

schiffner:
Thanks.

zmjones:
Sorry this took so long. I've now fixed the bug in the plot via `plotThreshVsPerf`. It was a somewhat subtle problem with the generation function and also a more obvious problem with the plotting functions.

Maybe we should now remove the dependency on `ROCR` though. Thoughts? I think it is weird that we rely on external measures via `ROCR` and then have our own set in mlr.

larskotthoff:
If we can remove the dependency easily, I'd be all for it.

zmjones:
As far as I am aware nothing depends on `ROCR` other than `generateROCRCurves` and `plotROCRCurves`. I will check all of this and push the fix and the dependency removal in separate commits. Maybe separate PRs as well?

larskotthoff:
Thanks, yes, separate PRs would be good if not too much hassle.

zmjones:
What are your thoughts on actually importing ggvis and shiny. Both are listed as imports in the description but are not actually loaded when mlr is attached. ggvis has one namespace conflict with another function "resolution" and so I was going to use `importFrom` and just import what is currently being used.

larskotthoff:
The latter sounds preferable (i.e. we only pull in what's necessary). Is that list still manageable or are we using so much stuff that it would become ridiculously long?

zmjones:
no it is fine i think

larskotthoff:
Ok cool then let's go for that.

schiffner:
[...]

PR #685, opened 2016-01-30
fixes issue #654

zmjones:
also added ggvis and shiny to imports on attach and fixed issues related to the ggvis threshvsperf function i also commented out the saving and temporary directory creating since we aren't testing the color of the saved svg files.


larskotthoff:
Thanks, looks good.

PR #686, opened 2016-01-30
remove ROCR plot/generate functions

zmjones:
removes ROCR for ROC curve generation and plotting, relying instead on generateThreshVsPerfData and plotThreshVsPerf/plotThreshVsPerfGGVIS and plotROCCurves. we still have ROCR in suggests because it is used for the AUC measure.


larskotthoff:
Thanks, looks good.

zmjones:
i guess i should have put this in news. would you call this an api change?

larskotthoff:
Yep, I'll update NEWS.

PR #689, opened 2016-01-30
fix typo and add test for threshold vs. performance plots (#654)

larskotthoff:


berndbischl:
This breaks travis, but the error seems unrelated. It occurs in tuneIrace.

```
1. Error: tuneIrace works with tune.threshold ----------------------------------
Assertion on 'exec.time' failed: All elements must be >= 0

```

A bit strange that the time seems to be < 0? Need to check locally

berndbischl:
I will restart travis now, to see what happens

berndbischl:
Hmm. The problem seems spurious. But I dont want this to break on CRAN. I will merge this but I want to remove the unit test based on exectime I guess, too

berndbischl:
Hmm. But it seems to be an assertion.  I will merge this now.

Issue #338, opened 2015-06-16
we should be able to tune randomForest just with the OOB estimator


berndbischl:
This is faster we do not need a CV here necesarrily. What other models provide that estimator?
Maybe we should generalize this.

@kerschke:  Relevant for gsoc.

jakob-r:
I suppose you are talking about the `se.method` from `regr.randomForest` right? So we should ad a `oobStandardError`?
At this opportunity a closer look at the old StandardError functions would be nice as well.

berndbischl:
> I suppose you are talking about the se.method

No, I am not, I am talking about evaluating the RF with its internal OOB estimator. This can be used for faster model selection

jakob-r:
I am still not totally clear about the internal OOB estimator.
My naive way would be
```r
oobStandardError = function(.learner, .model, .newdata, ...) {
  rf.preds = predict(.model$learner.model, newdata = .newdata, predict.all = TRUE)
  mean.responses = rf.preds$aggregate
  sd.responses = apply(rf.preds$individual, 1, sd)
  return(cbind(mean.responses, sd.responses))
}
```

zmjones:
you just mean that instead of doing further resampling on any learner which generates an OOB error estimate to just use that? i'd guess there are many methods other than rf that use resampling internally and can do this. that would indeed be much faster

berndbischl:
Yes, while you are tuning your params, instead of doing a crossval for each point evaluation, you just train the model. Then access its OOB estimator, which is in the learner.model. You return this as the performance measure. 

Actually this can already be done, by simply using Holdout, with all data for training, and coding your own perfmeasure, which accesses the learner.model oob info and returns this. 

tuneParams would then work without any changes.

berndbischl:
Here is some code that might help
```
library(mlr)

oobperf = makeMeasure("oobperf", minimize = TRUE, properties = c("classif", "classif.multi"),
  fun = function(task, model, pred, extra) {
  err = model$learner.model$err.rate
  err[nrow(err), "OOB"]
})

ps = makeParamSet(
  makeIntegerParam("ntree", lower = 1L, upper = 100L),
  makeIntegerParam("mtry", lower = 1L, upper = 4L)
)

ctrl = makeTuneControlGrid(resolution = 3)
lrn = makeTuneWrapper("classif.randomForest", par.set = ps, control = ctrl, measures = oobperf,
  resampling = makeResampleDesc("Holdout", split = 0.99))

[...]

PR #693, opened 2016-02-04
new function getOutOfBag, related to issue #338

florianfendt:
This is not really finished, since I'm still not sure if we should rather add 2 functions
````getOOBPredictions```` and ````getOOBError````.
Right now ````getOutOfBag```` returns both predictions and error rates in a list.
Opinions?


zmjones:
Just do the prediction extraction. Then we can apply whatever measure we
like.

berndbischl:
1) Lets be more specific with property names. lets use property name "oobpreds".

berndbischl:
Actually lets wait here. We need to decide what to to get probs. We need to define the API here first.

PhilippPro:
> what to to get probs
What do you mean with that? 
In Classification case probabilities would be nicer so we can calculate any measure. 
These are directly provided at least by rfsrc, ranger and randomForest. 
But maybe it would be nice to have also the opportunity to get only the bare predictions.

PhilippPro:
Ok, in my opinion, the result of the "getOutOfBag" function should be of the same format like the result of the "predict" function (could be with probabilities or not), so the same things (e.g. the function "performance") can be applied on this object.

larskotthoff:
@PhilippPro @berndbischl What's the status here?

berndbischl:
1) this needs to be transformed into a branch so we can all work on it

2) for the unit tests, we need a unit test helper function like in helpers_learners_all

berndbischl:
3) we really dont want to compute the oob error manually in each getOOBfunction, simply return the predictions

4) we also need a way to return OOB probabilities i guess

Issue #700, opened 2016-02-09
plotThreshVsPerf


schiffner:
Sorry, found another problem.

In the below plots there are separate lines drawn for each CV iteration, but the starting and end points are also connected, see example:

```
lrn1 = makeLearner("classif.lda", predict.type = "prob")
lrn2 = makeLearner("classif.ksvm", predict.type = "prob")
lrns = list(lrn1, lrn2)
rdesc = makeResampleDesc("CV", iters = 3)

## benchmark result
res = benchmark(lrns, tasks = sonar.task, resampling = rdesc, measures = list(mmce, auc), show.info = FALSE)
df = generateThreshVsPerfData(res, measures = list(fpr, tpr, mmce), aggregate = FALSE)
plotThreshVsPerf(df)

## list of resample results
r1 = resample(lrn1, sonar.task, rdesc, measures = ms, show.info = FALSE)
r2 = resample(lrn2, sonar.task, rdesc, measures = ms, show.info = FALSE)
df = generateThreshVsPerfData(list(r1, r2), measures = list(fpr, tpr, mmce), aggregate = FALSE)
plotThreshVsPerf(df)
```

@zmjones: Sorry, I'm playing around with the ROC tutorial page and thus found a number of issues at once. :(  Let me know if you want help here or for #699.

zmjones:
what is the `ms` measure. mmce?

zmjones:
well that looks weird. never seen that before.

schiffner:
> what is the ms measure. mmce?

Oops. I used `ms = list(auc, mmce)`, but I guess it doesn't matter.

Issue #699, opened 2016-02-09
plotROCCurves


schiffner:
Here is some nitpicky stuff I noticed about `plotROCCurves`:

* The docs for `obj` are wrong. It's an object of class `ThreshVsPerfData`.

* The docs for `measures` are misleading.
   It gives the impression that passing one single `Measure` is ok, although this makes no sense.
   It should say that you can only plot `measures` that were passed to `generateThreshVsPerfData`.

* The arg check that `measures` is a `list` of length 2, gives a confusing error message:

```
n = getTaskSize(sonar.task)
train.set = sample(n, size = round(2/3 * n))
test.set = setdiff(seq_len(n), train.set)

lrn1 = makeLearner("classif.lda", predict.type = "prob")
mod1 = train(lrn1, sonar.task, subset = train.set)
pred1 = predict(mod1, task = sonar.task, subset = test.set)

d = generateThreshVsPerfData(pred1, measures = list(mmce, fpr, tpr))
plotROCCurves(d, measures = mmce)
# Error in plotROCCurves(d, measures = mmce) : 
#   Assertion on 'measures' failed: Must have length 2, but has length 10
```

schiffner:
For threshold values exactly 0 or 1 some measure values will be `NA` / `NaN`.
At the moment `plotROCCurves` can't cope with that.

```
n = getTaskSize(sonar.task)
train.set = sample(n, size = round(2/3 * n))
test.set = setdiff(seq_len(n), train.set)

lrn1 = makeLearner("classif.lda", predict.type = "prob")
mod1 = train(lrn1, sonar.task, subset = train.set)
pred1 = predict(mod1, task = sonar.task, subset = test.set)
df = generateThreshVsPerfData(pred1, measures = list(ppv, tpr, tnr))
plotROCCurves(df)
# Error in if (diagonal & all(sapply(obj$data[, names(measures)], max) <=  : 
#   missing value where TRUE/FALSE needed

plotThreshVsPerf(df) # works

mlr:::measurePPV
# function (truth, response, positive) 
# {
#     measureTP(truth, response, positive)/sum(response == positive)
# }
```

schiffner:
A related question: `generateROCRCurvesData` etc. have been completely removed, right? I'm not sure what we agreed on. Shouldn't they be deprecated first?

zmjones:
Ok will fix the docs and will look into the other issue.

[...]

PR #701, opened 2016-02-09
fixes plotting issues in #699 and #700

zmjones:


schiffner:
Thank you very, very much for the quick fixes!!!
I installed this branch and all the tutorial examples run fine now.

Regarding #699 I noticed 2 mini things:

The axis labels in the case `pretty.names = FALSE` are not ok.
I guess the `obj$` in this line has to be removed, too:
https://github.com/zmjones/mlr/commit/82170df59302cb8bee14912c5265e47adba6d404#diff-51ee01993c76107746e819bd80050f89R371

In the docs of `plotROCCurves` there is a bracket missing:
https://github.com/zmjones/mlr/commit/82170df59302cb8bee14912c5265e47adba6d404#diff-51ee01993c76107746e819bd80050f89R329

Regarding #700 everything works fine.

zmjones:
thanks for checking it again, should be fixed now. i removed the bracket, unless you think the link to Measure should be bracketed. that did look consistent with how it is documented in arg_measures.R.

schiffner:
Thank you very much again and sorry that it took me a while to respond.

> i removed the bracket, unless you think the link to Measure should be bracketed. that did look consistent with how it is documented in arg_measures.R.

Really sorry to bother you with this, but the style guide is in favor of brackets. As far as I see
[arg_measures.R](https://github.com/mlr-org/mlr/blob/master/man-roxygen/arg_measures.R) has them, too. How about something like `[\code{list(2)} of \code{\link{Measure}}]\cr  all further explanations ...` ?

The only things open in my opinion are:
* The NA problem (https://github.com/mlr-org/mlr/issues/699#issuecomment-181937951), but there is no rush.
* Probably tests, but that does't seem to be easy and can also be done later.

@berndbischl: Is it ok if I merge once this is ready? Or does it interfere with the release?

zmjones:
No worries of course and it is no bother! I agree with you on the docs and I just pushed that change.

Further testing is on my list. The SVG tests would have caught this. I don't remember if you were on the hangout when we discussed it but @larskotthoff and I thought that the best thing to do would be to extract the default colors from the ggplot default theme, turn those into html color codes, and then insert those into the tests, rather than the hard-coded html color codes that are in there now (and broke with the latest ggplot2 release). However it is probably the case that they won't change again, so maybe this is much ado about nothing.

I forgot about the NA thing and will look into that now.

zmjones:
That is fixed now too,

berndbischl:
I will merge this in now.

- Can you close some of the issues that were dealt with here? If mini-points remain open, open up new precise ones. 

- Zach, can you please get together with Lars so we have tests again? this seems to become a problem,

schiffner:
@zmjones : Very cool, thank you so much.

@berndbischl : Both 699 and 700 are fixed completely, I will close them.

PR #704, opened 2016-02-11
re-enable hardcoded ggplot2 tests

zmjones:


berndbischl:
waiting for travis.
@zmjones if it passes, pls merge yourself

zmjones:
will do

larskotthoff:
@zmjones So the old colour definitions work again? I'm sure that they broke at some point.

zmjones:
yea they do. i guess they changed the defaults back?

larskotthoff:
Hmm ok. We'll just have to watch this space to see if they break again.

@berndbischl I don't think that disabling these tests is actually a huge drawback, as we will still find things like errors because of changed parameters without them.

zmjones:
Well the last two issues would have been caught by the tests but passed when they were disabled.

larskotthoff:
Which issues are that? #699 and #700? It doesn't sound to me like this would have been caught by checking the number of items, but maybe I'm misreading this?

zmjones:
Well it would have caught #700 anyway. There were some extra lines being drawn there which we do check.

berndbischl:
i dunno. I am much in favor of reenabling the tests, it there is an easy way to do this

PR #709, opened 2016-02-12
Fix #707: NoFeaturesModel when using ModelMultiplexer.

### LINKED ISSUE #707: ModelMultiplexer fails when learning from task with no features
When a ModelMultiplexer model is generated from a task with no features and then used to predict, an error is thrown:
```R
> emptyTask = makeClassifTask("emptyTask", data.frame(x=factor(c("a", "b"))), target='x')
> ml = makeModelMultiplexer(list(makeLearner("classif.probit")))
> tml = train(ml, emptyTask)
> predict(tml, emptyTask)
Error in UseMethod("isFailureModel") :
  no applicable method for 'isFailureModel' applied to an object of class "NULL"
```

mb706:


berndbischl:
thx, well spotted. 

This what you do here, is actually part of the structure in mlr I really hate.
quoting from ChainModel.R 
```
getLearnerModel.BaseWrapperModel = function(model, more.unwrap = FALSE) {
  # FIXME: this structure and special-cases really suck. FailureModel and NoFeaturesModel
  # should probably be redesigned at some point
  if (inherits(model$learner.model, "NoFeaturesModel"))
```

Am I right to assume that you got the idea how to fix this from BaseWrapper? 
Because I would suppose we fix this further up in BaseEnsemble, so this works for all inheriting Wrappers down the line.

berndbischl:
I looked at the code, in mlr im general. I guess your fix might be best, I dont see a much better way for now to put this in BaseEnsemble.  Also, the only other derived class from that is StackedLearner.
Which seems to work when run on a task with 0 features. 
We should still make sure that a unit test exists as well.

mb706:
The isFailureModel change is inspired by BaseWrapper, the getLearnerModel seemed liked the smallest additional change to solve this problem.

I don't know the class structure in mlr very well. My impression based on ModelMultiplexer is that everything that is supposed to happen inside trainLearner() and predictLearner() can easily be "virtualized" by just wrapping these functions, but everything that happens outside needs to be considered separately. NoFeaturesModel is such an outside-case, since it is caught by predict.Learner (with a dot) before predictLearner (without a dot) is called. Another case is FailureModel: If trainLearner.ModelMultiplexer itself throws an error then isFailureModel.ModelMultiplexerModel will also fail since model$learner.model$next.model won't exist (as model _itself_ will be a FailureModel) (I'll write the bug report shortly).

PR #710, opened 2016-02-12
Fix #708: omitted parameter in {classif,regr}.glmboost

### LINKED ISSUE #708: classif.glmboost parameters `trace` and `stopintern` lead to error when given.
The learner `classif.glmboost` has the parameters `stopintern` and `trace` in its ParamSet, but gives an error when these are given.
```R
> gb = makeLearner("classif.glmboost", trace=TRUE)
> train(gb, pid.task)
Error in mboost_fit(bl, response = response, weights = weights, control = control,  :
  unused argument (trace = TRUE)
Timing stopped at: 0.013 0 0.02
```
```R
> gb = makeLearner("classif.glmboost", stopintern=FALSE)
> train(gb, pid.task)
Error in mboost_fit(bl, response = response, weights = weights, control = control,  :
  unused argument (stopintern = FALSE)
Timing stopped at: 0.007 0.001 0.008
```

mb706:
I didn't write a test because it is my impression that arbitrary / mostly unimportant hyperparameters don't get tested anyways. Please tell me if there is an infrastructure that sets all hyperparameters to their default value and compares to the result of no given hyperparameter.

### DIFF
--- a/R/RLearner_classif_glmboost.R
+++ b/R/RLearner_classif_glmboost.R
@@ -25,8 +25,8 @@ makeRLearner.classif.glmboost = function() {
 }
 
 #' @export
-trainLearner.classif.glmboost = function(.learner, .task, .subset, .weights = NULL, mstop, nu, m, risk, ...) {
-  ctrl = learnerArgsToControl(mboost::boost_control, mstop, nu, risk)
+trainLearner.classif.glmboost = function(.learner, .task, .subset, .weights = NULL, mstop, nu, m, risk, stopintern, trace, ...) {
+  ctrl = learnerArgsToControl(mboost::boost_control, mstop, nu, risk, stopintern, trace)
   d = getTaskData(.task, .subset)
   if (.learner$predict.type == "prob") {
     td = getTaskDescription(.task)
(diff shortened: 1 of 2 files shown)


berndbischl:
1) If params were just forgotten and you add them, you maybe dont need to have the test.
I would write a test if something went wrong in the code with a specific setting. 
See an example here were I added such a test
 https://github.com/mlr-org/mlr/tree/fix_xgboost

2) We dont have such an infrastructure that you asked about. Acuatlly that might be a good idea. Do you want to open a issue for that please? Maybe even add a test?
(We host "vaguely similar" stuff in the "_learners_" subgroup of the unit tests

berndbischl:
Do these params also exist for regr.glmboost and surv.glmboost (would guess: yes)
Could you add these too pls?

If done, we can merge when 2.8 is on CRAN (I already submitted)

mb706:
I had fixed regr.glmboost in the same commit already, I guess the commit message was subpar, sorry about that. Should I change it now & push -f?

The ParamSet of surv.glmboost doesn't contain the parameters in question, so no bug here. (Maybe one could add them for consistency's sake but I'm not familiar with the learner to judge how much sense that would make.)

Issue #714, opened 2016-02-12
training regr.randomForest fails for data frame tbl. (dplyr)


ja-thomas:
When training `"regr.randomForest"` with an `tbl.df` data.frame from dplyr, the `train()` function crashes

mwp:
```
library(mlr)
data("mpg", package = "ggplot2")

#stupid factor preproc
mpg$model <- NULL
mpg$year <- as.factor(mpg$year)
mpg$manufacturer <- as.factor(mpg$manufacturer)
mpg$trans <- as.factor(mpg$trans)
mpg$fl <- as.factor(mpg$fl)
mpg$class <- as.factor(mpg$class)
mpg$drv <- as.factor(mpg$drv)


task_mpg <- makeRegrTask(data = mpg, target = "cty")
forest_learner <- makeLearner("regr.randomForest")


train(forest_learner, task_mpg)
```

```
Error in randomForest.default(x = data[["data"]], y = data[["target"]],  : 
  length of response must be the same as predictors
In addition: Warning messages:
1: drop ignored 
2: In randomForest.default(x = data[["data"]], y = data[["target"]],  :
  The response has five or fewer unique values.  Are you sure you want to do regression?
```

Seems to have something to do with the `[,drop=TRUE]` that is ignored by `tbl.df` data frames.

zmjones:
a `tbl.df` is not quite a regular `data.frame` for the reason you suggest. Cast your data to an actual `data.frame` in the task creation bit and it will work as expected. As you can [see](https://github.com/mlr-org/mlr/search?q=dplyr&type=Issues&utf8=%E2%9C%93) we have talked about this a bit and as far as I know we should be at least checking this.

berndbischl:
Its basically the same as in here
https://github.com/mlr-org/mlr/pull/253
https://github.com/mlr-org/mlr/issues/273

I guess what we should do for now, we call as.data.frame and warn about the conversion?

jakob-r:
Done in PR #723, waiting for Travis and Bernds approval

berndbischl:
just so you dont do it again, please add a space after "if" and before "("
sorry :)

Issue #273, opened 2015-03-31
dplyr drop generates bug in predict()


zmjones:
dplyr does not drop a one_column `data.frame` which also inherits from `tbl_df` to a vector, which generates a bug in [`predict`](https://github.com/berndbischl/mlr/blob/43ffa9bf994b956d72b13aa304e52b66b496c370/R/predict.R#L74-82).

see https://github.com/berndbischl/mlr/pull/253 and https://github.com/hadley/dplyr/issues/587 as @dickoa mentioned. as @berndbischl noted, it is unclear if this should be handled inside mlr or if it should just be checked for. the near silent failure (only the warning)  it generates with the below example is not great though.

```{r}
library(dplyr)
test = data_frame(x = 0:5, y = 5:10)
str(test[, 1, drop = TRUE])
## Classes 'tbl_df' and 'data.frame':	6 obs. of  1 variable:
## $ x: int  0 1 2 3 4 5
## Warning message:
## drop ignored 
> str(test[[1]])
## [1] 0 1 2 3 4 5
```

berndbischl:
I dont really care about this now. Reopen when it becomes important.

bkomboz:
I just encountered this issue (error from mlr that truth column is not there). It took me some time to figure out (mostly by chance) that this is an issue due to dplyr. Using `as.data.frame(task_data)` before creating a task solved the issue but it would be nice if this is somewhere documented.

larskotthoff:
@bkomboz Thanks for the report -- could you provide a complete example please? This may be something that we should check in mlr.

zmjones:
I think this is the same issue that I documented above (and in #253). As I recall @berndbischl argued for not doing anything about it because a `tbl_df` is not (simply) a `data.frame` and all of the mlr infrastructure which uses the data element of the task (in particular `performance`, here) expects this element to behave like a `data.frame`.

berndbischl:
Yes. But we should make life as simple as possible for users. We should really have an FAQ.
Also, dplyr is becoming so popular that we should either directly support it or add some check.

larskotthoff:
It also sounds like the problem occurred somewhere during tuning, i.e. deep within mlr and that obscures the true cause of the problem.

berndbischl:
As we dont really use and explot dplyr in mlr (yet), shouldnt we simply "as.data.frame" the data on task creation, if we are passed a dplyr object? and everything is fine?

larskotthoff:
Sounds good to me.

[...]

PR #723, opened 2016-02-15
Convert data if necessary (Fixes Issue #714, #253, #273)

jakob-r:


berndbischl:
just so you dont do it again, please add a space after "if" and before "("
sorry :)

berndbischl:
also we had a bug here. but you added no unit test. can you please do that?

jakob-r:
I am not so sure whether we had a bug or if dplyr is to blame for giving us a `data.frame` class which does not implement all the necessary methods.

However: We neither have `dplyr` in depends, imports nor suggests. So what is a reasonable way to test it?

berndbischl:
Ah, damn. I guess it does not hurt if we add it to suggests. We might use it anyway soon.

jakob-r:
cleanup and test added, your turn again

PR #727, opened 2016-02-16
skip some tests if a package is not available

mllg:


berndbischl:
I guess you know what you are doing. Merging

PR #738, opened 2016-02-17
disable test due to travis / cran problems

berndbischl:
keep your filthy hands away from this PR, just for repairing travis


berndbischl:
current state: 
- we need a issue to enable the test later again for PPWrapperCaret
- now bartmachine fails, we need to skip_if_not_installed in the learners_all tests

mllg:
Suggestion: Extend listLearners in a way to filter for learners whose required packages are installed.

larskotthoff:
Is this still relevant or can it be closed?

Issue #602, opened 2015-12-08
randomForestSRCSyn learners are broken


larskotthoff:
Tests fail:

    1. Error: classif_randomForestSRCSyn -------------------------------------------
     dim(X) must have a positive length
     1: withCallingHandlers(eval(code, new_test_environment), error = capture_calls, message = function(c) invokeRestart("muffleMessage"))
    2: eval(code, new_test_environment)
    3: eval(expr, envir, enclos)
    4: apply(mod$predicted, MARGIN = 1L, which.max) at test_classif_randomForestSRCSyn.R:21
    5: stop("dim(X) must have a positive length")

etc.

randomForestSRC package got updated, not obvious to me what changed.

berndbischl:
Like I said in the PR where we noticed the error first:

We predict on one observation, I guess. Now, in the new package version of rfSRCSyn, a the dimension gets dropped and we get a single vector instead of a df / matrix. Therefore our apply breaks.

It is annoying, but the fix is a simple "if" case.

berndbischl:
I will do this now

berndbischl:
I look at the code. They basically did a renaming of some slots.

BUT: The orig. code is really strange. We apparently cannot predict  predict on newdata, if newdata does not contain the target label col. So the person wrapping this, actually TRAINED the model during predict. 
I am very close to voting for learner removal as I currently cannot see how this can be properly integrated.

Can I get more opinions fast pls?

berndbischl:
Here code to look at:

```
library(mlbench)
require("randomForestSRC")
load_all()

ring = data.frame(mlbench.ringnorm(25, 2))
ringSyn1 = rfsrcSyn(classes ~., data = ring)
nd = ring[, 1:2]
p = rfsrcSyn(object = ringSyn1, newdata = nd)$rfSynPred
print(str(p))
```

berndbischl:
I will remove the learners to todofiles now, so travis runs again

schiffner:
I only had a short look:
With the previous version 1.6.1 you actually *could* predict on newdata *without* the target column. (So there was really no need to train the model in predictLearner in the first place).
I guess it's time for a bug report if we want this solved properly.

berndbischl:
@schiffner Would you do this and ask them about this?

schiffner:
I feared that you would say this. :)
I can ask them.

berndbischl:
I fixed more SRC problems.

A side issue is that we have hardcoded args again:
```
[...]

PR #741, opened 2016-02-18
clean up randomForestSRC learners (this is related to #602)

schiffner:
I basically did the stuff I listed [here](https://github.com/mlr-org/mlr/issues/602#issuecomment-171127957), i.e. added some missing parameters (most of minor importance), added dependencies, changed what parameters are hardcoded.

Questions regarding the `par.set`:
* Is this construction ok to cope with mixed logical and character options?

```
makeDiscreteLearnerParam(id = "var.used", default = FALSE, tunable = FALSE,
  values = list(`FALSE` = FALSE, "all.trees", "by.tree"))
```

* I'm not sure how to set the limits for `nodedepth`:
  Negative values of `nodedepth` mean that this parameter is ignored.
  0 corresponds to no splits, 1 results in a stump etc.
  So with regard to param checks allowing all negative integer values is ok.
  For tuning setting `lower = -1L` would be a better idea.
* Does anyone know what `split.null` does? I included it in the `par.set` for completeness sake, but I have no idea what it does, though I tried to find out.

train:
* `importance = "none"` was hardcoded.
  That doesn't hurt much because the importance can also later be extracted from the fitted
  tree by function `vimp`.
  But in case that someone wants to calculate the importance directly, I added it to the `par.set`,
  changed the default from `"permute"` to `"none"` in `par.vals` to save time and added a `note`.
* `proximity = FALSE` was hardcoded. In constrast to the importances this cannot be easily
  retrieved later on.
  I added it to the `par.set` and it defaults to `FALSE` anyway.
* I left `forest = TRUE` hardcoded and removed it from the `par.set`.
  (If `forest = FALSE` no fitted forest is returned and one can't make predictions.)

predict:
* `importance` was hardcoded, but this is not really necessary. It is ignored anyway
  because `.newdata` does not contain the target variable.
* I hardcoded `membership = FALSE` because it defaults to `TRUE`, but we ignore the calculated
  membership values anyway.

### DIFF
--- a/R/RLearner_classif_randomForestSRC.R
+++ b/R/RLearner_classif_randomForestSRC.R
@@ -9,35 +9,48 @@ makeRLearner.classif.randomForestSRC = function() {
         values = c("by.root", "by.node", "none")),
       makeIntegerLearnerParam(id = "mtry", lower = 1L),
       makeIntegerLearnerParam(id = "nodesize", lower = 1L, default = 1L),
-      makeIntegerLearnerParam(id = "nsplit", default = 0L),
+      makeIntegerLearnerParam(id = "nodedepth", default = -1L),
+      makeDiscreteLearnerParam(id = "splitrule", default = "gini",
+        values = c("gini", "gini.unwt", "gini.hvwt", "random")),
+      makeIntegerLearnerParam(id = "nsplit", lower = 0L, default = 0L,
+        requires = quote(splitrule != "random")), # nsplit is ignored and internally set to 1L for splitrule = "random"
+      makeLogicalLearnerParam(id = "split.null", default = FALSE),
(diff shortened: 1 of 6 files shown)


schiffner:
I rebased and squashed.
There is only one minor conflict with [this commit](https://github.com/mlr-org/mlr/commit/498815d903d13b99df41c67579e7fb15bd59dd47) that can easily be resolved.

larskotthoff:
Thanks Julia. Can this be merged now?

schiffner:
If you are ok with my changes, then yes. :)
Locally the tests pass, so I hope everything is ok.

larskotthoff:
Looks good to me. I'll merge it.

schiffner:
Thx.

PR #742, opened 2016-02-19
SMOTEWrapper: nn param is not being passed to `smote` function

tijoseymathew:
The number of nearest neighbors to be considered `nn` is now passed from `SMOTEWrapper` to `smote`.
I have an additional concern, shouldn't the `...` dots in the wrapper be passed to the `train` function? For additional learner parameters? However, this causes some unit test to fail.


berndbischl:
thx. I fixed this locally. and added a unit test to spot this.

> , shouldn't the ... dots in the wrapper be passed to the train function?
no, thats actually ok, the params are "peeled" of by the wrapper from the par.val member variable

PR #752, opened 2016-02-19
Fix #735, #755: Add requirements to params of RWeka learners, change defaults to FALSE

### LINKED ISSUE #735: Set classif.PART & classif.J48 parameter requirements
WEKA checks some of the parameters for these algorithms and throws an exception when they are conflicting. This should be reflected in the mlr ParameterSet for these learners.

For **classif.PART**, an error is thrown when:
- `R` is set and `C` is given
- `N` is given and `R` is not set

For **classif.J48**, an error is thrown when:
- `U` and `S` are both set
- `U` and `R` are both set
- `U` is set and `C` is given
- `R` is set and `C` is given
- `N` is given and `R` is not set

Also, for both, the feasible range for `C` is (0, 1) (exclusive!), not [0, Inf).

Another suggestion, it is my impression that all logical parameters default to `FALSE`, so it would make sense to set the `$default` properties if the Params to `FALSE`.

mb706:
... and also mark the logical parameters as default FALSE (since they are).


jakob-r:
Short questions: 
* The defaults you put are documented in Weka?
* Because you use `!u`, we don't run into problems you mentioned [here](https://github.com/berndbischl/ParamHelpers/issues/62). ?

mb706:
- The documentation is not entirely clear, but I would say it is implied by the fact that the flags are set to TRUE if the option (e.g. -U) is given and FALSE when absent. FALSE is the 'default' in the sense that it is the value used when the option is not given.
- `quote(!U)` works since internally it is the same as ``quote(`!`(U))`` (function call); only `quote(U)` would not work.

jakob-r:
Then I thank you for your research and from my side we can merge it.

mb706:
Updated to address #755

larskotthoff:
Thanks, merging.

PR #763, opened 2016-02-26
Add Multilabel Wrapper: Classifier Chains

QuayAu:


R/MultilabelClassifierChainsWrapper.R
```diff
@@ -0,0 +1,97 @@
...
+  chained_targets = targets
+  for (tn in targets[order]) {
+    data2 = dropNamed(data, setdiff(chained_targets, tn))
+    index = which(names(data2) %in% setdiff(targets, tn))
+    if (length(data2[, index]) != 0) {
+      data2[, index] = sapply(data2[, index], as.numeric)
+    }
+    ctask = makeClassifTask(id = tn, data = data2, target = tn)
+    models[[tn]] = train(.learner$next.learner, ctask, weights = .weights)
+    chained_targets = setdiff(chained_targets, tn)
```
PhilippPro: take this line to the beginning of the for loop, so it has not to be computed in lines 68 and 69 again

PR #764, opened 2016-02-26
Add Multlabel Wrapper: Stacking

QuayAu:
Add stacking (STA) multi label wrapper (from http://www.sciencedirect.com/science/article/pii/S0031320313004019)

In the training stage, STA trains two sets of learners:

    the level 1 learners
    the meta level learners

The level 1 learners are the same as the binary relevance learners. Simple binary classificator are trained for each label. The level 1 learners are used to produce predicted labels for augmenting the feature space at prediction time.

The meta level learners model the label dependencies by training classifiers on augmented feature spaces. For each label, all other labels (including the target label itself) are augmenting the feature space. The difference to DBR is that at training time predicted label information is used.
Predicted labels are obtained through an internal cross validation step (using the level 1 learner). Default is 2-fold-CV (again, this was arbitrarily chosen by me).


R/MultilabelStackingWrapper.R
```diff
@@ -0,0 +1,106 @@
...
+    function(m) as.logical(getPredictionResponse(predict(m, newdata = .newdata, ...)))
+  } else {
+    function(m) getPredictionProbabilities(predict(m, newdata = .newdata, ...), cl = "TRUE")
+  }
+  if (.learner$level1Learner$predict.type == "response") {
+  predLvl1 = sapply(data.frame(asMatrixCols(lapply(modelsLvl1, f))), as.numeric)
+  } else {
+  predLvl1 = data.frame(asMatrixCols(lapply(modelsLvl1, f)))
+  }
+  colnames(predLvl1) = paste(.model$task.desc$target, ".1", sep = "")
```
giuseppec: make this better? If feature is called the same, we have a problem

R/MultilabelStackingWrapper.R
```diff
@@ -0,0 +1,106 @@
...
+#' lrn = makeMultilabelStackingWrapper("classif.rpart", level1Learner = "classif.rpart")
+#' lrn = setPredictType(lrn, "prob")
+#' # train, predict and evaluate
+#' mod = train(lrn, task)
+#' pred = predict(mod, task)
+#' p = performance(pred)
+#' performance(pred, measure = hamloss)
+#' getMultilabelBinaryPerformances(pred, measures = list(mmce, auc))
+#' # above works also with predictions from resample!
+makeMultilabelStackingWrapper = function(learner, level1Learner = learner, cv.folds = 2) {
```
PhilippPro: Should we provide different learner for the two phases? I think rather not (and if yes we should maybe also do it for nested stacking)

R/MultilabelStackingWrapper.R
```diff
@@ -0,0 +1,106 @@
...
+  x$cv.folds = cv.folds
+  return(x)
+}
+
+#' @export
+trainLearner.MultilabelStackingWrapper = function(.learner, .task, .subset, .weights = NULL,...) {
+  targets = getTaskTargetNames(.task)
+  .task = subsetTask(.task, subset = .subset)
+  data = getTaskData(.task)
+  # train level 1 learners
```
PhilippPro: maybe use the MultilabelBinaryRelevanceWrapper for the process

R/MultilabelStackingWrapper.R
```diff
@@ -0,0 +1,106 @@
...
+  makeHomChainModel(.learner, c(modelsLvl1, modelsMeta))
+}
+
+#' @export
+predictLearner.MultilabelStackingWrapper = function(.learner, .model, .newdata, ...) {
+  models = getLearnerModel(.model, more.unwrap = FALSE)
+  # Level 1 prediction (binary relevance)
+  modelsLvl1 = models[1:length(.model$task.desc$target)]
+  f = if (.learner$level1Learner$predict.type == "response") {
+    function(m) as.logical(getPredictionResponse(predict(m, newdata = .newdata, ...)))
```
PhilippPro: would be easier with MultilabelBinaryRelevanceWrapper

PR #765, opened 2016-02-26
Add Multilabel Wrapper: Dependent binary relevance

QuayAu:
Add dependent binary relevance (DBR) multi label wrapper (from http://www.sciencedirect.com/science/article/pii/S0031320313004019)

Like the other multi label wrappers, the idea of this wrapper is the same:
Modeling label dependencies by augmenting the feature space with label information.

In the training stage, DBR trains two sets of learners:

    the level 1 learners
    the meta level learners

The level 1 learners are the same as the binary relevance learners. Simple binary classificators are trained for each labels. The level 1 learners are only used to produce predicted labels for augmenting the feature space at prediction time.

The meta level learners model the label dependencies by training classifiers on augmented feature spaces. For each label, all other labels (except the target label itself) are augmenting the feature space. At training time true label information is used.


jakob-r:
Hi,
I saw your numerous PR on Multilabel Wrappers. On a first glimpse it looks very well written, tested and documented so I appreciate your contribution a lot. I try to speak in the name of all mlr-devs (even though we have not discussed this topic so far): We are drowning in Pull Requests :ocean: as we are too slow and having to deal with all these Travis checks and everything. 

So here is my crazy proposal for which you are the chosen one: 
_Would you like to combine your methods in your very own R-package which extends mlr?_
This is not meant as an exclusion of your contribution but more as an honoring of your own work. I also choose your PRs because they pretty much stand for they own and do not tinker with the inner mlr methods.
So let me know what you are thinking. I will also try to be as helpful as I can in the process of making your package and I am quite sure we can promote R-packages for mlr in a special section of our github-page and in the tutorial.

berndbischl:
Jakob, Quay is my master student and all of what happens here is planned. we are just too slow with reviewing this - and it is not that urgent. Giuseppe and I will take care of this.

R/MultilabelDBRWrapper.R
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#' @export
+trainLearner.MultilabelDBRWrapper = function(.learner, .task, .subset, .weights = NULL,...) {
+  targets = getTaskTargetNames(.task)
+  .task = subsetTask(.task, subset = .subset)
+  data = getTaskData(.task)
+  modelsLvl1 = namedList(targets)
+  modelsMeta = namedList(targets)
+  # train level 1 learners (binary relevance)
```
PhilippPro: same as in stacking, maybe use BinaryRelevanceWrapper

R/MultilabelDBRWrapper.R
```diff
@@ -0,0 +1,95 @@
...
+#' lrn = makeMultilabelDBRWrapper("classif.rpart", level1Learner = "classif.rpart")
+#' lrn = setPredictType(lrn, "prob")
+#' # train, predict and evaluate
+#' mod = train(lrn, task)
+#' pred = predict(mod, task)
+#' p = performance(pred)
+#' performance(pred, measure = hamloss)
+#' getMultilabelBinaryPerformances(pred, measures = list(mmce, auc))
+#' # above works also with predictions from resample!
+makeMultilabelDBRWrapper = function(learner, level1Learner = learner) {
```
PhilippPro: Same as in stacking, do we need two learners? and maybe use MultilabelBinaryRelevanceWrapper?

PR #772, opened 2016-03-03
fix linear.output parameter for classif.neuralnet (fixes #736)

### LINKED ISSUE #736: classif.neuralnet parameter linear.output broken
Similar to #731, `linear.output` is set explicitly in `trainLearner.classif.neuralnet` and therefore leads to a bug if also given as hyperparameter:
```R
> train(makeLearner("classif.neuralnet", linear.output=TRUE), pid.task)
Error in neuralnet::neuralnet(f, data = taskdat, linear.output = FALSE,  :
  formal argument "linear.output" matched by multiple actual arguments
Timing stopped at: 0.007 0 0.017
```

larskotthoff:


mllg:
The formula stuff in this learner is ridiculous, but that is not your concern. The diff looks good.

larskotthoff:
Thanks, merging.

PR #776, opened 2016-03-03
fix travis

berndbischl:
so i can see if it runs now

### DIFF
--- a/R/getCaretParamSet.R
+++ b/R/getCaretParamSet.R
@@ -30,8 +30,8 @@
 #'   tuning parameters
 #' }
 #' @export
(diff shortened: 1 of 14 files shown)


berndbischl:
i have disabled the example it getCaretparamSet. We might convert this into a issue and copy the err msg there.

larskotthoff:
It looks like it's disabling the tests for the learners as well. Is that required?

larskotthoff:
Well, I've commented the tests back in and the previous error is gone. Now it fails because of missing adabag? Which is installed according to the log.

PR #778, opened 2016-03-03
fix double trafo

jakob-r:
unfortunately not all went smooth during refactoring, forgot to use the par.set without trafos


jakob-r:
don't merge for now!

PR #782, opened 2016-03-05
Fix #781: tuneControlRandom with only budget infers maxit from that.

### LINKED ISSUE #781: makeTuneControlRandom(budget=...) shouldn't give an error
Minor nitpick, but imho
```R
> makeTuneControlRandom(budget=20)
Error in makeTuneControlRandom(budget = 20) : 
  The parameters budget (20) and maxit (100) differ.
```
shouldn't be.

mb706:
Take it or leave it, but I like this more; it is also consitent with the behaviour of `makeTuneControlIrace` and `makeTuneControlGenSA`, which also assign the budget value to the corresponding other variable.

### DIFF
--- a/R/TuneControlRandom.R
+++ b/R/TuneControlRandom.R
@@ -3,15 +3,19 @@
 #'   Default is 100.
 #' @export
 #' @rdname TuneControl
-makeTuneControlRandom = function(same.resampling.instance = TRUE, maxit = 100L, tune.threshold = FALSE,
+makeTuneControlRandom = function(same.resampling.instance = TRUE, maxit = NULL, tune.threshold = FALSE,
   tune.threshold.args = list(), log.fun = NULL, final.dw.perc = NULL, budget = NULL) {
 
-  if (is.null(budget))
+  if (is.null(budget)) {
+    if (is.null(maxit))
+      maxit = 100L
     budget = maxit
-  else if (is.null(maxit))
-    maxit = budget
-  else if (budget != maxit)
-    stopf("The parameters budget (%i) and maxit (%i) differ.", budget, maxit)
+  } else {
+    if (is.null(maxit))
+      maxit = budget
+    else if (budget != maxit)
+      stopf("The parameters budget (%i) and maxit (%i) differ.", budget, maxit)
+  }
   maxit = asCount(maxit)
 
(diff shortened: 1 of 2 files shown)


jakob-r:
Why don't we simply just use budget and get rid off (aka deprecate) maxit?

R/TuneControlRandom.R
```diff
@@ -3,15 +3,19 @@
...
 
-  if (is.null(budget))
+  if (is.null(budget)) {
+    if (is.null(maxit))
+      maxit = 100L
     budget = maxit
-  else if (is.null(maxit))
-    maxit = budget
-  else if (budget != maxit)
-    stopf("The parameters budget (%i) and maxit (%i) differ.", budget, maxit)
```
jakob-r: Maybe `BBmisc::coalesce()` would do the trick with something like
```r
maxit = coalesce(budget, maxit)
if (is.null(budget) && budget != maxit)
  stopf(...)
```
but as i already mentioned I see maxit as totally useless here

mb706: [GenSA](https://github.com/mlr-org/mlr/blob/d7005d1408d8379258b9de8868cf98cf409ef555/R/TuneControlGenSA.R#L8-L21) and [IRace](https://github.com/mlr-org/mlr/blob/d7005d1408d8379258b9de8868cf98cf409ef555/R/TuneControlIrace.R#L21-L31) do it pretty much like this. What do you mean 'without having to change the defaults of makeTuneControlRandom'?

jakob-r: We could leave `maxit = 100L` like it is (if we still want to use maxit)

mb706: Ah, so you mean leaving it in the function head? If the function sees `maxit` as 100 and `budget` as 200 then, is there a way for it to detect whether the user made a mistake or whether he left out the `maxit` argument?

jakob-r: I wasn't making much sense. IMHO: Just remove the `maxit = 100L` and use `maxit = coalesce(budget, maxit, 100L)`. But maybe @berndbischl wants to say something.

berndbischl:
i fixed this slightly differently here #980

PR #785, opened 2016-03-09
new learner glinternet for classification and regression

schiffner:
`glinternet` is a penalized regression learner and can detect 2-way interactions.
The interface is very similar to `glmnet`.

* In contrast to `glmnet`, `glinternet` cannot interpolate when making predictions for
`lambda`s that were not used for training.
That's why I introduced a new parameter `predictLambda` which is mapped to the closest
value in the `lambda`-sequence from training and then passed to `predict.glinternet`.
Maybe long-term we cold do the interpolation ourselves.

* `glinternet` wants categorical variables as integers starting with 0.
I therefore extended `fixDataInfo` and `fixDataforLearner`.
There is now an additional option `factors.to.int` and a logical parameter `min.int.zero` indicating
if the integer values start at 0 or 1.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -79,6 +79,7 @@ Suggests:
     gbm,
     GenSA,
     glmnet,
+    glinternet,
     Hmisc,
     irace (>= 1.0.7),
     kernlab,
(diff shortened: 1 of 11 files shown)


schiffner:
Travis fail seems unrelated.

schiffner:
Closing this now because

* it's not really important,
* it currently relies on and extended version of `fixDataForLearner` that we want to get rid of longterm (
#926 ).

I can just send a new PR when we have solved #926.

berndbischl:
thx + agreed

PR #790, opened 2016-03-13
Re-add bartMachine

larskotthoff:


larskotthoff:
Need to wait for new bartMachine version to be on CRAN before merging this.

kapelner:
Is bartmachine support back in MLR? Or is there another problem? I don't see any issues open.

larskotthoff:
I think we're all set, but we're about to make a new mlr release and I'm not sure if we want to pull it in before then. @berndbischl ?

PR #805, opened 2016-03-28
some small corrections for randomForestSRC learners

schiffner:
* Added option `"na.omit"` for `na.action`
* Added parameter `ntime` for survival
* For survival from mlr's side there is an additional learner parameter `mtry.ratio` which indicates the *proportion*  (while `mtry` gives the *absolute number*) of variables considered for a split. Oddly, `mtry.ratio` is multiplied by `nrow(data)` (the number of *observations*) instead of the number of *features*. 
I changed that and for completeness sake added the `mtry.ratio` option also for regression and classification.


larskotthoff:
Thanks, merging.

Issue #803, opened 2016-03-26
Bug in `method = "hill.climb"` of `makeStackedLearner`


philippstats:
This code

```
 lrns = list(
    makeLearner("classif.ksvm", predict.type = "prob"),
    makeLearner("classif.randomForest", predict.type = "prob"),
    makeLearner("classif.kknn", id = "classif.knn2", predict.type = "prob", k = 2),
    makeLearner("classif.kknn", id = "classif.knn3", predict.type = "prob", k = 3),
    makeLearner("classif.kknn", id = "classif.knn4", predict.type = "prob", k = 4),
    makeLearner("classif.kknn", id = "classif.knn5", predict.type = "prob", k = 5))
m = makeStackedLearner(base.learners = lrns, predict.type = "prob", method = "hill.climb",
    parset = list(replace = TRUE, bagprob = 0.7, bagtime = 3, init = 2))
tmp = train(m, tsk)
```

throws that error message

```
Warning messages:
1: In weights[bagmodel] + weight :
  longer object length is not a multiple of shorter object length
2: In weights[bagmodel] = weights[bagmodel] + weight :
  number of items to replace is not a multiple of replacement length
```

`testthat` doesn't catch that because too few baselearner are used. I already implemented a clean version (incl. new tests).

philippstats:
This problem occurs if `init` in `makeStackedLearner` is larger than 0.

larskotthoff:
Thanks, good catch. Could you open a pull request with the fixed code please?

philippstats:
I know why the checks didn't pass. I will try it again soon.

philippstats:
I will do a PR some day after #808 is fixed as well. I need to ask @berndbischl how to solve 808 first.

philippstats:
Done in #1041

PR #807, opened 2016-03-29
Fix stackedLearner for method 'hillclimb'

philippstats:
Fix bug for issue #803.
Moreover this contains an improved test, new default parameters for method "hill.climb", fixes for parameter description, and better input checking.


philippstats:
I know why the checks didn't pass. I will try it again later when other issues are fixed, too.

PR #816, opened 2016-03-31
fixes #814 change permutation importance learner arg name

### LINKED ISSUE #814: Can not set the learner for permutation.importance filter
I understand that the permutation.importance filter requires a learner to be specified, like in this example:
```r
filterFeatures(pid.task, method = 'permutation.importance', abs = 5, learner = 'classif.logreg')
```

However, if I want to use to fuse a base learner with this filter, I cannot specify the learner for the permutation.importance because that argument name is already used for the base learner. I want to do something like this
```r
wrapped.lrn <- makeFilterWrapper(learner = 'classif.randomForest', fw.abs = 3,
   fw.method = 'permutation.importance', fw.learner = 'classif.logreg')
train(wrapped.lrn, pid.task)
```

zmjones:


larskotthoff:
Thanks, looks good.

Issue #703, opened 2016-02-10
reference ron kovavis paper for warpper feature selection


berndbischl:
http://ai.stanford.edu/~ronnyk/wrappersPrint.pdf

berndbischl:
done in PR by flow

PR #817, opened 2016-03-31
#703 reference ron kovavis paper in FeatSelControl

florianfendt:


berndbischl:
Hi, 

please if you do this, do very nice references with roxygen. see eg here

```
R/RLearner_regr_randomForest.R
14:#' @references [Joseph Sexton] and [Petter Laake],; [Standard errors for bagged and random forest estimators], Computational Statistics and Data Analysis Volume 53, 2009, [801-811].
```

The URL to the PDF of course is a nice extra. 

Maybe add 1-2 lines to the contributer wiki how a good example should look like when you are done and post feedback here

florianfendt:
Fixed the reference here.
I found out that it was actually already mentioned here https://github.com/mlr-org/mlr/wiki/mlr-Coding-Guidelines
under General Rules. I do think though, that this section is rather unstructured, as it contains lots of information on different topics.
Maybe we should put the two bullets about roxygen to the style guide page? (There's already a section about documentation on it)

jakob-r:
Looks good now

berndbischl:
thx!

PR #820, opened 2016-03-31
adds filter documentation for permutation importance

zmjones:
this is not ready for merging yet. is this the preferred doc structure for the filter constructors? if so i will go back and fill in some more of them. and travis generates the docs now right? so i should not generate/commit them?

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -3,7 +3,9 @@
 #' Create a feature filter.
 #'
 #' Creates and registers custom feature filters. Implemented filters
-#' can be listed with \code{\link{listFilterMethods}}.
+#' can be listed with \code{\link{listFilterMethods}}. Additional
+#' documentation for the \code{fun} parameter specific to each filter can
+#' be found in the description.
 #'
 #' @param name [\code{character(1)}]\cr
 #'  Identifier for the filter.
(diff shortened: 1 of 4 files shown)


berndbischl:
Talked to Zach on hangout. The docs look basically OK, but I dont know if @param works here.

zmjones:
I generated the docs last night. I can't see how to get each of the constructors to appear as its own section. The params just appear as if there was a "filter" function.

berndbischl:
That is what I thought would happen. 
Can you try to doc them "customly" in a reasonable way please?

zmjones:
ok so i think the best alternative is to put everything in the `makeFilter` doc page. In the description field we can have a list and for each item we describe the filter method and what additional arguments it takes. thoughts?

berndbischl:
Sound good.

zmjones:
Ok I've changed the docs in the manner I suggested. If you like it then I can do some more and put that in a separate PR. I added a test for the issue as well.

larskotthoff:
Is this ready to be merged?

zmjones:
what do you think about the way that i documented the permutation importance filter?

larskotthoff:
Reads fine to me.

zmjones:
then yep it can be merged

larskotthoff:
Great, thanks.

berndbischl:
thx a lot for doing this!
are you going to add more docs for the other filters? 
or do you need support here?

(we also should not accept stuff anymore in master without proper docs...)

zmjones:
i will issue another pr in the next day or two for some of the other filters. probably not all of them though.

berndbischl:
very nice, thx

Issue #716, opened 2016-02-13
subsetTask and getTaskData should also allow integer indices for features


berndbischl:
it is much nicer if we dont have to do this always
getTaskFeatureNames(task)[1:3]

berndbischl:
are there other functions to consider?

florianfendt:
@berndbischl 
Should I get rid of this FIXME in ````getTaskData()````:
````
#FIXME: argument checks currently not done for speed
````

berndbischl:
OK, do it.

berndbischl:
PR is here #825

PR #825, opened 2016-04-05
Enables integer indices for features in subsetTask and getTaskData (Issue #716)

florianfendt:
Please have a look at the diff in ````UnsupervisedTask````.
I had to make changes in there because my argument check for ````task```` in ````getTaskData```` wouldn't have worked otherwise. Hope I handled that right.


larskotthoff:
Thanks. I notice a few changes that seem unrelated, i.e. argument checks are now done that were disabled for speed according to the comments and you've changed one line to access task data directly rather than through `getTaskData()`. Is that intentional?

berndbischl:
> i.e. argument checks are now done that were disabled for speed according to the comments

Please always check the related issue thread. He asked about this....

larskotthoff:
Ugh. I'm starting to really dislike this separation of issues and PRs.

Anyway, the issue doesn't actually say anything substantial about it. Is speed no longer a concern? Is it not actually that much overhead?

berndbischl:
What seems wrong is the following:
a) you seem to allow logical indices. But this does is not documented?

b) you dont convert numeric integer vecs to integers with checkmates asInteger.

Shall we simply remove the logical stuff?

berndbischl:
@mllg  @larskotthoff @schiffner @jakob-r 

this is a "core" change. we agreed that 2 people review this then. i iterated over this with flo multiple times.

imho we can draw this in.

pls check

R/UnsupervisedTask.R
```diff
@@ -1,6 +1,6 @@
 makeUnsupervisedTask = function(type, data, weights = NULL, blocking = NULL, fixup.data = "warn", check.data = TRUE) {
   task = makeTask(type, data, weights, blocking, fixup.data = fixup.data, check.data = check.data)
-  checkTaskData(getTaskData(data))
+  checkTaskData(task$env$data)
```
jakob-r: how is this change motivated?

jakob-r:
I checked and added one line note. Besides that it looks okay to me.

R/UnsupervisedTask.R
berndbischl: read the PR thread? he explained this?

berndbischl: please really read the threads, issue + PR for stuff like this, to we dont waste time.

but more to the point: the object is not completely constructed at this point in time i guess. so the getter does not work. and here, as an exception, it should then be OK to directly access the slot.

errrrr: the previous line getTaskData(data) seeems to be complete bullshit? "data" is not a task here?

berndbischl:
@florianfendt 
For conistency, can you please make the same changes to "train" and "predict", the "subset" arg. 
and also add tests for this. thx.

R/UnsupervisedTask.R
jakob-r: >wouldn't have worked otherwise
is pretty vague. And as I am the second person to ask maybe a comment in the source would be usefull. But are we sure that `checkTaskData(getTaskData(task))` does not work? What would be the error?

berndbischl:
> But are we sure that checkTaskData(getTaskData(task)) does not work? What would be the error?

@florianfendt 
please comment the line, why the getter cannot be used here. so people reading the code know.

R/UnsupervisedTask.R
florianfendt: @jakob-r @berndbischl 
yes sorry I should have been more precise here.
````getTaskData(data)```` was simply a bug or a workaround of some sorts that worked before,
but did not do the trick anymore after I've done my changes to ````getTaskData()````.
I first tried ````checkTaskData(getTaskData(task))```` but then we run exactly into the problem
Bernd describes above, I guess. The traceback of the error is pretty tedious but as far is I understand we are then calling ````getTaskDescription(task)$target````, and the description object is not existing yet.
My next step was to look into ````makeSupervisedTask()````, where we also use ````task$env$data````, so I figured that's the right way to go here.
Thanks for looking through this with me!

florianfendt:
@berndbischl 
Should I do a new PR for train and predict (since this PR is still from my fork)

berndbischl:
> Should I do a new PR for train and predict (since this PR is still from my fork)

i think it is better do do this here, so have it "comlete"
so pls extend this PR

berndbischl:
hi guys.

1) especially @mllg and @jakob-r and @larskotthoff 
I would suggest not to do comments as "line notes" while reviewing code.
this seems not to work, now because flo adapted his PR, the whole conversation there is gone (?).
we have to be careful here I guess

2) @florianfendt 
comment is perfect now

jakob-r:
>the whole conversation there is gone (?).

It's still there. Just collapsed.

berndbischl:
@florianfendt 
you have introduced a template for arg_subset, but not for arg_features (copypaste).
and is still see some typos ("denots"). try to enable a spellchecker please.

also we dont need this: (thats a bit much and clear)
"logical vectors will be transformed to integer with \code{\link[base]{which}}.

berndbischl:
continued in #962

PR #827, opened 2016-04-07
improve regression randomForestSRCSyn

schiffner:
See #602. 
* extended and corrected parameter set
* changed `trainLearner` and `predictLearner` methods such that the synthetic forest is actually trained when calling `trainLearner`. (Before it was trained in `predictLearner`.)


berndbischl:
thx a lot. will merge

PR #833, opened 2016-04-13
Multilabel wrappers all

QuayAu:
I made one PR for all wrappers (if desired, i can update the 4 PRs).
I incorporated the latest comments into the files. The following changes were made:

- used "hasLearnerProperties" for check if "twoclass" is available
- there is no option for a different level 1 learner any more (for DBR and STA)
- used "makeMultilabelBinaryRelevanceWrapper" for level 1 learners (for DBR and STA)
- small corrections (deleted unnecessary "newdata = .newdata" or "paste", added "..." in predict,  small style changes)
- unit tests were pretty much copy and paste. I wrote a function, for a clearer and shorter code.


R/MultilabelClassifierChainsWrapper.R
```diff
@@ -0,0 +1,94 @@
...
+    model.subclass = "MultilabelClassifierChainsModel")
+  x$type = "multilabel"
+  x$order = order
+  return(x)
+}
+
+#' @export
+trainLearner.MultilabelClassifierChainsWrapper = function(.learner, .task, .subset, .weights = NULL, ...){
+  if (.learner$order[1] == "random") {
+    order = sample(1:length(.task$task.desc$target)) #random order
```
PhilippPro: use getTaskTargetNames

R/MultilabelClassifierChainsWrapper.R
```diff
@@ -0,0 +1,94 @@
...
+  } else {
+    order = .learner$order
+  }
+  if (length(order) != length(getTaskTargetNames(.task))) {
```
PhilippPro: better: `sort(order) != 1:length(getTaskTargetNames(.task))` and "order does not match number of targets"

R/MultilabelDBRWrapper.R
```diff
@@ -0,0 +1,88 @@
...
+    function(m) as.logical(predict(m, newdata = .newdata, ...)$data$response)
+  } else {
+    function(m) getPredictionProbabilities(predict(m, newdata = .newdata, ...), cl = "TRUE")
+  }
+  predLvl1 = sapply(data.frame(asMatrixCols(lapply(modelsLvl1, f))), as.numeric)
+  # Meta level prediction
+  modelsMeta = models[(length(.model$task.desc$target) + 1):(2 * length(.model$task.desc$target))]
+  newdataMeta = data.frame(.newdata, predLvl1)
+  f = if (.learner$predict.type == "response") {
+    function(m, tn) as.logical(getPredictionResponse(predict(m, newdata = newdataMeta[,which(colnames(newdataMeta) != tn)], ...)))
```
PhilippPro: Leerzeichen bei  `newdataMeata[, which` ;)

R/MultilabelDBRWrapper.R
```diff
@@ -0,0 +1,88 @@
...
+  makeHomChainModel(.learner, c(modelsLvl1, modelsMeta))
+}
+
+#' @export
+predictLearner.MultilabelDBRWrapper = function(.learner, .model, .newdata, ...) {
+  models = getLearnerModel(.model, more.unwrap = FALSE)
+  # Level 1 prediction (binary relevance)
+  modelsLvl1 = models[1:length(.model$task.desc$target)]
+  f = if (.learner$predict.type == "response") {
+    function(m) as.logical(predict(m, newdata = .newdata, ...)$data$response)
```
giuseppec: Ist hier sichergestellt, dass `predict(m, newdata = .newdata, ...)$data$response` Werte 0 und 1 annehmen? Wenn das z.b. ein Faktor ist, dann führt `as.logical` zu einem bug, siehe z.B. 
```r
> as.logical(as.factor(c(0,1)))
# [1] NA NA
```

Wenn das sichergestellt ist, dann würde ich `getPredictionResponse` verwenden:

```r
function(m) {
  pred = predict(m, newdata = .newdata, ...)
  as.logical(getPredictionResponse(pred))
}
```

QuayAu: When making a multilabel task, the binary relevance for each label must be coded in logical. So it is ensured that `predict(m, newdata = .newdata, ...)$data$response` has logical values `TRUE` and `FALSE`. That `as.logical` was too much.

However, I used `getPredictionResponse` in my newest commit now.

R/MultilabelDBRWrapper.R
QuayAu: done

R/MultilabelClassifierChainsWrapper.R
QuayAu: done

R/MultilabelClassifierChainsWrapper.R
QuayAu: done

R/measures.R
```diff
@@ -717,6 +717,133 @@ measureHAMLOSS = function(truth, response) {
...
+    measureF1MULT(getPredictionTruth.PredictionMultilabel(pred),
+    getPredictionResponse.PredictionMultilabel(pred))
+  }
+)
+
+#' @export measureF1MULT
+#' @rdname measures
+#' @format none
+measureF1MULT = function(truth, response) {
+  Fi = as.numeric()
```
giuseppec: We know the length of `Fi` right? Then it is much faster for the for-loop, if you use the preallocation `Fi = numeric(nrow(truth))`. Do this for all measures. Never use preallocations without specifying the vector length if you can avoid this.

QuayAu: ok, I've just done it

R/MultilabelDBRWrapper.R
```diff
@@ -0,0 +1,88 @@
...
+  targets = getTaskTargetNames(.task)
+  .task = subsetTask(.task, subset = .subset)
+  data = getTaskData(.task)
+  # train level 1 learners (binary relevance)
+  modelsLvl1 = getLearnerModel(train(makeMultilabelBinaryRelevanceWrapper(.learner$next.learner), .task, weights = .weights))
+  # train meta level learners
+  modelsMeta = namedList(targets)
+  for (tn in targets) {
+    data2Meta = dropNamed(data, setdiff(targets, tn))
+    trueLabelData = sapply(data.frame(data[, setdiff(targets, tn)]), as.numeric)
```
giuseppec: why do we write `data.frame(data[, setdiff(targets, tn)])`? Isn't `data[, setdiff(targets, tn)]` already a `data.frame`? And we have to keep in mind that `sapply(data.frame(data[, setdiff(targets, tn)]), as.numeric)` will be a  `matrix` and not a `data.frame`.
And maybe: Checkout if you can use the function `convertDataFrameCols` from `BBmisc` instead.

QuayAu: `data[, setdiff(targets, tn)]` is only a `data.frame`, when there are more than 2 columns. In case of 1 column, it will be a `logical`vector in this case (this happens, when there are only 2 labels).

giuseppec: Do you know this `data[, setdiff(targets, tn), drop=FALSE]`? It ensures
that it is always a data.frame, even if there is only one column. So please
use always drop=FALSE when you need data.frames for further computations.
Am 28.04.2016 18:52 schrieb "QuayAu" <notifications@github.com>:

berndbischl:
continued here: #977

PR #835, opened 2016-04-14
Load data into the calling environment

jimhester:
Fixes hadley/devtools#1146


larskotthoff:
Thanks! Build failures are because of the new version of testthat (fixed separately). Merging.

PR #846, opened 2016-04-22
Adding classification learner C5.0

MariaErdmann:
Hi, can you please have a look at the documentation of the C50 package. We didn't know what subset doeas and default for bands is problematic. Also see the FIXMEs. Thanks!

possible update for NEWS:
`````
- new learners
-- classif.C50
````


berndbischl:
OK, I will fix this now. This is mine

berndbischl:
@MariaErdmann 
Actually no, I will not. Maria, can you pls update all things discussed and rebase on master so we can pull this in.

ja-thomas:
#853 should do this with a correct rebase

PR #853, opened 2016-04-26
C50 fixed

ja-thomas:
This should be the same as #846 but with a correct rebase (hopefully)

NEWS:
```
- new learners
-- classif.C50
```


ja-thomas:
Can someone check why jenkins is failing? I don't think I can log in to Jenkins.

PR #854, opened 2016-04-26
Added Brier score for multiclass problems

PhilippPro:
(maybe the model.matrix( ~ . -1, data = as.data.frame(truth)))^2) could be done better, needs reshape2, which is imported in mlr)

### DIFF
--- a/tests/testthat/test_base_measures.R
+++ b/tests/testthat/test_base_measures.R
@@ -245,6 +245,11 @@ test_that("check measure calculations", {
   acc.perf = performance(pred.classif, measures = acc, model = mod.classif)
   expect_equal(acc.test, acc$fun(pred = pred.classif))
   expect_equal(acc.test, as.numeric(acc.perf))
+  #multiclass.brier
+  multiclass.brier.test = mean(rowSums((getPredictionProbabilities(pred.classif) - matrix(c(0,0,1,0,1,0,0,1,0,1,0,0), 4, 3))^2))
+  multiclass.brier.perf = performance(pred.classif, measures = multiclass.brier, model = mod.classif)
+  expect_equal(multiclass.brier.test, multiclass.brier$fun(pred = pred.classif))
+  expect_equal(multiclass.brier.test, as.numeric(multiclass.brier.test))
   #multiclass.auc
   n.cl = length(levels(tar.classif))
   pred.probs = getPredictionProbabilities(pred.classif)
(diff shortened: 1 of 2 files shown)


PhilippPro:
If it is ensured, that probabilities sum up to 1, the worst possible value would be 2. Is this ensured in mlr?

jakob-r:
It would be great if you can add a test in `test_base_measures.R` in the part `"check measure calculations"`

PhilippPro:
maybe let's first discuss, if to implement a new multiclass learner or implement this in the normal brier function, like discussed in #855

catastrophic-failure:
I see no problem with Brier Score assuming different values in multiclass or binary tasks. You could scale it to [0,1] or let it be after all we want to compare either ranks (doesn't depend on scale) or performance within a task (hypothesis testing) in a benchmark experiment.

Besides the original Brier Score definition is the one you coded, which adds to 2 with any number of classes.

Looking at this this way, it would make sense to have two measures: you can use `brier` when your tasks are all binary, but if you include multiclass tasks among them `multiclass.brier` will work.

tests/testthat/test_base_measures.R
```diff
@@ -245,6 +245,11 @@ test_that("check measure calculations", {
   acc.perf = performance(pred.classif, measures = acc, model = mod.classif)
   expect_equal(acc.test, acc$fun(pred = pred.classif))
   expect_equal(acc.test, as.numeric(acc.perf))
+  #multiclass.brier
+  multiclass.brier.test = mean(rowSums((getPredictionProbabilities(pred.classif) - matrix(c(0,0,1,0,1,0,0,1,0,1,0,0), 4, 3))^2))
+  multiclass.brier.perf = performance(pred.classif, measures = multiclass.brier, model = mod.classif)
+  expect_equal(multiclass.brier.test, multiclass.brier$fun(pred = pred.classif))
+  expect_equal(multiclass.brier.test, as.numeric(multiclass.brier.test))
```
catastrophic-failure: Should read `expect_equal(multiclass.brier.test, as.numeric(multiclass.brier.perf))` no?

PhilippPro:
Somebody can tell me, why the term probabilites instead of probabilities is used in some functions?

berndbischl:
> Somebody can tell me, why the term probabilites instead of probabilities is used in some functions?

because it is a typo i would guess?

Issue #859, opened 2016-04-29
`removeHyperPars` works only with one wrapper


giuseppec:
If we have a wrapper of wrappers, it does not work, see:

```r
lrn = makeLearner("classif.qda", nu = 2)
lrn = makeOversampleWrapper(makeFilterWrapper(lrn, fw.perc = 0.5), osw.rate = 1)
getHyperPars(lrn)
#$nu
#[1] 2
#
#$fw.method
#[1] "rf.importance"
#
#$fw.perc
#[1] 0.5
#
#$osw.rate
#[1] 1
#
removeHyperPars(lrn, ids = names(getHyperPars(lrn)))
#Learner classif.qda.filtered.oversampled from package mlr,randomForestSRC,MASS
#Type: classif
#Name: ; Short name: 
#Class: OversampleWrapper
#Properties: numerics,factors,prob,twoclass,multiclass
#Predict-Type: response
#Hyperparameters: nu=2
```
The hyperparameter `nu=2` has not been removed, i'll do a PR here #860.

berndbischl:
PR is here #860

berndbischl:
PR is merged. thx!

PR #860, opened 2016-04-29
fix removeHyperPars.BaseEnsamble for wrapped learners (Issue #859)

giuseppec:


berndbischl:
@giuseppec 
thx a lot! i would like to draw this in asap.

but can we try, if we are already at it, discuss the thing with the basensemble, too?

berndbischl:
i mean we are discussing a potential bug here for the BaseEns correct? so why not simply set up a few tests?

giuseppec:
I can do this. I also found this:

https://github.com/mlr-org/mlr/blob/removeHyperPars.BaseEnsamble/R/BaseEnsemble_operators.R#L12
https://github.com/mlr-org/mlr/blob/removeHyperPars.BaseEnsamble/R/BaseEnsemble_operators.R#L38
https://github.com/mlr-org/mlr/blob/removeHyperPars.BaseEnsamble/R/BaseEnsemble_operators.R#L67
https://github.com/mlr-org/mlr/blob/removeHyperPars.BaseEnsamble/R/BaseWrapper_operators.R#L50

Will look have a look each of those add add some tests. Currently, I don't see a reason why we, e.g. use `setPredictType.Learner` and not just `setPredictType`.

berndbischl:
improve the code and add more tests. then we will see.

giuseppec:
The closer I look at the code, the more I have to vomit. This bug seems to happen for several helper functions (especially for the getter and setter funs). Just to make sure this is a bug before I rewrite everything:

```r
bl1 = makeLearner("classif.rpart", minsplit = 2L, id = "rpart")
bl2 = makeLearner("classif.ksvm", C = 2, id = "ksvm")
# now make a wrapper around bl2 (ksvm)
bl2 = makeOversampleWrapper(makeFilterWrapper(bl2, fw.perc = 0.5), osw.rate = 1)
be = makeBaseEnsemble(id = "foo", base.learners = list(bl1, bl2), cl = "mywrapper")
# Does only return oversample hyperpars and NOT all hypepars from the wrapperd-ksvm:
getHyperPars(be)
```
This returns:

```
$rpart.xval
[1] 0

$rpart.minsplit
[1] 2

$ksvm.filtered.oversampled.osw.rate
[1] 1
```
Here `ksvm.filtered.fw.perc = 0.5` and `ksvm.C = 2` are missing, right? 
If this is correct, I will setup various tests that use different combinations of BaseWrappers and BaseEnsamble similar to this one above.

berndbischl:
@giuseppec 

i would REALLY REALLY like to pull this in soon / now

1) 
> Here ksvm.filtered.fw.perc = 0.5 and ksvm.C = 2 are missing, right?
what you are saying is correct

2) 
> I also found this:

this seems indeed buggy and we should indeed call the generic
please fix

3) 
> If this is correct, I will setup various tests that use different combinations of BaseWrappers and BaseEnsamble similar to this one above.

can you please do this, so i can them pull?

giuseppec:
Done. I added a few tests that did not work before. If you have questions, just talk to me.

berndbischl:
so nothing open / problematic from your point of view? thx!

giuseppec:
Can be merged from my point of view. Everything that worked still seems to work + things that did not work now work.

berndbischl:
thx a lot for doing this very properly. i read the code, seems fine, and it clearly fixes bugs. merge.

PR #861, opened 2016-04-29
Add binary classification measure Brier scaled

studerus:
Unlike, the original brier score whose upper bound is dependent on the incidence rate, this one is scaled to be between 0 and 1.


larskotthoff:
I think I would prefer the name to be more explicit, i.e. .scaled instead of .sc. Thoughts?

larskotthoff:
Thanks, merging.

berndbischl:
this merge is a bit problematic. 
- the measure does NOT contain docs for what it actually does
- there is NO REAL unit tests. a real unit test would actually compare a handcalculated result on a mini-data situation with the result of the measure.

pls REALLY watch out for this, will create issue now

larskotthoff:
@studerus, do you have a reference for the formula please?

berndbischl:
its also explained in this one here, erich linked to this before IIRC

http://www.ncbi.nlm.nih.gov/pmc/articles/PMC3575184/

@larskotthoff 
would you be so nice to add one sentence to the note, and copy add a mini unit test so we can close this?

Issue #812, opened 2016-03-30
Bug in convertBMRToRankMatrix


florianfendt:
The function is not preserving the learner id's as row names in the matrix if you only pass one task.
See for example this code:

````
lrns = list(makeLearner("classif.lda"), makeLearner("classif.rpart"))
tasks = list(iris.task, sonar.task)
rdesc = makeResampleDesc("CV", iters = 2L)
meas = list(acc, ber)
bmr1 = benchmark(lrns, tasks, rdesc, measures = meas)
rmat = convertBMRToRankMatrix(bmr1)
print(rmat)


tsk = multiclass.task
lrns = list(makeLearner("classif.rpart", predict.type = "prob"),
  makeLearner("classif.glmnet", predict.type = "prob"))
measures.lst = list(mmce, acc)
bmr2 = benchmark(lrns, tsk, cv2, measures = measures.lst)
rmat2 = convertBMRToRankMatrix(bmr2)
print(rmat2)
````

berndbischl:
was merged in PR https://github.com/mlr-org/mlr/pull/863

PR #863, opened 2016-04-30
#812 fixes bug in convertBMRToRankMatrix

florianfendt:
Fixes a bug we had where the dim names were not passed correctly, if the bmr had only one task


larskotthoff:
Thanks, merging.

Issue #837, opened 2016-04-14
using glm with mlr?


andreashandel:
Hi, it's me again :)

My question this time: I have some outcome data that decently follows a gamma distribution, so I want to fit a GLM with a gamma distribution. I would like to do it in the mlr framework to be able to use all the 'goodies' that come with mlr (specifically in this case automated feature selection). Unfortunately glm is not part of the learners. I'm surprised to not see it on the list. 

I searched around in the 'issues' posts a bit and found some related discussion about poisson distributed data and a 'hack' for glmnet. Is there a way I could get mlr to 'talk to' glm? Maybe somehow by defining a performance measure that would correspond to the gamma link function? Or some other way? Thanks!

larskotthoff:
There's `(classif|regr).glmboost` and `(classif|regr).glmnet`.

andreashandel:
Yes but they don't allow gamma distributed outcomes, as far as I'm aware?

berndbischl:
I guess we could add it. We do have regr.lm, which of course offers less flexibility.

Andreas, why not add this locally as a custom learner as explained in the tutorial. 
Then do a PR. And I can have someone iterate over it to improve it

andreashandel:
@berndbischl while I'm in principle all in favor of your suggestion that i build a custom learner, the combination of my available time and my coding prowess suggest that it's unlikely this could happen anytime soon.

berndbischl:
ok, we will take this over then. but it can take a couple of days as there is other stuff on the prio list.

the main arg of glm seems to the the family list? what kind of options do you think are reasonable to inlcude?

andreashandel:
@berndbischl That sounds great, and I'm happy to wait until it's available, no problem if it takes a while. The main 'functionality' I want to get from glm is having access to different families/link-functions. When I use glm() by itself, I almost never change any other possible settings and just use whatever defaults glm() sets. So all those could be (for my purpose anyway) be set by mlr to the glm() defaults.

Also, based on @larskotthoff comment above, I took a look at the (so far unfamiliar to me) regr.glmboost and the mboost package. I saw in the help file that glmboost/mboost also has a 'family' setting, with what seems support for more families/link functions than glm(). I tried setting different families within mlr like so:
learner_name = "regr.glmboost"
mylearner = makeLearner(learner_name, family = Laplace() )
But that didn't seem to work. I don't need it to work, just figured I let you know what I tried.

larskotthoff:
[...]

PR #864, opened 2016-04-30
Adds glm learner #837

MariaErdmann:
Hi,
working on the reg.glm learner I took a look at the regr.lm learner. glm() and lm() both have the parameter "offset", but in the mlr learner regr.lm it does not appear in the ParamSet. Does mlr support offset in another way probably? If yes, then I will need to delete the offset parameter from regr.glm. Thanks!

suggestion for NEWS
````
- new learners
-- regr.glm
````

### CODE REVIEW

**Comment by schiffner** [comment]
Hi,
thx very much, looks good.

I realize you are new to gitHub and mlr?
1) Welcome!
2) Usually every PR is supposed to be reviewed by another member of the mlr team, no one is supposed to merge his/her own PR and you usually let the Travis/Jenkins checks finish before merging. :)
The Wiki has some contribution guidelines: https://github.com/mlr-org/mlr/wiki/mlr-Coding-Guidelines
Apologies if you know all this and this was something you agreed upon with Bernd. :)

I had a short look at your code and have some comments:

* The conditions passed to the `requires` args (e.g. https://github.com/mlr-org/mlr/pull/864/commits/d34756cfd09811bc9638f32191570aaf94590261#diff-548f9ac269799ff1ad47ad33ff4ef4a2R10) don't work because `family` is a `character`.

For example I can do this without complaint from mlr:
```
## unsuitable link function
lrn = makeLearner("regr.glm", family = "Gamma", poisson.link = "sqrt")
lrn
Learner regr.glm from package stats
Type: regr
Name: Generalized Linear Regression; Short name: glm
Class: regr.glm
Properties: numerics,factors,se,weights
Predict-Type: response
Hyperparameters: family=Gamma,poisson.link=sqrt

m = train(lrn, bh.task)
Model for learner.id=regr.glm; learner.class=regr.glm
Trained on: task.id = BostonHousing-example; obs = 506; features = 13
Hyperparameters: family=Gamma,poisson.link=sqrt

## in truth the learner was trained with link = "inverse"
str(m$learner.model$family)
List of 12
 $ family    : chr "Gamma"
 $ link      : chr "inverse"
 $ linkfun   :function (mu)  
 $ linkinv   :function (eta)  
 $ variance  :function (mu)  
 $ dev.resids:function (y, mu, wt)  
 $ aic       :function (y, n, mu, wt, dev)  
 $ mu.eta    :function (eta)  
 $ initialize:  expression({     if (any(y <= 0))          stop("non-positive values not allowed for the 'gamma' family")     n <- rep.int(1, nobs)     mustart <- y })
 $ validmu   :function (mu)  
 $ valideta  :function (eta)  
 $ simulate  :function (object, nsim)  
 - attr(*, "class")= chr "family"
```

* When defining the learner you should use the `note` arg of `makeRLearnerRegr` to warn the user that mlr deviates from the original `glm` API, i.e., that the `family` arg must be a character and that there are a number of new `link` args.

* For maintenance reason you shouldn't generate two almost equal `parset.list`s for the tests.
I realize that you need `family` as a character and the `inverse.gaussian.link` arg for the mlr learner interface, but that you could also infer automatically from the first `parset.list`.

* About the `offset` arg/param: I don't think that we have a good way to support this currently.
I would delete it from the parameter set because it doesn't work with resampling, for example:

```
lrn = makeLearner("regr.glm", family = "Gamma", offset = rep(1,getTaskSize(bh.task)))
crossval(lrn, bh.task)
[Resample] cross-validation iter: 1
Error in model.frame.default(formula = f, data = d, offset = c(1, 1, 1,  : 
  variable lengths differ (found for '(offset)')
Timing stopped at: 0.032 0 0.032
```

Cheers,
Julia

berndbischl:
hi.

maria is one of my new student assistants. the merge was a misclick
(can we actually define these rights more fine-grained?)

@schiffner thx a lot for the detailed review

@MariaErdmann can you now pls fix quickly what julia said in a new PR?

MariaErdmann:
Hi,

as Bernd wrote I am new to mlr and github as well and this happened not by purpose. I apologize for the unconveniences caused.
@schiffner: Thank you a lot for taking a look at the code and for the detailed feedback.
@schiffner & @berndbischl: I will implement these things  and do a new PR.
Sorry again!
Maria

schiffner:
Hi,

@MariaErdmann: Not a problem and no need to apologize. :smile:

Thanks again,
Julia

Issue #857, opened 2016-04-27
Observation weights are not passed to classif.logreg


studerus:
The following gives exactly the same performance although it shouldn't

    lrn <- makeLearner('classif.logreg', predict.type = 'prob')
    lrn.w <- makeWeightedClassesWrapper(lrn, wcw.weight = 2)
    benchmark(list(lrn, lrn.w), pid.task, cv10, list(bac, auc, tpr, tnr))

studerus:
Observation weights are also not passed to the following learners
`classif.binomial, classif.ctree, classif.lqa, classif.probit`

studerus:
`classif.binomial, classif.ctree, classif.probit` are easy to change. 
However, I don't know why this gives the same performance;
    
    lrn <- makeLearner('classif.lqa', predict.type = 'prob')
    lrn.w <- makeWeightedClassesWrapper(lrn, wcw.weight = 2)
    benchmark(list(lrn, lrn.w), pid.task, cv10, list(bac, auc, tpr, tnr))

The weights seem to be passed to the learner with the following lines:
 
    if (!is.null(.weights))
      args$weights = .weights
    do.call(lqa::lqa.default, args)

berndbischl:
i have made EXACTLY the same observations as you, yesterday night. ;-)

i do have a general unit test for this. and fixed all learners where this was triggered (basically your list).
but lqa and plr where the only ones where i could not see an effect of the weights, although the seem to be passed down correctly. 

i  only need to solve this, then it will be done.

studerus:
I already wrote an email to the maintainer of `lqa` but didn't get a response so far. I think we should remove the `weights` property from `classif.lqa` for the time being.

berndbischl:
> I already wrote an email to the maintainer of lqa but didn't get a response so far. I think we should remove the weights property from classif.lqa for the time being.

thx for the email. and i do agree with your suggestion. will finish this NOW

berndbischl:
just for reference sake, i tried this. it fails:
```
library(mlbench)
library(lqa)
d = Sonar[59:61]
d$Class = as.numeric(d$Class == "M")
w = rep(1, nrow(d))
w[1:100] = 10000
m1 = lqa(Class ~., data = d, family = binomial(), penalty = ridge(lambda = 0.1), control = lqa.control())
p1 = predict(m1, newdata = d)
m2 = lqa(Class ~., data = d, family = binomial(), penalty = ridge(lambda = 0.1), control = lqa.control())
[...]

PR #865, opened 2016-05-01
fix weights 

berndbischl:
fix weights tags for learners and make sure that weights are indeed passed down correctly, and add a general unit test for this

please read issue #857 

can somebody pls review this when travis is finished?
@studerus @larskotthoff @schiffner  @mllg

### DIFF
--- a/R/RLearner_classif_ada.R
+++ b/R/RLearner_classif_ada.R
@@ -26,7 +26,7 @@ makeRLearner.classif.ada = function() {
       makeIntegerLearnerParam(id = "xval", default = 10L, lower = 0L, tunable = FALSE)
     ),
     par.vals = list(xval = 0L),
-    properties = c("twoclass", "numerics", "factors", "prob", "weights"),
+    properties = c("twoclass", "numerics", "factors", "prob"),
     name = "ada Boosting",
     short.name = "ada",
     note = "`xval` has been set to `0` by default for speed."
--- a/R/RLearner_classif_binomial.R
+++ b/R/RLearner_classif_binomial.R
@@ -18,8 +18,7 @@ makeRLearner.classif.binomial = function() {
 #' @export
 trainLearner.classif.binomial = function(.learner, .task, .subset, .weights = NULL, link = "logit", ...) {
   f = getTaskFormula(.task)
-  stats::glm(f, data = getTaskData(.task, .subset), model = FALSE,
-    family = binomial(link = link), ...)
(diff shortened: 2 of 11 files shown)


berndbischl:
for NEWS:
-- fixed weight handling and weight tag for some learners

berndbischl:
actually, we should probably also now add tests for clustering and surv

studerus:
Thanks. Looks good, but why did you take out the weights from `classif.ada`? Weighting seems to have an effect on this learner.

berndbischl:
can you pls post your example for this? i thought they did not, maybe i made mistake

studerus:
I just replaced the learner in my original example and increased the weight a little bit:

      lrn <- makeLearner('classif.ada', predict.type = 'prob')
      lrn.w <- makeWeightedClassesWrapper(lrn, wcw.weight = 5)
      benchmark(list(lrn, lrn.w), pid.task, cv10, list(bac, auc, tpr, tnr))

You can see that the true positive rate increases as it should

berndbischl:
but like I posted in the issue thread:
>>> classif.ada had the weights tag, but "weights" isnt even supported in its API

???

studerus:
Strange. It also does not seem to make a difference whether I set the weights to 10 or 5. I agree do take it out.

berndbischl:
erich, i also just ran your benchmark code in the master branch. i am getting exactly the same results for both learners.

berndbischl:
can you please post your output? because i am getting confused now

berndbischl:
> It also does not seem to make a difference whether I set the weights to 10 or 5. I agree do take it out.

also note in general that this is not enough for some learners. i had to set very extreme weights like 100 or 1000 in some cases to see an effect

studerus:
task.id                  learner.id bac.test.mean auc.test.mean tpr.test.mean tnr.test.mean
    1 PimaIndiansDiabetes-example                 classif.ada     0.7187919     0.8200593     0.6019030     0.8356809
    2 PimaIndiansDiabetes-example weightedclasses.classif.ada     0.7218202     0.8242875     0.6078131     0.8358272

studerus:
with set.seed(1)

berndbischl:
ok erich just saw random noise. ada does not support weights at the moment

jakob-r:
I checked carefully and it looks good. Only thing which bothered me was this unnamed argument [here](https://github.com/mlr-org/mlr/pull/865/commits/d4b327bac49a2a254d09e598d284baadf3b5e5f3#diff-ef2cc592db5a22d79a152642ab415a79R68) :wink:

jakob-r:
my mistake

Issue #810, opened 2016-03-29
normalizeFeatures and createDummyFeatures mini clean up


berndbischl:
1) normalizeFeatures is not S3 --> do this

2) check that checkTargetPreproc is used consistentlly everywhere

3) in capLargeVales and impute we positively specify "cols", here we have an "exlude" arg to exlude unwanted cols. Remove exlude and use "cols"

berndbischl:
@florianfendt @Coorsaa 

Can you please talk about this, so we can close this ASAP? Who can do this first?

florianfendt:
I'll do it

berndbischl:
fixed in #868

PR #868, opened 2016-05-03
#810 normalizeFeatures and createDummyFeatures clean up

florianfendt:


florianfendt:
@berndbischl 
I'll ping you here since you wanted this ASAP. Maybe you can review it over the weekend then we could wrap it up on Monday.

berndbischl:
in general:

1) if you use expect_error in a test, always match against a pattern of the actual err message

2) is this a new PR? pls really work in a branch next time, as we emailed

for createDummyFeatures:

3) pls really reread your API / signatures before you push. you say that the default is all features in the docs. you should say: "all numeric features". 
and worse: but in your signature you define the default as char(0). I guess the semantic here is not "no features", but "not specified", so all features. We always use NULL for this.

4) this also seems bad style. if i pass you a cols char vec with num features in there, you now silently drop them in line 30.
in conclusion: why not compute the factor.cols, work.cols first. then use this for the arg check?

5) also you dont have to do the arg check for "cols" in 3 different places. just do in once in the data.frame S3 method. everything else is too much code and also slower.

for normalizeFeatures:
6) for the target doc: there is the template arg_taskdf_target
but if you use the template ALWAYS read them, whether they truely match

7) see point 3).....

8) normalizeFeatures.data.frame: you have a dozen lines of code there, to check that "cols" is a subset of the numeric cols (without the target). just use assertSubset?

jakob-r:
I think he fixed all your points but what is missing is the NEWS and maybe check again if this API change does not brake the tutorial. Can you make sure that `normalizeFeatures` and `createDummyFeatures` are not used in the Tutorial?
But in general I think it is okay to change from exclude to inlcude. Makes the user more aware of what he is doing.

florianfendt:
possible NEWS update:
````
- API changes
-- For the functions 'normalizeFeatures' and 'createDummyFeatures the argument
   'exclude' was replaced by 'cols'. So, for subsetting variables one know specifically selects,
   instead of dropping, them.
````

PR #869, opened 2016-05-03
rename partial prediction funs to partial dependence

zmjones:
better reflects terminology in the literature


berndbischl:
ok. thx

@zmjones @schiffner 

zach, can you pls check that the tutorial still builds? and possibly do the renames there as well

zmjones:
it does break it but i'll issue pr to fix it today

Issue #788, opened 2016-03-13
predict.WrappedModel gives warning when on.learner.error == "quiet"


mb706:
(... if `on.learner.warning == "warn"`)
The culprit is [this](https://github.com/mlr-org/mlr/blob/1d2c1dfdf53e66d6c529880449775659291e68c8/R/predict.R#L117) line, it should be
```R
if (opts$on.learner.error == "warn")
```
(instead it is 'on.learner.**warning** == "warn"'.)

berndbischl:
thx to both parties involved!

PR #870, opened 2016-05-04
fix for issue #788

florianfendt:
Fixed the typo we had in ````predict.WrappedModel````and wrote a small test for it.


berndbischl:
thx. the test seems most important here and looks good. for later:
I usually mention the issue number for the reported bug in a comment before the test.

berndbischl:
@florianfendt 
and please use branches from now on

PR #873, opened 2016-05-05
#815 fix rank.correlation to only accept continuous features

### LINKED ISSUE #815: rank.correlation filter should support factors but doesn't
The rank.correlation filter throws an error when a task with factors is supplied
```
filterFeatures(bh.task, method = 'rank.correlation', abs = 2)
```

However, in the definition of this filter, it says that it supports factors
```
subset(listFilterMethods(features = T), id == 'rank.correlation')
````

MasonGallo:
Fixes #815 so rank.correlation correctly only accepts continuous features. This will also need to be fixed on the tutorial at https://mlr-org.github.io/mlr-tutorial/release/html/filter_methods/index.html but I'm not sure how to update that?

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -211,8 +211,8 @@ makeFilter(
   name = "rank.correlation",
   desc = "Spearman's correlation between feature and target",
   pkg  = "FSelector",
-  supported.tasks = c("regr"),
-  supported.features = c("numerics", "factors"),
+  supported.tasks = "regr",
+  supported.features = "numerics",
   fun = function(task, nselect, ...) {
     y = FSelector::rank.correlation(getTaskFormula(task), data = getTaskData(task))
     setNames(y[["attr_importance"]], getTaskFeatureNames(task))


schiffner:
Hi,

thanks, looks good.

You don't need to do anything about the tutorial.
The table on the filter_methods page is generated automatically using `mlr::listFilterMethods`. So once your PR is merged and we build a new version of the tutorial the information on this page should be correct.
If you want to know more have look here: https://github.com/mlr-org/mlr-tutorial/tree/gh-pages

MasonGallo:
Thanks! I'll take a look

berndbischl:
thx looks good

berndbischl:
@MasonGallo 

But please try to work in branches later on. then it is simpler i think to help you.

MasonGallo:
@berndbischl 

Thanks - can you clarify what you mean? Do you mean pull requests sent to a particular branch other than master in this case?

berndbischl:
what i mean is: work in a branch of the mlr repo, not in your own fork.
you need contributor rights for this. (just invited you)
(pls do not abuse these rights :) )

this has the following advantage: if you need help, we can simply take over your branch for a few minutes  and update it. and the PR will be auto-updated then, too. we cannot do this if you are in your own fork.

MasonGallo:
Got it thanks will do!

PR #879, opened 2016-05-07
Added new cluster method kernlab::kkmeans

catastrophic-failure:


larskotthoff:
Thanks! Could you have a look at a few issues though please:

- We follow the convention that the files have the same name as the learner, i.e. you should put the code in RLearner_cluster_kkmeans.R
- What does the documentation say about the "p" parameter?
- Could you add some unit tests as well please? See one of the other learners for examples.

catastrophic-failure:
> We follow the convention that the files have the same name as the learner, i.e. you should put the code in RLearner_cluster_kkmeans.R
Sorry. I always miss these small details.
> What does the documentation say about the "p" parameter?
p: a parameter used to keep the affinity matrix positive semidefinite. I don't know what kind of practical effect this would have though.
> Could you add some unit tests as well please? See one of the other learners for examples.
Yes, will do ASAP.

larskotthoff:
Thanks! There should be a file with name RLearner_cluster_kmeans.R already, so I'm not sure how git didn't pick this up.

berndbischl:
@catastrophic-failure 
hi, do you need some help with the unit tests? or are you able to copy the structure of the test from another cluster learner?

berndbischl:
also: 
		 +    #FIX: Not sure if alg parameter is being passed correctly to predict fun

1) please use FIXME and not FIX in general
2) the FIXME content still looks "bad" and dangerous?

catastrophic-failure:
Hi, I'll try editing other cluster methods unit tests, I should learn how to do it properly anyways, I'll just try to be more careful so you don't have to waste time correcting it all again :)

So, about the `alg` parameter: `the algorithm to use. Options currently include kkmeans and kerninghan`. The description says: `With a particular choice of weights this algorithm becomes equivalent to Kernighan-Lin, and the norm-cut graph partitioning algorithms.`. Have to investigate further, I've never worked with clustering algorithms so I have no clue what that implies.

I extracted the kernel and the centers from `kernlab` built in functions, as these uniquely determine the clusters.

berndbischl:
> Hi, I'll try editing other cluster methods unit tests, I

ok great.



> So, about the alg parameter: the algorithm to use.

ok, look into this. but IMHO you dont have to become an "expert" on this. but you simply have to make sure that the param is passed down correctly. what it "does" in the learner itself is more or less an issue that the authors of that package have to worry about (?)

catastrophic-failure:
Now I changed the cluster membership prediction, actually doing the kernel distance calculation. This solves the issue over `alg`. I'm also writing the testing unit for it.

catastrophic-failure:
Why is Jenkins failing on this one?

Also, I'll revisit this soon, currently talking to kernlab's maintainer to add predict methods to both kkmeans and specc.

larskotthoff:
Jenkins build failure seems to be because of some files left over from a previous build? I've changed it to delete those at the start now, so it should be fine.

berndbischl:
hi, @catastrophic-failure 

i would like to draw this in soon. some comments to finish this:

- centers: the default in the package is not 2, so please remove this from the LearnerParam. Also you need to set par.vals with centers = 2, so the learner is trainable without args

- p: we dont know whether this affects performance or not, so please set tunable = TRUE

- note: pls describe what you are doing the predict function (computing the nearest center with kernelized distance / in features space). this is your code and not from kernlab

- please add a copy-pasted version of a unit test

catastrophic-failure:
@berndbischl I paused this one because I went for Alexandros to get the predict function (exactly like I wrote here) implemented in kernlab, but I surely could do with how it's now. Give me a few days.

catastrophic-failure:
Seems like I can't let a `kernelMatrix` float around. Calling `as.matrix` doesn't help.

catastrophic-failure:
Jenkins is still failing, I saw something to do with h2o on the emails earlier today. Is it still the reason?

larskotthoff:
This should be fixed now.

berndbischl:
i guess we can merge soon. 

but can you please really note in the "note" what mlr is *changing* for that learner?
that we are assigning centers = 2 by default, as the learner does not have one? and what exactly we do in predict?

berndbischl:
thx. 

i dont want to be too nitpicky, but the which.min at the end has a problem if the minimum dist is not unique. in your case you simply take the best point in the "order" of the data. i would suggest to take a random one.

you could do this with BBmisc::getMinIndex

catastrophic-failure:
@berndbischl agreed, it looks better now, though I didn't test it.

PR #881, opened 2016-05-08
makeRemoveConstantFeaturesWrapper

mllg:


jakob-r:
I checked quite carefully and it looks fine to me.

berndbischl:
1) good addition thx

2) please really dont merge your own stuff, for non-trivial code extensions. 
you are at least missing the family eda_preprocess for the wrapper
i will set this now.

i had to review quickly now, more i cannot see atm

berndbischl:
also your file name is wrong, we never we have named files with the class names and not the contructor name, so without "make"

PR #883, opened 2016-05-11
Adds a check for learner properties...

MariaErdmann:
... and removes check for weights.
checkLearnerBeforeTrain.R could be adapted based upon this PR. This would make it possible to check multiple properties at once, but the error message would be a bit less concise.

### DIFF
--- a/R/CostSensWeightedPairsWrapper.R
+++ b/R/CostSensWeightedPairsWrapper.R
@@ -20,7 +20,7 @@
 #' @family costsens
 #' @aliases CostSensWeightedPairsWrapper CostSensWeightedPairsModel
 makeCostSensWeightedPairsWrapper = function(learner) {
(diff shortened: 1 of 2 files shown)


jakob-r:
Looks like a useful restructuring of the code to me. I would be interested to see the PR for checkLearnersBeforeTrain.R but let's see what @berndbischl says.

berndbischl:
1) In general OK.

2) Two things, pls change: 
a) Learner '%s' must support all properties '%s', but does not support %s 
is more informative. 2nd slot is props, 3rd is missing props
b) add an arg check for "props" pls. use getSupportedLearnerProperties and assertSubset!

3) checkLearnersBeforeTrain you can refactor in a new PR. Do that right after this is finished and pulled in.

jakob-r:
All requests are fulfilled. IMHO ready to be merged

PR #884, opened 2016-05-12
fix a bug in tuneParams for named list of measures

mllg:


berndbischl:
your unit test in in the wrong file. 

can you please put this into test_base_tuning? we only put "long running stuff" / exhaustive tests 
in the extra groups.

berndbischl:
again, please really dont merge yourself.

have you checked the docs of PH? 

```
#' @param minimize [\code{logical}]\cr
#'   Which of the performance measures in y.names should be minimized?
#'   Vector of booleans in the same order as \code{y.names}.

```

because this seems "weird". 
a) first of all we dont mention the names in docs
b) then in the arg check in the codes we check name equality with "setequal"

now do we rely on the names or on the order?

PR #885, opened 2016-05-12
Introduce nugget.stability for regr.km

jakob-r:
I decided to trust the requirements check and not add any extra code to check if we overwrite anything


larskotthoff:
Thanks, merging.

Issue #880, opened 2016-05-08
measures: Strictly proper scoring rules


catastrophic-failure:
There are three famous strictly proper scoring rules: Quadratic, Logarithmic and Spherical.
I added the spherical scoring rule, but I also have coded the other two. There's a issue, the Quadratic scoring rule is simply an affine tranformation of `multiclass.brier` and the Logarithim scoring rule is simply `-1*logloss`. 

Are those two too trivial to be added? I would be fine with that, anyone would be able to calculate those _a posteriori_ with a simple linear transformation. ML begginers could be unaware though.

What do you think?

PhilippPro:
I think it would be ok to have also the affine transformation versions. In fact we also have rmse and mse. 
Maybe in the tutorial we can cluster measures, that are closely related to each other instead of ordering them only alphabetically, for the better overview.

catastrophic-failure:
All right. It's quite trivial, really, Brier Score and Logloss are strictly proper scoring rules, too.

Now, about the implementation itself, should I make new and self contained `measure.quadraticscore` and `measure.logarithmicscore` functions or could I point these to `measure.multiclassbrier` and `measure.logloss`, when these are merged, like it's done on RMSE?

berndbischl:
a) i have no problem to add measures, which are trafos of each other. we also have mmce and acc = 1-mmce. 


b) 
> Now, about the implementation itself, should I make new and self contained measure.quadraticscore and measure.logarithmicscore functions or could I point these to measureMulticlassBrier and measureLogloss, when these are merged, like it's done on RMSE?

either ways is OK, but better style IMHO is to call into the existing measure function. they exist, so we can do exactly that.

repopen if you have more questions. otherwise do the PRs.

AND: THX A LOT FOR ACTIVELY CONTRIBUTING! MUCH APPRECIATED!

PR #887, opened 2016-05-12
measures: Proper Scoring Rules

catastrophic-failure:
As I asked on #880, here are the three PSRs.

I need help with the name of measures though, they are too long I think.

### DIFF
--- a/R/measures.R
+++ b/R/measures.R
@@ -495,6 +495,7 @@ qsr = makeMeasure(id = "qsr", minimize = FALSE, best = 1, worst = -1,
 measureQSR = function(probabilities, truth){
   #We add this line because binary tasks only output one probability column
   if (is.null(dim(probabilities))) probabilities = cbind(probabilities,1 - probabilities)
+  truth = factor(truth, levels = colnames(probabilities))
   1 - mean(rowSums((probabilities - model.matrix( ~ as.factor(truth) + 0))^2))
 }
 
--- a/R/measures.R
+++ b/R/measures.R
@@ -471,6 +471,7 @@ ssr = makeMeasure(id = "ssr", minimize = FALSE, best = 1, worst = 0,
 #' @rdname measures
 #' @format none
 measureSSR = function(probabilities, truth){
+  truth = match(as.character(truth), colnames(probabilities))
   p = getRowEls(probabilities, truth)
   mean(p/sqrt(rowSums(probabilities^2)))
 }
@@ -493,7 +494,7 @@ qsr = makeMeasure(id = "qsr", minimize = FALSE, best = 1, worst = -1,
 #' @format none
 measureQSR = function(probabilities, truth){
   #We add this line because binary tasks only output one probability column
-  if (nlevels(truth) == 2L) probabilities = cbind(probabilities,1 - probabilities)
+  if (is.null(dim(probabilities))) probabilities = cbind(probabilities,1 - probabilities)
   1 - mean(rowSums((probabilities - model.matrix( ~ as.factor(truth) + 0))^2))
 }
 
--- a/R/measures.R
+++ b/R/measures.R
@@ -454,6 +454,69 @@ measureLogloss = function(probabilities, truth){
   -1*mean(log(p))
 }
 
+#' @export ssr
+#' @rdname measures
+#' @format none
+ssr = makeMeasure(id = "ssr", minimize = FALSE, best = 1, worst = 0,
+  properties = c("classif", "classif.multi", "req.truth", "req.prob"),
+  name = "Spherical Scoring Rule",
+  note = "Defined as: mean(p_i(sum_j(p_ij))), where p_i is the predicted probability of the true class of observation i and p_ij is the predicted probablity of observation i for class j.
+	See: Bickel, J. E. (2007). Some comparisons among quadratic, spherical, and logarithmic scoring rules. Decision Analysis, 4(2), 49-65.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureSSR(getPredictionProbabilities(pred, cl = pred$task.desc$class.levels), pred$data$truth)
+  }
+)
+
+#' @export measureSSR
+#' @rdname measures
+#' @format none
+measureSSR = function(probabilities, truth){
+  p = getRowEls(probabilities, truth)
+  mean(p/sqrt(rowSums(probabilities^2)))
+}
+
+#' @export qsr
+#' @rdname measures
+#' @format none
+qsr = makeMeasure(id = "qsr", minimize = FALSE, best = 1, worst = -1,
+  properties = c("classif", "classif.multi", "req.truth", "req.prob"),
+  name = "Quadratic Scoring Rule",
+  note = "Defined as: 1 - (1/n) sum_i sum_j (y_ij - p_ij)^2, where y_ij = 1 if observation i has class j (else 0), and p_ij is the predicted probablity of observation i for class j.
+	See: Bickel, J. E. (2007). Some comparisons among quadratic, spherical, and logarithmic scoring rules. Decision Analysis, 4(2), 49-65.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureQSR(getPredictionProbabilities(pred, cl = pred$task.desc$class.levels), pred$data$truth)
+  }
+)
+
+#' @export measureQSR
+#' @rdname measures
+#' @format none
+measureQSR = function(probabilities, truth){
+  #We add this line because binary tasks only output one probability column
+  if (nlevels(truth) == 2L) probabilities = cbind(probabilities,1 - probabilities)
+  1 - mean(rowSums((probabilities - model.matrix( ~ as.factor(truth) + 0))^2))
+}
+
+#' @export lsr
+#' @rdname measures
+#' @format none
+lsr = makeMeasure(id = "lsr", minimize = FALSE, best = 0, worst = -Inf,
+  properties = c("classif", "classif.multi", "req.truth", "req.prob"),
+  name = "Logarithmic Scoring Rule",
+  note = "Defined as: mean(log(p_i)), where p_i is the predicted probability of the true class of observation i.
+	See: Bickel, J. E. (2007). Some comparisons among quadratic, spherical, and logarithmic scoring rules. Decision Analysis, 4(2), 49-65.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureLSR(getPredictionProbabilities(pred, cl = pred$task.desc$class.levels), pred$data$truth)
+  }
+)
+
+#' @export measureLSR
+#' @rdname measures
+#' @format none
+measureLSR = function(probabilities, truth){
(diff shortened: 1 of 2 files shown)


studerus:
The proper quadratic scoring rule is the same as the scaled brier score and the logarithmic scoring rule is the same as Nagelkerke's R-squared, correct? Since some people might not know that these are the same, it could be helpful to add the alternative names to the notes. I recently made a PR (#861) for the scaled brier score. I assume that this is now obsolete.

catastrophic-failure:
@studerus I'm not sure.The (most common, there are infinite possibilities for this) QSR goes from [-1,1] and it's the trafo QSR = 1 - `multiclass.brier`. How does this compare to `brier.scaled`? I think it's [0,1] right?
 
LSR goes from (-Inf, 0], and it's the trafo LSR = -1*`logloss`, I've read elsewhere Negelkerke's R^2 is related to it, but I can't find a source that says it's exactly the same, to any number of classes, to what I'm pushing here. Could you help regarding this?

studerus:
Ok, than it's probably not exactly the same. The scaled brier score and Nagelkerke's R squared are both scaled to range between 0 and 1.

In [this](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC3575184/) paper, it says:

> Explained variation (R2) is the most common performance measure for continuous outcomes. For    generalized linear models, Nagelkerke’s R2 is often used 1,33. This is a logarithmic scoring rule. For binary outcomes Y, we score a model with the logarithm of predictions p: Y*log(p) + (Y−1)*(log(1 – p)). Nagelkerke’s R2 can also be calculated for survival outcomes, based on the difference in −2 log likelihood of a model without and a model with one or more predictors.
> 
> The Brier score is a quadratic scoring rule, where the squared differences between actual binary outcomes Y and predictions p are calculated: (Y - p)2,34. We can also write this similar to the logarithmic score: Y*(1 – p)2 + (1 – Y)*p2. The Brier score for a model can range from 0 for a perfect model to 0.25 for a non-informative model with a 50% incidence of the outcome. When the outcome incidence is lower, the maximum score for a non-informative model is lower, e.g. for 10%: 0.1*(1–0.1)2 + (1–0.1)*0.12 =0.090. Similar to Nagelkerke’s approach to the LR statistic, we could scale Brier by its maximum score under a non-informative model: Brierscaled = 1 – Brier / Briermax, where Briermax = mean(p)*(1 – mean(p)), to let it range between 0% and 100%. This scaled Brier score happens to be very similar to Pearson’s R2 statistic 35.

catastrophic-failure:
I have no idea what's failing Travis.

larskotthoff:
Looks like this was some stuff that was fixed in master. Rebase on master and push again and it should be fine.

ja-thomas:
1) Please use branches instead of forks, otherwise it is way harder to help with the code

2) Use short names for the measures 

`sphericalScoring` -> `sphsc`
`quadraticScoring` -> `qdrsc` etc.

3) Please explain what the measures do (in one short sentence) instead of linking papers in the note.

4) resolve the merge conflict. We cannot do that for you since you work on a fork.

berndbischl:
> 2) Use short names for the measures
> 
> sphericalScoring -> sphsc
> quadraticScoring -> qdrsc etc.

the names can be a bit longer. the abbreviations janek posted might be a bit confusing. but the style is wrong in the PR. dont use camel case for objects. so for longer names pls use sphericalscoring

berndbischl:
> 3) Please explain what the measures do (in one short sentence) instead of linking papers in the note.

"instead" is wrong in janek's comment. the link to the paper is good. but the measures can hopefully be explained in 1-2 sentences. please add this in the note

catastrophic-failure:
Am I supposed to open a branch from `master`? How to do that?
I'm actually currently working on a branch: https://github.com/catastrophic-failure/mlr/tree/patch-3, can you confirm you can't do PRs there?
Will work on the rest.

berndbischl:
> Am I supposed to open a branch from master? How to do that?
> I'm actually currently working on a branch: https://github.com/catastrophic-failure/mlr/tree/patch-3, can you confirm you can't do PRs there?
> Will work on the rest.

i think we got confused here, because erich is also in the thread. you could only open branches in the mlr repo, if you would be a member of the team. 

if you do stuff like this regulary, we could talk about granting you these rights.

for now, please simply follow our requests and hints and update from your side.

catastrophic-failure:
About the names, we could abbreviate to QSR, SSR and LSR (I've seen it before in the literature) or "quadratic score", "spherical score" and "logarithmic score".

berndbischl:
> "quadratic score", "spherical score" and "logarithmic score".

sounds good

berndbischl:
sorry, that i am so insistent here, and slow, but the measures must be really "perfect"
i also looked into the paper you linked.

1) for all 3 measures you can write down a very short formula that defines them precisely. 
see how i did that for the logloss. pls do the same here.
(your explaining remarks are good, i am not saying you should remove them)

2) the qsr seems to be 1 - multiclass.brier?
that needs to be said in the note then and also why do we duplicate code then? call into that function then.

3) important: we need handcrafted uni tests in test_base_measures. 
see e.g. lines from 309 on there, where we test this for brier and logloss.

```
  p1 = p2 = matrix(c(0.1, 0.9, 0.2, 0.8), 2, 2, byrow = TRUE)
  colnames(p1) = c("a", "b")
  colnames(p2) = c("b", "a")
  y1 = factor(c("a", "b"))
  y2 = factor(c("b", "b"))
  # multiclass.brier
  expect_equal(measureMulticlassBrier(p1, y1), 0.5 * ((1-0.1)^2 + (0-0.9)^2 + (0-0.2)^2 + (1-0.8)^2))
  expect_equal(measureMulticlassBrier(p1, y2), 0.5 * ((0-0.1)^2 + (1-0.9)^2 + (0-0.2)^2 + (1-0.8)^2))
  expect_equal(measureMulticlassBrier(p2, y1), 0.5 * ((1-0.9)^2 + (0-0.1)^2 + (1-0.2)^2 + (0-0.8)^2))
  # logloss
  expect_equal(measureLogloss(p1, y1), -mean(log(c(0.1, 0.8))))
  expect_equal(measureLogloss(p1, y2), -mean(log(c(0.9, 0.8))))
  expect_equal(measureLogloss(p2, y1), -mean(log(c(0.9, 0.2))))

  pred.probs = getPredictionProbabilities(pred.classif)
  pred.probs[pred.probs > 1-1e-15] = 1-1e-15
  pred.probs[pred.probs < 1e-15] = 1e-15
  logloss.test = -1*mean(log(pred.probs[model.matrix(~ . + 0, data = as.data.frame(tar.classif)) - pred.probs > 0]))
  logloss.perf = performance(pred.classif, measures = logloss, model = mod.classif)
  expect_equal(logloss.test, logloss$fun(pred = pred.classif))
  expect_equal(logloss.test, as.numeric(logloss.perf))

``


also try to think about corner cases here, like only 1 observation, and so on.

**Comment by catastrophic-failure** [comment]
@berndbischl no problem with the requests at all, those really help to improve the code.

1) Will do.
2) I think it is indeed, but it wasn't merged when I wrote this. multiclass.brier works for binary tasks, right?
3) yeah, that's a leftover, I haven't checked much the code the past months, I'll look into what you did in logloss.

**Comment by berndbischl** [comment]
there seems to be merge conflict. can you resolve this on your own?
(sorry, we are currently at the mlr workshop and heavily working on the codebase)

**Comment by berndbischl** [comment]
> ) I think it is indeed, but it wasn't merged when I wrote this. multiclass.brier works for binary tasks, right?

yes it works also for binary classification. there is a small difference between multiclass.brier and our normal brier. and we should explain this in a note as well. 
i will open up an issue about this.

**Comment by catastrophic-failure** [comment]
@berndbischl `measureMulticlassBrier` doesn't work with binary tasks because there's a `truth = factor(truth, levels = colnames(probabilities))` call while `probabilities` has only one column in binary tasks.

**Comment by catastrophic-failure** [comment]
**Please don't merge yet**, I'm going to add some handcrafted tests, probably tomorrow.

**Comment by catastrophic-failure** [comment]
Something broke the manual it seems.

**Comment by berndbischl** [comment]
> About the names, we could abbreviate to QSR, SSR and LSR (I've seen it before in the literature) or "quadratic score", "spherical score" and "logarithmic score".

sorry, my quote / reply was wrong. if the short names exist in the literature, please use those.

we are going to fix the travis problems on friday

**Comment by catastrophic-failure** [comment]
`getRowEls` doesn't work as one might expect when there's a single factor level represented in `truth`, it treats the factor level as the first in a sequence.

**Comment by berndbischl** [comment]
@ja-thomas and me will merge this now.
@catastrophic-failure you dont have to do anything we will do this ourselfs.

1) we should add some notes here which measures are equivalent

2) whats wrong with getRowEls?

3) whats wrong with measureMultiClassBrier?

**Comment by berndbischl** [comment]
abilities)) call while probabilities has only one column in binary tasks.

@catastrophic-failure 
we have merged all of your stuff now. thanks a lot. 

a) maybe you wanna proof-read it?

b) 
>  measureMulticlassBrier doesn't work with binary tasks because there's a truth = factor(truth, levels = colnames(prob)

but we do not understand this? measureMulticlassBrier also works for 2 classes? it simply that the formula is slightly different here and compute 2*brier?

**Comment by catastrophic-failure** [comment]
@berndbischl the `qsr` is defined for multiclass and binary tasks while multiclass.brier only work in multiclass and brier only work in binary tasks. We could surely edit multiclass.brier to work with two classes as you say, it's just that I wasn't sure I should do that in this PR.

**Comment by berndbischl** [comment]
> @berndbischl the qsr is defined for multiclass and binary tasks while multiclass.brier only work in multiclass and brier only work in binary tasks. We could surely edit multiclass.brier to work with two classes as you say, it's just that I wasn't sure I should do that in this PR.

i dont get what you are saying. here:

```
load_all()
task = sonar.task
r = holdout(lrn, task, measures = list(brier, multiclass.brier))
print(r$aggr)
```

```
           brier.test.mean multiclass.brier.test.mean 
                    0.2433                     0.4865 

```

the factor-2 difference is also documented in the notes now

**Comment by catastrophic-failure** [comment]
@berndbischl Oh I know about that, that works because you call `getPredictionProbabilities(pred, pred$task.desc$class.levels)` in the measure object, but if you intend to use `measureMulticlassBrier` directly on a `data.frame` or anything other than a task it won't work.

**Comment by berndbischl** [comment]
i guess i now understand what you mean

```
  #We add this line because binary tasks only output one probability column
  if (is.null(dim(probabilities))) probabilities = cbind(probabilities,1 - probabilities)
```

are you talking about this line in qsr?
because then i find it a bit confusing how you expressed this :)

you are not talking about binary or multiclass, really, but about the data type we expect and accept in the basic measure-function.

 and for qsr we do this little trick to expand a single vector p to (p, 1-p).

if thats it, you are right we should handle this consistently in the same manner

catastrophic-failure:
@berndbischl well, yeah, that's what I meant haha. That's why I first said `measureMulticlassBrier` (not `multiclass.brier`) wouldn't work with two classes (one probability column output), but then I made it confusing mixing the names.

PR #891, opened 2016-05-13
Update support of `getClassWeightParam()`

### LINKED ISSUE #525: Question about getClassWeightParam
I can do this:

```
getClassWeightParam("classif.ksvm")
           Type len Def   Constr Req Tunable Trafo
1 numericvector  NA   - 0 to Inf   -    TRUE     -
Used: train.
```

Is this also supposed to work?

```
lrn = makeWeightedClassesWrapper("classif.ksvm", wcw.weight = 5)
getClassWeightParam(lrn)
Error in learner$par.set$pars[[learner$class.weights.param]] : 
  attempt to select less than one element
```

ja-thomas:
This should fix #525, if we merge PR #883 before this, the check for `class.weights` can be done right (see FIXME).


larskotthoff:
Does this depend on #883 or should the tests pass?

ja-thomas:
no, I think I did something wrong with the tests, I'll check.

berndbischl:
@ja-thomas 
did you run the respective tests locally before you pushed?

berndbischl:
@ja-thomas 
this seems like a mini issue: can you please get back here soon?

berndbischl:
@ja-thomas 

Please always think about API changes. the docs of getClassWeightParam say that you return a single numeric Param object. For your getClassWeightParam.BaseEnsemble you now return a list of params.

This is neither documented nor mentioned here in the thread....

berndbischl:
The other thing is that in principle we wanna avoid this. Because now the function API does not tell us what type is returned. which is usually not good. But I also dont really see how to make this better? 

But lets think about this.

berndbischl:
Can we maybe get aways without returning that param? Or simply being allowed to ask for it for a base learner? I mean all other operation like set it and get it should be abstracted away through API?

ja-thomas:
@berndbischl I think we talked about how to handle ensembles (But that was some time ago). 

So for wrapper and learner this shouldn't be a problem. For ensembles I don't see another possibility, either return all possible base weights in a list or give a meaningful error? I haven't thought about the API change, so the list may be the worst option...

Or maybe we add a new optional argument to `getWeightParam()` to select an element in the ensemble from which to get the weight param?

larskotthoff:
What's the status here?

pat-s:
@ja-thomas 	Is this PR still valid?

PR #896, opened 2016-05-17
Add ordered property to ranger learners (fixes #888)

### LINKED ISSUE #888: Why do the ranger-learners not support odered factors?
Is there a specific reason, why the ranger learners do not have the `ordered` property set? It seems that the `ranger` random forest implentation correctly handles ordered factors the same way as numeric variables, as you can see in the following example:

    library(ranger)
    dat2 <- dat <- getTaskData(bc.task)
    i <- which(sapply(dat, is.ordered))
    dat2[,i] <- lapply(dat[,i], as.numeric)
   
    set.seed(1)
    p1 <- ranger(Class ~ ., dat)$predictions
    set.seed(1)
    p2 <- ranger(Class ~ ., dat2)$predictions
    all.equal(p1, p2)

studerus:
#888


larskotthoff:
Thanks, merging.

PR #902, opened 2016-05-18
Implementation of getOutOfBag

PhilippPro:
This is based on #693. 
I just transformed it, so that a standardized "predict" object is provided, so that e.g. `performance` can be applied on the resulting object. 

Further development could be changing it, so it is usable for Tuning and Bagging Wrapper. 

Code example:

```
 set.seed(123)
 training.set = sample(1:150, 50)
 lrn = makeLearner("classif.ranger", predict.type = "prob", predict.threshold = 0.6)
 mod = train(lrn, sonar.task, subset = training.set)
 oob = getOutOfBag(mod, sonar.task)
 oob
 performance(oob, measures = list(auc, mmce))
```

### DIFF
--- a/R/getOutOfBag.R
+++ b/R/getOutOfBag.R
@@ -33,6 +33,8 @@
 #' performance(oob, measures = list(auc, mmce))
 
 getOutOfBag = function(object, task) {
+  assertClass(object, classes = "WrappedModel")
+  assertClass(task, classes = "Task")
   td = object$task.desc
   # extract truth column
   subset = object$subset
--- a/R/getOutOfBag.R
+++ b/R/getOutOfBag.R
@@ -1,0 +1,58 @@
+#' Extracts out of bag predictions from trained models
+#'
+#' Learners like \code{randomForest} produce out of bag predictions. 
+#' \code{getOutOfBag} extracts this information from trained models and builds a 
+#' prediction object like provided by predict. 
+#' See \sQuote{Details} for a list of learners for which this is implemented.
+#'
+#' The following learners support out of bag predictions:
+#' \itemize{
+#' \item{randomForest} \cr
+#' {Support for classification and regression.}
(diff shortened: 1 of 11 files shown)


R/getOutOfBag.R
```diff
@@ -0,0 +1,60 @@
...
+  
+  st = system.time(p <-  getOutOfBagS3(object$learner, object))
+  time.predict = as.numeric(st[3L])
+  
+  makePrediction(task.desc = td, row.names = rownames(data), id = subset, truth = truth,
+                 predict.type = object$learner$predict.type, predict.threshold = object$learner$predict.threshold, y = p, time = time.predict)
+}
+
+getOutOfBagS3 = function(.learner, .model) {
+  UseMethod("getOutOfBag")
```
ja-thomas: This is quite similiar to the FeatureImportance PR #995. There the analog function is called `getFeatureImportanceLearner` instead of `getFeatureImportanceS3`, for consistency we should use the same convention, I dont really care which we use, since it is not exported anyways.

R/getOutOfBag.R
```diff
@@ -0,0 +1,60 @@
...
+  t.col = match(td$target, colnames(data))
+  if (!all(is.na(t.col))) {
+    truth = data[, t.col, drop = TRUE]
+    if (is.list(truth))
+      truth = data.frame(truth)
+  } else {
+    truth = NULL
+  }
+  
+  st = system.time(p <-  getOutOfBagS3(object$learner, object))
```
ja-thomas: evil `<-`

PhilippPro: no, this is one of the cases, where only <- is possible, = does not work in system.time. ;)

mllg: This generates NAs if there are NAs in `t.col`.

PhilippPro:
new PR in #1084

PR #910, opened 2016-05-23
Multiclass AUC measures

PhilippPro:
This is the implementation of four multiclass AUC measures as defined in https://www.math.ucdavis.edu/~saito/data/roc/ferri-class-perf-metrics.pdf

Please suggest better naming if you are unsatisfied with it. 

File  R/measures_colAUC.R is used for the implementation of colAUC of the caTools package (but for avoiding a new dependency). It calculates columnwise AUC with all possible class combinations and is used for the calculation of the multiclass auc measures.

### DIFF
--- a/R/measures_colAUC.R
+++ b/R/measures_colAUC.R
@@ -1,0 +1,66 @@
+# mainly copied from the caTools package
+colAUC = function(probabilities, truth) {
+  y = as.factor(truth)
+  X = as.matrix(probabilities)
+  if (nrow(X) == 1)
+    X = t(X)
+  nR = nrow(X)
+  nC = ncol(X)
+  nY = table(y)
+  uL = as.factor(rownames(nY))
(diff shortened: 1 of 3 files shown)


R/measures_colAUC.R
```diff
@@ -0,0 +1,66 @@
...
+  if (nrow(X) == 1)
+    X = t(X)
+  nR = nrow(X)
+  nC = ncol(X)
+  nY = table(y)
+  uL = as.factor(rownames(nY))
+  nL = length(nY)
+  if (nL <= 1) 
+    stop("colAUC: List of labels 'y' have to contain at least 2 class labels.")
+  if (!is.numeric(X))
```
jakob-r: maybe better `assert_matrix(X, mode = "numeric")`

berndbischl:
PR continued here: #960 

will close this

PR #911, opened 2016-05-25
use getS3method in listLearners

mllg:


berndbischl:
that change implies that people will not be able to list their locally added learners, except they register the S3 methods manually?

mllg:
Before only learners found by `methods()` have been considered by `listLearners()`. This behavior is unchanged (except S4 methods are not considered any more, but I guess that is no biggie?).

berndbischl:
merging this here, but we should communicate that people should registerS3 there local learners which are not packaged in mlr. see here:

https://github.com/mlr-org/mlr-tutorial/issues/39

@schiffner maybe you could be so nice to add 2 sentences about this in "how to extend mlr learners"?
pretty please?

schiffner:
But only because you're asking so nicely. ;)

PR #915, opened 2016-05-30
remove upper limit for glmnet's parameter s

mllg:
As far as I can tell parameter `s` of glmnet may have values `> 1`. Don't know why there is an upper limit in the param set.


schiffner:
Agreed.

larskotthoff:
Thanks, merging.

PR #916, opened 2016-05-30
learner for GPfit

Dotterbart:


larskotthoff:
Thanks. Could you check the parameters again please? It looks like you have a typo in "control", no lower and upper values specified, "trace" shouldn't be tunable (set "tunable = FALSE") and "optim_start" is not numeric.

Dotterbart:
optim_start expects a matrix as input, i wasn´t sure whether i should use makeNumericVectorLearnerParam or makeUntypedLearnerParam, i took the first one.

berndbischl:
if optim-start really is a matrix, you have to use "Untyped".

Also I dont think we can merge this in this form, if all features are expected to be in [0,1] as your note says. Are you sure that this is a fact? 

Because if so, we need to enable the user to also apply this on "normal" data and you have to add an extra internal scaling option.

Dotterbart:
Even if the features are not in [0,1], the function GP_fit() works and provides a result, but it also returns a comment, that the optimization expects scaled data. So im not sure about it, i´ve implemented a scaling for now, but i think ill have to talk to Jakob about it.

berndbischl:
First of all, I would like it better if we had a new option "scale" that turns on what you are doing.
With a default = TRUE. 

I do like how you correctly use attachTrainingInfo, to propagate the info to "predict"!

Another complicating thing is that low[j] = high[j] might happen. In these cases you divide by 0.
You should remove such features instead, as they are constant. But you also need to propgate this info to "predict".

Also you have to adapt the note so it correctly tells the users what we have changed in the method now.

catastrophic-failure:
I never used this function, but you scale using estimates of maxima and minima feature values (what happens to be in your training set), what would happen when you predict over features outside [0,1]?

larskotthoff:
You would apply the scaling on training and prediction data separately. Yes, this may distort things.

catastrophic-failure:
How would you scale a single instance?

larskotthoff:
For a single instance, you could apply the scale factors from the training set.

Dotterbart:
I added the scaling as option now. As i am not sure which version is better here, i used the one in which the variables are scaled separately for the prediction data, but i added the version where the extrema of the training are used uncommented.

larskotthoff:
In principle it should be the version that uses the scaling parameters from the training data, but if it blows up completely if values outside [0,1] are passed in, we would need to work around that.

catastrophic-failure:
> For a single instance, you could apply the scale factors from the training set.
What if the features of that instance are over the maximum values or under the minimum values of the training data?

And I'm not saying `predict` won't work with data outside [0,1], that was a question actually.

larskotthoff:
Could you investigate what happens if you have data outside [0,1] please?

Dotterbart:
So i reversed the uncommentation now. 
As i wrote above the function returns results even if the data is outside of [0,1], but i don´t know about the quality of the results.

larskotthoff:
Well, could you evaluate the quality please?

R/RLearner_regr_gpfit.R
```diff
@@ -0,0 +1,72 @@
...
+}
+#' @export
+trainLearner.regr.gpfit = function(.learner, .task, .subset, scale = TRUE, ...) {
+  d = getTaskData(.task, .subset, target.extra = TRUE)
+  if (scale) {
+    low = apply(d$data, 2, min)
+    high = apply(d$data, 2, max)
+    const = which(high == low)
+    if (length(const > 0)) {
+      d$data = apply(d$data[,-const], 2, function(x) x = (x - min(x)) / (max(x) - min(x)))
```
jakob-r: i now it has been already stated in the comments I guess but here we definitely want to use `BBmisc::normalize()`

berndbischl:
hi, please lets not make this a cacophony

1) the option "scale" should scale the training data. min vals, max vals, and removed features need to be stored.

2) during prediction, remove the same features as in train. scale the other features with the same min and max vals. yes, this can potentially produce values outside of [0,1] but this is how it is done.

3) it would be cool if we could use BBmisc::normalize for this, but we cannot. because of the "predict" complications. the R function "scale" e.g. allows to use the precomputed means and stds, BBmisc::normalize should allow to use the precomputed mins and maxs, but currently it does not, so do it manually.

4) dont use <- in your code, use = at all places.

5) expose "scale" also in the ParamSet

6) remove the commented lines at the end when you are done.

berndbischl:
we should actually improve BBmisc::normalize for crap like this, as I outlined above, but this is an orthogonal bonus issue for later and should not block the addition of this learner

R/RLearner_regr_gpfit.R
```diff
@@ -0,0 +1,61 @@
...
+  makeRLearnerRegr(
+    cl = "regr.gpfit",
+    package = "GPfit",
+    par.set = makeParamSet(
+      makeNumericVectorLearnerParam(id = "control", len = 3, lower = c(1, 1, 1)),
+      makeNumericLearnerParam(id = "nug_thres", default = 20, lower = 10, upper = 25),
+      makeLogicalLearnerParam(id = "trace", default = FALSE, tunable = FALSE),
+      makeIntegerLearnerParam(id = "maxit", default = 100, lower = 0),
+      makeUntypedLearnerParam(id = "optim_start", tunable = FALSE),
+      makeLogicalLearnerParam(id = "scale", default = TRUE)
```
jakob-r: better `makeLogicalLearnerParam(id = "scale", default = TRUE, tunable = FALSE)`

berndbischl:
better makeLogicalLearnerParam(id = "scale", default = TRUE, tunable = FALSE)

if it effects performance it is certainly tunable.

berndbischl:
what is still relevant is that the code contains too much copypaste.
i dont want to make this too annoying, but please dont code this with 1 million "ifs".

you need the if (scale). for the rest, remove the const features, then scale. thats it. no "if".

berndbischl:
and note that your "const" is much better a chat vector. which could also be empty. you dont have to leave it out if its empty

R/RLearner_regr_gpfit.R
Dotterbart: Soll ich das tunable jetzt doch drin lassen, nachdem Bernd Bischl das
angemerkt hat? Oder rausnehmen? Du bist (in Bezug auf mich) ja der Chef
hier :)

berndbischl: es ist ziemlich egal wer dein chef in dortmund ist für diese frage

Dotterbart: O.k. :)

berndbischl: zu weiteren erklärung, auch wenn hier der falsche platz für diese diskussionen ist:
es ist ja (vermutlich) richtig, dass du alles mit jakob absprichst, denn ich denke mal, dass du sein hiwi bist. was heißt dass er (vermutlich) bestimmen kann was du so alles tun musst.

was in die software gemergt wird, bestimmt aber eine gruppe von leuten. wir versuchen das hier halbwegs im konsens mit einigen "älteren" leuten zu machen, indem wir uns gegenseitig helfen, durch reviews. am ende zählt vorallem das was "richtig" ist. (so wie hier)

und manchmal bestimme ich oder ein anderer der core devels (was hoffenltich sehr selten notwendig ist, weil klar sein sollte was richtig ist)

berndbischl: und zu der ganzen "scale" sache: 
(irgendwie nervig, dass der einzelne lerner hier so kompliziert zu machen ist)

das doofe hier scheint mir zu sein, dass irgendwie unklar ist, ob man diese option jetzt haben will, oder immer macht. die software doku scheint mir zu sagen, dass ich sie eigentlich "will", wenn meine daten außerhalb von [0,1] liegen. frage ist, wann man sie jemals *nicht* wollen würde.

aber bei sowas ist es vielleicht besser, auf der seite zu "irren" wo man dem user letzlich eine option gibt, mit einem guten default. was wir ja hier getan haben.

smilesun:
1. "note that your "const" is much better a chat vector", you want to see cheat vector which might be null, yes? So when this is null, it is well handled?
2. in the comments, someone mentioned the performance of GPfit should be evaluated in the case of scaling and non-scaling, but this is done already? Shall I work on it?
3. Shall we add more to the Unit test for GPfit? 
4. "I do like how you correctly use attachTrainingInfo, to propagate the info to "predict"!"  I don't quite get the subtleness of "attachTrainingInfo" and what does it do, need explanation

smilesun:
This pull request is continued at #1004 with name: "improve GPfit learner from Dotterbart master (learner for GPfit #916) #1004"

PR #917, opened 2016-05-30
add functional anova

zmjones:
this is ready for a little reviewing now. the only major thing i think worth discussing is how i reused some of the existing code. i gave the output two classes, one new "FunctionalANOVAData" and the other "PartialDependenceData". The output is so similar (unsurprising given the former uses the latter heavily) that the plot functions can be re-used with essentially no modification. i could, however, not do this, and instead write a wrapper with a name that matches the generation function., `plotFunctionalANOVA`.


zmjones:
i've rebased this now

zmjones:
ping @berndbischl and @larskotthoff 

could we merge this relatively soon? this is one of the things i need to write my job market paper and so is somewhat time sensitive!

larskotthoff:
Thanks, looks good. Sorry, missed this earlier.

zmjones:
no worries! thanks!

berndbischl:
@zmjones 
if this is merged now, can you please also explain this at least BRIEFLY in the tutorial pretty soon?
otherwise people just miss this.

do you need a new issue for this?

zmjones:
No I will get to it asap

PR #918, opened 2016-05-31
set respect.unordered.factors for ranger learners

mllg:
According to http://www.r-bloggers.com/on-ranger-respect-unordered-factors/, `respect.unordered.factors` set to `TRUE` seems to be a better default.

Or is the blog post wrong, @mnwright?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -35,6 +35,7 @@ mlr_2.9:
 - subsetTask, getTaskData: arg "features" now also accepts logical and integer
 - normalizeFeatures, createDummyFeatures: arg 'exclude' was replaced by 'cols'
 - normalizeFeatures is now S3 and can be called also on data.frames
+- classif.ranger, regr.ranger and surv.ranger: now respect unordered factors by default
 
 - new learners
 -- classif.C50
--- a/R/RLearner_classif_ranger.R
+++ b/R/RLearner_classif_ranger.R
@@ -22,11 +22,11 @@ makeRLearner.classif.ranger = function() {
       makeLogicalLearnerParam(id = "verbose", default = TRUE, when = "both", tunable = FALSE),
       makeIntegerLearnerParam(id = "seed", when = "both", tunable = FALSE)
     ),
-    par.vals = list(num.threads = 1L, verbose = FALSE),
+    par.vals = list(num.threads = 1L, verbose = FALSE, respect.unordered.factors = TRUE),
     properties = c("twoclass", "multiclass", "prob", "numerics", "factors", "ordered"),
     name = "Random Forests",
     short.name = "ranger",
-    note = "By default, internal parallelization is switched off (`num.threads = 1`) and `verbose` output is disabled. Both settings are changeable."
+    note = "By default, internal parallelization is switched off (`num.threads = 1`), `verbose` output is disabled and `respect.unordered.factors` is set to `TRUE`. All settings are changeable."
   )
 }
 
--- a/R/RLearner_regr_ranger.R
+++ b/R/RLearner_regr_ranger.R
@@ -21,11 +21,11 @@ makeRLearner.regr.ranger = function() {
       makeLogicalLearnerParam(id = "verbose", default = TRUE, when = "both", tunable = FALSE),
       makeIntegerLearnerParam(id = "seed", when = "both", tunable = FALSE)
     ),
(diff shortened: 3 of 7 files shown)


mnwright:
In the blog post is a perfect example for a case in which one should set this option. However, I'm not sure this is always the case. There are also many examples where there is no improvement by unordered splitting. Ordered factors are often not coded as such and users would be disappointed by the factor level limit or the long runtimes. Finally, in large high dimensional datasets, it's just impossible to do in reasonable time. 

In https://github.com/imbs-hl/ranger/issues/36#issuecomment-203967512 there is a note by Philipp on a simplification. That shouldn't be hard to implement (have to check multicategorical and survival outcomes) and I will take a look in the next days.

mnwright:
The approach described above is now included in the ranger master (v 0.4.5). Now there are three options for `respect.unordered.factors`: `'order'`, `'partition'` and `'ignore'`. For `'order'`, the new approach is used, for `'partition'` the old unordered mode and for `'ignore'`, the factors are considered ordered. The default is now `'order'`. 
`TRUE` and `FALSE` still work and are translated to `'order'` and `'ignore'`, respectively. 

How do we handle this in `mlr`? We should set the default to `FALSE` for older versions and to `'order'` from 0.4.5. But is this possible? Other ideas?

larskotthoff:
We don't distinguish between different learner versions. The code is implemented to work with the latest version. So we would wait for the next ranger release and then update the code. Does that sound ok?

mnwright:
Fine with me. I'll comment again when the feature is on CRAN.

berndbischl:
ok. if this is updated, we need to inc the version requirement in SUGGESTS for ranger then

berndbischl:
@mnwright 
I am not sure I agree with your three options for ranger. Also, we might possibly take this discussion someplace else.

First of all, thanks for implementing the more efficient rules for MSE regression and binary classification. I  always wondered why some packages leave those out. What I dont get is why there are now three options to handle this.

Why not simply keep the respect.unordered.factors? And internally use  the "ordered" implementation for binary classification and MSE regression. Noboby has *any* reason whatsoever to use the "partition" rule here! As you can prove that this "trick" is equivalent to full partition, in the sense that you find the best split.

You now effectively 
a) break  your old API
b) by making it more complex
c) give users a third option they have no reason to ever use, as it simply makes their code slower

berndbischl:
@mnwright 

```
##' Unordered factor covariates can be handled in 3 different ways by using \code{respect.unordered.factors}: 
##' For 'ignore' all factors are regarded ordered, for 'partition' all possible 2-partitions are considered for splitting and for 'order' the factor levels are ordered by their mean response, as described in Hastie et al. (2009), chapter 9.2.4.
##' The default 'order' is generally recommended, as it computationally fast and can handle an unlimited number of factor levels. 
##' Note that the factors are only reordered once and not again in each split. 
```


And in any case you should update the docs. It is a bit unclear what "mean respsonse" is for binary classfication.  @PhilippPro stated this more clearly in his post. Also one should say that for binary classification (with the usual criterions!) and MSE splitting in regression this is equivalent to normal partitioning. And that this equivalency does NOT hold for multiclass. 
Come to think of it: what DO you do for multiclass in the "order" case? as there is no class "1"?

mnwright:
Bernd, you are right. Your second post gives the reason for the first post. My first idea was to just change the behaviour for `respect.unordered.factors=TRUE`, but there are two reasons I didn't: 
* In my understanding the factors need to be re-sorted in each split for equivalence and I sort only once (see the note in the docs)
* I'm not yet sure what to do for multi class classification (as you mentioned in your second post) and for survival outcomes

I must admit it was a little rushed to merge it in the master in this state. 

For now, I will set the default back to `FALSE` or `'ignore'` and update the docs. I will then check options for multi class and survival outcomes.

berndbischl:
@mnwright Lets please continue to talk in the ranger tracker please. I have added a comment in your old issue thread
https://github.com/imbs-hl/ranger/issues/36
Or possibly open a new one.

I do that think that this is an important topic to get (w)right in ranger. (sorry :) )

berndbischl:
@mllg pls dont chane NEWS.

post the new contents here, pls.
reason: the (trivial) conflict that i now have to resolve on the console.

and: thx, agree with the change

PR #919, opened 2016-05-31
added max.number.of.layers to classif.nnTrain

mllg:
Added extra argument to `classif.nnTrain` in order to tune the number of hidden layers (as mlr cannot tune over vector length).


larskotthoff:
So wait, how exactly does that work? What's the interaction between the two layer parameters?

jakob-r:
The main thing whcih happens is
`hidden = head(hidden, max.number.of.layers)` but I am not sure if I like that we now use the default of `hidden = 10` in the train-call.

berndbischl:
i guess this is ok, thx @mllg 

will merge now

Issue #874, opened 2016-05-06
provide "learners_all" type unit tests for multilabel and surv learners


berndbischl:
copy parts of the structure from the test files for the other groups

berndbischl:
we have these now

Issue #840, opened 2016-04-19
Learners from penalized package seem to be broken


larskotthoff:
See https://hagakure.cs.ubc.ca:2893/job/mlr/370/console

Unit tests are failing after upgrading the penalized package:

1. Error: regr_penalized_lasso (@test_regr_penalized_lasso.R#24) 
2. Error: regr_penalized_ridge (@test_regr_penalized_ridge.R#25)

florianfendt:
@berndbischl
Since I already uploaded quite a big PR on this a while ago I think it might be a good time now to review it and have this fixed.
What do you think?

berndbischl:
@florianfendt 
Please quickly link it here. Can you also try to repair the penalized learners now. 
This is completely blocking us

florianfendt:
@berndbischl 
https://github.com/mlr-org/mlr/pull/665
will look over it again today!

berndbischl:
The problem is that this simple code does not run:

```
library(penalized)
library(mlbench)
data(BostonHousing)

m = penalized(medv ~ ., data = BostonHousing)
p = predict(m, data = BostonHousing[1:10,])
```

output 

```
$unpenalized
response ~ 1
<environment: 0xc5fdc60>

$penalized
medv ~ .
<environment: 0xed22da8>

Error in .local(object, ...) : 
  row counts of "penalized", "unpenalized" and/or "data" do not match
```

One can see that the formula in $unpenalized seems wrong....
I will report this to the maintainer now and add a local fix in mlr.
I also dont get why they dont have unit tests for this....

berndbischl:
I mailed the maintiainer Jelle Goeman and he said he would look into this.

berndbischl:
new PR is #850

berndbischl:
PR is merged. We should leave this open, to remove the FIXMEs from the penalized package. 

If this takes a long time: We can later find the relevant lines where i patch the penfit models by searching for issue number 840 in R and tests files.

berndbischl:
who cares. lets close this now.

PR #921, opened 2016-06-02
Fixes survival learners from penalized package (somewhat related to #840)

florianfendt:
The survival learners from the penalized package were broken.
E.g: They were throwing an error for factor features.
@MariaErdmann detected this while she was working on #874 and we fixed it together.


larskotthoff:
For the tests, since you're modifying the data in all of them, could you move that to one of the helper files so we don't have that much duplicated code please?

florianfendt:
Yes ok, will do. I wasn't so sure if we wanted that. Thanks

berndbischl:
please dont merge this, there are general problems here

berndbischl:
handled better in #931

PR #925, opened 2016-06-09
add params to regr_km (update)

jakob-r:
Quite specific stuff but we definitely want `multistart` to be available in *mlr*


larskotthoff:
Thanks, merging.

PR #927, opened 2016-06-13
Deleted param mtry.ratio

MariaErdmann:
... and adapted the note and the test.

### DIFF
--- a/R/RLearner_surv_randomForestSRC.R
+++ b/R/RLearner_surv_randomForestSRC.R
@@ -8,7 +8,6 @@ makeRLearner.surv.randomForestSRC = function() {
       makeDiscreteLearnerParam(id = "bootstrap", default = "by.root",
         values = c("by.root", "by.node", "none")),
       makeIntegerLearnerParam(id = "mtry", lower = 1L),
-      makeNumericLearnerParam(id = "mtry.ratio", lower = 0L, upper = 1L),
       makeIntegerLearnerParam(id = "nodesize", lower = 1L, default = 3L),
       makeIntegerLearnerParam(id = "nodedepth", default = -1L),
       makeDiscreteLearnerParam(id = "splitrule", default = "logrank",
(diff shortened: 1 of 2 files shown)


schiffner:
Hi,

why exactly do we want to delete `try.ratio`? Did it cause any problems?

berndbischl:
@schiffner 
because this is a "hacky" learner patch. since when do we introduce these new learner params? for how many learners do you want to do this?

berndbischl:
i am *not* saying we should not possibly support this in *general*, such "param recodings".  but then we have to talk about how to do this properly, and not "hack it in" in the places that we currently would like to see this.

schiffner:
@berndbischl:
Thx. Wasn't sure how we handle such things as this isn't my hack, there are other learners with such extra parameters, and there are other PRs like #919 that introduce similar things as well.

For consistency we have to delete this param also in other forest learners. I can do a PR.

berndbischl:
hi, @schiffner 

> Thx. Wasn't sure how we handle such things as this isn't my hack, there are other learners with such extra parameter

i know that this wasnt your code. and probably we should have written more "explaining" text in the PR. 
we need to talk about this in the next devel meetings. we *REALLY* have to discuss things like this. 
and NEVER merge them without agreeing on a standard here.

also paging @mllg @larskotthoff 


> and there are other PRs like #919 that introduce similar things as well.
> For consistency we have to delete this param also in other forest learners. I can do a PR.

ok, please help to clean this up. shall we open up a quick issue now?

schiffner:
Hey Bernd,

didn't need an issue for the other forest learners because this was a 5 min thing.

Opened issue #929 for general discussion on this topic.

larskotthoff:
I may be missing some context here, but how is this making things better? It doesn't seem like the code becomes much easier to maintain or something like this because of the removed parameter. What's the definition of a "hacky" patch here? Anything that adds functionality on top of what the learner provides?

berndbischl:
> I may be missing some context here, but how is this making things better? It doesn't seem like the code becomes much easier to maintain or something like this because of the removed parameter. What's the definition of a "hacky" patch here? Anything that adds functionality on top of what the learner provides

the point is this: (i kinda hoped this was somehow obvious..)
adding "mtry.ratio" as a param allows you to optimize mtry from 1 to p, by going from 0 to 1. so it allows a rescaling of the param, w.r.t. to some property of the data. 

yes, this is something that mlr should really allow, it comes up all the time in experiments. 
but adding this here, manually, cannot be a solution. because you could do the same for many learners, for multiple params. we can never maintain this in a reasonable way.

so my point is: 
- the problem is optimizing params on different scales. mlr should somehow support this
- would is done here is IMHO not a solution to the problem, as it does not scale.

what do you think?

larskotthoff:
Ok, sounds good to me.

PR #930, opened 2016-06-13
add learners gausspr

Coorsaa:
suggestion for NEWS:

````
new learners
-- classif.gausspr
-- regr.gausspr
````


larskotthoff:
Thanks. I see that there are quite a few FIXMEs in the code -- is this something that should be done before merging?

berndbischl:
these look like copy-paste FIXMEs from other kernlab code. please remove this

catastrophic-failure:
At least the `"stringdot"` one should remain, it's the same among every `kernlab` learner.

berndbischl:
@Coorsaa 
delete the other copy-pasta stuff so we can be done here please, the stringkernel stuff FIXME stuff can remain

Coorsaa:
@berndbischl and @larskotthoff: Thanks for the input, wasn't sure about the FIXMEs so I left them in in first place. Should be fine now.

larskotthoff:
Thanks, looks good. Can be merged as far as I'm concerned. @berndbischl ?

PR #931, opened 2016-06-13
Fix survival learners from penalized package II (shall replace #921)

MariaErdmann:
After taking a closer look at #921, we figured a better way to fix penalized survival learners:
In the predict()-function we use data = instead of penalized =. The tests were adapted accordingly.
This change requires the penalized package version >= 0.9-47 (see DESCRIPTION).
Additionally, in surv_penalized_fusedlasso trace = FALSE in par.vals() was missing. This disables output logging by default (see note).
What I did not do (yet): changing the tests in a way that they test if the penalized survival learners can handle factors. This is what @larskotthoff commented on #921. This will be implemented in a script called test_learners_all_surv (which I want to do a PR on asap).
So my question is: Shall the helper objects for surv.task be adapted such that they contain a factor and then adapt the tests for the penalized survival learners? Or will a test implemented in test_learners_all_surv (which will test all survial learners with property "factors") be enough?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -104,7 +104,7 @@ Suggests:
     numDeriv,
     pamr,
     party,
-    penalized (>= 0.9-46),
+    penalized (>= 0.9-47),
     pls,
     PMCMR (>= 4.1),
     pROC (>= 1.8),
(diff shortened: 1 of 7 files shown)


berndbischl:
> changing the tests in a way that they test if the penalized survival learners can handle factors. This is what @larskotthoff commented on #921. This will be implemented in a script called test_learners_all_surv (which I want to do a PR on asap).

yes, please do this in a general fashion, in a different PR.

berndbischl:
> So my question is: Shall the helper objects for surv.task be adapted such that they contain a factor and then adapt the tests for the penalized survival learners? Or will a test implemented in test_learners_all_surv (which will test all survial learners with property "factors") be enough?

for the new PR, create a nice function that add a factor variable to a given dataset, then calls the learner on it and checks that it runs without error (perf != NA). that function we can then call in every test_learners_all script.

PR #932, opened 2016-06-14
Added parameter 'grouped' in surv_cvglmnet 

MariaErdmann:
... and changed tests accordingly
I recognized that this parameter was missing. It is set to TRUE by default (then the "CV partial likelihood for the kth fold is obtained by substraction"). Setting it to FALSE will compute the log partial likelihood on the kth fold. 
But when there are less than three observations in a fold, grouped = TRUE will be enforced by cv.glmnet itself. Actually, this also happens in the test (because of the 'small' training set??). Therefore, one gets following warning:
1. surv_cvglmnet (@test_surv_cvglmnet.R#29) - Option grouped=TRUE enforced for cv.coxnet, since < 3 observations per fold
So I am not sure, if we want this parameter then or if I shall adapt the test (increasing training set or number of obs) or if everything is just alright.

### DIFF
--- a/R/RLearner_surv_cvglmnet.R
+++ b/R/RLearner_surv_cvglmnet.R
@@ -8,6 +8,7 @@ makeRLearner.surv.cvglmnet = function() {
       makeIntegerLearnerParam(id = "nfolds", default = 10L, lower = 3L),
       makeDiscreteLearnerParam(id = "type.measure", values = c("deviance"), default = "deviance"),
       makeLogicalLearnerParam(id = "exact", default = FALSE, when = "predict"),
+      makeLogicalLearnerParam(id = "grouped", default = TRUE),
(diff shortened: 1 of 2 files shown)


berndbischl:
1) please dont produce warnings in your test

2) i really fail to see why you introduced the grouped=F in the unit test?
i mean we dont test all the other params exhaustively....?

i have removed this from the test, so we have no warning. 

please merge this when travis is green.

MariaErdmann:
1) Thank you. I was not sure about that. So now I know.
2) The reason why I included it, was, that I recognized from different PRs that if one changes a learner, it always should be testet in the test and therefore I testet the parameter. But please correct me, if I got that wrong!

berndbischl:
> ) The reason why I included it, was, that I recognized from different PRs that if one changes a learner, it always should be testet in the test and therefore I testet the parameter. But please correct me, if I got that wrong!

in general you are correct, but like i said: for the param descriptions we cannot really do that.

Issue #933, opened 2016-06-14
Shall surv.coxph have property "missings"?


MariaErdmann:
Hi,

preparing a script that tests all survival learners I stumbled over the coxph learner, which has property "missings". Coxph as it is implemented now in mlr handles NAs in following manner: Calling predict() function reveals that for the according observation a NA is output (instead of a prediction). Therefore, the default performance measure, cindex, cannot be calculated. Hence, shall surv.coxph have property "missings"? Or am I missing something (implemented something wrong?)?
Below a code for reproduction:
````
## Preparation: either load_all() or the helper_objects need to be sourced after loading mlr library
load_all()
#library(mlr)
#source("helper_objects.R")

# Create data with missing value
data = getTaskData(surv.task)
data[1,3] = NA
task = makeSurvTask(data = data, target = surv.target)
head(getTaskData(task)) # x1 has a missing value in first obs

# surv.coxph has property "missings"
listLearners("surv", properties = "missings")$class

# Create learner, train, and predict
lrn = makeLearner("surv.coxph", par.vals = list(ties = "efron"))
m = train(lrn, task)
# But in the predict function nothing sophisticated happens with the missing value
p = predict(m, task)
perf = performance(p)
# cindex cannot be calculated
````

berndbischl:
In this case remove the missing property from coxph now. And make a note in your general unit tests for learners-all to test theses cases. That learners that claim to work with missing do indeed work.

berndbischl:
fixed in #934

PR #934, opened 2016-06-14
Deleted property 'missings' from surv_coxph

MariaErdmann:
This PR is related to Issue #933.


annette987:
The coxph function in R handles missing values according to the parameter na.action, which can take values such as  na.omit. (https://www.rdocumentation.org/packages/survival/versions/2.43-1/topics/coxph) But now that you have removed the property 'missings' surv.coxph cannot take advantage of this parameter. Would it be better to leave the 'missings' property in and add the na.action parameter to surv.coxph?

PR #936, opened 2016-06-14
note in listLearners defaults to empty string

### LINKED ISSUE #774: recheck listLearners refactoring by michel

mllg:
Should fix #774


jakob-r:
sorry to bother for such a small thing but might `character()` not be the better default?

mllg:
Columns are vectors, therefore rows cannot have length 0.

PR #939, opened 2016-06-14
Fixes for new PH

mllg:


berndbischl:
travis runs. i will merge this now.

@larskotthoff can you have a look whether we can also make jenkins runs after this pls?

larskotthoff:
It looks like I'll have to update R to a version that isn't in the repo yet for this to work again.

Issue #872, opened 2016-05-05
Tests fail for 'featsel'


MasonGallo:
I just setup my dev environment on a new mac with a fresh copy of mlr cloned this morning to fix #815 . Running the test for featsel below fails:

`devtools::test(filter = "featsel")`

The first failed dialog is - looks like it has to do with how filterFeatures is called:

```
1. Error: filterFeatures (@test_featsel_filterFeatures.R#5) ----------------------------
An error has occurred in the grow algorithm.  Please turn trace on for further analysis.
1: filterFeatures(binaryclass.task, select = "threshold", threshold = -Inf) at /Users/mason/projects/mlr/tests/testthat/test_featsel_filterFeatures.R:5

```
Is this reproducible on other machines?

Output of R.version on my local machine:

```
platform      x86_64-apple-darwin13.4.0   
arch           x86_64                      
os             darwin13.4.0                
system         x86_64, darwin13.4.0        
status                                     
major          3                           
minor          3.0                         
year           2016                        
month          05                          
day            03                          
svn rev        70573                       
language       R                           
version.string R version 3.3.0 (2016-05-03)
nickname       Supposedly Educational
```

schiffner:
Hi,

I couldn't reproduce on my mac, but I'm still on R 3.2.4. Updating now.

When running the featsel tests do you see only the one error you posted, or are there more?

MasonGallo:
Thanks! I only posted a snippet to avoid the wall of text in case it was something obvious. I will look into testing on another machine/VM. The full output is below:

```
analyzeFeatSelResult: ..
FeatSelWrapper: ...
[...]

PR #945, opened 2016-06-15
#872 remove rfsrc filters for sequential testing

MasonGallo:
It seems the only way to avoid this error is to remove rf.* from filters which I would assume we wouldn't want to do...putting this here just in case if not we can delete branch


berndbischl:
no, i dont want to remove this. you can always run the test with rt, that works without problems

Issue #898, opened 2016-05-18
plotBMRBoxplots - facetting should be configurable


berndbischl:
i want at least to set the number of rows and columns for this

berndbischl:
use facet_wrap. create new args facet.wrap.rows, facet.wrap.cols both with defaults = NULL

berndbischl:
solved in PR #950

PR #950, opened 2016-06-16
#898 facet_wrap for plotBMRBoxplots

florianfendt:
I don't know if we want parameter checks and/or mention that logicals are also possible,
since ggplot is already taking care of it?

Possible update for NEWS
````
- The function "plotBMRBoxplots" got two new arguments "facet.wrap.nrow" and
  "facet.wrap.ncol" which enable arrangement of facets in rows and columns.
````

### DIFF
--- a/R/generateCalibration.R
+++ b/R/generateCalibration.R
@@ -145,6 +145,7 @@ generateCalibrationData.list = function(obj, breaks = "Sturges", groups = NULL, 
 #'   Whether to include a rag plot which shows a rug plot on the top which pertains to
 #'   positive cases and on the bottom which pertains to negative cases.
 #'   Default is \code{TRUE}.
+#' @template arg_facet_nrow_ncol
 #' @template ret_gg2
 #' @export
 #' @examples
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by larskotthoff** [comment]
Hmm, we should probably consistently do this for all plots that do facetting. Could you have a look at the other plots as well please?

**Comment by florianfendt** [comment]
Ok, did it.
Why exactly did roxygen delete that line in NAMESPACE? I updated the package but the diff remains...

**Comment by berndbischl** [comment]
Pls remove those lines and doc what the default does

```

+    if (is.null(facet.wrap.nrow) && is.null(facet.wrap.ncol))
 +      facet.wrap.nrow = 1L

```

florianfendt:
@berndbischl 
did what you asked for. I think we can merge.

larskotthoff:
Thanks, merging.

berndbischl:
@larskotthoff 
you requested that this API change to be performed for *multiple* functions. but now the entry you just added to NEWS is wrong. please fix!

larskotthoff:
Done.

berndbischl:
> Done.

thx. lets be careful here pls.

PR #955, opened 2016-06-17
fix partial dependence parallel test

### LINKED ISSUE #952: generatePartialPredictionData test
There is still some tests for `generatePartialPredictionData` in `test_parallel_all.R` which seems to be overseen with #912 

1. How can it be there and not fail?
2. Is `Sys.info()["sysname"] != "Windows"` really the way to go, when we have `testthat::skip_on_os("windows")`?
3. Should this test be kept and renamed or deleted?

zmjones:
fixes #952


larskotthoff:
Hmm, Jenkins failure seems unrelated (something to do with paramhelpers)?

PR #959, opened 2016-06-20
Don't merge, still in review

MariaErdmann:
creates a test for all survival learners with more "generic" functions


MariaErdmann:
In helper_learners_all I created some helper functions in the style of testThatLearnerRespectsWeights(). The first one checkPerformance() is basically checking whether the learners produces a reasonale performance measure. Right now the code of this function appears in every single test_learners_all_* file at the beginning of the test. testThatLearnerHandlesFactors() introduces afactor in the data and then tests whether the learner can handle factors. And testThatLearnerHandlesMissings() introduces a NA value and tests whether the learner can handle missings.
Regarding the names: I tried my best... Another suggestion for testThatLearnerHandlesFactors() would be testThatLearnerRespectsFactors() and for testThatLearnerHandlesMissings(): testThatLearnerRespectsFactors().
In the test_learners_all_surv all these helper functions used to test whether the survival learners do what they claim to do in style of the former test_learners_all_* scripts.
Cheers, Maria

berndbischl:
naming: checkPerformance  --> testThatLearnerCanTrainPredict

berndbischl:
document all args which are used multiple times in the header of the files

MariaErdmann:
This is what happened:

helper_learners_all: 

- arguments that are used multiple times in header now
- changed name
- added some more functionality to testThatLearnerCanTrainPredict(): now it tests also whether learner can predict probs and se. Maybe the part with setPredictType.Learner() should better be setPredictType()?

test_learners_all_regr, test_learners_all_classif, test_learners_all_cluster were adapted:

- test_learners_all_cluster:
 There was a FIXME line saying "Remove this if DBScan runs stable". Since !inherits(lrn, "cluster.DBScan") was always TRUE I removed the FIXME and line. Please tell me, if this lines should be kept.
- The tests do produce some warnings. Some of them cannot be avoided, but some of them may be resolved by manipulating the data a little bit more (i.e. glm.fit fitted probabilities numerically 0 or 1)?
- When testing probabilities in the od scripts, there was a line getPredictionProbabilities(). I don't think this line is necessary, so I removed it.


Added test_learners_all_multilabel

Did some layout adaption on test_learners_all_surv

tests/testthat/helper_learners_all.R
```diff
@@ -34,3 +38,79 @@ testThatLearnerRespectsWeights = function(lrn, task, train.inds, test.inds, weig
...
+  if (lrn$id %in% names(hyperpars))
+    lrn = setHyperPars(lrn, par.vals = hyperpars[[lrn$id]])
+  
+  if (pred.type == "response") {
+    lrn = setPredictType.Learner(lrn, "response")
+  } else if (pred.type == "prob") {
+    lrn = setPredictType.Learner(lrn, "prob")
+  } else if (pred.type == "se") {
+    lrn = setPredictType.Learner(lrn, "se")
+  }
```
larskotthoff: The if statement seems unnecessary here. This could be just `lrn = setPredictType(lrn, pred.type)`.

berndbischl:
> The if statement seems unnecessary here. This could be just lrn = setPredictType(lrn, pred.type).

+1. Please try to avoid such unnecessary "if statements".

berndbischl:
> Maybe the part with setPredictType.Learner() should better be setPredictType()?

yes, please use the generic here. it really makes usually no sense to use the "method". dont do this unless you really HAVE to (ie when you have to manually select the method for a specific class)



> test_learners_all_cluster: There was a FIXME line saying "Remove this if DBScan runs stable". Since !inherits(lrn, "cluster.DBScan") was always TRUE I removed the FIXME and line. Please tell me, if this lines should be kept.

dont touch this at all, has nothing to do with this PR


> The tests do produce some warnings. Some of them cannot be avoided, but some of them may be resolved by manipulating the data a little bit more (i.e. glm.fit fitted probabilities numerically 0 or 1)?

warnings in tests suck. try to avoid them by changing the data a bit. but if you really cannot, thats how it is and we could live with it.



> When testing probabilities in the od scripts, there was a line getPredictionProbabilities(). I don't think this line is necessary, so I removed it.

 well, here is a better suggestion. in your testTrainPredict, if type = "prob" really DO acces the p object. with getPredictionProbabilities. and check the result! the line HAD its use! to really check that probs are actually contained in there! you now effectively disabled this check. 
think about like this: in your current code state, where do you check that probs are really produced? so add in back in, better add 1-2 MORE to inspect the output of that getter,.

Please resolve this ASAP so we can merge it in

berndbischl:
> The tests do produce some warnings. Some of them cannot be avoided, but some of them may be resolved by manipulating the data a little bit more (i.e. glm.fit fitted probabilities numerically 0 or 1)?
> warnings in tests suck. try to avoid them by changing the data a bit. but if you really cannot, thats how it is and we could live with it.

the learners_all tests throw 1000 warnings. you will never be able to fix all of them quickly. disregard this please, and focus on the rest here!

ja-thomas:
> > test_learners_all_cluster: There was a FIXME line saying "Remove this if DBScan runs stable". Since !inherits(lrn, "cluster.DBScan") was always TRUE I removed the FIXME and line. Please tell me, if this lines should be kept.

> dont touch this at all, has nothing to do with this PR

Well we have to at least rewite the statement. And I dont think the class "cluster.DBScan" exists, the object of `fpc` object is `dbscan` (all lower case). What was the "unstable" dbscan function that we have to exclude in the tests?

berndbischl:
there is a DBScan algo from RWeka. we had problems with this, so this in todo-files now, and not included in mlr at all. 

so all code and comments that refer to it, can and should be removed.

MariaErdmann:
I think this is ready to be checked and merged.

> there is a DBScan algo from RWeka. we had problems with this, so this in todo-files now, and not included in mlr at all.
> so all code and comments that refer to it, can and should be removed.

I did not keep the code refering to DBScan, so this is in my opinion resolved.

>the learners_all tests throw 1000 warnings. you will never be able to fix all of them quickly. disregard this please, and focus on the rest here!

Ok. I did not resolve the warnings.

>well, here is a better suggestion. in your testTrainPredict, if type = "prob" really DO acces the p object. with getPredictionProbabilities. and check the result! the line HAD its use! to really check that probs are actually contained in there! you now effectively disabled this check. 
think about like this: in your current code state, where do you check that probs are really produced? so add in back in, better add 1-2 MORE to inspect the output of that getter,.

I tried my best in my last commit. Can you please check if this is what you were thinking of?

> Please try to avoid such unnecessary "if statements".

Removed if statements and only used lrn = setPredictType(lrn, pred.type).

tests/testthat/helper_learners_all.R
```diff
@@ -34,3 +38,78 @@ testThatLearnerRespectsWeights = function(lrn, task, train.inds, test.inds, weig
...
+  
+  lrn = setPredictType(lrn, pred.type)
+  
+  expect_output(print(lrn), lrn$id)
+  m = train(lrn, task)
+  p = predict(m, task)
+  expect_true(!is.na(performance(pred = p, task = task)))
+  
+  if (pred.type == "se")
+    expect_equal(length(p$data$se), task$task.desc$size)
```
larskotthoff: Use `getTaskSize()` instead of `task$task.desc$size` please.

tests/testthat/helper_learners_all.R
```diff
@@ -1,11 +1,15 @@
...
+
 # test that a given learner respects its weights tag. we do this:
 # train without weights, with weights = 1, and with changed weights
 # then we check that changed weights actually change the output
 # note: it is not always easy to find a good data and weights combo to reliably achieve this,
 # as for some learners we need very extreme weights and for others exteme weights cause error.
 #
-# args: we pass a learner, task, train and test.inds, and weight vec for train inds
-# hyperpars is the list we set up in learners all where we need to deviate from the defaults for stability
+# further args: train and test.inds, and weight vec for train inds
```
larskotthoff: Could you keep the explanation what the `hyperpars` argument is please?

berndbischl:
@larskotthoff 
> Could you keep the explanation what the hyperpars argument is please?

thats doced in the header of the file for all following functions?

berndbischl:
ok i will take this over. the branch is mine now and i will merge it then

berndbischl:
@MariaErdmann  done?

MariaErdmann:
Yes, done!
Sorry, I said that there was no stuff with the DBScan in the cluster script, but there was. Removed it in my last commit.

MariaErdmann:
Don't know why Jenkins is failing (on monday I need to ask you were to look at in jenkins to see what happend), locally tests and check went fine. Travis is also ok.

PR #963, opened 2016-06-20
trying to fix the master build (again)

larskotthoff:


berndbischl:
this was an obvious oversight from me before

we removed the surv.penalized learner (which NEWS already stated as removed) but i forgot the test

PR #964, opened 2016-06-20
Container travis build (lets see what happens)

berndbischl:


berndbischl:
IMPORTANT:

the classiflabelswitch test sometimes breaks with h2o.(deeplearning?)

i have now set a seed in helpers_zzz for the unit tests and removed the h2o learners from classiflabelswitch. when then this here runs, we can merge this, but should create an issue out this thing here.

PR #965, opened 2016-06-22
Remove unnecessary ggplot2:: calls

MasonGallo:
I noticed some `ggplot2::` calls which shouldn't be necessary since it's already imported.

- [x] check passed
- [x] base tests passed


catastrophic-failure:
Is it good practice though? Genuinely asking.

zmjones:
I mean R's namespace management is just not great generally. I think we are avoiding this by only importing a few packages. I agree that using `::` isn't necessarily bad, but for ggplot2 it gets very tedious.

MasonGallo:
Agreed with @zmjones. 

@catastrophic-failure I'm not sure what you mean in the edit? I don't think that should be an issue for ggplot2. In this case we have already imported ggplot and shouldn't have any namespace issues (not to mention we call just ggplot elsewhere in the codebase).

larskotthoff:
I think @catastrophic-failure means that there could be issues if another package has functions of the same name as the ggplot functions. Yes, this is possible, but unlikely as it would break things for pretty much everybody else as well.

Merging, thanks.

Issue #969, opened 2016-06-27
remove stop("Unknown predict type") from all learners


berndbischl:
this is simply code bloat, we dont want this

Issue #966, opened 2016-06-23
surv.coxph and surv.ranger can't do probabilites


MariaErdmann:
Hi,
surv.coxph and surv.ranger do have properties "prob" but they actually can't do probabilities (yet).
While surv.coxph is throwing an error when predicting that survival cannot do probabilities yet, surv.ranger throws an error when training that the predict type is unknown.
Should the property "prob" be removed from these two learners for now? Or is there something wrong in the example script?

````
# load_all()
# source("/tests/testthat/helper_objects.R")

listLearners("surv", properties = "prob")$class
lrn.coxph = makeLearner("surv.coxph", predict.type = "prob")
m.coxph = train(lrn.coxph, surv.task)
p.coxph = predict(m.coxph, surv.task)
# Error in checkPredictLearnerOutput(.learner, .model, p) : 
# Survival does not support prediction of probabilites yet.


lrn.ranger = makeLearner("surv.ranger", predict.type = "prob")
m.ranger = train(lrn.ranger, surv.task)
# Error in trainLearner.surv.ranger(.learner = list(id = "surv.ranger",  : 
# Unsupported predict type 

````

Cheers,
Maria

intiveda:
Please, in case you remove property 'probs' from the package could you update the documentation?

Do not make people lose their time trying to debug code.

larskotthoff:
Sorry, which documentation are you referring to?

intiveda:
You could find the reference here: 

http://mlr-org.github.io/mlr-tutorial/release/html/integrated_learners/ 
at Survival analysis (13), 

> Additional learner properties:
prob: Probabilities can be predicted

Thank you in advance

larskotthoff:
Ah, sorry about that -- the property got reintroduced accidentally! I've fixed that and you can find the updated tutorial at https://mlr-org.github.io/mlr-tutorial/devel/html/integrated_learners/ once it's been regenerated.

PR #970, opened 2016-06-27
Removes property 'probs' from surv.coxph and surv.ranger

MariaErdmann:
This solves Issue #966.
In surv.coxph I  outcommented the lines where predict.type == "prob" since it is not ensured if surv.coxph can do probabilities. According to a line in predictLearner.R "Survival does not support prediction of probabilites yet."
Furthermore, lines with stop("Unknown predict type") were deleted (solving Issue #969 for these two learners.


schiffner:
Just a quick comment/question:
Why are these `if(.learner$predict.type == "response")` lines necessary?
If the learner does not have the property `"probs"` I can't even construct a learner with `predict.type = "probs"` and therefore don't need to query this in `trainLearner` or `predictLearner`?

MariaErdmann:
I thought the same right now, I will remove this line as well, with the PR resolving #969 :-)

berndbischl:
> Just a quick comment/question:
> Why are these if(.learner$predict.type == "response") lines necessary?
> If the learner does not have the property "probs" I can't even construct a learner with predict.type = "probs" and therefore don't need to query this in trainLearner or predictLearner?

they are not necessary, but become helpful if we later add code for the "prob" case :)
conclusion: i dont really care if that "if" is in there. either remove it or let it stay.

what i really DO care about is this here:#971 
which NEVER makes sense and only bloats the code.

PR #615, opened 2015-12-15
parallel ensembles

zmjones:
this is not ready for merge but is here for review. the relevant issue is #603.

left to do are:

CostSensRegrWrapper
MultilabelBinaryRelevanceWrapper
CostSensWeightedPairsWrapper
StackedLearner

any others? I am currently parallelizing training and prediction using the same level.

i have written but haven't pushed tests for all of this.

PR #977, opened 2016-07-01
Multilabel wrapper all

giuseppec:
- [x] we need to add a reference + short description in the `note` field for the measures

- [x] we need manual tests for the measures like here: https://github.com/mlr-org/mlr/blob/master/tests/testthat/test_base_measures.R#L271 where we create a small matrix and compare the measure function with its expected output

- [x] write 1-2 pages for the tutorial

- [x] use `checkLearner(, props = ...)` instead of `hasLearnerProperties(...)`

- [x] check styleguide (also check if the data type is mentioned in the API)

- [x] set the `order` arg to `NULL` and explain that by default this does "random" sampling. Also, use a vector with label names instead of indices for the `order` arg. 

- [x] Make multilabel-randomForestSRC ready (see #991)

- [ ] implement Parallelization (Zachs PR #615)

- [ ] Review paper and send it to the RJournal


berndbischl:
please note that this PR #959 already contains generic code to test all basic multilabel RLearners 
(learners_all style)

giuseppec:
@PhilippPro could you help here and look at points 1-3 from the list above? I'll do 4-6.

PhilippPro:
Yes. next week.

PR #979, opened 2016-07-01
fixes xgboost missing property, fixes #978

### LINKED ISSUE #978: Learner properties of "regr.xgboost" not updated
Hi,

I am trying to use MLR package for xgboost regression and my data has **missing data points**.

The learner i have used is "regr.xgboost"

Xgboost has the capability to handle missing values but train function of MLR gives me an error that data has missing values and learner does not support that. I am using mlr package instead of xgboost directly because i need partial dependence plots which is not available in xgboost package.

On going deep into the function i discovered that properties associated with learner are only mentioned as - **numeric, factors and weights**.

i edited the function using trace command and added "missings" property and now the function is working fine.

Let me know if MLR package is built in a way to not allow missing data, also if there is any alternative way of running xgboost on missing data using MLR package.

berndbischl:
NEWS:
-- classif.xgboost, regr.xgboost: can now properly handle NAs (property was missing)

### DIFF
--- a/R/RLearner_classif_xgboost.R
+++ b/R/RLearner_classif_xgboost.R
@@ -22,7 +22,7 @@ makeRLearner.classif.xgboost = function() {
       makeUntypedLearnerParam(id = "eval_metric", default = "error"),
       makeNumericLearnerParam(id = "base_score", default = 0.5),
 
-      makeNumericLearnerParam(id = "missing", default = 0, tunable = FALSE),
+      makeNumericLearnerParam(id = "missing", default = 0, tunable = FALSE, when = "both"),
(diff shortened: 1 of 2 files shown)


larskotthoff:
This breaks the tests. And could you add a unit test for the specific bug please?

berndbischl:
> This breaks the tests. And could you add a unit test for the specific bug please?

i know, still trying to figure out why.

> This breaks the tests. And could you add a unit test for the specific bug please?

no, because this is handled in general in learners_all tests. and improved here: #r959
also note that this is not really a bug, but a missing property

berndbischl:
rebased on master, this should dump logs now

berndbischl:
handled in PR #982

PR #980, opened 2016-07-01
fix TuneControlRandom arg_check for maxit / budget, fixes 781

berndbischl:


berndbischl:
this is so trvial i will merge myself

Issue #972, opened 2016-06-27
randomforest: sampsize and strata


jakob-r:
From the manpage:

> strata	
> A (factor) variable that is used for stratified sampling.

> sampsize	
> Size(s) of sample to draw. For classification, if sampsize is a vector of the length the number of strata, then sampling is stratified by strata, and the elements of sampsize indicate the numbers to be drawn from the strata.

Maybe check for further missing params.

berndbischl:
@MariaErdmann 
can you please add these 2 params and check the help page for other missing params?

berndbischl:
strata: add them to classif.rf and regr.rf. make it "untyped"

sampsize: i guess make it an intvec param then. for classif and regr.

berndbischl:
see discussion in PR.
this needs to be closed ASAP so we dont forget what happened

PR #984, opened 2016-07-04
Added and adapted params in randomForest (#972)

MariaErdmann:
Solving #972 

- Added param "strata": I was not sure which kind of LearnerParam to take. I wanted to go for makeCharacterLearnerParam() but I think this one is depricated, so I chose Untyped. I chose this one because you specify strata by putting the name of the varaible you want to stratify for in. I looked here for an example of strata: https://stat.ethz.ch/pipermail/r-help/2005-October/081557.html

- Added param "proximity" and "oob.prox", which produces a (nxn) proximity measures matrix 
- Added param "nPerm" for regression.
-changed sampsize to IntegerVectorParam

- I did not add "keep.forest", since this one needs always to be true  in order to get predictions.


MariaErdmann:
One of my commits said "added properties strata, proximity, oob.prox" --> of course I did not add properties. This should be "added params strata, proximity, oob.prox". Sorry for the confusion.

MariaErdmann:
Regarding the "strata" argument: I still need to take a closer look at it. I will then create also a test. Will do this in the course of the week.

larskotthoff:
Thanks, merging.

berndbischl:
> Regarding the "strata" argument: I still need to take a closer look at it. I will then create also a test. Will do this in the course of the week.

@larskotthoff @MariaErdmann 
why is this merged then? was this now done or not?

berndbischl:
> I did not add "keep.forest", since this one needs always to be true in order to get predictions.

this could be added. mlr could also be used with just training the model, but not predicting from it

berndbischl:
also the issue is still open. i dont know how to proceed

berndbischl:
> Well, this PR successfully closes the Issue. For everything else make a new issue if necessary.

if the PR creator says: "please wait, i am not done": we *wait*. an alternative would be to ask the creator. 

@MariaErdmann: please add comments what should be done from your side in the issue. ASAP. and also add keep.forest as a new param. do this in a new PR. 

i would like to resolve this soon.

jakob-r:
>also add keep.forest as a new param. do this in a new PR. 

Then we also want this for ranger (and maybe others where predicting is optional)

larskotthoff:
I didn't sound to me like "wait" (and this PR is completely standalone).

Shall we have a "ready-to-be-merged" label that we require to be added to avoid this?

berndbischl:
> I didn't sound to me like "wait" (and this PR is completely standalone).
> 
> Shall we have a "ready-to-be-merged" label that we require to be added to avoid this?

our conclusion was: the PR creator should explicitly say whether he is done. otherwise we ask if we are unsure.

(at least for me this PR is an example of: "i am not done", at least it we should agree that it is not obvious whether maria is done here, so we should ask)

MariaErdmann:
Sorry for the confusion and getting back a bit late.
The learners were ready to be merged (provided that I did everything right, which I was not sure about). But I would want to make a test testing whether the strata argument works properly. So this still needs to be done although you say there is no need.
Regarding the "keep.forest" param: Shall I add it?
I will do a new PR with the test and the "keep.forest" param when you say "go" ;-)

berndbischl:
> But I would want to make a test testing whether the strata argument works properly. So this still needs to be done except you say there is no need.

you dont have to test every new learner param (i think we talked about this before?).
of course, mlr  options should be tested in general, but for learner params there are simply too many and also you dont do much more then simply pass "strata" down to the rf. 


> Regarding the "keep.forest" param: I will add it then, except there are objections.

ok, go. you can BRIEFLY look at ranger and randomforestsrc whether there we are missing the same param and add that too

MariaErdmann:
> ok, go. you can BRIEFLY look at ranger and randomforestsrc whether there we are missing the same param and add that too

Resolved in PR #992

PR #987, opened 2016-07-04
run all examples in travis

larskotthoff:


larskotthoff:
Ok, this fails because Travis doesn't get enough output. Closing.

berndbischl:
lars what are you doing here? we should run "dont run"?

larskotthoff:
That doesn't work in Travis (and was disabled). I tried enabling it, but it doesn't produce enough output for Travis.

berndbischl:
we discussed this, the result was this issue #989 
we are done here

PR #990, opened 2016-07-06
adds data overplotting argument to plotPartialDependence

zmjones:


R/generatePartialDependence.R
```diff
@@ -443,46 +445,67 @@ print.PartialDependenceData = function(x, ...) {
...
 #' @template ret_gg2
 #' @export
-plotPartialDependence = function(obj, geom = "line", facet = NULL, p = 1) {
+plotPartialDependence = function(obj, geom = "line", facet = NULL, p = 1, data = NULL) {
   assertClass(obj, "PartialDependenceData")
-  if (length(obj$features) > 2L & geom != "tile" & obj$interaction)
-    stop("To plot more than 2 features geom must be 'tile'!")
   assertChoice(geom, c("tile", "line"))
-  if (geom == "tile" & !obj$interaction)
+  if (obj$interaction & length(obj$features) > 2L & !geom == "tile")
```
larskotthoff: Could you use `geom != "tile"` please?

R/generatePartialDependence.R
```diff
@@ -443,46 +445,67 @@ print.PartialDependenceData = function(x, ...) {
...
+    assertDataFrame(data, col.names = "unique", min.rows = 1L,
+                    min.cols = length(obj$features) + length(obj$td$target))
+    assertSubset(obj$features, colnames(data), empty.ok = FALSE)
+  }
 
   if (!is.null(facet)) {
     assertChoice(facet, obj$features)
     if (!length(obj$features) %in% 2:3)
-      stop("generatePartialDependence must be called with two or three features to use this argument!")
+      stop("generatePartialDependenceData must be called with two or three features to use this argument!")
```
larskotthoff: Wrong function name here.

zmjones: how so? that looks correct to me.

R/generatePartialDependence.R
zmjones: fixed

R/generatePartialDependence.R
larskotthoff: `generatePartialDependenceData` in the error message in the `generatePartialDependence` function.

zmjones: i'm confused. this error message is in plotPartialDependence if the obj argument, which has class PartialDependenceData but is generated by generatePartialDependenceData, wasn't called with arguments that comport with what is requested in plotPartialDependence. this change is to fix an error, which was referencing generatePartialDependence, which isn't a function at all.

larskotthoff: My worry is that the user calls function `plotPartialDependence` (this is what I meant in the previous message) and sees an error message about `generatePartialDependenceData`. Especially if this is in a script it may not be obvious where the error is coming from and how to fix it.

zmjones: I think removing that bit would make it much more uninformative. This tells you how to fix it.

larskotthoff: I agree, but I'm suggesting something along the lines of "`plotPartialDependence` must be called with data with two or three....". Does that sound ok?

zmjones: so you want to remove the reference to the generation function? i don't think that will make things easier to debug.

larskotthoff: Yes, that would work as well I think. I'm only worried that people will be confused if they call one function and get an error message about another function.

zmjones: ok i'll do that. i don't think it is confusing since you can only input output from one function here.

zmjones: how is something like this

"obj argument created by generatePartialDependenceData was called with interaction = FALSE!"

larskotthoff: Yep, sounds good.

zmjones: fixed

larskotthoff:
Thanks, merging.

larskotthoff:
And as a final remark, please put additions for NEWS in the thread rather than the PR.

zmjones:
there is an "add to NEWS" thread?

larskotthoff:
No, I mean post it to the the relevant PR thread rather than the PR code.

zmjones:
ah so post the news addition in the thread and then the merger will add it?

zmjones:
@berndbischl the data argument to plotPartialDependence we talked about yesterday is now merged.

larskotthoff:
Yep, exactly. As NEWS would be touched by almost everything, there would be conflicts all over the place otherwise.

PR #991, opened 2016-07-06
Multilabel randomForestSRC

PhilippPro:
new branch for multilabel randomForestSRC.

Followed recommendations of #684.

### DIFF
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -191,7 +191,7 @@ getTaskFormula = function(x, target = getTaskTargetNames(x), explicit.features =
 #' @family task
 #' @export
 #' @examples
-#' task = makeClassifTask(data = iris, target = "Species")
+#' task = makeClassifTask(data = iris, target = "Species")m
 #' getTaskTargets(task)
 getTaskTargets = function(task, recode.target = "no") {
   UseMethod("getTaskTargets")
@@ -323,7 +323,7 @@ recodeY = function(y, type, td) {
     return(as.numeric(2L * (y == td$positive) - 1L))
   if (type %in% c("lcens", "rcens", "icens"))
     return(recodeSurvivalTimes(y, from = td$censoring, to = type))
-  if (type == "multilabel_factor")
+  if (type == "multilabel.factor")
     return(lapply(y, function(x) factor(x, levels = c("TRUE", "FALSE"))))
   stopf("Unknown value for 'type': %s", type)
 }
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -227,11 +227,12 @@ getTaskTargets.CostSensTask = function(task, recode.target = "no") {
 #'   with the input data.frame and an extra vector or data.frame for the targets.
 #'   Default is \code{FALSE}.
 #' @param recode.target [\code{character(1)}]\cr
-#'   Should target classes be recoded? Supported are binary classification and survival.
+#'   Should target classes be recoded? Supported are binary and multilabel classification and survival.
 #'   Possible values for binary classification are \dQuote{01}, \dQuote{-1+1} and \dQuote{drop.levels}.
 #'   In the two latter cases the target vector is converted into a numeric vector.
 #'   The positive class is coded as \dQuote{+1} and the negative class either as \dQuote{0} or \dQuote{-1}.
 #'   \dQuote{drop.levels} will remove empty factor levels in the target column.
+#'   In the multilabel case the logical targets can be converted to factors with \dQuote{multilabel.factor}.
 #'   For survival, you may choose to recode the survival times to \dQuote{left}, \dQuote{right} or \dQuote{interval2} censored times
 #'   using \dQuote{lcens}, \dQuote{rcens} or \dQuote{icens}, respectively.
 #'   See \code{\link[survival]{Surv}} for the format specification.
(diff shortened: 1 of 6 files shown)


larskotthoff:
Is this ready to be merged?

Please use `stri_paste` instead of `paste`. Your test doesn't seem to check anything?

PhilippPro:
Yes there is missing something. I just copied the test from https://github.com/mlr-org/mlr/blob/master/tests/testthat/test_classif_randomForestSRC.R.
Can I just apply `testSimpleParset` analogously to the randomForestSRC case (see below)? 
I do not know where this function comes from... 

```
  testSimpleParsets("classif.randomForestSRC", binaryclass.df, binaryclass.target, binaryclass.train.inds, old.predicts.list, parset.list)
  testProbParsets ("classif.randomForestSRC", binaryclass.df, binaryclass.target, binaryclass.train.inds, old.probs.list, parset.list)
```

larskotthoff:
The functions are defined in helper_helpers.R. You should be able to just use them.

berndbischl:
> f = as.formula(paste("cbind(", paste(targets, collapse = ",", sep = " "), ")  ~ .", sep = ""))

I dont like this. Can you please help to extend getTaskFormula so such a formula is created?
That function already supports "multilabel" but i guess you need an extra option to include a "cbind" here for the left hand side, where the targets are.

berndbischl:
```
+  d = getTaskData(.task, .subset)
 +  for(i in targets)
 +    d[i] = factor(d[[i]], levels = c("TRUE", "FALSE"))
```

I also dislike this here. we probably want to have an option in getTaskData to do that conversion there.

PhilippPro:
@berndbischl 
1. Do you really think we need a cbind option in getTaskFormula? This is a very special case, isn't it?
2. Maybe the same for getTaskData

Can somebody tell me, what is doing jenkins in comparison to travis? (and how I can login to jenkins)

schiffner:
Two mini things about the learner params:
* For consistency the `forest` parameter should also be included (see @MariaErdmann 's PR #992 )
* In the newest version of randomForestSRC observation weights are supported.

larskotthoff:
@PhilippPro Email me to get a Jenkins account.

berndbischl:
> 1. Do you really think we need a cbind option in getTaskFormula? This is a very special case, isn't it?
> 2. Maybe the same for getTaskData

for 1) i might agree. for 2) this seems pretty "expectable" that some methods would like to have the target as factors?


> In the newest version of randomForestSRC observation weights are supported.

yup, see "case.wt". so you need to set the property and pass .weights to this.

berndbischl:
@PhilippPro 
Philipp:

```
  if (type == "multilabel_factor")
    return(lapply(y, function(x) factor(x, levels = c("TRUE", "FALSE"))))
```

You have performed a change to one of mlr's most important core functions:
- you did not document this
- all other options for recode.target used a "dot" as word separator, you now used a "_". please call it 
- you did not unit test this function itself

please always thing about stuff like this when you do API changes

PhilippPro:
All the others are "drop.levels" ?;). 

I just did this based on your recommendation, so I thought this is obvious, that I will change something there. 
Where are the tests for TaskOperators?

berndbischl:
> All the others are "drop.levels" ?;).

yes, but also true for other functions in mlr! (its basically the style guide)

> I just did this based on your recommendation, so I thought this is obvious, that I will change something there.

this sentence does not "parse" for me. do you mean: 
"my PR was not  done, it was just a first try"
in general, please communicate a bit after more commits to the PR so we know whats going on.
either say: "ok did some stuff, but note done" or "hey, i guess this is OK from my side, pls review!"

> Where are the tests for TaskOperators?

i guess test_base_getTaskData is the best place!

thx!

PhilippPro:
Ok, I think it is ready to be merged (although Bernd will surely find some mistakes ;)). 
Sorry for the amount of commits, I sometimes just see the mistakes after committing.

berndbischl:
> Sorry for the amount of commits, I sometimes just see the mistakes after committing.

commits can now be squashed from our side very easily on merge, so i dont care if you 100 of those, where in each one you convert a single underscore in a dot :)

MasonGallo:
@PhilippPro I sometimes do this too...I edited the wiki to @berndbischl's point so hopefully it's clear now 👍

berndbischl:
> @PhilippPro I sometimes do this too...I edited the wiki to @berndbischl's point so hopefully it's clear now

thx!

R/Task_operators.R
```diff
@@ -323,6 +323,8 @@ recodeY = function(y, type, td) {
     return(as.numeric(2L * (y == td$positive) - 1L))
   if (type %in% c("lcens", "rcens", "icens"))
     return(recodeSurvivalTimes(y, from = td$censoring, to = type))
+  if (type == "multilabel.factor")
```
schiffner: Ideally, this option should also be documented (`recode.target` param of `getTaskData`, l. 229`)

berndbischl:
1) pls doc the API change as Julia (and I before...) requested

2) expect_true(all(sapply(df[, multilabel.target], is.factor)))

this is really not enough for a proper test. you also need to check that the conversion has the correct contents.  this is most simply if you create a mini-task with very few rows and cols for the test...

3) multilabel.formula = as.formula(paste("cbind(", paste(multilabel.target, collapse = ",", sep = " "), ")  ~ .", sep = ""))
 
*YOU* were the one who told me that this formula structure does LIKELY ONLY holds for this learner. that was your argument not to change the getTaskFormula getter. Now you put it in a place where it would need to hold for many other learners to be of use.....?
At least call it maybe multilabel.formula.cbind

PhilippPro:
In my opinion it is mergeable now.

PR #667, opened 2016-01-17
new function getFeatureImportance

florianfendt:
What I did:
1. I cloned PR #395, ended up changing quite a lot in there, though.
2. Added the regression and survival counterparts for the learners in #395.
3. Added methods for ranger and randomForestSRC.

There was some documentation in #395 that explained how to implement new methods. I put it in an own section and left it in there for now, although I think this should go to the tutorial, right?

Possible update for NEWS:
````
- Added a new function getFeatureImportance to extract feature importance measures for
  certain learners. So far the following learners are supported:
  boosting, cforest, gbm, randomForest, randomForestSRC, ranger, rpart 

- new functions
-- getFeatureImportance
````

PR #995, opened 2016-07-08
Feature Importance fix merge conflict and move to branch (MIGHT BE DONE, BUT BASE CHANGE. REVIEW PROPERLY NOW!)

ja-thomas:
I resolved the merge conlict in #667  and moved it from a fork to a branch.

Hopefully we can merge this now..


larskotthoff:
Thanks, looks good! I've noticed a few things:

- `classif.gbm` says it supports feature importance but doesn't have the method.
- The hard-coded list of classifiers that support this seems brittle. Is there a way to auto-generate this?
- Not all the learners that need special parameters set are checking this in their feature importance implementations. What happens when "invalid" parameters are set there?

Thanks for your work on this!

R/RLearner_classif_cforest.R
```diff
@@ -65,3 +65,14 @@ predictLearner.classif.cforest = function(.learner, .model, .newdata, ...) {
   }
   p
 }
+
+getFeatureImportance.classif.cforest = function(.learner, .model, AUC = FALSE, ...) {
```
berndbischl: AUC --> auc

R/RLearner_classif_randomForest.R
```diff
@@ -50,3 +50,28 @@ predictLearner.classif.randomForest = function(.learner, .model, .newdata, ...)
   type = ifelse(.learner$predict.type=="response", "response", "prob")
   predict(.model$learner.model, newdata = .newdata, type = type, ...)
 }
+
+getFeatureImportance.classif.randomForest = function(.learner, .model, ...) {
```
berndbischl: i am 100% we want to generalize this!

berndbischl:
> The hard-coded list of classifiers that support this seems brittle. Is there a way to auto-generate this?

I talked to Lars about this. He's referring to the list of the featimp procedures in the roxygen doc page. The list is in principle very good, but one has to remember that it exists later and it is also a maintenance burden. But we still agreed that for now we keep it like it is.

berndbischl:
Wrt the unit test: test_base_getFeatureImportance.R
In "base" unit tests, test that the GENERAL getFeatureImportance works. Do NOT iterate over all learners here. Do that in learners_all. And create a general function for that test (kinda like you did) in helpers_learners_all.

R/RLearner_classif_randomForest.R
ja-thomas: Not sure if I understand you here. So we have parameters that are required to be a certain value (e.g. for ranger `importance = "impurity"` or `"permutation"`). And we should set an additional flag in the learner when the conditions to calculate variable importance are fullfilled, so that when we actually call `getFeatureImportance` we can just assert the flag?

ja-thomas:
Just as a note, we also need importance calculations for `xgboost` and `rrf`. I'll implement them tonight or tomorrow.

ja-thomas:
We should also add a function to plot the importance.

berndbischl:
> We should also add a function to plot the importance.

i would suggest not to muddle this PR. create an extra issue out of this please

berndbischl:
> Just as a note, we also need importance calculations for xgboost and rrf. I'll implement them tonight or tomorrow.

tell me whether you really want to do this now, or also create an extra issue.

ja-thomas:
>> Just as a note, we also need importance calculations for xgboost and rrf. I'll implement them tonight or tomorrow.

> tell me whether you really want to do this now, or also create an extra issue. 

That is already done ;)

berndbischl:
unit test: I am seeing multiple times:

"testThatLearnerHandlesMissings"

that is wrong? you want to test for featimp!

ja-thomas:
> unit test: I am seeing multiple times:
> 
> "testThatLearnerHandlesMissings"
> 
> that is wrong? you want to test for featimp!

damn, stupid copy paste error, fixing this asap...

ja-thomas:
This should be ready to merge now

berndbischl:
this must be reviewed by multiple people

paging @larskotthoff @schiffner @zmjones @jakob-r

larskotthoff:
Looks good to me, with the caveats mentioned above.

ja-thomas:
> Not all the learners that need special parameters set are checking this in their feature importance implementations. What happens when "invalid" parameters are set there?

I talked with Bernd about this. The problem is that this really depends on the method, which and how parameters have to be set. E.g. the  importance  of `randomForest` works with `type=1` even if `importance=FALSE`, but for `type=2` the importance has to be set to `TRUE` in the learner.

I'm not sure if there is an easy way to implement these checks in the learner itself. Currently the easiest/best solutions seems to catch them in the importance methods, customized for every method, and then give a meaningfull (!) error. 


For the handwritten list I'm not sure if that can be automatized. I think we just have to keep in mind that in every PR of a new learner with an importance method, the general documentation has to be adapted as well. Which is not ideal but doable.

R/Learner_properties.R
```diff
@@ -57,11 +57,11 @@ hasProperties = function(learner, props) {
 
 getSupportedLearnerProperties = function(type = NA_character_) {
   p = list(
-    classif    = c("numerics", "factors", "ordered", "missings", "weights", "prob", "oneclass", "twoclass", "multiclass", "class.weights"),
+    classif    = c("numerics", "factors", "ordered", "missings", "weights", "prob", "oneclass", "twoclass", "multiclass", "class.weights", "featimp"),
```
schiffner: Property `featimp` needs also to be documented (see l. 8 in this file)

R/RLearner_classif_RRF.R
```diff
@@ -49,3 +49,28 @@ predictLearner.classif.RRF <- function(.learner, .model, .newdata, ...) {
   p = predict(object = .model$learner.model, newdata = .newdata, type = type, ...)
   return(p)
 }
+
```
schiffner: Don't we need an `#'export` here to register the S3 method in the NAMESPACE?
(The same for all other learners.)

berndbischl:
also why do we add ipred to SUGGESTS?

R/RLearner_classif_RRF.R
berndbischl: > Don't we need an #' @export here to register the S3 method in the NAMESPACE?
> (The same for all other learners.)

yes of course

R/RLearner_regr_cforest.R
```diff
@@ -59,3 +59,14 @@ trainLearner.regr.cforest = function(.learner, .task, .subset, .weights = NULL,
 predictLearner.regr.cforest = function(.learner, .model, .newdata, ...) {
   as.vector(predict(.model$learner.model, newdata = .newdata, ...))
 }
+
+getFeatureImportance.regr.cforest = function(.learner, .model, AUC = FALSE, ...) {
```
schiffner: The AUC option is not needed for regression.
What I found puzzling at first is that I could call `varimpAUC` for a regression forest without getting an error or a warning.
But when you look at the source code of `varimp` and `varimpAUC` you see that the AUC option is only relevant for binary regression and that both functions do exactly the same in the regression case (decrease in MSE).

berndbischl: also AUC --> auc, again

schiffner:
> For the handwritten list I'm not sure if that can be automatized. I think we just have to keep in mind that in every PR of a new learner with an importance method, the general documentation has to be adapted as well. Which is not ideal but doable.
Alternatively, we could put the parts of the documentation that refer to `getFeatureImportance` methods for single learners in the learner files (with `@rdname getFeatureImportance`).
Then it's less likely that someone forgets to update the docs.

ja-thomas:
> also why do we add ipred to SUGGESTS?

the importance for `cforest` requires `ipred`.

ja-thomas:
> Don't we need an #' @export here to register the S3 method in the NAMESPACE?
(The same for all other learners.)

So the thing is that we use a general `getFeatureImportance`, which uses only a `model` as input, but the S3 methods have to be defined by the learner `model$learner`. Currently we use  `getFeatureImportance` to call a  (unexported) function `getFeatureImportanceS3()`, which uses the actual learner as the first input. The methods of the different learners are then based on this function, not the original  `getFeatureImportance()`. But this yields problems when we export the methods, since the learner argument is not defined there. 


I'm not 100% sure whats the correct approach to handle something like that.

@schiffner Thanks, I fixed the other stuff.

schiffner:
Thanks @ja-thomas for your work on this.

> So the thing is that we use a general getFeatureImportance, which uses only a model as input, but the S3 methods have to be defined by the learner model$learner. Currently we use getFeatureImportance to call a (unexported) function getFeatureImportanceS3(), which uses the actual learner as the first input. The methods of the different learners are then based on this function, not the original getFeatureImportance(). But this yields problems when we export the methods, since the learner argument is not defined there.

> I'm not 100% sure whats the correct approach to handle something like that.

Thanks for the explanation, I overlooked `getFeatureImportanceS3`...
Maybe, we could handle this analogously to `train`/`trainLearner` and `predict`/`predictLearner`:

* `getFeatureImportanceS3` (one could also rename it to `getFeatureImportanceLearner` for consistency) is exported and documented (you kinda did that already in the "implementing new methods section").
Exporting would also allow users to write their own methods for custom learners.
* For the individual learners one writes S3 methods for `getFeatureImportanceS3` (with export tag).

@berndbischl , @larskotthoff : What do you think?

ja-thomas:
@larskotthoff  @schiffner @berndbischl @jakob-r @mllg @zmjones 

If you have time, please take a look. I added the suggestion by @schiffner

larskotthoff:
Just for the record, I like @schiffner 's suggestion to have S3 methods for the feature importance.

Please also document the `featimp` flag. I also like the suggestion of having the documentation be part of the learner, e.g. as a new note-type thing.

man/getFeatureImportance.Rd
```diff
@@ -0,0 +1,79 @@
...
+\details{
+The following learners have been implemented with the associated importance measures.
+
+\itemize{
+   \item{boosting} \cr
+   {Measure which accounts the gain of Gini index given by a feature 
+   in a tree and the weight of that tree.}
+   \item{cforest} \cr
+   {Permutation principle of the 'mean decrease in accuracy' principle
+   in randomForest. If \code{AUC=TRUE} (only for classification and regression),
```
schiffner: AUC is only relevant for binary classification.

R/getFeatureImportance.R
```diff
@@ -0,0 +1,89 @@
...
+#' 
+#' The following learners have been implemented with the associated importance measures.
+#' 
+#' \itemize{
+#'    \item{boosting} \cr
+#'    {Measure which accounts the gain of Gini index given by a feature 
+#'    in a tree and the weight of that tree.}
+#'    \item{cforest} \cr
+#'    {Permutation principle of the 'mean decrease in accuracy' principle
+#'    in randomForest. If \code{AUC=TRUE} (only for classification and regression),
```
schiffner: The AUC is only relevant for (binary) classification.

R/RLearner_regr_cforest.R
```diff
@@ -59,3 +59,16 @@ trainLearner.regr.cforest = function(.learner, .task, .subset, .weights = NULL,
 predictLearner.regr.cforest = function(.learner, .model, .newdata, ...) {
   as.vector(predict(.model$learner.model, newdata = .newdata, ...))
 }
+
+#' @export
+#' @rdname getFeatureImportanceLearner
+getFeatureImportance.regr.cforest = function(.learner, .model, auc = FALSE, ...) {
```
schiffner: The AUC option is unnecessary.

R/getFeatureImportance.R
schiffner: And It think it must be auc (lower case).

R/RLearner_classif_RRF.R
```diff
@@ -49,3 +49,30 @@ predictLearner.classif.RRF <- function(.learner, .model, .newdata, ...) {
...
+#' @rdname getFeatureImportanceLearner
+getFeatureImportance.classif.RRF = function(.learner, .model, ...) {
+  mod = getLearnerModel(.model)
+  ctrl = list(...)
+  if (is.null(ctrl$type)) {
+    ctrl$type = 2L
+  } else {
+    if (ctrl$type == 1L) {
+      has.fiv = .learner$par.vals$importance
+      if (is.null(has.fiv) || has.fiv != TRUE)
```
jakob-r: better: isFALSE(has.fiv)

R/RLearner_classif_RRF.R
```diff
@@ -49,3 +49,30 @@ predictLearner.classif.RRF <- function(.learner, .model, .newdata, ...) {
...
 }
+
+#' @export
+#' @rdname getFeatureImportanceLearner
+getFeatureImportance.classif.RRF = function(.learner, .model, ...) {
+  mod = getLearnerModel(.model)
+  ctrl = list(...)
+  if (is.null(ctrl$type)) {
+    ctrl$type = 2L
+  } else {
```
jakob-r: better: `} else if (ctrl$type == 1L) {`

R/RLearner_classif_boosting.R
```diff
@@ -47,3 +47,12 @@ predictLearner.classif.boosting = function(.learner, .model, .newdata, ...) {
     return(as.factor(p$class))
   }
 }
+
+#' @export
+#' @rdname getFeatureImportanceLearner
+getFeatureImportance.classif.boosting = function(.learner, .model, ...) {
+  mod = getLearnerModel(.model)
+  fiv = as.numeric(mod$importance[.model$features])
+  names(fiv) = .model$features
```
jakob-r: a bit shorter in the last line: `setNames(fiv, .models$features)` return would not be needed then

R/RLearner_classif_cforest.R
```diff
@@ -65,3 +65,16 @@ predictLearner.classif.cforest = function(.learner, .model, .newdata, ...) {
...
+#' @export
+#' @rdname getFeatureImportanceLearner
+getFeatureImportance.classif.cforest = function(.learner, .model, auc = FALSE, ...) {
+  mod = getLearnerModel(.model)
+  if (auc) {
+    fiv = as.numeric(party::varimpAUC(mod, ...))
+  } else {
+    fiv = as.numeric(party::varimp(mod, ...))
+  }
+  names(fiv) = .model$features
```
jakob-r: again: `setNames()` is a bit shorter / better to read IMHI

jakob-r:
In general I like this idea a lot. Sometimes the code could be a bit more compact and maybe we can avoid some copy/paste for the classif/regr/surv sub methods.

zmjones:
so @ja-thomas talked and we are going to add a class (`FeatureImportance`) to this output that will match that of `generateFeatureImportanceData` (in PR #1083). rows are feature names and columns are the loss function, or for class specific importance there is a column for each class level. @ja-thomas also is going to add support for local importance, where rows will be the number of observations and there is a column for each feature.

zmjones:
@ja-thomas and i settled on column names for features, class levels or observations (for local importance) as rows.

R/RLearner_surv_rpart.R
```diff
@@ -40,6 +41,12 @@ predictLearner.surv.rpart = function(.learner, .model, .newdata, ...) {
   if (.learner$predict.type == "response") {
     predict(.model$learner.model, newdata = .newdata, type = "vector", ...)
   } else {
-    stop("Unsuporrted predict type")
+    stop("Unsuported predict type")
```
jakobbossek: It should be "Unsupported ..." 😉

larskotthoff:
Looks good to me, I think this can be merged.

zmjones:
maybe we can just do it after we merge this but we need to agree on a class definition for `FeatureImportance`. I have one in #1083 but it might be excessive. after we have that i will write a print method and put it in my PR.

berndbischl:
1) 
#' @section Implementing new methods:

this should be moved to the tutorial. and only briefly linked here,

2) check for rox template, especially WrappedModel

3) the s3 methods for the learners should be called getFeatureImportanceLearner.classif.foo

4)  getFeatureImportanceLearner = function(.learner, .model, ...) {
the docs can stay, but that learner can be a string is wrong. have minimal docs here. 
and say tha this is an INTERNAL functions which is there for the LEARNER API.

5) getFeatureImportanceLearner should have a "minimal" return type. this is a named vec.

6) remove the auc docs with roxygen. just doc it list where you explain all featimp methods.
(already there)

7) in getFeatureImportance briefly check what you got the model (cheaply).
so basically that it is a named vec. you can also compare the names to the features in the wrappedmodel i guess.

8) test_base_getFeatureImportance
you have copy paste here. just test the functionality with a very basic learner. randomforest

ja-thomas:
Section to preserve for later use in tutorial
```
#' @section Implementing new methods:
#' You can add more learners by implementing methods for them.
#' Your implementation must adhere to the following: \cr
#' The method must take arguments \code{.learner}, \code{.model} and \code{...}.
#' Where, the model was fitted on the subset of \code{.task} given by \code{.subset}. 
#' The parameters in \code{...} should be parameters of the underlying importance value
#' generating function, so they just need to be passed to it directly. 
```

zmjones:
@berndbischl i disagree with 5. although it isn't implemented yet here, this function will return "local" (observation specific) importance as well, which will require a data.frame in the data return slot. So I think it makes more sense to always return a data.frame, just one with a single row when there is only one importance for each feature. Also `randomForest` can return class-specific importance where we'd need one row per class level.

zmjones:
ping @berndbischl @larskotthoff this is ready for merge i think

i will add some features to this later to make it nicer but i would like to work on the tutorial with it as is now.

larskotthoff:
Another top-quality pull request that I just *love* merging.

schiffner:
@larskotthoff : Don't you think you are going a little overboard here? :wink:

larskotthoff:
@schiffner It was pointed out to me by a certain person (and that could have been *any* Zach involved with the project) that your merge messages for the tutorial are much nicer than mine. So I'm trying to improve.

schiffner:
@larskotthoff: This was pointed out to me by a certain Zach as well...
@zmjones: Please be assured that I will always love merging your PRs more than anyone else, regardless of what a certain Lars involved with the project is trying to imply... :innocent:

Sorry for the spam, will stop now.

zmjones:
:) such a happy project

Issue #996, opened 2016-07-08
Facetting in plotPartialDependence


schiffner:
Hi @zmjones 

I noticed this mini bug while fiddling with the tutorial:
Due to a missing `obj$` in this line:
https://github.com/mlr-org/mlr/blob/master/R/generatePartialDependence.R#L501
 `plotPartialDependence` currently isn't working with a factor facetting variable.

I also wanted to ask you about this example in the tutorial:
```
lrn.classif = makeLearner("classif.ksvm", predict.type = "prob")
fit.classif = train(lrn.classif, iris.task)
pd.int = generatePartialDependenceData(fit.classif, iris.task, c("Petal.Width", "Petal.Length"), TRUE)
plotPartialDependence(pd.int, facet = "Petal.Length")
# Error in plotPartialDependence(pd.int, facet = "Petal.Length") : 
#  The feature set as the facetting variable must be an integer or a factor.
```
This is currently broken and I can simply use some other data that contain factors or integers.
But was it really your intention that this doesn't work anymore?

zmjones:
I think what I intended was that you should use tile plots to do this sort of thing instead. However that wouldn't work with multiclass classification. 

My problem with using a numeric variable as a facet is that you can easily end up with a large number of facets which makes things less interpretable. Ideally, if you wanted to facet a numeric variable, you'd pick a smaller number of values for that variable when you are generating the partial dependence in the first place, but there currently isn't a way to do that, and that is separate from the step of plotting.

berndbischl:
> Do you think on balance that I should put this back in?

well, this really seems like a small thing an my 2cents are: does it hurt if we allow this? 
there are cases where it CAN help a user (if the num var does not have too many values).

i would argue: allow it, and doc in one sentence that this might not be the most perfect thing in some cases. but we also dont support an alternative at moment :)

in any case, would like to close this soon.

zmjones:
ok i will re-add it

schiffner:
Thank you very much for the fix @zmjones, and sorry for not getting back to this sooner.
I agree with Bernd on this :)

larskotthoff:
Is this worth mentioning in NEWS? I feel that it's probably not.

berndbischl:
> Is this worth mentioning in NEWS? I feel that it's probably not.

nope

PR #997, opened 2016-07-08
fix for issue #996

zmjones:
i didn't catch this in the tests because of the new data overplotting argument. now we have a case where that arg is omitted which would have caught this.


larskotthoff:
Thanks, merging.

Issue #472, opened 2015-09-12
irace: should support parallel tuning


berndbischl:
use hookRunParallel for this

larskotthoff:
Breaks tests, reverted. Please use PRs.

berndbischl:
hmm, i tested this pretty intensly locally, even looked at it under top. I will do a PR.

do you have a proposal for a unit test to check that paralleliazation REALLY works?

berndbischl:
(also: i was on the train)

larskotthoff:
From my quick look it seemed that something unrelated was breaking (in particular the test that checks that irace allows to specify the number of digits). I don't know if this only happens if there's no parallelization, but I was able to reproduce this locally.

berndbischl:
Argh, then you are perfectl right and I did NOT test this enough, sorry, my bad.

Still: Have an idea how to test that it indeed runs parallel?

larskotthoff:
Wallclock time vs cpu time? That probably won't work on Travis though.

berndbischl:
Which rules it then out.

berndbischl:
we should REALLY check that this line is correct

 # the instance is always the same for all different param setting
 +    rin = experiment[[1]]$instance

berndbischl:
fixed in PR #1001

PR #1001, opened 2016-07-11
#472 irace now parallel

florianfendt:
NEWS
````
- irace-tuning now also supports parallelization
````

### CODE REVIEW

**Comment by berndbischl** [comment]
as you can see this breaks travis. also look in the diff, you simply omitted some of the special cases I handled before, which is probably now the cause for the bugs.

please debug

**Comment by berndbischl** [comment]
you need this here. 

```

  hookRunParallel = function(experiment, hook.run, config = list()) {
    # get our param settings that irace should try
    cands = extractSubList(experiment, "candidate", simplify = FALSE)
    # the instance is always the same for all different param setting
    rin = experiment[[1L]]$instance
    cands = lapply(cands, dfRowToList, i = 1, par.set = par.set, enforce.col.types = TRUE)

    tunerFitnFunVectorized(cands, learner = learner, task = task, resampling = rin, measures = measures,
      par.set = par.set, ctrl = control, opt.path = opt.path, show.info = show.info,
      convertx = cx, remove.nas = TRUE)
  }

  n.instances = control$extra.args$n.instances
```

make sure irace is tested with one discrete and also 2 discrete params

PR #1004, opened 2016-07-11
improve GPfit learner  from Dotterbart master (learner for GPfit #916)

smilesun:
This pull request aims at solves some minor problems with <learner for GPfit #916>

with 
"
git checkout -b Dotterbart-master master
git pull https://github.com/Dotterbart/mlr.git master
"
After merging the conflict, I have made the following modifications are made:

1. makeNumericLearnerParam(id = "nug_thres", default = 20, lower = 10, upper = 25),
I have removed the upper=25

2.  makeUntypedLearnerParam(id = "optim_start", default = NULL),   I have removed the tunable=FALSE

3. makeIntegerLearnerParam(id = "maxit", default = 100, lower = 0), I have changed to lowwer=1

4. The most significant change is to the branching structure, the previous one is 

if (scale) {
    d$data[,not.const] = apply(d$data[,not.const], 2, function(x) x = (x - min(x)) / (max(x) - min(x)))
    res = GPfit::GP_fit(d$data[, not.const], d$target, ...)
    res = attachTrainingInfo(res, list(scaled = TRUE, not.const = not.const, high = high, low = low))
    return(res)
  } else {
    res = GPfit::GP_fit(d$data[, not.const], d$target, ...)
    res = attachTrainingInfo(res, list(scaled = FALSE, not.const = not.const))
    return(res)
  }

I have changed that to 

if (scale) {
    d$data[,not.const] = apply(d$data[,not.const], 2, function(x) x = (x - min(x)) / (max(x) - min(x)))
    mlist = list(scaled = TRUE, not.const = not.const, high = high, low = low)
  } else {
    mlist = list(scaled = FALSE, not.const = not.const)
  }
  res = GPfit::GP_fit(d$data[, not.const], d$target, ...)
  res = attachTrainingInfo(res, mlist)


berndbischl:
hi, i am not sure why you had to create a new PR.

but the code might look good now. 2 things:

a) can you please rename the learner to "GPfit"?

b) can you please figure out if GPfit supports posterior sd estimation, so predict.type = "se"?

smilesun:
@berndbischl 
1. OK.
2. Yes,  posterior sd = result$mse.

berndbischl:
hi
1) set name = "Gaussian Process" (shorter)
2) You also need to set property "se"

@jakob-r You also wanted to have this learner, can you also review once?

smilesun:
After reading through the two papers below ,besides the above modification, I want to discuss also to set the bound of "nug_thres" parameter to be [10,25] as what @Dotterbart  has done or just [1,100] or bigger.  This is based on the following arguments.
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Reference:
Ranjan, P., Haynes, R., & Karsten, R. (2011). A Computationally Stable Approach to Gaussian Process Interpolation of Deterministic Computer Simulation Data. Technometrics, 53(4), 366–378. http://doi.org/10.1198/TECH.2011.09141

Peng, R. D., & Dominici, F. (2015). GPfit: An R package for fitting a Gaussian Process Model to Deterministic Simulator outputs
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
For background knowledge, please first read some tutorial in GP and then the above papers

The nugget parameter come from the reason of computing the ML (theta) parameter for R can be unstable.  In the author’s words “the key problem here is the inaccurate computa-
tion of |R| and R−1 due to ill-conditioning of R”  “The computational stability of GP estimation algorithms can depend critically on the set of design points and corresponding simulator outputs that are used to build a GP model. If any pair of design points in the input space are close together, the spatial correlation matrix R may become near-singular and hence the GP model fitting procedure computationally unstable”

But the nugget could not be two big, otherwise, it is over-smooting the data and deviates from the true design. So it is necessary to set nugget to be the lower bound.

The ill-conditioning of R matrix (The correlation matrix of the initial design point) is defined by the condition number of R=\div{\lambda_n}{\lambda_1}, and in the paper, the author use a=25 as a upper bound for the log{condition number}, the author further states that "The lower bound on the nugget is only a sufficient condition and not a necessary one for Rδ to be well-conditioned" AND "If the correlation matrix is well- conditioned, R should be used instead of Rδ"That means when R itself is not ill-conditioned and the condition number is small enough compared to a=log(Cond)=25, there is no need to use the nugget parameter. 

According to the paper:
"
nug_thres: A threshold parameter used in the calculation of the lower bound of the nugget, δlb. Although Ranjan et al. (2011) suggest nug_thres = 25 for space-filling designs, we use a conservative default value nug_thres = 20. This value might change for different design schemes."

In fact, the number 25 is decided by the author in numerical experiment using Matlab. When a=25, the proportion of R matrix being near-singular is big. To make this safer, in GPfit, the author set a=20(nug_thres) instead of 25. Now the problem is, shall we set 25 as a upper bound for nug_thres? or let it be arbitrary? since this is a tunable parameter.

smilesun:
There is another problem, in the paper "Ranjan, P., Haynes, R., & Karsten, R. (2011). A Computationally Stable Approach to Gaussian Process Interpolation of Deterministic Computer Simulation Data. Technometrics, 53(4), 366–378. http://doi.org/10.1198/TECH.2011.09141"
In the experiment, the author said 
"For convenience, we rescale the input variables in [0,1]" and in fact, the empirical parameter nug_thres is found out by LHS in the cube of [0,1]^d, which is 25, so it remains a problem if this parameter still applies if the user use SCALE=FLASE

jakob-r:
@smilesun Thanks for digging deeper. However I defend the position that in *mlr* we should just give an interface to a learner. It's the duty of the user to have the complete understanding of what is going on inside. From this perspective the definition of the parameter set is more from the technical point of view. Which values can be handled? Answering this question might lead to nug_thres should be (0, Inf).
However we could add a warning, or at least a node. But we never did this in any other learner so we should not start with such a thing here.

berndbischl:
> @smilesun Thanks for digging deeper. However I defend the position that in mlr we should just give an interface to a learner. It's the duty of the user to have the complete understanding of what is going on inside. From this perspective the definition of the parameter set is more from the technical point of view. Which values can be handled? 
> Answering this question leads me to: nug_thres should be (0, Inf).

agree.  and just discussed exactly the same thing with xudong :)

> However we could add a warning, or at least a node. But we never did this in any other learner so we should not start with such a thing here.

nope. just set the technical limits and be done with it.

berndbischl:
> "For convenience, we rescale the input variables in [0,1]" and in fact, the empirical parameter nug_thres is found out by LHS in the cube of [0,1]^d, which is 25, so it remains a problem if this parameter still applies if the user use SCALE=FLASE

i would simply ignore this.

are we done here?

smilesun:
Then, I am done with my part

berndbischl:
i dont want to overdo this, but should for nug_thres lower = 0 then? you still have 10?

PR #1008, opened 2016-07-12
addOptPath always checks for feasibility (fixes #719)

### LINKED ISSUE #719: tuning: We store transformed x-vals in opt.paths.
But some part's in the code indicate that we actually want to store the original values.
What should we agree on?

jakob-r:


larskotthoff:
Thanks, merging.

PR #1018, opened 2016-07-14
make function signature and default parameter setting consistent

larskotthoff:
GPfit was implemented inconsistently with the rest of mlr. This causes warnings when checking -- it's not clear to me why Travis didn't catch this.


berndbischl:
> his causes warnings when checking -- it's not clear to me why Travis didn't catch this.

huh? i also dont get how this passed R CMD check?

but you need scale = TRUE in the signature as well

Issue #1016, opened 2016-07-14
measures PPV, TPR, FPR corner cases


giuseppec:
For PPV, TPR and FPR the value NaN is returned when the denominator is zero. 
For Multilabel PPV return 1 when denominator is zero, see:

```
measurePPV(c(1,1,1), rep(0, 3), positive = 1)
# [1] NaN
measureMultilabelPPV(cbind(c(1,1,1), c(1,1,1)), cbind(rep(0, 3), rep(0, 3)))
# [1] 1
```

should we keep this as it is or should we unify this and when the denominator is zero return NaN in case of the multilabel PPV (or return 1 in case of the PPV)? Maybe this discussion helps http://stats.stackexchange.com/questions/1773/what-are-correct-values-for-precision-and-recall-in-edge-cases

same question holds for tpr, fpr...

berndbischl:
code for better checks, needs to be finished

```

load_all()

getData = function(true1 = 0.5, pred1 = 0.5, n = 10L) {
  list(
    y = rbinom(n, prob = true1, size = 1L), levels = c("0", "1"),
    yhat = rbinom(n, prob = pred1, size = 1L), levels = c("0", "1")
  )
}

doSeq = function(arg, measfun, s, ...) {
  vnapply(s, function(r) {
    args = namedList(arg, r)
    d = do.call(getData, args)
    measfun(d$y, d$yhat, ...)
  })
}

checkMeasure = function(measfun, lower = 0, upper = 1, n = 10L, ...) {
  s = seq(0, 1, length.out = n)
  print("true1")
  z = doSeq("true1", measfun, s, ...)
  print(z)
  # expect_numeric(z, lower = lower, upper = upper, finite = TRUE, any.missing = FALSE)
  print("pred1")
  z = doSeq("pred1", measfun, s, ...)
  print(z)
  # expect_numeric(z, lower = lower, upper = upper, finite = TRUE, any.missing = FALSE)
}


n = 10L
# checkMeasure(measureTP, upper = n, positive = "0")
# checkMeasure(measureFP, upper = n, positive = "0")

# checkMeasure(measureTPR, positive = "1")
# checkMeasure(measureFPR, positive = "1", negative = "0")

checkMeasure(measurePPV, positive = "1")

# checkMeasure(measureTP)
# checkMeasure(measureTP)
# checkMeasure(measureFP)
# checkMeasure(measurePPV)
# checkMeasure(measurePPV)

```

smilesun:
[...]

Issue #1005, opened 2016-07-11
clean up multilabel code


berndbischl:
1) some multilabel measures sport "for-loops". Pls remove them

2) all multilabel wrappers basically have the same copy-paste example! please try to spot this. it is also on a dontrun block. lets have one example and reference this one from the other places. and the example should be without dontrun if possible.

@giuseppec 
@PhilippPro 

can you please resolve this asap?

giuseppec:
I started a branch https://github.com/mlr-org/mlr/tree/cleanup_multilabel and will do the first iteration, @PhilippPro I'll give you a sign when you can do the second iteration, hope this is okay.

PhilippPro:
yes

giuseppec:
@PhilippPro I did some changes regarding 1) in https://github.com/mlr-org/mlr/pull/1019. Could you do 2) and remove the examaples as bernd mentioned?

PhilippPro:
Done.

berndbischl:
i think the PRs for this are still not merged?

PR #1019, opened 2016-07-15
cleanup measures and extend tests

giuseppec:
This is a pull-request for: https://github.com/mlr-org/mlr/issues/1005

We might need to discuss this issue before merge https://github.com/mlr-org/mlr/issues/1016. For TPR, for example:

``` r
bincombo = matrix(c(TRUE, FALSE, TRUE, TRUE, FALSE, TRUE, FALSE, FALSE), ncol = 2, byrow = TRUE)
bincombo
#      [,1]  [,2]
#[1,]  TRUE FALSE
#[2,]  TRUE  TRUE
#[3,] FALSE  TRUE
#[4,] FALSE FALSE

# without the 4th line, TPR corresponds to the worst value
measureMultilabelTPR(bincombo[-4,], !bincombo[-4,])
#[1] 0

# 4th line yields a denominator of 0 for the computation of the TPR
measureMultilabelTPR(bincombo, !bincombo)
# [1] 0.25
```

For `measureMultilabelTPR(bincombo, !bincombo)` one would expect that either the worst TPR value from `multilabel.tpr$worst` (which is 0) is returned or `NaN` (when denominator is zero).


larskotthoff:
Is this ready to be merged?

Do I see it correctly that most of the new unit tests are simply comparing the old implementation to the new implementation? Do we really need this?

berndbischl:
the code cleanup is good. 
but please, instead of the for loops in the tests, add handcrafted simple examples.

see what i did around lines 309 for brier score and logloss

berndbischl:
@giuseppec 
what is the status here? i want to merge this soon

giuseppec:
Status: still waiting for an answer in https://github.com/mlr-org/mlr/issues/1016 regarding what to do in the corner cases for **multilabel** TPR and PPV. 

Because currently, we are doing this:

```r
(y = matrix(c(TRUE, FALSE, FALSE, TRUE, TRUE, TRUE, FALSE, TRUE), 4, 2))
#      [,1]  [,2]
#[1,]  TRUE  TRUE
#[2,] FALSE  TRUE
#[3,] FALSE FALSE
#[4,]  TRUE  TRUE
(pred1 = matrix(c(TRUE, FALSE, FALSE, TRUE, TRUE, TRUE, TRUE, TRUE), 4, 2))
#      [,1] [,2]
#[1,]  TRUE TRUE
#[2,] FALSE TRUE
#[3,] FALSE TRUE
#[4,]  TRUE TRUE
(pred2 = matrix(c(TRUE, FALSE, FALSE, TRUE, TRUE, TRUE, FALSE, FALSE), 4, 2))
#      [,1]  [,2]
#[1,]  TRUE  TRUE
#[2,] FALSE  TRUE
#[3,] FALSE FALSE
#[4,]  TRUE FALSE

measureMultilabelTPR(y, pred1)
# [1] 1

measureMultilabelTPR(y, !y) # the worst tpr value should be 0, right?
# [1] 0.25

measureMultilabelPPV(y, pred2) 
# [1] 1

measureMultilabelPPV(y, !y) # the worst ppv vlaue should be 0, right?
# [1] 0.5
```

Is it okay that we return the values above? If yes, I will push my local comit, otherwise we should decide what to do.

smilesun:
if we change the code like this, the result will make sense
```
bincombo = matrix(c(TRUE, FALSE, TRUE, TRUE, FALSE, TRUE, FALSE, FALSE), ncol = 2, byrow = TRUE)
truth = bincombo
response = !bincombo
numerator = rowSums(truth & response)  # 1&1 is 1, which is the only possible case
denominator = rowSums(truth)
t = numerator/denominator # I prefer to use apply here  

mean(ifelse(denominator == 0, 1, t)) # generate 0.25
ifelse(sum(denominator) == 0, 1, mean(t)) # generate NA
```

After this, we need to deal with the NA case

smilesun:
Another option would be` ifelse(sum(denominator) == 0, 1, mean(t, na.rm = TRUE))` in this case it generate 0 as expected by ignoring the label with only negative samples.

smilesun:
@berndbischl  ping

giuseppec:
@berndbischl and I agreed on letting 0/0 be NaN and using na.rm = TRUE when taking the mean (currently we replace the NaN with 1). This is also what the guys from the mldr package are doing (see https://cran.r-project.org/web/packages/mldr/vignettes/mldr.pdf). 
We also need to write a test that compares our measures from the ones from the mldr package.

berndbischl:
@giuseppec 
are the corner cases for the measures documented? 
otherwise can be merged

giuseppec:
@berndbischl I added a sentence (in the note of the measure) stating what happens when the denominator becomes 0.

PR #1020, opened 2016-07-15
Deleted the examples in all except BinaryRelevanceWrapper.

PhilippPro:
This is a pull-request for: #1005

- Deleted the examples in all except BinaryRelevanceWrapper.
- Referenced to the BinaryRelevanceWrapper example.
- Made some small changes in this example.


PhilippPro:
We could also have an external example and reference to it from all learners with 
@example path/relative/to/packge/root

berndbischl:
> We could also have an external example and reference to it from all wrappers with 
> @example path/relative/to/packge/roo

ok i have not used this before. but you cannot do \code\link in at-examples. 
that are roxygen / rdoc tags

berndbischl:
if you know how this works put it in an extra file and include / link that

PhilippPro:
and the file, where should I put it?

berndbischl:
you need to out this in inst/examples, then link to this with a relative path

in general this is better then simply "telling" the user where the example code is

you can see how we did this in openmlr here


https://github.com/openml/openml-r/blob/master/R/convertOMLTaskToMlr.R
https://github.com/openml/openml-r/blob/master/inst/examples/convertOMLTaskToMlr.R

PhilippPro:
this should be ready now

PR #1021, opened 2016-07-15
generateFilterValuesData: Order filter values correctly

### LINKED ISSUE #940: Filter mrmr and generateFilterValuesData
* In "mrmr" `nselect` is misleading because one gets one score less than the supplied `nselect`.

* mrmr does not return scores in the original order of the features in the data set.
This is not taken into account when writing the results into the data.frame returned by `generateFilterValuesData` and therefore gives wrong results when more than one filter is used.

I guess it's easiest if mrmr returns the scores in the original order. But maybe this should also be checked in `generateFilterValuesData`.

```
generateFilterValuesData(iris.task, method = c("mrmr", "anova.test"), nselect = 3)
FilterValues:
Task: iris-example
          name    type       mrmr anova.test
1  Sepal.Width numeric 0.03281607  119.26450
2 Sepal.Length numeric         NA   49.16004
3  Petal.Width numeric 0.33922212 1180.16118
4 Petal.Length numeric         NA  960.00715

generateFilterValuesData(iris.task, method = "anova.test")
FilterValues:
Task: iris-example
          name    type anova.test
1 Sepal.Length numeric  119.26450
2  Sepal.Width numeric   49.16004
3 Petal.Length numeric 1180.16118
4  Petal.Width numeric  960.00715
```

schiffner:
This partly fixes #940. I will do a separate PR for the mrmr filter.

For some reason (looks like a git accident) we had two files test_base_filterFeatures.R and test_base_generateFilterValuesData.R with almost the same content. I tried to clean that up and added a test to check that the filter values are ordered correctly.


schiffner:
I think I'm done here. Can be reviewed.

larskotthoff:
Looks good to me.

Issue #662, opened 2016-01-14
mboost: add support for various families


giuseppec:
Here is a list of families that can be used in mboost: http://www.inside-r.org/packages/cran/mboost/docs/Family
Currently, we only support `Binomial()` and `AdaExp()` for classification (but don't support `Binomial(link = "probit")`) or families that have an additional arg. Also we should think of a solution how to handle additional args in mboost families (as they can be hyperparameters as well, as in the `Huber` family).
For regression we should at least also support `Poisson()` which we currently don't support. 

``` r
library(mlr)
library(mboost)

# works
makeLearner("classif.glmboost", family = mboost::Binomial())
# does not work
makeLearner("classif.glmboost", family = mboost::Binomial(link = "probit"))

# furthermore we do not support poisson regression
makeLearner("regr.glmboost", family = Poisson())
```

berndbischl:
I guess we add the families with hyperpars in the following way:
We set a default for their hyperpars.

We also introduce some a new param family2 (or whatever) which is untypped and you can pass what you want.

ok?

could you do that?

berndbischl:
introduce family.custom as an untyped param. if this is set, it has preference over "family"

MariaErdmann:
These are the families supported by boost_family and some of my thoughts with an arrow behind:
AdaExp()  --> already implemented in the classif-learners
AUC()  --> not implemented yet. Shall we implement this one for mboost's classif learners?
Binomial(link = c("logit", "probit"), ...)  --> right now we have Binomial() in classif_glmboost meaning that the loss function is the negative binomial log-likelihood
Gaussian()  --> default for regression
Huber(d = NULL)  --> for continous response. added this one in PR #1023.
Laplace()  --> for regression. already implemented in regr_glmboost
Poisson()  -->  We talked about adding this into the regression learner (see PR #1023 )
GammaReg(nuirange = c(0, 100))  --> for non-negative responses. I added this one in PR #1023 
CoxPH()  --> for survivial. alread implemented
QuantReg(tau = 0.5, qoffset = 0.5)  --> for regression. I added this one in #1023 
ExpectReg(tau = 0.5)  --> for regression. I added this one in #1023 
NBinomial(nuirange = c(0, 100))
 PropOdds(nuirange = c(-0.5, -1), offrange = c(-5, 5)) 
Weibull(nuirange = c(0, 100))  --> for survival
Loglog(nuirange = c(0, 100)) 
Lognormal(nuirange = c(0, 100))
[...]

Issue #1022, opened 2016-07-17
Quantile regression


bhvieira:
Basic questions, I'm adding Quantile Regression Forests (`quantregForest` from package of same name).  Working on my fork for now (https://github.com/catastrophic-failure/mlr/tree/quantregForest)
It's functionally the same as `randomForest` regression, but offers quantile estimation at prediction.

Now, two questions regarding the implementation:
- What should I actually predict? The default of the packages are `c(0.1, 0.5, 0.9)` quantiles. For now I set it at the median, like in `caret`. This question is also pertinent to further quantile regression methods we add.
- Should I add all those other functionalities `mlr` supports for `randomForest`?

PhilippPro:
Hi,
my first impression is, that that would be a complete new class of learners, cause maybe the user wants several quantiles at once and not only the median. In any case the user should be able to set the quantile to the value he wants. Do you know other quantile regression methods in R? (I've got a quantregRanger repo, but not actualised)

catastrophic-failure:
Linear Quantile Regression and Nonlinear Quantile Regression (`rq` and `nlrq` in `quantreg`).

There are also Kernel Quantile Regression (`kqr` in `kernlab`) and Quantile Regression Neural Network (`qrnn.fit` in `qrnn`, you ).

In all these you set the desired quantile `tau` during training.

PhilippPro:
I just discovered the robust lqr package

PR #1023, opened 2016-07-18
Support more families and a custom.family in mboost learners (refers to #662)

MariaErdmann:
Hi,

This is the first part of the PR that solve #662 .

Before changing the other mboost learners, please have a look at this commit and tell me if this is what you were thinking about @berndbischl 

I added these families to regr.glmboost: Huber, Poisson, QuantReg, ExpectReg, GammaReg
I added an untyped parameter: custom.family (sorry for turning around your suggestion: family.custom, @berndbischl , but I typed it the other way around a 100 times and found it more intuitive that way)
If custom.family is specified, then this one will be preferenced.
Added a test for custom.family. And added a test for family Poisso. Test is simple copy and paste from test_regr_glmnet.R


MariaErdmann:
Travis was failing, because regr_glmboost was not working if no family was set via the parameter set. Therefore, I added family = mboost::Gaussian() into par.vals in the learner. But I am not very happy with it, since it does not look very nicely when I specifiy a custom.family.
mini-example:

````
lrn = makeLearner("regr.glmboost", par.vals = list(family = mboost::Gaussian(),
  custom.family = mboost::Family(ngradient = function(y, f, w = 1) y - f,
    loss = function(y, f) (y - f)^2,
    name = "My Gauss Variant")))
lrn
````
produces:
Learner regr.glmboost from package mboost
Type: regr
Name: Boosting for GLMs; Short name: glmboost
Class: regr.glmboost
Properties: numerics,factors,weights
Predict-Type: response
Hyperparameters: m=mstop,family=Gaussian,custom.family=\<boost_family\>

and the last line, is IMHO not so good-looking. But for now, I do not have any better idea to implement the custum.family parameter like I did. But I am glad for new idesas/suggestions.

File: R/RLearner_regr_glmboost.R [hunk 1]
```diff
@@ -22,14 +26,21 @@ makeRLearner.regr.glmboost = function() {
...
   f = getTaskFormula(.task)
+  if (!is.null(getHyperPars(.learner)$custom.family)) {
+    family = custom.family
+  } else {
+    family = family
+  }
   if (is.null(.weights)) {
-    model = mboost::glmboost(f, data = d, control = ctrl, ...)
+    model = mboost::glmboost(f, data = d, control = ctrl, family = family, ...)
+    print(model)
```
ja-thomas: why print the model?

R/RLearner_regr_glmboost.R
```diff
@@ -22,14 +26,21 @@ makeRLearner.regr.glmboost = function() {
...
 #' @export
-trainLearner.regr.glmboost = function(.learner, .task, .subset, .weights = NULL, mstop, nu, m, risk, trace, stopintern, ...) {
+trainLearner.regr.glmboost = function(.learner, .task, .subset, .weights = NULL, mstop, nu, m, risk, trace, stopintern, family, custom.family, ...) {
   ctrl = learnerArgsToControl(mboost::boost_control, mstop, nu, risk, trace, stopintern)
   d = getTaskData(.task, .subset)
   f = getTaskFormula(.task)
+  if (!is.null(getHyperPars(.learner)$custom.family)) {
+    family = custom.family
+  } else {
+    family = family
```
ja-thomas: I dont really like that we overwrite the family automatically and dont give a message when costum family is used. Can we add a dummy value for family that is required to be set to use a costum family?

ja-thomas:
Do the families and links similar to glm.

Give a meaningfull error when both costum family and a general family are set.

R/RLearner_regr_glmboost.R
MariaErdmann: Deleted it

MariaErdmann:
So I removed the crossvalidation stuff from the fitting process and added/ adapted the missing families.
Regression: Added Hurdle() and NBinomial, deleted QuantReg() and ExpectReg() (which I previously did add, but relating to #1022 I removed them once again)
Classification: Added ProppOdds(), family that requires the response to be an ordered factor. And added AUC(). I did not add Multinomial (see #662).
Survival: Added Gehan()

Regarding:

> 
Do the families and links similar to glm.
Give a meaningfull error when both costum family and a general family are set.

I am sorry but trying for a quite long time I don't know how to do that. In regr.glm the family is an DiscreateLearnerParam, looking like that

`makeDiscreteLearnerParam(id = "family", default = "gaussian",
        values = c("gaussian", "Gamma", "inverse.gaussian", "poisson"))`

So here the values are characters and in the train function I can easily do: if (family == "Gamma") .... (acutally there is no if statement, but a switch statement).

but in regr.glmboost, the DiscreateLearnerParam is looking like that:

`makeDiscreteLearnerParam(id = "family", default = mboost::Gaussian(), values = list(Gaussian = mboost::Gaussian(), Laplace = mboost::Laplace()))`

Here the values are not characters, are they? And therefore I cannot do something like
if (family == Binomial) or if (family == mboost::Binomial()) 

Would be glad to get a hint.

Also I am still quite confused about how to define the 'custom.family' as a value of family

I tried:

```` 
makeDiscreteLearnerParam(id = "family", default = mboost::Gaussian(),
        values = list(Gaussian = mboost::Gaussian(), Laplace = mboost::Laplace(),
          Huber = mboost::Huber(), Poisson = mboost::Poisson(), GammaReg = mboost::GammaReg(nuirange = c(0,100)),
          NBinomial = mboost::NBinomial(), Hurdle = mboost::Hurdle(), "custom.family"))
````

```` 
makeDiscreteLearnerParam(id = "family", default = mboost::Gaussian(),
        values = list(Gaussian = mboost::Gaussian(), Laplace = mboost::Laplace(),
          Huber = mboost::Huber(), Poisson = mboost::Poisson(), GammaReg = mboost::GammaReg(nuirange = c(0,100)),
          NBinomial = mboost::NBinomial(), Hurdle = mboost::Hurdle(), custom.family = "custom.family"))
````

```` 
makeDiscreteLearnerParam(id = "family", default = mboost::Gaussian(),
        values = list(Gaussian = mboost::Gaussian(), Laplace = mboost::Laplace(),
          Huber = mboost::Huber(), Poisson = mboost::Poisson(), GammaReg = mboost::GammaReg(nuirange = c(0,100)),
          NBinomial = mboost::NBinomial(), Hurdle = mboost::Hurdle(), custom.family))
````

I am sorry that I need to ask that  but I am missing experience here and also would be glad to get a hint in order to continue my work.

**Comment by MariaErdmann** [comment]
Hi,
I implemented the families, just like it is in  regr_glm.
So now, it is possible to have a custom family. Moreover, parameters like `nuirange` can be set for families like NBinomial, GammaReg,
and Hurdle, parameter `d` can be set for Huber.
Please review and tell me if it is ok, then I will adapt the other learners as well (classif and surv).

Shall I suggest something for NEWS?
I did not make an extra error message, since mlr tells the user that "custom.family.definition" is missing, when the user chose `family = "custom.family"` and forgot the definition in "custom.family.definition.

File: R/RLearner_classif_glmboost.R [hunk 3]
```diff
@@ -4,27 +4,37 @@ makeRLearner.classif.glmboost = function() {
     cl = "classif.glmboost",
     package = "mboost",
     par.set = makeParamSet(
-      makeDiscreteLearnerParam(id = "family", default = mboost::Binomial(), values = list(AdaExp = mboost::AdaExp(), Binomial = mboost::Binomial())),
+      makeDiscreteLearnerParam(id = "family", default = "Binomial",
+        values = c("Binomial", "AdaExp", "AUC", "PropOdds", "custom.family")),
       # FIXME default of glmboost() for family is Gaussian()
```
ja-thomas: Do we really need that FIXME? I would guess not, since we are using classification and already have a note about the default

R/RLearner_classif_glmboost.R
```diff
@@ -4,27 +4,37 @@ makeRLearner.classif.glmboost = function() {
...
+      makeUntypedLearnerParam(id = "custom.family.definition", requires = quote(family == "custom.family")),
+      makeNumericVectorLearnerParam(id = "nuirange", default = c(-0.5,-1), requires = quote(family == "PropOdds")),
+      makeNumericVectorLearnerParam(id = "offrange", default = c(-5,5), requires = quote(family == "PropOdds")),
+      makeDiscreteLearnerParam(id = "Binomial.link", default = "logit", values = c("logit", "probit")),
       makeIntegerLearnerParam(id = "mstop", default = 100L, lower = 1L),
       makeNumericLearnerParam(id = "nu", default = 0.1, lower = 0, upper = 1),
       makeDiscreteLearnerParam(id = "risk", values = c("inbag", "oobag", "none")),
       makeLogicalLearnerParam(id = "stopintern", default = FALSE),
```
ja-thomas: stopintern does not make sense becaus we do not do crossvalidation in the fitting anymore. Should we remove it? Not sure what the general procedure is, but this parameter doesn't do anything in our case

R/RLearner_classif_glmboost.R
```diff
@@ -4,27 +4,37 @@ makeRLearner.classif.glmboost = function() {
...
+      makeDiscreteLearnerParam(id = "family", default = "Binomial",
+        values = c("Binomial", "AdaExp", "AUC", "PropOdds", "custom.family")),
       # FIXME default of glmboost() for family is Gaussian()
+      makeUntypedLearnerParam(id = "custom.family.definition", requires = quote(family == "custom.family")),
+      makeNumericVectorLearnerParam(id = "nuirange", default = c(-0.5,-1), requires = quote(family == "PropOdds")),
+      makeNumericVectorLearnerParam(id = "offrange", default = c(-5,5), requires = quote(family == "PropOdds")),
+      makeDiscreteLearnerParam(id = "Binomial.link", default = "logit", values = c("logit", "probit")),
       makeIntegerLearnerParam(id = "mstop", default = 100L, lower = 1L),
       makeNumericLearnerParam(id = "nu", default = 0.1, lower = 0, upper = 1),
       makeDiscreteLearnerParam(id = "risk", values = c("inbag", "oobag", "none")),
```
ja-thomas: Same as stopintern, as long as we do not do early stopping we do not have an oobag error.

R/RLearner_classif_glmboost.R
```diff
@@ -49,12 +54,17 @@ trainLearner.classif.glmboost = function(.learner, .task, .subset, .weights = NU
...
+  fam = getLearnerParVals(.learner)$family
   if (.learner$predict.type  == "prob") {
-    td = .model$task.desc
-    p = p[, 1L]
-    levs = c(td$negative, td$positive)
-    return(propVectorToMatrix(p, levs))
+    if (fam %in% c("AdaExp", "AUC")) {
+      stopf("prediction.type = 'prob' not implemented for family %s", fam)
+    } else {
+      td = .model$task.desc
```
ja-thomas: What is going on in this else block, please add a short comment what is done here

ja-thomas:
There is also `blackboost` which is also a mboost function like glmboost, please adapt these function (`classif.blackboost`, `regr.blackboost`, `surv.blackboost`) analogously.

R/RLearner_classif_glmboost.R
MariaErdmann: Well, I am not completely sure how to handle this: glmboost with family Binomal can predict probabilities (this is why the learner has the "prob" property) But if you switch family to AUC or AdaExp only classes can be predicted. This is why I implemented an error message.
Will add a note if throwing an error for this purpose is ok!

R/RLearner_classif_glmboost.R
MariaErdmann: I unfortunately did not find a single example where stopintern was set to TRUE or where the risk parameter was used. But like you say, these parameters do nothing in our case. So the only reason to keep them is for the sake of completeness.

R/RLearner_classif_glmboost.R
berndbischl: > Well, I am not completely sure how to handle this: glmboost with family Binomal can predict probabilities (this is why the learner has the "prob" property) But if you switch family to AUC or AdaExp only classes can be predicted. This is why I implemented an error message.
> Will add a note if throwing an error for this purpose is ok!

yes, just add a note. also you dont even have to implement a custom error msg in mlr

R/RLearner_classif_glmboost.R
berndbischl: pls just keep the params for completeness sake. you never know what the user wants to do with the fitted model aftwerwards.

MariaErdmann:
Ok, I mixed up branches, tried to reset to a commit before the mix up happened but still there are files changed (like regr_xgboost) that were changed by the commits I reseted. Don't know how to solve this right now and will take it up again, when I am back from holdiays.

MariaErdmann:
Ok, this is ready for a (final) review. I implemented the changes, as discussed and added the missing families and the custom family option to blackboost. Working on the blackboost learner I figured that there was a bug in th tree_controls part in the train function (see #1244 ). I fixed this also in this branch. Tests are adapted (they test the default settings, some of the new families (i.e. for PropOdds and the count data families extra tests have been added).
Some families (like AUC and AdaExp cannot predict probabilities (we discussed this earlier and decided to point this out in the note)). But I am not completely sure about the error message thrown, if somebody regardless of the note tries to predict probs. See the small code snippet below:

````
lrn = makeLearner("classif.blackboost", par.vals = list(family = "AdaExp"), predict.type = "prob")
pred = predict(mod, binaryclass.task)
#  Error in p[, 1L] : incorrect number of dimensions 
````

larskotthoff:
If the families don't support this, please add a check and throw a suitable error message.

MariaErdmann:
> If the families don't support this, please add a check and throw a suitable error message.

Implemented. It works, but not sure if the check is good enough?


This PR fixes #1023, #662 , #1244 :-)

I chenged the filename of RLearner_regr_blackBoost.R to RLearner_regr_blackboost.R but unfortunately I cannot commit this change as it does not appear as a change to commit.

larskotthoff:
I'm almost happy now, but could you add a test for the families check?

If you rename a file, you need to tell git that you've done so by adding it `git add RLearner_regr_blackboost.R`. Then it will show up in the commit. You can also do this in one step with `git mv RLearner_regr_blackBoost.R RLearner_regr_blackboost.R`.

MariaErdmann:
> I'm almost happy now, but could you add a test for the families check?

Yes, of cource. See my last commit. While implementing the test I recognized that for `family = "custom.family"` it depends on the family one implements if probabilitites can be predicted or not. But for this case I don't think we can implement a reasonable check and throw an error if probs cannot be predicted. In my opinion the user that implements a family himself/herself will know if prediction of probs is possible or not. What do you think?

One question on the test: when you run i.e. test_classif_blackboost.R now then there will be this information:
```` 
Timing stopped at: 0.03 0 0.03
 ````
because of expect_error(). Is this just locally happening? Can I switch that off? Do we care about this?

> If you rename a file, you need to tell git that you've done so by adding it git add RLearner_regr_blackboost.R. Then it will show up in the commit. You can also do this in one step with git mv RLearner_regr_blackBoost.R RLearner_regr_blackboost.R.

Thank you :-)

larskotthoff:
Sorry, I should have phrased this differently -- I would add a generic test that checks whether the predictions have the expected/necessary structure (i.e. there are probabilities) and otherwise gives a "the selected family does not support probabilities" error. This will work with custom families as well.

We don't care about the "timing stopped" messages.

Thanks!

MariaErdmann:
Sorry, I need to ask you one more time since I am not completely sure if I understand you correctly. When you say

>  I would add a generic test that checks whether the predictions have the expected/necessary structure (i.e. there are probabilities) and otherwise gives a "the selected family does not support probabilities" error.

do you mean, adding the generic test in the predict funtion instead of asking whether the family is AUC or AdaExp?

larskotthoff:
Yes, that's correct.

MariaErdmann:
Is this what you meant? ;-) (see my last commits).

I also changed one line in the prediction function that has nothig to do with the probability issue.
This new line catches a special case, which is may be very unlikely in practise, but might occur: 
Usually, the predict() function of mboost outputs a matrix (with one column), also if one predicts only one observation. But for family "PropOdds" the predict() function results in a matrix with two columns (if one predicts > 1 obs) and a numeric vector with two entries (if one predicts just one obs). This latter case (That the prediction object is a numeric vector) is catched with the new line.

MariaErdmann:
While adding a test, I stumbled over this. And I don't know why this is happening and where to change it. When predicting probabilites for family PropOdds, which requires an ordered target, then mlr's predict switches the classes (i.e. when it should be class 'M', it is class 'R' and vice versa). Note, that this is not happening when using the Binomial family. See the small example code below:
````
# Creation of data/task with ordered target
data = getTaskData(binaryclass.task)
data[,binaryclass.target] = as.ordered(data[,binaryclass.target])
ordered.task = makeClassifTask(data = data, target = binaryclass.target)

lrn = makeLearner("classif.glmboost", par.vals = list(family = "PropOdds"),
  predict.type = "prob")
mod = train(lrn, ordered.task)
pred = predict(mod, ordered.task)

orig.mod = mboost::glmboost(binaryclass.formula, data = data, family = mboost::PropOdds())
orig.pred = predict(orig.mod, newdata = data, type = "response")
orig.pred.class = predict(orig.mod, newdata =data, type = "class")

# Compare
pred
head(orig.pred)
head(orig.pred.class)
````

larskotthoff:
mlr reorders the levels so that the negative class comes first. Does that correspond to what you have in the task?

MariaErdmann:
Thanks for the answer! Good to know :)
We were talking about my problem with the test and ascertained that ordered targets are not yet implemented in mlr but will come in the future. Therefore, I outcommented everything in the classif learners and tests that had to do with family "PropOdds", which is a family for ordered targets only.

MariaErdmann:
Something bad happend here when I tried to rebase. @ja-thomas and me tried to resolve the problem but were not able to find what I did wrong. Thus, I will close this PR and open a new one with the changes I have made to the 10 files concerning the mboost learners.

PR #1032, opened 2016-07-20
Introduce mlr.defaults as par.vals

jakob-r:
This is an quite important change.
From now on when implementing a new Learner
1. overwriting or manually set param.values go on `mlr.defaults` !
2. `par.vals` just store temporary param settings
3. getHyperParams gives joint param.values
4. in the test files of each learner, when you define the parset.list for `test*Parsets` you also have to put 

the values you put in mlr.defaults (this is open to debate)

(more comments comming soon)


larskotthoff:
What does the "reset" parameter do? And I don't like the idea of having to replicate the defaults in the tests.

jakob-r:
Let me know if we should extend this part of the manpage
```r
#' @param reset [\code{character}]\cr
#'   Can take values \code{soft} for setting the hyper parameters while keeping the \code{mlr.defaults} as far as they are feasible. 
#'   \code{no} for updating the old parameter values with new ones without dropping old ones. A feasibility error then might occur later.
#'   \code{hard} for completely dropping old parameter values and just unsing the new ones.
```

We will have to replicate the defaults anyway because we have to call the original function with those defaults (remember: the mlr.defaults are those that overwrite or initially set the original behaviour). Unless! We first construct the learner and read the mlr.defaults and put them into the par.vals. But this might get very tideous for all those weird special cases.

larskotthoff:
I don't really understand the documentation. So with "soft" mlr tries to "repair" infeasible settings? And "hard" is just the new ones, whereas "no" is (old \ new) \setunion new?

jakob-r:
- As we now have the `mlr.defaults` we usually want to stick with these hyperparameter settings unless a) the user specifically overwrites them or b) they conflict with the par.vals of the user because of the requirements. This is taken car of by "soft"
- As the mlr.defaults are now stored in an extra slot they are ought not to be changed. With any new call of `setHyperPars` they can be restored but during train we only take the values which are stored in `par.vals`. "hard" sets those values explicitly. Also helpfull if you want the behaviour of the original learner.
- "no" is the dumb version of "soft" as this only inserts the new par.vals and keeps the old ones. This is what we do now as we initally store the mlr.defaults in the par.vals

I have to note that I am not 100% certain which modes we will need in the future.

berndbischl:
@jakob-r 
bb, jr and mllg shoud do a team review here.
also travis fails and must be rebased

R/removeHyperPars.R
```diff
@@ -23,6 +23,15 @@ removeHyperPars = function(learner, ids = character(0L)) {
 #' @export
 removeHyperPars.Learner = function(learner, ids = character(0L)) {
   learner$par.vals[ids] = NULL
+  on.par.out.of.bounds = coalesce(learner$config$on.par.out.of.bounds, getMlrOptions()$on.par.out.of.bounds)
```
berndbischl: this whole code block seems strage. discuss this with BB and mllg in review. but please:
never do this! please ALWAYS add comments here. i am not even sure whether you should have also extended the docs.
problem is: we dont get at all what happens here.

larskotthoff:
Is this still relevant?

jakob-r:
Yes.

pat-s:
@jakob-r 	Is this still relevant?

jakob-r:
@pat-s Well, it's an architectural change that is probably not worth to be changed anymore.

PR #1036, opened 2016-07-27
Subset data to rows where target is notNA (Fixes #1035)

### LINKED ISSUE #1035: imputeLearner seems not to work
Something is wrong here (or am I using it wrong?)
```r
> df = data.frame(x1 = runif(10), x2 = runif(10), x3 = c(runif(9), NA))
> lrn = makeLearner("regr.km")
> impute(df, cols = list(x3 = imputeLearner(lrn, features = c("x1", "x2"))))
Error in checkLearnerBeforeTrain(task, learner, weights) : 
  Task 'impute' has missing values in 'x3', but learner 'regr.km' does not support that!
```

jakob-r:
Before we trained with the complete data set including the observations where we had NAs in the column we wanted to impute.
This does not make so much sense to me.
Now we only train on the data, where the column to impute is not NA.

### DIFF
--- a/R/ImputeMethods.R
+++ b/R/ImputeMethods.R
@@ -283,7 +283,9 @@ imputeLearner = function(learner, features = NULL) {
         if (col %nin% features)
           features = c(col, features)
       }
-      ind = !is.na(data[[col]]) # we train only on data where the column is not NA
+      # Remove all observations with missing values in the target for training. This is required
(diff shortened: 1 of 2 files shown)


berndbischl:
> we train only on data where the column is not NA

column --> target.

otherwise seems ok, can be merged after fix

berndbischl:
actually both the unit test and the code change should have a better comment explaining what happens there

PR #1037, opened 2016-07-27
imputeLearner:return data directly if no NAs (Fixes #848)

### LINKED ISSUE #848: impute() does not handle non missing values in columns
impute should be able to do nothing if no missing values are present:
```r
library(mlr)
iris.m = iris
imp = impute(data = iris.m, 
  target = "Species", 
  cols = list(Sepal.Width = imputeLearner(makeLearner("regr.rpart"))))
#Error in impute(data = iris.m, target = "Species", cols = list(Sepal.Width = imputeLearner(makeLearner("regr.rpart")))) : 
#  Assertion on 'newdata' failed: Must have at least 1 rows, but has 0 rows.

jakob-r:
We could also debate if we allow predict for empty newdata. Until then this seems to be the best solution.

### DIFF
--- a/R/ImputeMethods.R
+++ b/R/ImputeMethods.R
@@ -144,8 +144,8 @@ imputeMax = function(multiplier = 1) {
 #'   If NA (default), it will be estimated from the data.
 #' @rdname imputations
 imputeUniform = function(min = NA_real_, max = NA_real_) {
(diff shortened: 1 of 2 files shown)


berndbischl:
BB: can be merged if travis is done

Issue #732, opened 2016-02-16
Some parameters of classif.ada do not work.


mb706:
Because of something that might be a bug in `ada::ada`, it throws an error whenever there are extra args in the `...` that don't match `formals(rpart::rpart)`. That means `rpart.control` parameters can not be used, since they fall inside the `...` of `rpart::rpart`.

The parameters in question are
- minsplit
- minbucket
- cp
- maxcompete
- maxsurrogate
- usesurrogate
- surrogatestyle
- maxdepth
- xval

```R
> l = makeLearner("classif.ada", minsplit=20, minbucket=20, cp=0.01, maxcompete=4, maxs
urrogate=5, usesurrogate=2, surrogatestyle=0, maxdepth=30, xval=10)
> train(l, pid.task)
Error in ada.default(x, y, ..., na.action = na.action) :
  Error:  Argument minbucket not matchedError:  Argument maxcompete not matchedError:  Argument maxsurrogate not matchedError:  Argument usesurrogate not matchedError:  Argument surrogatestyle not matched
Timing stopped at: 0.009 0.001 0.013
```

One notices there is no error for some of the parameters I listed. This is because **additionally** there is a bug in `trainLearner.ada` where some of these parameters are collected in a list and given as unnamed parameter:
```R
#R/RLearner_classif_ada.R
  to.list = c("maxdepth", "cp", "minsplit", "xval")
  list = filterNull(dots[to.list])
  dots = dropNamed(dots, to.list)
  args = c(dots, list(list))
```
Maybe at some point in time this was a way to circumvent the ada bug, but as I see it now, ada assigns the list to the first parameter that is not given otherwise ("`test.x`").

If there is no way to give these parameters to `rpart::rpart.control` without hiccup from ada, then these parameters should be removed from the ParamSet.

mb706:
The given code doesn't fix the issue. The error won't be thrown any more, but instead the given values are ignored by ada.

This is because `trainLearner.classif.ada` assigns all the values that I list above to a list and gives this list as an unnamed argument to `ada::ada`. Therefore, `ada` interprets this list as the first argument that is not otherwise named in the call, which happens to be `test.x`. As can be seen in `ada:::ada.default`, `test.x` [will be removed since `test.y` is `NULL`](https://github.com/cran/ada/blob/3c3917b4e3a53894d43752bcf5f28015fd6921fe/R/ada.default.R#L30-L41), hence the arguments are ignored.

Proof:
```R
[...]

PR #1052, opened 2016-08-02
Issue #732 put all rpart.control parameters into a list named 'control'

jklanger:
Working on issue #732 I found the following problem:
The rpart.control parameters were collected in an unnamed list, which is then given to the ada function. Like mentioned by @mb706 this unnamed list is recognized as the parameter 'test.x', which during the execution of ada is set to NULL iff the argument 'test.y' is NULL.
To fix this I put all rpart.control parameters in a list named 'control', which is then recognized by ada.


ja-thomas:
but now you only pass the parameters for `rpart.control` and disregard all `ada` parameters.

example:
```
resample(makeLearner("classif.ada", max.iter = 1L), binaryclass.task, cv5)
resample(makeLearner("classif.ada", max.iter = 10^6L), binaryclass.task, cv5)
```
Does the same thing. You still have to pass all non-matching arguments to `ada`

jklanger:
All arguments are passed!

There are two problems with this test:

- ada does not use 'max.iter' at all if 'loss="exponential"' and 'type="discrete"', you have to pick other values here. Another option to NOT use 'max.iter' would be to set "model.coef=FALSE".

- also this loop breaks after the 'delta' value is reached.

Try this:
``resample(makeLearner("classif.ada", max.iter = 1L, type = "gentle", delta = 0), binaryclass.task, cv5)
``resample(makeLearner("classif.ada", max.iter = 10^6L, type = "gentle", delta = 0), binaryclass.task, cv5)

ja-thomas:
Yes you are right, sorry.

The name `max.iter` kind of gives it away.

giuseppec:
I created a branch out of this and extended the tests, the new PR is here https://github.com/mlr-org/mlr/pull/1081, therefore close.

PR #1055, opened 2016-08-03
3D implementation

BoyaoZhang:
The plotLearnerPredictionPloty() function works relatively well, while the API of plotPartialPredictionPlotly() function will be added in the following days.


berndbischl:
@BoyaoZhang 
you can certainly not do a PR with > 100 changed files. you have not rebased on master.
please do a PR only with the relevant stuff

Issue #1050, opened 2016-08-01
hyper par settings: partial matching hints


berndbischl:
if a user enters a wrongly typed param name, display, instead of only the error, a hint which other name matches most closely.

like:
"did you mean ntrees instead of n.tree?"

ja-thomas:
I really like that!

berndbischl:
me too :)
dont know why we didnt see this before. this was a suggesting from a participant of my ml course
(thanks, wolfgang!)

wolfganglederer:
You are welcome.

wolfganglederer:
You are welcome.

wolfganglederer:
You are welcome.

berndbischl:
> You are welcome.

:)

ja-thomas:
we could use the [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), which is already implemented in `adist` in Base-r.

e.g.

```
pars = names(getParamSet("classif.ranger")$pars)

pars[which.min(adist(pars), "n.tree"))]
pars[which.min(adist(pars), "trees"))]
pars[which.min(adist(pars), "nrounds"))]
pars[which.min(adist(pars), "iters"))]
pars[which.min(adist(pars), "numberoftree"))]
```

MasonGallo:
+1 

`adist` in {utils} does fuzzy matching - I can do a PR soon if no one else does

berndbischl:
@MasonGallo 
do a PR pls.

berndbischl:
@MasonGallo 
solved this (nearly) in PR #1060

PR #1060, opened 2016-08-03
ENH: hyperpar fuzzy matching suggestion (#1050)

MasonGallo:
If mlr can't find the hyperparameter, suggest the 3 closest matches to the user. Given that most learners probably don't have more than 10 hyperparameters that are frequently tuned, I figured 3 was a safe number here.

- [x] Fix tests
- [x] Convert suggestion to a message() instead?


MasonGallo:
Here's an example:

```
lrn = makeLearner("classif.ksvm", Cc = 2L)
 Error in setHyperPars2.Learner(learner, insert(par.vals, args)) : 
  classif.ksvm: parameter Cc has no available description object!
Did you mean one of these: C nu tol
You can switch off this check by using configureMlr! 
```

berndbischl:
@MasonGallo 
1) you cannot so easily call configureMlr in the tests. that will change the gobal state. you would need to save the old config options and reset them at the end
2) for your test, why dont you simply use the arg "config" of makeLearner? that seems much easier...?

MasonGallo:
@berndbischl Should be fixed now. Thanks!

berndbischl:
> classif.ksvm: couldn't find Cc!

i would probably make the err msg a bit more specfic, so like 
"could find hyperparam 'CC'" 
(and the name in quotes)

also @larskotthoff @jakob-r @ja-thomas 
what do you think? done?

any downside to always select the 3 best nearest neighbors?

larskotthoff:
Sounds like a great idea and I think we should have this.

berndbischl:
> Sounds like a great idea and I think we should have this.

any comment on my question whether always the 3 best options should be display? ok?

larskotthoff:
Sounds good to me. 

Lars

jakobbossek:
Three seems to be a good choice. Why not make add an mlr option with default set to 3?

berndbischl:
> Three seems to be a good choice. Why not make add an mlr option with default set to 3?

beacause we dont want feature creep and option overload :)
thx, if you like this, too, like me, we can merge this soon

jakobbossek:
> beacause we dont want feature creep and option overload :)
Ok :-)

However, we could outsource the helper function to BBmisc/HELL. I would like to make use of it in other packages as well.

berndbischl:
> However, we could outsource the helper function to BBmisc/HELL. I would like to make use of it in other packages as well. Sure, I can write this quite fast by myself, but we have even shorter functions in BBmisc ;-)

ok i see your point. but maybe more for PH? or do you really want this independent  of a paramset? 
also pls propose a signature then.

jakobbossek:
This would be helpful in non-PH context as well. E.g., fuzzy checking of passed EA algorithm parameters in ecr. A S3 method would be nice. Maybe 

```splus
proposeParams.character = function(x, input.pars) {...}
proposeParams.Param = function(x, input.pars) {...}
proposeParams.ParamSet = function(x, input.pars) {...}
```

MasonGallo:
I see your point @jakobbossek but I am slightly weary of "over-generalizing" this. IMO this works in this particular instance because the set hyperparameters for a given learner tend to have a very small number of elements.

MasonGallo:
> i would probably make the err msg a bit more specfic, so like 
"could find hyperparam 'CC'" 
(and the name in quotes)

@berndbischl I agree I'll change to that shortly

berndbischl:
> This would be helpful in non-PH context as well. E.g., fuzzy checking of passed EA algorithm parameters in ecr. A S3 method would be nice. Maybe

@jakobbossek 
thought about this:
a) i will simply merge this in mlr now, when mason is done here. 
b) i dont want s3 dependencies over multiple packages again. that sucked donkeys balls last time... i can see the point of having a helper in PH. do a PR if you want that. or best open an issue up now, and do  the PR later.

PR #1062, opened 2016-08-03
Implementation of multilabel.cforest

MariaErdmann:
I started to implement multilabel,cforest when mutlilabel.randomForestSRC wasn't merged yet. Because I also need to do changes in helper_helpers.R to test multilabel.cforest I had a merge conflict with my last git pull. I deleted my changes and overtook all changes from the pull and tried to resolve it but it still shows some awkward diffs. I'm sorry but I don't know what to do about this.
Some remarks: 
The test is only implemented for probabilities since cforest in the multivariate case can only predict probabilities. (in the manual it says: "Note that for multivariate responses predict does not convert predictions to the type of the response, i.e., type = "prob" is used."). But the multilabel.cforest in mlr can predict responses!
For the test I needed to add another multilabel.formula, since the one that was working for multilabel.randomForestSRC wasn't working for me.
Please review and if it passes Jenkins/Travis and you're ok then this is able to merge.


berndbischl:
> but it still shows some awkward diffs. I'm sorry but I don't know what to do about this.

what exactly do you mean?

MariaErdmann:
The diff says that I deleted a line and then inserted one at the same place (anyhow that is what I think it is). And this is happening at multiple places.

berndbischl:
you have whitespace changes there
try this in you branch, as the file has no real changes

git checkout origin/master filename

then commit it, and push. this should revert it to the state of the master brnach

MariaErdmann:
Thanks! Is fine now :-)
One more remark: I also adapted learners_all_multilabel since multilabel.cforest (in contrast to multilabel.rFerns) can do factors, weights, etc. I added the associated tests.

berndbischl:
BB: can be merged if travis is done

PR #1063, opened 2016-08-04
resolve issue 712, add classif_featureless

smilesun:
This is  a smarter implementation for the sklearn-dummy learner since our implementation could allow the user to specify the measure to choose an optimal constant prediction, the core codes is done by Bernd


R/RLearner_classif_featureless.R
```diff
@@ -0,0 +1,45 @@
+#' @export
+makeRLearner.classif.featureless = function() {
+  makeRLearnerClassif(
+    cl = "classif.featureless",
+    package = "mlr",
+    par.set = makeParamSet(
+      makeUntypedLearnerParam(
+        id = "measure",
+        default = mmce,
+        tunable = TRUE
```
ja-thomas: code style, this can be written in one line

R/RLearner_classif_featureless.R
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+#' @export
+trainLearner.classif.featureless = function(.learner,
+                                            .task,
+                                            .subset,
+                                            .weights = NULL,
+                                            measure = mmce,
+                                            ...) {
```
ja-thomas: here as well

tests/testthat/test_classif_featureless.R
```diff
@@ -0,0 +1,48 @@
+context("classif_featureless")
+
+test_that("classif_featureless", {
+  costs = matrix(c(0, 10, 10, 1, 0, 1, 1, 1, 0), 3, 3, byrow = TRUE)
+  rownames(costs) = colnames(costs) = getTaskClassLevels(iris.task)
+  print(costs)
```
ja-thomas: why print it?

ja-thomas:
```
lrn = makeLearner("classif.featureless", measure = mmce)

mod = train(lrn, iris.task)

> mod
Model for learner.id=classif.featureless; learner.class=classif.featureless
Trained on: task.id = iris-example; obs = 150; features = 4
Hyperparameters: measure=<Measure>
```

the measure is not printed, not sure how to fix this

ja-thomas:
Not sure why the test is failing now, everything runs locally on my machine.

berndbischl:
1) you dont have to set tunable=TRUE, that the default
2) when you  specify the dwfsult measure, use getDefaultMeasure("classif")
(which is then mmce)
3) short name is bad. "featureless"?
4) the properties seem wrong? a featureless learner can also work on ordered, factors....
5) it seems there is code to basically check a measure vs a task. 
pls put this code in helper function and call it also in performance
6) the optimization of the learner output itself should be a helper function
7) classif.featureless also needs to support prob output

ja-thomas:
1) 2) 3) 4) 7) Done

5) 6) Is that not the same thing? In both cases we need some data (which is a task) and a measure to optimize it. In 5) there would only be a call to 6). But you can convince me otherwise :)

I still want to have a look at `constantClassWrapper` because this could essentially be just a call to `constantPrediction`. Maybe we can also do something with `NoFeaturesModel`.

Also I just realized that the name `constantPrediction` might not be so good. Maybe `constantEstimation`?

florianfendt:
@ja-thomas 
branch is rebased. Should I continue with this now or do you take back over?

giuseppec:
I talked to @berndbischl because I found this branch over here https://github.com/mlr-org/mlr/compare/create_featureless_dummy_models . You might also look into this branch and use some code from there.

larskotthoff:
What's the status here? Can this be merged?

ja-thomas:
nope, we won't merge this. The simpler solution by @giuseppec is more than enough. This is still buggy for some cases and introduces a lot of overhead for "optimal" featureless prediction.

PR #1064, opened 2016-08-04
DOC: add notes for also generating performance on train set (#1044)

### LINKED ISSUE #1044: Add support for displaying train performance when comparing models/sets
IMO Most of our methods that we use to compare multiple learners or multiple hyperparameter sets would benefit from displaying train error out of the box, rather than just validation error, in order to see if a model/set is under/overfitting.

For example, I would expect [learning curve](http://mlr-org.github.io/mlr-tutorial/devel/html/learning_curve/index.html) to at least have an option to also draw a line for the training performance. Same goes for my GSoC work on hyperparameter tuning.

Thoughts?

MasonGallo:
I added note for `resample` and `tuneParams` along with an example. I also cleaned up references that should use the new `generateHyperParsEffectData`.

Closes #1044

### DIFF
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -1,31 +1,31 @@
 #' @title Generate hyperparameter effect data.
 #'
-#' @description 
-#' Generate cleaned hyperparameter effect data from a tuning result or from a 
-#' nested cross-validation tuning result. The object returned can be used for 
-#' custom visualization or passed downstream to an out of the box mlr method, 
+#' @description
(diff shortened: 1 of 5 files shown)


berndbischl:
this is mine now. i will change and merge this.

@MasonGallo 
2 things which are improvable: 
- in your note you could either say how to reach that goal exactly, or better: simply refer to the example.
- i dislike that in the simply printing of the optpath you have changed the as.data.frame.
the user should know about this function IMHO

berndbischl:
can be merged if travis is done

PR #1075, opened 2016-08-07
GSoC WIP: Add support for partial dependency on hyperpar effects tasks

MasonGallo:
This will add support for performing partial dependency on hyperparameter effects data. Partial dependency will allow us to better visualize the effect of a given hyperparameter against a performance measure, especially when 2+ hyperparameters were tuned simultaneously.

TODO:

- [ ] Build out usecases and tutorial for partial dependency
- [ ] Decide on "facetting" support and usecases
- [ ] Integrate train and test on same plot
- [ ] Feature selection for hyperparameter importance
- [ ] Tests

Updated examples will be [HERE](https://github.com/MasonGallo/data-sci-scripts/blob/master/tutorial_draft.md#visualizing-the-effect-of-more-than-2-hyperparameters)

### DIFF
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -269,7 +269,7 @@ plotHyperParsEffect = function(hyperpars.effect.data, x = NULL, y = NULL,
   # if single hyper and multiple measures, 1 graph/multiple colored lines or many graphs/1 line
   # if multiple hypers & single measure, pull from defaults and facet to get graph per hyper
   # if multiple hypers & multiple measures, graph for each hyper/multiple colored lines or column for each measure
-  if (!is.null(hyperpars.effect.data$partial) && 
+  if ((hyperpars.effect.data$partial) && 
       !("iteration" %in% c(x,y,z))) {
     partial = hyperpars.effect.data$partial.data
     # just x, y
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -1,31 +1,33 @@
 #' @title Generate hyperparameter effect data.
 #'
-#' @description
-#' Generate cleaned hyperparameter effect data from a tuning result or from a
-#' nested cross-validation tuning result. The object returned can be used for
-#' custom visualization or passed downstream to an out of the box mlr method,
+#' @description 
+#' Generate cleaned hyperparameter effect data from a tuning result or from a 
+#' nested cross-validation tuning result. The object returned can be used for 
+#' custom visualization or passed downstream to an out of the box mlr method, 
 #' \code{\link{plotHyperParsEffect}}.
-#'
+#' 
 #' @param tune.result [\code{\link{TuneResult}} | \code{\link{ResampleResult}}]\cr
 #'  Result of \code{\link{tuneParams}} (or \code{\link{resample}} ONLY when used
-#'  for nested cross-validation). The tuning result (or results if the
-#'  output is from nested cross-validation), also containing the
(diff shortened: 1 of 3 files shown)


R/generateHyperParsEffect.R
```diff
@@ -223,130 +239,157 @@ plotHyperParsEffect = function(hyperpars.effect.data, x = NULL, y = NULL,
...
-          d[,col][is.na(d[,col])] = max(d[,col], na.rm = TRUE)
+  
+  # partial dep
+  # if heatcontour w k hypers, need interaction=T
+  # if iteration asked, need to use original hyperpars data
+  # if single hyper and single measure, pull from defaults
+  # if single hyper and multiple measures, 1 graph/multiple colored lines or many graphs/1 line
+  # if multiple hypers & single measure, pull from defaults and facet to get graph per hyper
+  # if multiple hypers & multiple measures, graph for each hyper/multiple colored lines or column for each measure
+  if ((hyperpars.effect.data$partial) &&
```
MasonGallo: @zmjones am I using partial.dep correctly here?

R/generateHyperParsEffect.R
```diff
@@ -37,28 +39,28 @@
...
 #'                extract = getTuneResult)
 #' data = generateHyperParsEffectData(res)
 #' plotHyperParsEffect(data, x = "C", y = "mmce.test.mean", plot.type = "line")
 #' }
 #' @export
 #' @importFrom utils type.convert
 generateHyperParsEffectData = function(tune.result, include.diagnostics = FALSE,
-                                       trafo = FALSE) {
+                                       trafo = FALSE, partial.dep = FALSE) {
   assert(
```
zmjones: should probably use two `assertClass` statements here

R/generateHyperParsEffect.R
zmjones: yes basically.

i think you should match the resolution of the grid you use for interpolation with the size of the grid you use when calling partial dependence. i think you could unify the grid creation by using `generateFeatureGrid` which is currently in `generatePartialDependence.R` but should be moved to `utils.R`. It looks like this would serve your purposes without modification but if not I can make it more general so that it can. 

If you did that then you could call `doAggregatePartialDependence` directly which would have lower overhead. 

This would also allow you to remove points of extrapolation in the same way you are doing when not using partial dependence.

Also when plotting the points that were actually evaluated against the background which shows the interpolation/extrapolation, it might be good to color the points on the same scale as the background (you can do this by making the scales have the same name and range). Then you could see the difference between the performance value at the points and what the interpolation/extrapolation predicts them to be.

Otherwise looks really good!

zmjones:
also you don't have to explicitly reference plyr since we import the relevant functions (or maybe the whole package). we are going to remove the plyr and reshape2 dependencies anyway in favor of data.table, so you don't actually need to change this, but any package in imports doesn't need `::`.

R/generateHyperParsEffect.R
```diff
@@ -125,14 +139,16 @@ print.HyperParsEffectData = function(x, ...) {
   catf("Measures: %s", collapse(x$measures))
   catf("Optimizer: %s", collapse(x$optimization))
   catf("Nested CV Used: %s", collapse(x$nested))
+  if (!is.null(x$partial))
+    print("Partial dependence generated")
   catf("Snapshot of $data:")
```
zmjones: random $ here

MasonGallo:
Thanks @zmjones ! Sorry I've been sick and starting to catch up on work

PR #1079, opened 2016-08-08
move dependencies (except for ParamHelpers) to imports

### LINKED ISSUE #1078: Move stuff from depends into imports
- [ ] ggplot2 
- [ ] BBmisc
- [ ] stats 
- [ ] stringi

zmjones:
should close #1078


berndbischl:
@zmjones why habe you added the library(BBmisc) to the example in model multiplexer?

zmjones:
because we are importing now after this modification so we need to load it explicitly then.

PR #1082, opened 2016-08-08
Bugfix: fix #1066, argument pass-down in rf.min.depth filter

### LINKED ISSUE #1066: Feature selection - parameter naming conflict
I really enjoy working with mlr, but I've run into a little problem, and I'm not sure how to get around it. I'm hoping you can help. I'm trying to do feature selection with the rf.min.depth function. I want to specify a value for the "method" parameter within the underlying library ( randomForestSRC::var.select). But there is also a "method" parameter in the generateFilterValuesData function for the learner name. Below is some example code along with the error I'm getting. I'm running R-3.3.1 and mlr version 2.9.

```
library(mlr)

parameterValues <- c(method="vh", conservative="low")
task <- makeClassifTask(data = iris, target = "Species")
fv <- generateFilterValuesData(task, method = "rf.min.depth", more.args=list("rf.min.depth"=parameterValues))
```

Error in randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),  : 
  formal argument "method" matched by multiple actual arguments

Any suggestions?

jakobbossek:
The `method` argument was passed twice to `randomForestSRC::var.select` directly by passing `method` argument and via `...` resulting in a `formal argument "method" matched by multiple actual arguments` error.

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -161,9 +161,9 @@ makeFilter(
   pkg  = "randomForestSRC",
   supported.tasks = c("classif", "regr", "surv"),
   supported.features = c("numerics", "factors"),
-  fun = function(task, nselect, ...) {
+  fun = function(task, nselect, method = "md", ...) {
     im = randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),
-      method = "md", verbose = FALSE, ...)$md.obj$order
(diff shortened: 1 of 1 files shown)


larskotthoff:
Could you add a unit test for the error please?

PR #1083, opened 2016-08-08
new function generateFeatureImportanceData

zmjones:
new generic feature importance method not for use as a filter.

this takes arbitrary learners, measures (appropriate to task), contrast functions (measuring the difference between permuted predictions and un-permuted predictions), and aggregation functions.

we can also jointly permute features. we can construct interaction importance by subtracting the individual feature importances from this. you can currently compute the joint importance but i'm not subtracting the individual importances. for computing the importance of a 3 way interaction we'd need to compute the depth 2 interactions and the individual importances and subtract all of that off. the functional anova code currently does a similar thing and i'll try to abstract that to use here as well.

i think i can refactor the "permutation.importance" filter to just use this code which is better.

i want to add the ability to compute local importance. the reason this doesn't work right now is because there doesn't seem to be any way to not aggregate a measure. does anyone know of a way to get around this without upsetting everything? 

this would allow us to compute class-specific importance and, analogously for regression, to compute the density of the loss across the empirical distribution of the target.

thoughts on these directions or other related things?

```{r}
generateFeatureImportanceData(bh.task, "regr.lm", c("crim", "lstat"), interaction = TRUE, list(mse, mae))
```


zmjones:
the function `generateFeatureImportance` now generates output with class `FeatureImportance`.

this now computes "local" feature importance, which is the per-observation change in the performance. i did this without changing how we compute measures by subsetting the prediction object and calling performance on it repeatedly.

local importance allows us to compute the loss per class as well as the density of the loss across the distribution of the target.

```{r}
library(ggplot2)

imp = generateFeatureImportanceData(bh.task, "permutation.importance",
                                    "regr.randomForest",
                                    features = "crim", interaction = FALSE,
                                    measure = mse,
                                    contrast = function(x, y) x - y,
                                    aggregation = mean,
                                    nperm = 50L,
                                    replace = TRUE,
                                    local = TRUE)
head(imp$data)
##          crim
## 1  3.36125568
## 2  2.62024115
## 3  0.04291650
## 4  0.51869797
## 5  1.31532479
## 6 -0.08098203

data = getTaskData(bh.task)

plt = data.frame("local.imp" = imp$data[["crim"]], "crim" = data[["crim"]])
ggplot(plt, aes(crim, local.imp)) + geom_point() + geom_smooth()
ggsave("regr_ex.png", width = 6, height = 4)

imp = generateFeatureImportanceData(iris.task, "permutation.importance",
                                    makeLearner("classif.randomForest", predict.type = "prob"),
                                    features = "Petal.Width", interaction = FALSE,
                                    measure = ber,
                                    contrast = function(x, y) x - y,
                                    aggregation = mean,
                                    nperm = 50L,
                                    replace = TRUE,
                                    local = TRUE)

head(imp$data)
##   Petal.Width
## 1  0.08666667
## 2  0.17333333
## 3  0.02000000
## 4  0.06000000
## 5  0.12666667
## 6  0.12000000

data = getTaskData(iris.task)

plt = data.frame("local.imp" = imp$data[["Petal.Width"]], "class" = data$Species,
                 "Petal.Width" = data[["Petal.Width"]])

ggplot(plt, aes(Petal.Width, local.imp, color = class)) + geom_point()
ggsave("classif_ex.png", width = 6, height = 4)
```
classif_ex
regr_ex

R/generateFeatureImportance.R
```diff
@@ -0,0 +1,144 @@
...
+    stop("aggregation function must either return 1 number or a numeric vector of the same length as the number of rows in the task data.frame.")
+
+  out = switch(method,
+    "permutation.importance" = doPermutationImportance(
+      task, learner, features, interaction, measure, contrast, aggregation, nperm, replace, local)
+    )
+
+  makeS3Obj(
+    "FeatureImportance",
+    data = out,
```
larskotthoff: I would prefer "res" instead of "data" here to make it clearer that this is the result of the operation.

R/generateFeatureImportance.R
```diff
@@ -0,0 +1,144 @@
...
+  contrast, aggregation, nperm, replace, local) {
+  
+  ## train learner to get baseline performance
+  fit = train(learner, task)
+
+  # compute unpermuted performance
+  pred = predict(fit, task = task)
+  if (local) {
+    # subset the prediction data element to compute the per-observation performance
+    perf = lapply(1:getTaskSize(task), function(i) {
```
larskotthoff: Could use `vnapply` instead of `lapply` here.

R/generateFeatureImportance.R
```diff
@@ -0,0 +1,144 @@
...
+    })
+    perf = as.numeric(perf)
+  } else {
+    perf = performance(pred, measure)    
+  }
+  
+  data = getTaskData(task)
+
+  ## indices for resampled data to be used for permuting features
+  idx = replicate(nperm, sample.int(getTaskSize(task), replace = replace))
```
larskotthoff: `idx` implies singular -> `indices`?

zmjones:
all this stuff you mentioned @larskotthoff is fixed

zmjones:
closing this to issue a pr from a branch in the mlr-org repo

PR #1087, opened 2016-08-09
Fix #1025: remove minprob, pvalue and randomsplits params from cforest

### LINKED ISSUE #1025: pvalue and randomsplits is not a parameter of cforest
Hi,

implementing a multilabel.cforest I stumbled over the parameter pvalue in the paramset of our classification, regression and survival learners. I searched for this parameter in conjunction with cforest on the net but did not find it. Indeed, it does not work (creating the learner does work, but training not):

````
clrn = makeLearner("classif.cforest", par.vals = list(pvalue = TRUE))
mod = train(clrn, binaryclass.task)

rlrn = makeLearner("regr.cforest", par.vals = list(pvalue = FALSE))
mod2 = train(rlrn, regr.task)

slrn = makeLearner("surv.cforest", par.vals = list(pvalue = TRUE))
mod3 = train(slrn, surv.task)
````
Do I miss something or can I remove the parameter from the learners?

Cheers, Maria

jakobbossek:
@berndbischl and me decided to get rid of the parameters `minprob`, `pvalue` and `randomsplits`, since these are set internally to fixed values inside `cforest_control()` and cannot be changed by the user.

### DIFF
--- a/R/RLearner_classif_cforest.R
+++ b/R/RLearner_classif_cforest.R
@@ -10,16 +10,13 @@ makeRLearner.classif.cforest = function() {
       makeNumericLearnerParam(id = "fraction", lower = 0, upper = 1, default = 0.632),
       makeLogicalLearnerParam(id = "trace", default = FALSE, tunable = FALSE),
       makeDiscreteLearnerParam(id = "teststat", values = c("quad", "max"), default = "quad"),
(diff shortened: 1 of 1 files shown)


larskotthoff:
Thanks, merging.

PR #1089, opened 2016-08-09
silence setHyperPars

mllg:


berndbischl:
@giuseppec 
michel might be busy. pls take over.

but this must be done:
1) travis must run

2) when on.par.without.desc = quiet we dont want to see the hyperpar suggestion message from the string matching

berndbischl:
this still procduces the message

configureMlr(on.par.without.desc = "quiet")
lrn = makeLearner("classif.ksvm", sigm=1, show.info = T)

also make sure to unit test properly. i would guess you also want to use a mocklearner here to make sure the passed param actually reaches the learner.

and you need to test that the message is NOT generated, too. i hope test that can do that.
(i think so)

giuseppec:
I think adding `show.info` arg in `setHyperPars` is not good because you can already set `show.info` on learner creation and something like this would be possible:
```
wl = makeLearner("classif.ksvm", config = list(on.par.without.desc = "warn", show.info = TRUE), 
  sigm = 1, show.info = FALSE)
```
@mllg why do you want the show.info in setHyperPars? Isn't it sufficient to set the `show.info` using the `config` when creating the learner?

R/makeLearner.R
```diff
@@ -55,16 +55,19 @@ makeLearner = function(cl, id = cl, predict.type = "response", predict.threshold
 
   assertString(cl)
   assertFlag(fix.factors.prediction)
+  assertList(config, names = "named")
+  # FIXME: maybe forbit show.info here issue #1098:
+  assertSubset(names(config), choices = names(formals(configureMlr)))
```
mllg: names(getMlrOptions())?

tests/testthat/test_base_hyperpars.R
```diff
@@ -98,5 +133,8 @@ test_that("options are respected", {
   configureMlr(on.par.out.of.bounds = "quiet")
   expect_is(setHyperPars(lrn, alpha = 2), "Learner")
 
+  configureMlr(on.par.without.desc = "quiet")
+  lrn = makeLearner("classif.ksvm", sigm = 1, show.info = TRUE)
```
mllg: expect_silent()?

tests/testthat/test_base_hyperpars.R
```diff
@@ -63,12 +63,51 @@ test_that("setting 'when' works for hyperpars", {
...
+  cs = list(on.par.without.desc = "stop")
+  # never print message when quiet, no matter hat show.info is
+  expect_silent(makeLearner("classif.ksvm", config = cq, sigm = 1, show.info = TRUE))
+  expect_silent(makeLearner("classif.ksvm", config = cq, sigm = 1, show.info = FALSE))
+  configureMlr(on.par.without.desc = "quiet")
+  expect_silent(makeLearner("classif.ksvm", sigm = 1, show.info = TRUE))
+  expect_silent(makeLearner("classif.ksvm", sigm = 1, show.info = FALSE))
+
+  # show.info on/off and warn
+  expect_message(expect_warning(makeLearner("classif.ksvm", config = cw, sigm = 1, show.info = TRUE),
```
larskotthoff: Why are `expect_message`/`expectWarning` etc nested here? Is this necessary?

larskotthoff:
Looks good to me apart from the nested expectations that I mentioned above.

berndbischl:
should be merged when travis is done.

but we need NEWS then for the show.info change.

PR #1092, opened 2016-08-09
Quickcheck for testing

larskotthoff:


DESCRIPTION
```diff
@@ -112,6 +112,7 @@ Suggests:
     pls,
     PMCMR (>= 4.1),
     pROC (>= 1.8),
+    quickcheck,
```
mllg: quickcheck is not on CRAN, I don't think that we can put it into the DESCRIPTION w/o getting into trouble on the next release.

berndbischl:
would better bounds for the test be "best" and "worst" from the measure?
otherwise i think we could merge this

larskotthoff:
I didn't use `best` and `worst` in the spirit of having a hand-coded example. The `best` and `worst` values could be wrong.

berndbischl:
you suck

Issue #1090, opened 2016-08-09
surv.ranger does not support importance = impurity


ja-thomas:
Mini fix.

```
> lrn = makeLearner("surv.ranger", importance = "impurity")

> task = surv.task

> mod = train(lrn, task)
Error in ranger::ranger(formula = NULL, dependent.variable.name = tn[1L],  : 
  Node impurity variable importance not supported for survival forests. 
```

berndbischl:
solved in PR #1094

PR #1094, opened 2016-08-09
Remove impurity importance from surv.ranger

ja-thomas:
This is a just a mini fix.

For `surv.ranger` node impurity for importance calculations is not supported, see #1090


berndbischl:
pls: look at the unit tests and make a proposal which would have caught this.

berndbischl:
> pls: look at the unit tests and make a proposal which would have caught this.

hmm. no. disregard.

PR #1103, opened 2016-08-10
Better print for predictions (fixes #1049)

### LINKED ISSUE #1049: Prediction object (and others?) better printer
we currently display the data with a call to "head" in the printer.
can be confusing to users, when they think this is the complete data.

we should end with something like

... (100 rows, 5 cols)

jakobbossek:
The `head` output is now followed by a single line mentioning the number of rows and columns in the data.frame as proposed by @berndbischl in #1049.

### DIFF
--- a/R/Prediction.R
+++ b/R/Prediction.R
@@ -189,8 +189,6 @@ print.Prediction = function(x, ...) {
   catf("threshold: %s", collapse(sprintf("%s=%.2f", names(x$threshold), x$threshold)))
   catf("time: %.2f", x$time)
   if (!is.na(x$error)) catf("errors: %s", x$error)
-  df = as.data.frame(x)
-  print(head(x))
-  catf("... (%i rows, %i cols)", nrow(x), ncol(x))
+  printHead(as.data.frame(x))
 }
 
--- a/R/Prediction.R
+++ b/R/Prediction.R
@@ -190,5 +190,6 @@ print.Prediction = function(x, ...) {
   catf("time: %.2f", x$time)
   if (!is.na(x$error)) catf("errors: %s", x$error)
   print(head(as.data.frame(x)))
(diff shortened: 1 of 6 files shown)


jakobbossek:
- Added helper `printHead` in utils.R.
- Use it in printer for Prediction and LearningCurveData

berndbischl:
ok, i like this.

i have also found this. can you take a look?

R/Prediction.R
195:  print(head(as.data.frame(x)))

R/generatePartialDependence.R
475:  print(head(x$data))
605:  print(head(x$data))

R/generateLearningCurve.R
100:  print(head(x$data))

R/ModelMultiplexer.R
48:#' print(head(as.data.frame(res$opt.path)))

R/ResamplePrediction.R
54:  print(head(as.data.frame(x)))

R/generateFilterValues.R
93:  print(head(x$data))

R/tuneParams.R
47:#' print(head(as.data.frame(res$opt.path)))
48:#' print(head(as.data.frame(res$opt.path, trafo = TRUE)))
50:#' print(head(generateHyperParsEffectData(res)))
51:#' print(head(generateHyperParsEffectData(res, trafo = TRUE)))
69:#' print(head(as.data.frame(res$opt.path)))
76:#' print(head(as.data.frame(res$opt.path)))

R/generateHyperParsEffect.R
129:  print(head(x$data))

jakobbossek:
Sure

jakobbossek:
The calls in R/tuneParams.R and R/ModelMultiplexer.R are located in the @examples section. Not changing this, since `printHead` is not exported.

berndbischl:
> The calls in R/tuneParams.R and R/ModelMultiplexer.R are located in the @examples section. Not changing this, since printHead is not exported.

of course, this was simply a hasty output from my ack.... sorry

PR #1105, opened 2016-08-10
export makePrediction

giuseppec:
Need this for OpenML.


berndbischl:
this seems wrong. you need to export the function makeLearner, not the NULL

jakobbossek:
Docs are missing for makePrediction. This causes a warning on travis.

PR #1106, opened 2016-08-10
Fix 1070 name suggestion for make learner

jakobbossek:
- wrapped fuzzy string/name proposal in function `getNameProposals`. Placed that function in utils.R. Ok? Or is there a better place for that?
- Adapted `setHyperPars` accordingly.
- Added learner name suggestions to `makeLearner` if user mistyped learner name.
- Added a small test for that.


larskotthoff:
Thanks, Travis failure is unrelated (and not sure how it can happen at all). Merging.

PR #1109, opened 2016-08-10
Fix #1046: plotLearnerPrediction now works if a single learner is passed

### LINKED ISSUE #1046: plotLearnerPrediction: error with just one learner (not wrapped in a list)
```
lrn = makeLearner("classif.rpart")

d = generateLearningCurveData(lrn, iris.task)
plotLearningCurve(d)
```

# Error in getS3method("makeRLearner", class = cl) :
  # S3 method 'makeRLearner.classif' not found

jakobbossek:
See #1046.
Added a check for `learners` being a single learner and a corresponding unit test.

### DIFF
--- a/R/generateLearningCurve.R
+++ b/R/generateLearningCurve.R
@@ -46,6 +46,9 @@
 generateLearningCurveData = function(learners, task, resampling = NULL,
   percs = seq(0.1, 1, by = 0.1), measures, stratify = FALSE, show.info = getMlrOption("show.info"))  {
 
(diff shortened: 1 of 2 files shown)


berndbischl:
pls use BBmisc::ensureVector

jakobbossek:
Ah, did not know about that function.

jakobbossek:
`generateLearningCurveData` now works for single learner or learner name.
Did not use `BBmisc::ensureVector` (@berndbischl we talked about that).

PR #1111, opened 2016-08-10
remove plyr dependency in favor of dplyr

zmjones:
- only used rbind.fill and ldply: replaced with bind_rows
 - new util function to match factor levels for data.frames in a single
   level list (rbind.fill & ldply would silently cast them to characters)


berndbischl:
should this be a helper in utils.R?

+    if (length(res) > 1L)
 +      res = matchFactorLevels(res)
 +    res = bind_rows(res)

zmjones:
matchFactorLevels is. I don't see the point I adding the switch to
matchFactorLevels though. If the list only has one element we don't need to
call it at all.

berndbischl:
i asked because i saw that you coded exactly these 3 lines in multiple places (copypaste).
so to me it looked like an "idiom" which might be converted into an extra little helper?

zmjones:
i think it is just twice. i guess i can put the switch in there and it wouldn't affect the generality of the function. will do that and re-push.

mllg:
`bind_rows` does return a tibble instead of a regular data.frame. I think for now we definitely want an explicit conversation via `as.data.frame` afterwards.

zmjones:
is this one of these cases where it has the same classes but has different behavior?

mllg:
Do you also plan to replace the `ldply`/`ddply` stuff with functions from `tidyr`? If we go this way, I will consider removing the `data.table` dependency tomorrow.

zmjones:
well we use data.table in listLearners at least. elsewhere?

i removed all of the ldply stuff already in this code. 

i forgot about ddply but yes i'll do that as well.

berndbischl:
lets please discuss this general question with a few people and make an informed decision here

zmjones:
do you mean the data.table dependency? or the removal of the plyr dependency. i think the latter is for sure a good idea, as it has been deprecated in favor of dplyr/tidyr for a while (also reshape2).

zmjones:
all the `bind_rows` are now wrapped with `as.data.frame`

zmjones:
this is to be closed because we are going with `data.table` instead. not sure why we are leaving it open actually.

PR #1112, opened 2016-08-10
new fun generateFeatureImportanceData

zmjones:
- new generic feature importance method not for use as a filter
 - adds local importance computation
 - refactors permutation.importance filter to use this new code

see #1083 for examples & plots


larskotthoff:
Thank you very much dearest Zach, I will merge this with great pleasure.

PR #1114, opened 2016-08-10
add n.trees parameter to allow for prediction with less trees

smilesun:


R/RLearner_classif_gbm.R
```diff
@@ -44,7 +44,13 @@ trainLearner.classif.gbm = function(.learner, .task, .subset, .weights = NULL,
 predictLearner.classif.gbm = function(.learner, .model, .newdata, ...) {
   td = .model$task.desc
   m = .model$learner.model
-  p = gbm::predict.gbm(m, newdata = .newdata, type = "response", n.trees = m$n.trees, single.tree = FALSE, ...)
+  tridots = list(...)
+  if(is.null(tridots$n.trees))
+    m.n.trees = m$n.trees
+  else{
+    m.n.trees = tridots$n.trees
+  }
```
jakobbossek:
```splus
m.n.trees = coalesce(tridots$n.trees, m$n.trees)
```

tests/testthat/test_classif_gbm.R
```diff
@@ -44,3 +44,44 @@ test_that("classif_gbm keep.data is passed correctly", {
   train(makeLearner("classif.gbm", keep.data = FALSE), binaryclass.task)
   train(makeLearner("classif.gbm", keep.data = TRUE), binaryclass.task)
 })
+
```
jakobbossek: The following lines you added are just copy&paste!

berndbischl:
@smilesun 
i will shortly take this over now. there is a much easier way to "bind" the n.tree parameter:
just add it to the signature. pls watch this!

berndbischl:
@smilesun 
also please always watch out when a further method exists: here we have regr.gbm. the same changes need to be applied.

berndbischl:
and you added an empty comment line in your test. it should say instead what happens in the next code block.

smilesun:
@berndbischl , it seems that you did not makes any modifications to the code?

larskotthoff:
ping @berndbischl

pat-s:
Any motivation to finish this up?

pat-s:
> also please always watch out when a further method exists: here we have regr.gbm. the same changes need to be applied.

I guess to prove the correct implementation here, testing on one learner type is sufficient?

@larskotthoff If you approve, I think we can merge.

larskotthoff:
This needs to be documented better. What exactly are the semantics of this?

pat-s:
> This needs to be documented better. What exactly are the semantics of this?

I'll ask Xudong what the advantage is here.
But in any case, it is an official parameter in the predict function of `predict.gbm` and we should support it. 

Oh dear, yes, I completely overlooked that the same should be done for `regr.gbm` and we should add a sentence to the learner info about this new option.

berndbischl:
it is clear why this is very useful, correct?

pat-s:
I never got in touch with that yet, so no - a short explanation would be good here :)

jakob-r:
> I never got in touch with that yet, so no - a short explanation would be good here :)

You could train with 1000 trees and then check afterwards how many do you realy need to get the good performance without having to retrain the model. The same would be practical with all ensamble learners (e.g. rf) -- if the ensambles are independent, but I guess that is always the case when you talk about ensambles?

berndbischl:
Jakob is right and being able to predict on smaller ensembles posthoc, a prefix of the sequence so to speak, is the only thing we want to support

berndbischl:
That's basically allowing to pass 1 param down during predict

pat-s:
> You could train with 1000 trees and then check afterwards how many do you realy need to get the good performance without having to retrain the model. 

Ok. That was my assumption but I did not get why one would favor this instead of doing a normal tuning for the number of trees. The way of using a subset of the trained model needs manual action, right? I.e. a manual change of `n.trees.predict` in the `predict()` call. 
During a nestedCV or a tuning we would fit the same amount of models anyway (with the same value of `n.trees` even if `n.trees.predict` would be different - so we would not really save anything?
So while I understand that it is cheaper not refitting the model again, I would prefer the automated way during tuning and just tune `n.trees`?

## about this PR
Ok, back to the implementation part: It sounds like that this PR is not usable at all in its current state and there would be the need for a substantial rewrite of this, right?
I have the feeling that the effort of doing this does not weight the added value of the implementation? I.e. probably neither Jakob or Bernd wants to do that now?

If so, I would open an issue that this feature is missing and close this PR for now?

jakob-r:
> The way of using a subset of the trained model needs manual action, right?

Yes, mlr internal tuning does not benefit from prediction only params. This is an interesting topic for `mlr3` we should have in mind. However, even the manual way would be a nice to have. Some people are writing their own optimizers anyhow.

> It sounds like that this PR is not usable at all in its current state and there would be the need for a substantial rewrite of this, right?

I would highjack this PR, revert everything and just change this line:
```r
makeIntegerLearnerParam(id = "n.trees", default = 100L, lower = 1L, when = "both")
```
Because this technically right, minimal effort and allows for manual tuning at least. @berndbischl ?

pat-s:
> Yes, mlr internal tuning does not benefit from prediction only params. This is an interesting topic for mlr3 we should have in mind. However, even the manual way would be a nice to have. Some people are writing their own optimizers anyhow.

This sounds interesting but a but niche as well - I guess only few people would do this and the effort implementing it in a robust way is not really worth the effort for _mlr_ **right now**? Referring to our mlr <-> mlr3 time policy.

> I would highjack this PR, revert everything and just change this line:

Much appreciated (!) but IMO then a clean PR would be better than changing everything in here. I'll close here for now.

Issue #767, opened 2016-03-02
expressions in learner params


kerschke:
As mentioned in [PR #82 from ParamHelpers](https://github.com/berndbischl/ParamHelpers/pull/82) I started to use expressions within the mlr learners. The first steps have already been made, as you can [see here](https://github.com/mlr-org/mlr/compare/master...kerschke:expression).
In general, it is now possible to define expressions within the parameters of a learner. However, due to the high complexity, this now raises new issues on other ends of mlr. Here are some problems that came up:
- `weightedClassWrapper` did **not** update the `learner` to the `task`; why does this not work? I mean, `trainLearner.WeightedClassesWrapper` internally calls `train`, which actually adapts the `learner`
- in `test_base_Learner.R` the following call fails `expect_error(makeLearner("classif.__mlrmocklearners__5"), "used 'expression'")`, because PH now wants to have `"a"` in its dictionary (because parameter `"b"` requires `expression(a == "x")`).. do you have an idea, how to avoid this?

There are probably a lot more bugs coming up afterwards, but I thought it's best to start at some point and tackle them step-by-step..

mb706:
Do you plan this to work with wrappers that change properties of a task (e.g. a feature selection wrapper removing features), or will you drop the expressions when building a wrapper? I guess it wouldn't be such a problem if you have `default=expression(sqrt(p))` (since the default doesn't influence actual behaviour), but `len=expression(p)` could be a problem.

pat-s:
Apparently solved in #1126.

PR #1126, opened 2016-08-11
Expression handling

kerschke:
This PR allows to finally use task dependent expressions in the learner (see #767).


larskotthoff:
Thanks Pascal.

- Please document prominently what parameters are available. This should appear not only in the documentation for `evaluateLearner`, but also `makeLearner`, `setHyperPars` and anywhere else where people can set hyperparameters.
- Why is the task itself in the dictionary?
- In the unit tests, instead of checking the values for randomForest (which are specified somewhere else), could you please use a hand-constructed expression? Otherwise this test will fail for no good reason if the values in randomForest are changed.

kerschke:
>    Please document prominently what parameters are available. This should appear not only in the documentation for evaluateLearner, but also makeLearner, setHyperPars and anywhere else where people can set hyperparameters.
Ok, I'll do that :)

> Why is the task itself in the dictionary?

That's a decision we've made with @mllg and @berndbischl. The main reason is that now the user can use any expression which takes any information from the task. I mean, you probably wouldn't want to do `expression(task)` but `expression(task$...)`.

> In the unit tests, instead of checking the values for randomForest (which are specified somewhere else), could you please use a hand-constructed expression? Otherwise this test will fail for no good reason if the values in randomForest are changed.

The tests already contain hand-constructed expressions, e.g. the rpart with hand-constructed hyperparameters.. The reason for having the randomForest in there, is actually to check whether it also works with the parameter setup that is defined from our side within the definition of the learner. And if those values are changed by someone of us, well then that person should also update the unit tests accordingly. Right?

larskotthoff:
> > Why is the task itself in the dictionary?  
> 
> That's a decision we've made with @mllg and @berndbischl. The main reason is
> that now the user can use any expression which takes any information from the
> task. I mean, you probably wouldn't want to do `expression(task)` but
> `expression(task$...)`.

Ok, but then this should be documented as well.

> > In the unit tests, instead of checking the values for randomForest (which
> > are specified somewhere else), could you please use a hand-constructed
> > expression? Otherwise this test will fail for no good reason if the values
> > in randomForest are changed.  
> 
> The tests already contain hand-constructed expressions, e.g. the rpart with
> hand-constructed hyperparameters.. The reason for having the randomForest in
> there, is actually to check whether it also works with the parameter setup
> that is defined from our side within the definition of the learner. And if
> those values are changed by someone of us, well then that person should also
> update the unit tests accordingly. Right?

I don't see the benefit of testing something external -- what does this test
that the hand-constructed example doesn't test? It introduces unnecessary
coupling and I'm sure that we'll always forget this and then have to come back
to a PR later and fix things.

kerschke:
the documentation has been updated.. please review again and then merge ;)

mllg:
Reviewed. Everything okay. @berndbischl -> ok to merge?

kerschke:
@berndbischl: ping..

berndbischl:
you do know that i am on holiday right?

kerschke:
yes, and I see your working on all other ends of the package during your holiday ;)

berndbischl:
evaluateParset should be called evaluateParamSet. We have to be consistent with the names

kerschke:
Then, we're mixing up the names from ParamHelpers (evaluateParamSet) and mlr (evaluateParSet). But if that's the only thing that you don't like, I can fix that tomorrow..

berndbischl:
actually, thinking about this, the whole name scheme seems not good as too general.
evaluateLearner can mean a million things. especially in the context of mlr. 
evaluateExpressions would be a lot better. probably this should be S3 then.

berndbischl:
evaluateLearner.R: the arg "parset" should use a rox template. it adds no new documentation. its also wrong as "requires" can now contain both "quote" and "expression".

furthemore: evaluateParset and evaluateLearner do not have any arg checks with checkmate

berndbischl:
> Learners can contain task dependent expressions, see \code{\link{evaluateLearner}} for more information.

it's not really the learner, but the paramset of the learner that can contain these expressions

kerschke:
> evaluateLearner.R: the arg "parset" should use a rox template. it adds no new documentation. its also wrong as "requires" can now contain both "quote" and "expression".

so `mlr::evaluateLearner` should be `mlr::evaluateExpression.learner`, `mlr::evaluateParSet` should be `mlr::evaluateExpression.ParamSet`. And `ParamHelpers::evaluateParamSet` stays the way it is?

> furthemore: evaluateParset and evaluateLearner do not have any arg checks with checkmate

I avoided those checks, as it is only called from within `train` and `tuneParams` - and in both cases the arguments are checked prior to calling `evaluateLearner`.

kerschke:
> it's not really the learner, but the paramset of the learner that can contain these expressions

The hyperparameters can contain expressions as well..

berndbischl:
> The hyperparameters can contain expressions as well..

which is pretty much underdocumented. that must be precisely documented. saying "somewhere in the learner" does not help.

kerschke:
Did you look at the examples?

berndbischl:
> Did you look at the examples?

for what?

kerschke:
Well, if you configure a learner, you most likely define the values via the hyper parameters. And there are examples given, e.g. `makeLearner("classif.rpart", minsplit = expression(k))`.
But if it's not obvious enough, I can re-iterate over it..

berndbischl:
> Well, if you configure a learner, you most likely define the values via the hyper parameters. And there are examples given, e.g. makeLearner("classif.rpart", minsplit = expression(k)).
> But if it's not obvious enough, I can re-iterate over it..

a) i am not saying i dislike the feature of having expression in the par.vals of makeLearner and setHyperPars

b) of course i am also reading the examples

c) i REALLY do hope that you dont confuse API docs with examples. the former must be complete, while the latter should demonstrate interesting use cases. why do we have to discuss this?

berndbischl:
> furthemore: evaluateParset and evaluateLearner do not have any arg checks with checkmate
> I avoided those checks, as it is only called from within train and tuneParams - and in both cases the arguments are checked prior to calling evaluateLearner.

you have exported the functions, please add the checks

kerschke:
> a) i am not saying i dislike the feature of having expression in the par.vals of makeLearner and setHyperPars

Me neither. I just wanted to respond to your comment ("it's not really the learner, but the paramset of the learner that can contain these expressions") by clarifying that from my point of view it's actually the learner and not just it's paramset.

> b) of course i am also reading the examples

good :)

> c) i REALLY do hope that you dont confuse API docs with examples. the former must be complete, while the latter should demonstrate interesting use cases. why do we have to discuss this?

I am not mixing it up. But from my point of view, examples are useful to clarify exactly those things. But as I said before, I can re-iterate over it and add a sentence or two..

berndbischl:
example comment: tuning with expressions seems to be one of the most important use cases, so we should add an example in tuneParams. must execute quickly.

kerschke:
Ok, I might take the example that's already mentioned in the tutorial documentation..

berndbischl:
> so mlr::evaluateLearner should be mlr::evaluateExpression.learner, mlr::evaluateParSet should be mlr::evaluateExpression.ParamSet. And ParamHelpers::evaluateParamSet stays the way it is?

hmm. this is hard and needs to be defined now. 
evaluateParamExpressions? 
and make this an S3 generic in PH.

signatures would then be

PH::evaluateParamExpressions(obj, ....)
PH::evaluateParamExpressions.ParamSet(obj, dict, ....)

mlr::evaluateParamExpressions.LearnerParamSet(obj, task, ....)
mlr::evaluateParamExpressions.Learner(obj, task, ....)

@larskotthoff  @mllg i guess i want some help here

the thing abobe with the generic over multiple packages and multiple classes that dont inherit from the same base class now violates what i said and wanted before. but to me it seems to be still the best fitting option....

berndbischl:
> mlr::evaluateParamExpressions.LearnerParamSet(obj, task, ....)

fuck that does not work. i wanted to attach a more specific class LearnerParamSet to the par.set in makeRLearner. that would work.
but we also need this for the tuning param set.

larskotthoff:
Hmm. I'm not a fan of this -- it couples PH and mlr more tightly and I think that this is something that should be avoided. Could we simply have a non-S3 function `evaluateParamExpression` that is passed a dictionary/environment? Then we could use this in other places as well (e.g. for resampling reduce number of folds for small numbers of data points).

berndbischl:
ok new try:
PH: evaluateParamSetExpressions(par.set, dict)
this we already have. (must be simply renamed)

mlr:
we export makeTaskDisctionary. better call it getTaskDictionary
mlr::evaluateLearnerExpressions(lrn, task)
mlr::evaluateParamSetExpression with a task we then dont have. the user has to call
getTaskDictionary first, then evaluateParamSetExpressions
(we could later or now introduce evaluateLearnerParamSet(par.set, task)?)

berndbischl:
or this is all really bad design and this is much simpler:
we ONLY have in PH:
evaluateParamExpressions.Par(obj, dict)
evaluateParamExpressions.ParamSet(obj, dict)
evaluateParamExpressions.list(obj, dict)

in mlr we have zero extra function, except for getTaskDictionary.
we tell the user anyhow that he does not have to call these new evaluateBla function. so we dont really need to save the one line with the getTaskDictionary. 

I think i like this best.

larskotthoff:
Yep, I like you last suggestion best.

berndbischl:
@kerschke 
give @mllg a chance to comment before you start changes

mllg:
```{r}
evaluateParamExpressions.Par(obj, dict)
evaluateParamExpressions.ParamSet(obj, dict)
evaluateParamExpressions.list(obj, dict)
```
This is good. And we additionally define `evaluateParamExpressions.Learner(obj, dict)` in mlr? I don't see why this should not work.

kerschke:
Lets summarize it:

1) Code changes to PH:
```{r}
evaluateParamExpressions.Par(obj, dict)
evaluateParamExpressions.ParamSet(obj, dict)
evaluateParamExpressions.list(obj, dict)
```

2) Code changes to mlr:
```{r}
evaluateParamExpressions.learner(obj, dict)
getTaskDictionary(task) # just renaming makeTaskDictionary

```

3) Documentation should be more explicit on where the expressions come from, i.e. do we evaluate a Param, ParamSet, Learner (containing ParamSet and HyperPars), etc.

4) Add a short example to `tuneParams`.

5) Update the tutorial accordingly..

Is that all or am I missing something?
By the way, where / when do we need to `evaluateParamExpressions.list(obj, dict)`?

berndbischl:
> By the way, where / when do we need to evaluateParamExpressions.list(obj, dict)?

to eval the epressions in the par.vals list?

berndbischl:
otherwise i think i agree with the summary list. go.

kerschke:
>> By the way, where / when do we need to evaluateParamExpressions.list(obj, dict)?

> to eval the epressions in the par.vals list? 

Ok, that makes sense :)

> otherwise i think i agree with the summary list. go.

cool! then, i'll work on it and let you know, once i'm done. in the meantime, enjoy your holidays ;)

kerschke:
>  i'll work on it and let you know, once i'm done.

Part 1 (updating PH) should be done. The [PR](https://github.com/berndbischl/ParamHelpers/pull/136) is open / ready for review. Let me know, if there are things that you don't like about it. Otherwise, I will go on and update mlr accordingly.

kerschke:
Solved 2) and 4).
Let me know, whether there are still functions remaining, where the documentation should be more explicit.

Once everything's ok from your side (and ready to merge), I can update the tutorial accordingly.

mllg:
Everything looks good. Will merge soon if Travis agrees.

mllg:
@berndbischl @larskotthoff Last call. :wink:

@kerschke Where can I find your tutorial stuff? Can you link it here?

berndbischl:
> @berndbischl @larskotthoff Last call

i am on holiday pls wait at least until monday.

schiffner:
> Where can I find your tutorial stuff? Can you link it here?

In case that you haven't found it already: https://github.com/mlr-org/mlr-tutorial/pull/49

larskotthoff:
Looks good to me.

mllg:
@berndbischl ping ping ping

kerschke:
@berndbischl what's your status on this one? i mean ping, ping, ping ;)

jakob-r:
Test fail because of roxygen2 6.0.0

jakob-r:
@berndbischl please review

mllg:
@berndbischl ping

jakob-r:
@berndbischl ping

R/train.R
```diff
@@ -31,6 +31,10 @@
 train = function(learner, task, subset, weights = NULL) {
   learner = checkLearner(learner)
   assertClass(task, classes = "Task")
+  if (hasExpression(learner)) {
+    dict = getTaskDictionary(task = task)
```
mb706: The `dict` should somehow contain information about `subset`. The learner only sees the subset task, and probably expects the parameters to behave accordingly.

jakob-r: I would assume that the task is already subsetted?

mb706: Subsetting happens in `trainLearner` ([e.g.](https://github.com/mlr-org/mlr/blob/c2d40a3264ef564391260e38acf5be4d8c4f7473/R/RLearner_classif_logreg.R#L22))

jakob-r: Of course. I remember... what I don't remember is the motivation behind this though.

jakob-r:
All reviews were postive. @mb706 comment can be implemented later in another PR.

berndbischl:
@jakob-r @mllg 
can we please comment on why:
a) you merged something that apparently seems wrong? or please tell me why @mb706 is not correct.
if the the subsetting is done later, the the "n" symbol will we incorrect in the dict?

if thats the case, that creates invalid results. 

b) this is a base change. how where the 2 core members who reviewed positively?

berndbischl:
@jakob-r 
i am sorry, i wanted to avoid this, but if it takes 3 min (!) to check EXACTLY what @mb706 said in a test.

```
load_all()
task = makeClassifTask(data = iris, target = "Species")
dict = getTaskDictionary(task = task)
lrn1 = makeLearner("classif.rpart", minsplit = expression(n))
print(getHyperPars(lrn1))
m = train(lrn1, task)
print(m$learner.model$control$minsplit)
m = train(lrn1, task, subset = 1:100)
print(m$learner.model$control$minsplit # is 150! wrong
```

what do you think happens now to people who use that code, and generate results from it?

will revert now

jakob-r:
`n` is documented to be the size of the Task and not of the subset. 

R/evaluateParamExpressions.R
> \item{\code{n}:} the number of observations in the task

berndbischl:
> In other words this example just assumes something that is never mentioned in the documentation.

we can discuss this on hangout, but you cannot be serious

berndbischl:
maybe as reviewer it is also your obligation to think about whether the docs make sense? whether the semantics are good and proper? 

do you want to argue they are, here? that the example i posted is OK? you would like to have that behavior in your own experiments? that the defaults in the other underlying packages work this way?

PR #1127, opened 2016-08-12
getDefaultMeasure now actually works with strings

ja-thomas:
The documentation to `getDefaultMeasure` states that you can use a character like `"classif"` to get the default measure but that didn't work.

```
> getDefaultMeasure("classif")
Error in switch(type, classif = mmce, cluster = db, regr = mse, surv = cindex,  : 
  EXPR must be a length 1 vector
```

Now it works with strings of the type as well as strings of learners

```
> getDefaultMeasure("classif")
Name: Mean misclassification error
Performance measure: mmce
Properties: classif,classif.multi,req.pred,req.truth
Minimize: TRUE
Best: 0; Worst: 1
Aggregated by: test.mean
Note: 
> getDefaultMeasure("classif.rpart")
Name: Mean misclassification error
Performance measure: mmce
Properties: classif,classif.multi,req.pred,req.truth
Minimize: TRUE
Best: 0; Worst: 1
Aggregated by: test.mean
Note: 
```

I also added a test.

### DIFF
--- a/R/Measure.R
+++ b/R/Measure.R
@@ -131,7 +131,7 @@ makeMeasure = function(id, minimize, properties = character(0L),
 #' }
 #'
 #' @param x [\code{character(1)} | \code{\link{Task}} | \code{\link{TaskDesc}} | \code{\link{Learner}}]\cr
-#'  Task type, task, task description, learner name, or a learner.
+#'  Task type, task, task description, learner name, a learner, or a type of learner (e.g. "classif").
 #' @return [\code{\link{Measure}}].
 #' @export
 getDefaultMeasure = function(x) {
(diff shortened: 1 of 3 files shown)


R/Measure.R
```diff
@@ -141,6 +141,10 @@ getDefaultMeasure = function(x) {
     x$task.desc$type
   else if (inherits(x, "Learner"))
     x$type
+  else if (!inherits(try(checkLearner(x), silent = TRUE), "try-error"))
```
ja-thomas: I really dislike this line, if someone has a better solution I'm happy to hear it

larskotthoff:
Great, thanks.

PR #1135, opened 2016-08-12
fix for FeatureImportance print method

zmjones:
- the get function returns na and breaks the measure field in the print method


R/generateFeatureImportance.R
```diff
@@ -210,7 +210,7 @@ print.FeatureImportance = function(x, ...) {
   catf("Task: %s", x$task.desc$id)
   catf("Interaction: %s", x$interaction)
   catf("Learner: %s", x$learner$id)
-  catf("Measure: %s", x$measure$id)
+  catf("Measure: %s", if (!is.na(imp$measure)) imp$measure$id else NA)
```
larskotthoff: Please use `ifelse` here.

Issue #926, opened 2016-06-13
fixDataForLearner is bad


berndbischl:
Look in fixDataForLearnner.R.

Status quo:
- this function is used to "enhance" learner capabilities. 
-  it can do 3 things: 
  -- restore.levels in features. so if factor levels are dropped from features during eg resampling this options restores. 
  verdict: this seems 100% superfluous as we have fix.factors.predictions in makeLearner?
  -- factors.to.dummies: can encode a factor with dummies, so a learner can work on factor features, as well. 
  verdict: this is in principle nice, but: it renders the "factors" property of a learner useless, as we could call this thing now in each base learner. and the latter sucks, you have to add this code in dozens of places. 
  -- ordered.to.int can encode an ordered  factor as int. 
  verdict: this is in principle nice, but see dummy.factors
- the whole code is undocumented 
- we mainly use this function in survival learners currently

jakob-r:
?

berndbischl:
my 2 cents:

- we should remove this now. it would not affect much of the codebase. and like i said, the design seems bad.

- after removal, we can talk about the functionality of  "factors.to.dummies" and "ordered.to.int". mlr should support this. but please in *one*  place. as added functionality for general base learners.
IMHO I would add these as options to makeLearner. as we already have "fix.factors.prediction" here. 
but @mllg seemed to prefer a wrapper. 

please help to make the design better here.

@larskotthoff @schiffner @mllg @jakob-r @giuseppec @ja-thomas

larskotthoff:
I'm ok with removing that code.

A wrapper sounds like a good option for this.

berndbischl:
> A wrapper sounds like a good option for this.

i new that you would say this :)  
i dunno, i am not going to force my opinion here and we could kinda "vote", but my opinion on this is:
- we already have fix.factors.predictions in makeLearner. adding more options to makeLearner here seems more consistent to me 
- for me this whole thing is more closer to the base learner and less about "adding more stuff on top of the learner" (yes, more a "feeling" like a proper argument)
- wrappers are kinda complicated and more verbose

schiffner:
I also don't have totally valid reasons, but:
While "fix.factors.prediction" is, like the name says, a necessary fix, treatment/encoding of categorical features "feels" more like preprocessing to me. Usually, one has several possible options (e.g. treat ordered factors as factors and therefore use dummies vs. encode ordered factors as integers) and either has good reasons to use one option or just try out what works best.
[...]

PR #1139, opened 2016-08-12
Make dummy features wrapper

giuseppec:
this was something that raffael needed for his openml study


schiffner:
I don't know if this is time-sensitive because of openML.
If not I would opt for a more general data-recoding wrapper (see #926)

berndbischl:
this is really low prio at the moment. lets please plan this a bit better.

berndbischl:
we really need to discuss the preprocessing things in a new devel hangout call.
i kinda think we should severly rewrite this, and there are other packages out there, that might do most of the work for us. 
but this is easier to explain in a call....

berndbischl:
(i iam not saying dont try to merge these things in now, but we should IMHO discuss and plan this very soon)

NAMESPACE
```diff
@@ -916,6 +916,7 @@ export(makeCostSensTask)
 export(makeCostSensWeightedPairsWrapper)
 export(makeCustomResampledMeasure)
 export(makeDownsampleWrapper)
+export(makeDummyFeaturesWrapper)
```
ja-thomas: Do not commit NAMESPACE and .RD files

R/DummyFeaturesWrapper.R
```diff
@@ -0,0 +1,35 @@
...
+#'
+#' @template arg_learner
+#' @inheritParams createDummyFeatures
+#' @template ret_learner
+#' @family wrapper
+#' @export
+makeDummyFeaturesWrapper = function (learner, method = "1-of-n", cols = NULL) {
+  learner = checkLearner(learner)
+  args = list(method = method, cols = cols)
+  rm(list = names(args))
```
ja-thomas: This is not really required and confuses more than it helps I guess

giuseppec: We are using this also in other wrappers (e.g. impute wrapper). 
I talked to @mllg and @berndbischl at the last workshop because I also found this confusing. They explained why this makes sense, but I forgot the reason...

Issue #1110, opened 2016-08-10
mini issue: brier score. better note


berndbischl:
we should have a note for "brier" which explains the measure and the formula behind it.
and the difference between brier and multiclass.brier should be explained in both notes.

PR #1143, opened 2016-08-14
Mini fixes measures

ja-thomas:
Some more notes for the new measures see PR #887 and also fix issue #1110 

We also fixed a problem with `getRowEls` and factors.

Also the Rd. files are now checked in


berndbischl:
1) qsr:  This scoring rule is the same as 1 - Brier score.
unit test this then and also use the name of the mlr measure

2) mention in both notes that lsr and logloss are nearly the same, add a test

3) do a test for binary and multiclass brier

PR #1153, opened 2016-08-14
fix printer and add test

ja-thomas:
There was a bug in the printer of `FeatureImportance` objects.

This fixes the undefined global warning warning in master.

I also added a test for that.

And for some reason the measure slot is a list so we have to use the first element. It seems that that `makeS3obj` does this for some reason


zmjones:
ah so you think this is generated by its insertion into the list when returning? that is weird. the input measure is required not to be a list.

berndbischl:
@zmjones 
you used variable "imp" which was not defined in the code block. a typo i guess.
r cmd check caught this though.

zmjones:
ah ok.

Issue #937, opened 2016-06-14
Truncated precision - recall plots


enricoferrero:
Hello

I'd like to visualise the performance of my classifiers using a precision - recall plot.
The problem is that, for some classifiers, these plots appear to be truncated for high precision and low recall values.

Is this a an expected or unexpected behaviour?
If it is expected, how can I interpret these truncated precision - recall curves?

This is the code I'm using:

``` r
bmrk <- benchmark(lrns, subsetTask(classif.task, subset=train.set), rdesc, measures=list(mmce, acc, auc, tpr, tnr, ppv, f1))
pr <- generateThreshVsPerfData(bmrk, measures=list(tpr, ppv))
png(file.path("../data/BenchmarkPR.png"), height=10*300, width=10*300, res=300)
print(
      ggplot(data=pr$data, aes(x=tpr, y=ppv)) +
          geom_path(aes(colour=learner), size=1.5) +
          xlab("Recall") +
          ylab("Precision") +
          scale_colour_brewer(palette="Set1", name="Classifier") +
          theme_bw(base_size=14)
)
dev.off()
```

where `lrns` is a list of learners and `classif.task` is a classification task.

The call to ggplot generates this warning message:

``` r
Warning message:
Removed 6 rows containing missing values (geom_path).
```

And indeed I can see that some classifiers have `NaN` in the `ppv` column corresponding to zero or low values of `tpr`:

``` r
summary(pr$data)
                      learner         tpr              ppv           threshold
 Random Forest            :100   Min.   :0.0000   Min.   :0.5079   Min.   :0.00
 Neural Network           :100   1st Qu.:0.5503   1st Qu.:0.6800   1st Qu.:0.25
 Support Vector Machine   :100   Median :0.6443   Median :0.7306   Median :0.50
 Gradient Boosting Machine:100   Mean   :0.6180   Mean   :0.7311   Mean   :0.50
                                 3rd Qu.:0.7167   3rd Qu.:0.7821   3rd Qu.:0.75
                                 Max.   :1.0000   Max.   :0.9333   Max.   :1.00
                                                  NA's   :6
```

And finally, this is what the plot looks like, you can see that 3 of the 4 curves are truncated:
![benchmarkpr](https://cloud.githubusercontent.com/assets/4481814/16043363/3c520a70-3237-11e6-92c8-bad5da45d7a9.png)

Thank you,

berndbischl:
hi, 

1) we cannot reproduce this.

[...]

Issue #1006, opened 2016-07-12
clean up measure brier.sc code


berndbischl:
see my comments in this PR #861

berndbischl:
```

the measure does NOT contain docs for what it actually does
there is NO REAL unit tests. a real unit test would actually compare a handcalculated result on a mini-data situation with the result of the measure.

```

berndbischl:
also I do agree that brier.scaled is really a better name

berndbischl:
@larskotthoff 
would you please clean this up so we can close this quickly

larskotthoff:
Well, both documentation and test are exactly like for the other measures. I asked about the name 2 months ago and never got a reply.

I'm not sure what you're looking for here. Renaming is easily done, but if we start documenting measures and/or testing them differently, this should be done consistently.

berndbischl:
> Well, both documentation and test are exactly like for the other measures. I asked about the name 2 months ago and never got a reply.

simply ask again then. sometimes we overlook stuff..... too many issues.

> I'm not sure what you're looking for here. Renaming is easily done, but if we start documenting measures and/or testing them differently, this should be done consistentl

we kinda do this consistently. 
a) check measures.R. all non-standard measures contain a little note that gives a hint to the user what happens.
b) check test_base_measures
test_that("check measure calculations"
this contains at least one handcrafted check per measure
(note that we had bugs here for some esoteric measures, and this is something thats REALLY bad for users)

berndbischl:
regarding the docs: 
if its called "accuracy" everbody in ML knows what we compute here. if its called brier.sc, nobody knows what it is. and you also have no chance of figuring this out (except for reading sources). which makes the object pretty useless....

berndbischl:
(and with the last comment i am not referring mainly to the name of the measure. if its called brier.scaled, a better name, i still dont know what it is)

larskotthoff:
Regarding docs: The note for the existing Brier score says "Brier score" -- based on that, "Brier scaled" seems acceptable (although I guess it could say what it's scaled to).

Regarding the tests: Same thing there. Brier.sc uses the same test as Brier.

Like I said, this is probably something that can be improved in general.

PR #1154, opened 2016-08-14
fix issue #937 and issue #1006,  edge NA case for ppv, leave tpr, fpr unchanged 

smilesun:
1. issue #937 said in the plot of ROC like curves, there are truncated part because of the NA values in computing PPV
2. issue #1006 give an example of the edge case, which basically means when the output are all negative, TP+FP =0,  and there are 0/0 .  

3.In this edge case, I set the 0/0 to be either 0 or 1 according whether the highest predicted probability is negative(0) or positive instance(1). 

4.Now both the ROC plotting and the PPV NA valued are eliminated
1. Here I leave the code for TPR and FPR unchanged, since if there are NA value, which means there is only one class in test set, they should always be NA regardless of the threshold. The user should be responsible for this.


berndbischl:
@smilesun 
is this your final code? it looks strange?
can you please clean this up? you dont follow the style guide.

also please add documentation for ppv what you are doing here and comments would also help.

and this check is done automatically by mlr, you dont need it, pls remove

```
 if(!is.null(dim(df))) {
        stop('more than one column probability, currently ppv does not support multiclass')
      }

```

larskotthoff:
ping @smilesun

PR #1157, opened 2016-08-16
adds measures arg to plotThreshVsPerf

### LINKED ISSUE #1061: plotThreshVsPerf: why cant i subset the displayed the measures
assume you have resampled with 5 measures and only want to display 2.

zmjones:
should fix #1061


berndbischl:
you have doced measures twice. i would recommend to run rcheck locally before you push

larskotthoff:
Thanks, merging. Double doc will be fixed by travis.

zmjones:
what is that from?

larskotthoff:
I think you didn't regenerate the documentation with the same version of the file that was committed. You don't need to do that anyway as Travis will update automatically.

zmjones:
ah ok. so i can just change the roxygen doc and let travis do it?

larskotthoff:
Yep. As documented in the wiki :)

berndbischl:
short code review:

1) please really add (at least) one unit test if you introduce a new feature. 

2) 
```
+  measures = checkMeasures(measures, obj)
 +  if (class(measures) == "Measure") {
 +    measures = list(measures)
 +  }
```

this is double-bad. 
a) for the last 3 lines, BBmisc has ensureVector
b) which is already called in checkMeasures, so your code is completely uneeded bloat

Issue #1141, opened 2016-08-13
plotPartialDependency: mini bug in facet title


berndbischl:
look here

https://mlr-org.github.io/exploring-learner-predictions-with-partial-dependence/

at the plot above this text

```

When interaction = TRUE, plotPartialDependence can either facet over one feature, showing the conditional relationship between the other feature and f̂ f^ in each panel, or a tile plot. The latter is, however, not possible with multiclass classification (an example of a tile plot will be shown later).

```

some number are displayed in the facet title with one million digits. 
we had this now a couple of times. we need a simple,. configurable way to avoid this.

PR #1159, opened 2016-08-16
fix for #1141

zmjones:
use significant digits for numeric type facet variables in plotPartialDependence

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -660,10 +660,15 @@ plotPartialDependence = function(obj, geom = "line", facet = NULL, facet.wrap.nr
 
     features = obj$features[which(obj$features != facet)]
 
-    if (!is.factor(obj$data[[facet]]))
-      obj$data[[facet]] = stri_paste(facet, "=", as.factor(obj$data[[facet]]), sep = " ")
-    else
+    if (is.factor(obj$data[[facet]])) {
       obj$data[[facet]] = stri_paste(facet, "=", obj$data[[facet]], sep = " ")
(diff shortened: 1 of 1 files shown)


larskotthoff:
Sorry, master with families was broken. I've rebased the branch.

zmjones:
ah thanks

larskotthoff:
Hmm, you're assuming that it's a number in the else branch. Could you check this explicitly please, or are there disadvantages to doing so?

zmjones:
No disadvantages other than a little bloat but I don't see why it is necessary really. If it isn't a factor or character type then it is a numeric. i think this is enforced by either the task creation or by the assertdataframe in this generationPartialDependenceData when either a task or a dataframe is passed as input.

larskotthoff:
You could create a task without checking, then it wouldn't be checked in `generatePartialDependenceData` either. I know that the chances of this happening are every small, but if it does happen it'll lead to hard-to-debug bugs.

zmjones:
ok i'll change that then.

i apparently have lost my jenkins login. so i can't see what that build failed.

larskotthoff:
That's something unrelated (fixed in #1162). Don't worry about that.

If you can't find your Jenkins password, email me and I'll reset it :)

larskotthoff:
It looks like you didn't pull my master rebase. Could you rebase on master again please?

zmjones:
yep.

larskotthoff:
Ok thanks, I'll merge this. I'll open an issue for making this configurable for plots in general.

PR #1162, opened 2016-08-16
check whether default learner params are actually in param set of learner

larskotthoff:


larskotthoff:
The bug fixed with this causes Jenkins to fail.

berndbischl:
lars, thx. very good. pls merge when you think its ok

Issue #1167, opened 2016-08-17
Problem in  `plotThreshVsPerf` if `pretty.names = FALSE`


schiffner:
Dear plot master @zmjones 

I have a problem with `plotThreshVsPerf` if `pretty.names = FALSE`:
```
lrn = makeLearner("classif.lda", predict.type = "prob")
mod = train(lrn, task = sonar.task)
pred = predict(mod, task = sonar.task)
d = generateThreshVsPerfData(pred, measures = list(fpr, fnr, mmce))
plotThreshVsPerf(d, pretty.names = FALSE)
# Error in melt.data.table(as.data.table(obj$data), measure.vars = mnames,  : 
#   object 'mnames' not found
```
My guess is that in [your latest commit](https://github.com/mlr-org/mlr/commit/1525ff0504fc11bf810de3ffa0d10d13616aef4c) the else case corresponding to
https://github.com/mlr-org/mlr/blob/master/R/generateThreshVsPerf.R#L155
vanished accidentally, but am not sure if I'm missing something.

zmjones:
yep that looks correct. PR incoming

schiffner:
Thx!

PR #1168, opened 2016-08-18
fix pretty.names = FALSE in plotThreshVsPerf

zmjones:
fix for #1167


larskotthoff:
Could you make the test an additional one please so that pretty names are also tested?

zmjones:
so i modified a test so that this bug would have caused a failure. you would prefer that i duplicate the test with only pretty.names changed so the failure would be unambiguous?

larskotthoff:
Yes. In particular the modified test doesn't test what happens for the other value of pretty names :)

zmjones:
No but since it is the default it is checked in all of the other calls to plotThreshVsPerf.

larskotthoff:
Ok, fair enough.

PR #1172, opened 2016-08-21
Relative error measures

catastrophic-failure:
Added the Root relative squared error measure and the Relative absolute error based on this answer http://stats.stackexchange.com/a/203187/60613.

These metrics are good to have in the package because they establish a baseline to compare results, i.e. if the relative error is above 1 you are doing worse than simply predicting the mean result.

- More measures can be added (e.g. Relative squared error)

- I could have used the already existing functions `measureSAE` and `measureSSE`, but I guess that wouldn't be optimal

- The measures won't work, for obvious reasons, with single instances (e.g. doing leave one out)

### DIFF
--- a/R/measures.R
+++ b/R/measures.R
@@ -271,6 +271,77 @@ arsq = makeMeasure(id = "adjrsq", minimize = FALSE, best = 1, worst = 0,
   }
 )
 
+#' @export rrse
+#' @rdname measures
+#' @format none
+rrse = makeMeasure(id = "rrse", minimize = TRUE, best = 0, worst = Inf,
+  properties = c("regr", "req.pred", "req.truth"),
+  name = "Root relative squared error",
+  note = "Defined as sqrt (sum_of_squared_errors / total_sum_of_squares). Undefined for single instances and when every truth value is identical. In this case the output will be NA.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureRRSE(pred$data$truth, pred$data$response)
+  }
+)
+
+#' @export measureRRSE
+#' @rdname measures
+#' @format none
+measureRRSE = function(truth, response){
+  tss = sum((truth-mean(truth))^2L)
+  if (tss == 0){
+    warning("RAE is undefined if all truth values are equal.")
+    return(NA)
+  }
+  sqrt(measureSSE(truth, response)/tss)
+}
+
+#' @export rae
+#' @rdname measures
+#' @format none
+rae = makeMeasure(id = "rae", minimize = TRUE, best = 0, worst = Inf,
+  properties = c("regr", "req.pred", "req.truth"),
+  name = "Relative absolute error",
+  note = "Defined as sum_of_absolute_errors / mean_absolute_deviation. Undefined for single instances and when every truth value is identical. In this case the output will be NA.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureRAE(pred$data$truth, pred$data$response)
+  }
+)
+
+#' @export measureRAE
+#' @rdname measures
+#' @format none
+measureRAE = function(truth, response){
+  meanad = sum(abs(truth-mean(truth)))
+  if (meanad == 0){
+    warning("RAE is undefined if all truth values are equal.")
+    return(NA)
+  }
+  return(measureSAE(truth, response)/meanad)
+}
+
+#' @export mape
+#' @rdname measures
+#' @format none
+mape = makeMeasure(id = "mape", minimize = TRUE, best = 0, worst = Inf,
+  properties = c("regr", "req.pred", "req.truth"),
+  name = "Mean absolute percentage error",
+  note = "Defined as the abs(truth_i - response_i)/truth_i. Won't work if any truth value is equal to zero. In this case the output will be NA.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureMAPE(pred$data$truth, pred$data$response)
+  }
+)
+
+#' @export measureMAPE
+#' @rdname measures
+#' @format none
+measureMAPE = function(truth, response){
+  if (any(truth == 0)){
+    warning("MAPE is undefined if any truth value is equal to 0.")
+    return(NA)
+  }
+  return(mean(abs((truth-response)/truth)))
+}
+
 ###############################################################################
 ### classif multi ###
 ###############################################################################
--- a/tests/testthat/test_base_measures.R
+++ b/tests/testthat/test_base_measures.R
@@ -260,8 +260,8 @@ test_that("check measure calculations", {
   pred.regr.mape = pred.regr
   pred.regr.mape$data$truth = c(5, 10, 1, 5) #we change the 0 target because mape is undefined
   mape.perf = performance(pred.regr.mape, measures = mape, model = mod.regr)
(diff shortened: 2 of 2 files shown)


larskotthoff:
Thanks. Why didn't you want to use the existing functions and then normalise? It seems to me that this would result in less duplicated code. And I may be missing something obvious, but why won't this work with single instances?

catastrophic-failure:
> Thanks. Why didn't you want to use the existing functions and then normalise? It seems to me that this would result in less duplicated code. 

Yes, I could do this, but while it's surely more readable, is it optimal? While I'm not sure how much measure evaluation is a resource drain, these aren't really complicated measures, I guessed calling functions wasn't necessary. But as I said I could just as well have called `measureSAE` and `measureSSE`, if it's the preferred path.

> And I may be missing something obvious, but why won't this work with single instances?

The denominator is proportional to the standard/mean absolute deviation of the `truth` vector. If it's a single stance it becomes zero.

larskotthoff:
Not sure what you mean by optimal. I don't think a function call is going to cause a noticeable slowdown. My preference is to reduce code duplication as much as possible.

I got what you mean about the single instances now, but this is something that can occur in other cases as well. It would be great if you could check for that and provide a suitable error message in that case (and a test).

catastrophic-failure:
@larskotthoff 

> Not sure what you mean by optimal. I don't think a function call is going to cause a noticeable slowdown. My preference is to reduce code duplication as much as possible.

Will do.

> I got what you mean about the single instances now, but this is something that can occur in other cases as well. It would be great if you could check for that and provide a suitable error message in that case (and a test).

The only other way it might happen is if there's a single unique `truth` value. I know it might happen by chance during resampling, but other measures breakdown upon that as well, such as `expvar`. What kind of error message should the measure output?

larskotthoff:
It can happen if all truth values are equal to the mean.

The error message could be something like "Cannot normalise <measure> as all truth values are equal."

catastrophic-failure:
@larskotthoff 
> It can happen if all truth values are equal to the mean.
Exactly, like I said.

> The error message could be something like "Cannot normalise as all truth values are equal."
Will do. Should we do it to Expvar too?

larskotthoff:
The current code base can certainly be improved :)

If you want to, feel free to add similar error messages for the relevant other measures (in a separate PR please). But for this I consider it necessary.

catastrophic-failure:
@larskotthoff I guess it should be a warning instead of an error then? Do resampling continue if the measure function return an error? As of now it accepts the NAs and report them as is.

larskotthoff:
Ah, good point. Yes, let's make it a warning.

berndbischl:
some comments:

1) if some measures fail, e.g. when executed on a single instance, MOST important is that we doc that.
i dont think we need much more here. you could output a warning. but note that warnings usually suck!

2) please doc the precise formula of the measure in the note. or explain it in words so that it is precisely clear. but i guess a formula is shorter.
the note could also mention the (orig) mlr measure that is normalized here. 

3) wrt the tests: i usually prefer REALLY manualy tests, without calling the copy-pasted R functions here. 
i can live with this, but would rather prefer a really handcrafted test.

4) if we are at it: can you add at least the MAPE as well? thats also used very often (sorry)

catastrophic-failure:
@berndbischl 

> 1) if some measures fail, e.g. when executed on a single instance, MOST important is that we doc that.
i dont think we need much more here. you could output a warning. but note that warnings usually suck!

Will add warnings. About doc'ing, you mean the note, or the roxygen description, or something else?

> 2) please doc the precise formula of the measure in the note. or explain it in words so that it is precisely clear. but i guess a formula is shorter.

Could do either, the formula is quite short though, so perhaps it's preferable.

the note could also mention the (orig) mlr measure that is normalized here.

My local has it already, just need to push to the remote.

> 3) wrt the tests: i usually prefer REALLY manualy tests, without calling the copy-pasted R functions here.
i can live with this, but would rather prefer a really handcrafted test.

I'll add manual tests as well.

> 4) if we are at it: can you add at least the MAPE as well? thats also used very often (sorry)

Will do.

catastrophic-failure:
@larskotthoff @berndbischl 
So, is this good enough?
```splus
measureMAPE = function(truth, response){
  if(all(truth != 0))  return(mean((abs(truth-response))/truth))
  warning("MAPE is undefined for actual values equal to 0!")
  return(NA)
}
```

larskotthoff:
Either is fine with me, although I would prefer "truth values all equal" over "actual values equal".

catastrophic-failure:
@larskotthoff 

> I got what you mean about the single instances now, but this is something that can occur in other cases as well. It would be great if you could check for that and provide a suitable error message in that case (and a test).

Now I'm writing this test, along with mape's test. What exactly should I test here? e.g test if it outputs `NA`s? Because the formula will actually return a `NaN` (`0/0`)

larskotthoff:
Also test that you're getting the warning please.

catastrophic-failure:
All right, what you think about this test?

```splus
expect_true(tryCatch(measureMAPE(0,0), warning = function(w) TRUE))
expect_true(tryCatch(measureMAPE(c(1,1,1,0),c(2,2,2,2)), warning = function(w) TRUE))
```

But I'm in doubt how to test that a valid entry doesn't output a warning.

larskotthoff:
See https://stackoverflow.com/questions/22003306/is-there-something-in-testthat-like-expect-no-warnings

catastrophic-failure:
If I output `NA`'s, they aren't equal to `NaN`'s, should we output `NaN`'s then?

larskotthoff:
I think `NA`s are more suitable here.

catastrophic-failure:
Then the handcrafted test and the measure won't agree: `expect_equal(NA, NaN)` throws an error.

larskotthoff:
Well, you can change the handcrafted test accordingly and say that in a comment.

catastrophic-failure:
The tests are
`expect_equal(mape.test, mape$fun(pred = pred.regr))`

And

` expect_equal(1/4*(abs((4-5)/5)+abs((11-10)/10)+abs((0-0)/0)+abs((4-5)/5)), measureMAPE(c(5, 10, 0, 5),c(4, 11, 0, 4)))`

I have no idea how to change the test without making it trivial. But obviously they aren't equal: one entry's a NA the other a NaN.

larskotthoff:
So set the left hand side to `NA`. I don't see how this would be worse than the current test -- you're checking the value in a particular case.

berndbischl:
> 1) if some measures fail, e.g. when executed on a single instance, MOST important is that we doc that.
> i dont think we need much more here. you could output a warning. but note that warnings usually suck!
> Will add warnings. About doc'ing, you mean the note, or the roxygen description, or something else?

i meant the note. that seems to be the only place where we can currently doc such objects. i am not too happy about this, but thats how it is for now.

i see you already have those notes in there. but its good to be precise. say that the measure will become NA in the "undefined case". dont leave that info out.




> 2) please doc the precise formula of the measure in the note. or explain it in words so that it is precisely clear. but i guess a formula is shorter.
> Could do either, the formula is quite short though, so perhaps it's preferable.

yes, formulas seem best.

berndbischl:
style comment: after an "if" there goes a SPACE so, "if (foo)"

larskotthoff:
Thanks, merging.

Issue #1174, opened 2016-08-23
plotPartialDependence: bug with factor input


berndbischl:
```
task = makeRegrTask(data = iris, target = "Sepal.Length")
lrn = makeLearner("regr.lm")
mod = train(lrn, task)
pd1 = generatePartialDependenceData(mod, task)
print(pd1)
print(plotPartialDependence(pd1))


pd2 = generatePartialDependenceData(mod, task, features = getTaskFeatureNames(task)[4])
print(pd2)
print(plotPartialDependence(pd2))
```

pd1 is totally broken, hence the plot of course too
```
PartialDependenceData
Task: iris
Features: Sepal.Width, Petal.Length, Petal.Width, Species
Target: Sepal.Length
Derivative: FALSE
Interaction: FALSE
Individual: FALSE
  Sepal.Length Sepal.Width Petal.Length Petal.Width Species
1        5.319       2.000           NA          NA    <NA>
2        5.451       2.267           NA          NA    <NA>
3        5.583       2.533           NA          NA    <NA>
4        5.716       2.800           NA          NA    <NA>
5        5.848       3.067           NA          NA    <NA>
6        5.980       3.333           NA          NA    <NA>
... (33 rows, 5 cols)

```

pd2 looks better, but the plot is completely empty
```
PartialDependenceData
Task: iris
Features: Species
Target: Sepal.Length
Derivative: FALSE
Interaction: FALSE
Individual: FALSE
  Sepal.Length    Species
1        6.426     setosa
2        5.702 versicolor
3        5.402  virginica
```

zmjones:
this is a weird case no? you are doing regression on multilevel factor?

berndbischl:
> this is a weird case no? you are doing regression on multilevel factor?

i dont understand that sentence? did you misread something? the factor is simply an input feature here...

zmjones:
the generation function output isn't broken at all as far as i can see. in the first case the data element of the container has a column for each of the features and the target, and since interaction = FALSE the column of each feature is NA except for gridsize rows, unless there are fewer than gridsize unique values in the input data, as is the case with the Species feature. 

The plot is messed up though. Checking that out now.

zmjones:
i just saw that you were using the iris data and thought at first you were using regression on that. it was early in the morning...

berndbischl:
> the generation function output isn't broken at all as far as i can see.

ah i always get confused by this weird "rbind.fill-format"....
[...]

PR #1180, opened 2016-08-23
fix for #1174

zmjones:
melt.data.table doesn't implicitly cast factors to numerics but
everything to character when the measure.vars have mixed type

also re-added geom_point which someone removed and added some comments.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -540,7 +540,7 @@ doAggregatePartialDependence = function(out, td, target, features, rng) {
       stop("function argument must return a sorted numeric vector ordered lowest to highest.")
 
   if (all(target %in% td$class.levels)) {
-    out = melt(out, id.vars = features, variable = "Class", value.name = "Probability")
+    out = melt(out, id.vars = features, variable = "Class", value.name = "Probability", variable.factor = TRUE)
     out$Class = stri_replace_all(out$Class, "", regex = "^prob\\.")
   }
   out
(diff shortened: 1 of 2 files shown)


berndbischl:
@zmjones 
i hate it that we cannot test this.
i will look at my problematic examples now manually then merge

zmjones:
yea me too. wasn't lars going to help on one of the hiwis to reenable the plot tests? they would have caught this when we made the switch to data.table.

berndbischl:
> yea me too. wasn't lars going to help on one of the hiwis to reenable the plot tests? they would have caught this when we made the switch to data.table.

@larskotthoff 

please just tell me what task i have to assign to someone....
we need the tests

larskotthoff:
There are lots of expectations in the plot tests that are commented. Uncomment, make tests pass.

berndbischl:
@zmjones 

i went into your branch the code for pd1 doesnt even run through?
i am seeing this here:
Error: (list) object cannot be coerced to type 'double'

here is a full example

```
task = makeRegrTask(data = iris, target = "Sepal.Length")
lrn = makeLearner("regr.lm")
mod = train(lrn, task)
pd1 = generatePartialDependenceData(mod, task)
print(pd1)
print(plotPartialDependence(pd1))
```

that, as travis seems to run thru, seems to imply we are missing a unit test here, too

zmjones:
trying to figure out why they don't fail now.

zmjones:
figured out why the test missed it. new test added. i think i've also finally solved my emacs indenting problem so there is some movement from that.

Issue #1045, opened 2016-08-01
corner case: resample  printer output on train data when no train preds are there


berndbischl:
```
meas = setAggregation(mmce, train.mean)
r = resample("classif.rpart", iris.task, cv2, measures = meas)
print(r)

```

this creates NA for the aggregated value, but strangely displays a mean and an sd?

```
Resample Result
Task: iris-example
Learner: classif.rpart
mmce.aggr: NA
mmce.mean: 0.05
mmce.sd: 0.01
Runtime: 0.0179381
```

berndbischl:
i also dont know whether i really like that long output. "aggr" and "mean" are usually redundant?

PhilippPro:
I just looked over it. The only measure, where it is different is rmse, where we make a \sqrt after the mean, all other have test.mean which is just the mean. 
Example:
```
lrn = makeLearner("regr.rpart")
rdesc = makeResampleDesc("CV", iters = 30)
r = resample(lrn, bh.task, rdesc, measures = list(rmse, mse))
print(r)
```
For me, I would delete aggr (or just not print it, maybe somebody is using it in real application), the one who is interested can make his/her own measure. 

Maybe it would be also interesting to measure on all predicted observations and not only on the specific folds. (e.g. auc over all observations, and not only mean on the folds; I am not sure now, which is the more exact measure... (drawback would be no sd))

berndbischl:
PR is here #1187

giuseppec:
Seems that https://github.com/mlr-org/mlr/pull/1187 has been merged, can this issue be closed?

PR #1187, opened 2016-08-24
introduce Aggregation class properties and clean up ResampleResult printer

berndbischl:
this addresses issue #1045 
it became a lot longer than i had hoped. 
summary of what i did:

1) Aggregation objects now have properties. they define whether they need the trains set evals, test set evals, or both. this allows to generate an informative error msg for the NA problem in the issue.

2) i removed the "mean" and "sd" display from the ResampleResult printer. this was kinda hacked in before, confusing as it used different aggregation functions, and also did not reliably work. 
it also nearly never provided more info but ate up extra lines in a useless manner.
i printed the aggreated measure performance in a standard way instead.

3) some tests

### DIFF
--- a/R/checkAggrBeforeResample.R
+++ b/R/checkAggrBeforeResample.R
@@ -1,0 +1,25 @@
+# check whether rdesc$predict is set, so that the requiring properties of the measure are satisfied
+# called the beginning of resample
+checkAggrBeforeResample = function(measure, rdesc) {
+  a = measure$aggr
+  p = a$properties
+  pred = rdesc$predict
+  p.allowed = if (all(c("req.train", "req.test") %in% p)) {
+    "both"
+  } else if ("req.train" %in% p) {
+    c("train", "both")
+  } else if ("req.test" %in% p) {
+    c("test", "both")
+  } else {
+    c("train", "test", "both")
+  }
+  if (pred %nin% p.allowed)
+    stopf("Aggregation '%s' not compatible with resampling! You have to set arg 'predict' to %s in your resample object, instead it is '%s'!", a$id, stri_paste("'", p.allowed, "'", collapse = " or "), pred)
+}
+
+# map the checker over multiple measures
+checkAggrsBeforeResample = function(measures, rdesc) {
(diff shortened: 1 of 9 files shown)


berndbischl:
this is more or less a core change. so please review very carefully!

R/checkAggrBeforeResample.R
```diff
@@ -0,0 +1,25 @@
+# check whether rdesc$predict is set, so that the requiring properties of the measure are satisfied
+# called the beginning of resample
+checkAggrBeforeResample = function(measure, rdesc) {
+  a = measure$aggr
+  p = a$properties
+  pred = rdesc$predict
+  p.allowed = if (all(c("req.train", "req.test") %in% p)) {
```
larskotthoff: Why not `setequal` here?

larskotthoff:
Looks good to me, but somebody else should review as well.

R/checkAggrBeforeResample.R
berndbischl: > Why not setequal here?

because my code expresses what i want to say, and is more robust them setequal. consider that the aggregation properties might grow larger. and contain other stuff.

berndbischl:
> Looks good to me, but somebody else should review as well.

thx + agree

berndbischl:
@jakob-r @mllg @zmjones  @schiffner 
can somebody review so i can merge

zmjones:
looks good to me. 

is there a reason for the expect_false -> is.na setup? could this ever be NaN or something other than a length 1 numeric?

berndbischl:
> is there a reason for the expect_false -> is.na setup? could this ever be NaN or something other than a length 1 numeric?

do you mean the code cleanup i did test_base_resample?
i simply cleaned up some (not so good) code here, which is unrelated.

resampleresult$aggr can be NA when the training or something else failed. 
it is one of our "idioms" in tests to check that everything correctly ran through.

zmjones:
yes that is what i meant. then merge away! i fixed a grammatical error in the docs but that is it.

berndbischl:
thx. someone pls merge when travis is done

zmjones:
should i wait for jenkins?

PR #1194, opened 2016-08-25
add weight fun arg to generatePartialDependenceData and generateFunctionalANOVAData

zmjones:
for use with methods to detect regions of extrapolation. their use here down weights predictions made at regions of low data density. this could be used with a density estimator, some sort of classifier (e.g., [Hooker 2004](http://dl.acm.org/citation.cfm?id=1014121)) or by using [data depth](https://www.jstor.org/stable/120138?seq=1#page_scan_tab_contents) which is what I've been testing this with.

this is particularly important with partial dependence since the algorithm assumes the features come from a product distribution (i.e., p(x_u, x_-u) = p(x_u)p(x_-u). since they usually don't functional behavior can be displayed at parts of the joint distribution with little or no density/mass. this means that the output reflects the learner's properties more than the data.

will add this to the tutorial and so on as well. i thought of integrating the extrapolation detection into mlr but as far as i know now this is the only use. i think i may write a package which has a bunch of methods for extrapolation detection in it and if we want to integrate that we can do so then.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -60,6 +60,10 @@
 #'   target feature.
 #'   The default is the mean, unless \code{obj} is classification with \code{predict.type = "response"}
 #'   in which case the default is the proportion of observations predicted to be in each class.
+#' @param weight.fun [\code{function}]\cr
+#'   A function which takes a numeric vector \code{x} and a \code{data.frame} \code{data} and returns a
+#'   numeric weight. The length of \code{x} must match the column dimension of \code{data}.
+#'   By default \code{weight.fun} returns a weight of 1 for every input.
 #' @param bounds [\code{numeric(2)}]\cr
 #'   The value (lower, upper) the estimated standard error is multiplied by to estimate the bound on a
 #'   confidence region for a partial dependence. Ignored if \code{predict.type != "se"} for the learner.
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks. Could you add a test for this please? And it looks like some of the existing tests are breaking.

zmjones:
yea will be fixed shortly i think.

so i am not sure about the test. the default argument to weight.fun = function(x, data) 1 which means that every time either the partial dependence or functional anova generation functions are called this tests that weights can be applied. i can't see how substituting in another function which does something else would increase our probability of catching any bugs.

the only edge case i can think of is if weight.fun returned NA or NaN. i think the behavior here is appropriate though. the partial dependence or functional anova effect is undefined here and should be omitted from the plot. the data depth methods i am working on now do this, though it wouldn't be an issue with an extrapolation classifier or a density estimator which would return 0 in an extreme case.

larskotthoff:
Is there a way to test if the weights work at all?

zmjones:
that is what is being tested now. in every case all the points get equal weight. that is why the tests are failing now also: the weights code isn't working correctly for some classification cases.

larskotthoff:
Well what I mean is that ideally we would test if changing the weights has the desired effect.

zmjones:
Ah ok. Yes that is theoretically possible. I'll have to think of a example to do that. We can wait on that before merging it.

zmjones:
so the case that is broken is with classification when predict.type = "response". i'm not really sure what to do with this case. before the addition of the weight.fun argument, the user would have to have specified the argument fun to reduce the class predictions to a vector of the same length as the number of classes, probably the proportion of each class, which aggregates the predictions for each iteration of the partial dependence computation. 

there is a conflict with the weighting though. basically in this case the user would need to specify a fun which accepts weights, e.g., a weighted table. i guess i can check to see whether the fun arg has an argument called weights in this case and throw an error if not?

larskotthoff:
Hmm, could you apply the weights function to the vector of responses before aggregation?

zmjones:
no because they are factors in this case

larskotthoff:
I would throw an error in this case then.

berndbischl:
> i did think about removing this functionality entirely. i.e., you could only compute the partial dependence with classifiers that could output probabilities, but that seems too drastic.

i am not sure whether this is too drastic. how DO you compute the PD curve if you dont have probs but only classes? and what are you expecting to see here?

ah well, ok, i might see your point. you can see the points where the classification changes. 
like if x1 < 7, then it's y = 0, but afterwards it's y=1.

zmjones:
yea it is not terribly interesting, but since we have so many learners which don't output probabilities or scores i thought we should keep it in

berndbischl:
> yea it is not terribly interesting, but since we have so many learners which don't output probabilities or scores i thought we should keep it in

ok agreed

zmjones:
test has been added now which is an example where the partial dependence is computed over an empty part of the joint distribution, with the joint distribution as the weight function. in this case the weights are all 0, and in this case i return NA.

larskotthoff:
Thanks, merging.

berndbischl:
this broke the tutorial, see #1199 
pls fix

PR #1202, opened 2016-08-28
fix small bug in listLearners

berndbischl:
defaults were missing for the "obj" arg in the s3 methods. 

i have added a test were this lead to a bug before when i used the method.

@larskotthoff @mllg @jakob-r 

pls review


berndbischl:
@zmjones

PR #1206, opened 2016-08-29
require weight.fun to be vectorized

zmjones:
accept a data.frame for x arg to weight.fun in
generatePartialDependenceData and generateFunctionalANOVAData which
allows the use of faster vectorized weighting functions. for
nonvectorized operations users can write a loop or use apply. also now
checking that weights are nonnegative.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -61,9 +61,10 @@
 #'   The default is the mean, unless \code{obj} is classification with \code{predict.type = "response"}
 #'   in which case the default is the proportion of observations predicted to be in each class.
 #' @param weight.fun [\code{function}]\cr
-#'   A function which takes a numeric vector \code{x} and a \code{data.frame} \code{data} and returns a
-#'   numeric weight. The length of \code{x} must match the column dimension of \code{data}.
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, merging.

larskotthoff:
@zmjones Looks like this broke the tutorial again...

```
Knitting file 'partial_dependence.Rmd' ...
Quitting from lines 163-166 () 
Error in lapply(list.files(src, pattern = "\\.Rmd$"), knitIt, "pdf") : 
  Assertion on 'weights' failed: Must have length 150, but has length 1.
Calls: lapply ... doPartialDependenceIteration -> assertNumeric -> makeAssertion -> mstop
Execution halted
```

berndbischl:
can you please, in cases like these, you when fixing this, always add a unit test?

zmjones:
you mean with fixing this new tutorial issue? i have been doing that (e.g., my last pr). this pr wasn't a bug fix though.

berndbischl:
i mean: apparently the tutorial is running other code examples then we in our tests. so we should extend our own tests with the code examples that apparently trigger the breakage. 

and as they are in the tutorial, i would assume they are very valid use cases....

zmjones:
fair enough

PR #1207, opened 2016-08-29
bug in weight.fun default for centered partial dependence

zmjones:
the default weight.fun was inappropriately set to its older
nonvectorized identity weight function. bug was identified in pr #1206.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -262,7 +262,7 @@ generatePartialDependenceData = function(obj, input, features,
         out = parallelMap(doPartialDependenceIteration, i = seq_len(nrow(rng)), more.args = args)
         if (!is.null(center) & individual)
           centerpred = doPartialDependenceIteration(obj, data, center[, x, drop = FALSE],
-            x, fun, td, 1, bounds = bounds, weight.fun = function(x, data) 1)
+            x, fun, td, 1, bounds = bounds, weight.fun = function(x, data) rep(1, nrow(x)))
         else
           centerpred = NULL
       }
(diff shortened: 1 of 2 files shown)


zmjones:
is there any way we could automate a devel tutorial build here?

berndbischl:
> is there any way we could automate a devel tutorial build here?

i talked to @larskotthoff about this during the WS. I dont think there is an easy convenient way.

zmjones:
i guess i could just add that in addition to local testing it would make sense to rebuild the tutorial as well. that definitely makes small changes a bit more time consuming though. i guess it isn't that big of a deal though.

berndbischl:
the problem is:
we would need to clone the tutorial repo turing the master mlr travis build for the R package. then build the tutorial. (I dont think we can trigger it the other way around AND display the results in such a PR)

but the former WOULD work. unfortunately I am certain we will then hit the dreaded time limit again.

zmjones:
i guess this brings up the question of why we have a tutorial rather than a vignette as well. i assume that has been discussed before? a vignette would certainly be more discoverable i would think. that would also solve this problem.

larskotthoff:
I think the main problem wouldn't be any Travis time limit (there is no global limit), but that a single build would then take 2-3 hours. I like @berndbischl's idea of having the code that breaks in the tutorial in unit tests/examples better.

zmjones:
in the limit these are the same things though. if code coverage is better in the tutorial (and we aren't repeating ourselves too much there) then, if we introduced bugs randomly, we'd eventually end up with every line of code that is in the tutorial in the unit tests, except we'd have had a whole bunch of things like this occur as well.

larskotthoff:
But the build wouldn't be significantly slower. That's the main point I worry about.

zmjones:
it wouldn't be? i thought the expensive part of the tutorial build was the code execution, not the pdf generation, though i realize they are inseparable when we are actually doing it.

if we had a vignette we could have less duplication across it and the unit tests. so yes the mlr build process would be slower but it could be a good bit faster than mlr + mlr-tutorial. if we have every case that is in the tutorial in the unit tests the mlr build woudn't be much faster than mlr + mlr tutorial i would think.

berndbischl:
@zmjones 
if you want to have a vignette, i would suggest to discuss this in another thread, no here.
i now did what you should have done in the 1st place ;-)
https://github.com/mlr-org/mlr/issues/1208

berndbischl:
but one final point you are REALLY wrong about (IMHO):
If something breaks in the tutorial, but the package tests pass, the unit test are incomplete!
thats IMHO a fact. which is not solved by saying "uuuh, I guess we should copy the tutorial into the unit tests then".
the unit tests already cover most of what is in the tutorial. cases like these here point us to the few places where they are incomplete. so we should react to this, in the proper way, and extend the tests.

Issue #1026, opened 2016-07-18
add learner regr.cvglmnet


schiffner:
We have learners classif.cvglmnet and surv.cvglmnet.
We could add this for regression, too.

berndbischl:
learner was added. closing, PR #1209

PR #1209, opened 2016-08-31
#1026 regr.cvglmnet

florianfendt:
Update for NEWS:

````
## learners - new
* regr.cvglmnet
````


berndbischl:
you have hardcoded "family". this is not good (and should also in general be documented in "note"!)

but here, also "poisson" is possible, pls make this a normal param

florianfendt:
done

mllg:
Thanks!

Issue #1195, opened 2016-08-25
Measures that fail on single/all identical samples


catastrophic-failure:
Per #1172 now measures that don't work on  single/all identical samples may include a warning. I'm willing to add the warnings and the tests if there's a consensus on which measures should have them.

For starters, `expvar` fails just like `rrse`.

---

There might be overlap with #1154, so an unified approach needs to be defined.

larskotthoff:
I think you've just said which measures should have them -- those that fail on single/all identical truth/prediction values :)

We can expand this definition later, but it would be very helpful if you could do a PR for those.

catastrophic-failure:
@larskotthoff Sorry, I meant some kind of list. Inspecting the functions reveal some problematic ones, but I'm not used to all of them, so I'm probably going to miss some.

larskotthoff:
Start with the ones you can easily identify.

berndbischl:
> There might be overlap with #1154, so an unified approach needs to be defined.

that PR is about something else, there will be no overlap here is guess (hope?)
at least what the topic concerns. if you touch measures like TPR and so on (ROC) there might be (hopefully fixable) conflicts.

and dont forget to write one unit test then where you test all measures on a single instance.

catastrophic-failure:
@berndbischl @larskotthoff I just started working on this in #1211 . I see `ber` also handles special cases, and returns an `NA_real_` in case of missing predictions.

catastrophic-failure:
There's a possible issue we should solve. I'm adding warnings to the `measureX` functions, but I think it should be better located at the `x` objects. The reason is simple: if, for any reason, we add a measure that uses the same `measureX` the warning won't make sense to the user. What do you think?

larskotthoff:
I would make the error message suitable for all calling measures.

catastrophic-failure:
Hey, long time no see. I'm finishing my Master's Dissertation, going to defend near the end of February, so I revisited the repository to see what's changed.

I'm closing this issue as the commit has been merged already, but if any measure is missing tests and warnings (and tests for warnings), ping me here and I'll add those in new PRs, so I continue to contribute to the repo.

PR #1211, opened 2016-09-01
DON'T MERGE YET: Added warning and tests to EXPVAR

catastrophic-failure:
Just as we were discussing in #1195 I'm starting to add these warnings to measures. Please refer to the issue to suggest other measures that need the warnings in singularity cases.

For now, I edited `expvar` and included tests (I'm busy these days in my Masters). Expect other measures the following weeks.

And also please don't merge yet.


catastrophic-failure:
#1228 added conflicts to things I was working here, any other day I'll look into this.

berndbischl:
> #1228 added conflicts to things I was working here, any other day I'll look into this.

i am really sorry. i saw your note in the other thread..... :(
but we needed this for the new mlr release.

catastrophic-failure:
@berndbischl No need to worry. I meant is that I don't have much time to keep looking at this everyday now. And I do understand the speed up to make it to 2.10. Next measures I'll look into other days are probably AUCs, but these need special attention I think.

catastrophic-failure:
I was going to start with the many `AUCs` but it seems `colAUC` throws errors already if less than 2 classes labels are entered. Should I still edit the measures, so throwing warnings and returning `NA_real_` before the call is made?

larskotthoff:
Yes please.

catastrophic-failure:
`au1u` and `au1p` do not fail in multiclass classification when at least two classes are represented in the prediction (even if K >= 3). You get bad results because the class not represented push the average towards 0.5, but I'm not sure that's the intended behavior.

@Coorsaa  could you enlighten me on this please? When not every class is present at prediction, should `au1u` and `au1p` fail?

PhilippPro:
In k-fold CV this can happen very often, so if anything I would print a warning. In aunp and au1p the auc's are also weighted according to the a-priori distribution of the classes, so the effect should not be so severe in these cases.

catastrophic-failure:
Hey, sorry for the long halt. I'm reaching a few deadlines on my Masters right now, but I'm probably coming back to this at the second half of November. See I allowed edits from maintainers aswell. See ya.

catastrophic-failure:
I think this might be complete now. What you think of the way I dealt with Adjusted R-squared, as it has no `measure` function?

larskotthoff:
Looks good. A few minor points:

- In the error message for adjusted r-squared I would say what n and p are explicitly.
- Please have a look at the style guide, in particular space after comma and the indentation of 2 spaces.

Thanks!

catastrophic-failure:
Thanks for the review @larskotthoff I'll look into this tomorrow, it's already 2:30 A.M. here haha

Issue #1201, opened 2016-08-28
listLearners: give output a s3 class and have printer


berndbischl:
everytime i call listLearner, the (table) output looks really crappy on the console. 
expecially as the not is included. 

shall we simply give the output an s3 class, which we simply add to the data.frame, so it looks better?

returning data.table would (of course) already mitigate the problem a lot. but the "note" column is still bad.

we could also have an option not to include this by default?

or: my last sentences are bad design and we really should handle this in a printer?

@mllg  @larskotthoff what do you think?

berndbischl:
@zmjones  @jakob-r

larskotthoff:
I would do this with a printer.

mllg:
:+1: for S3 printer to generate something more compact if this bugs you.

PR #1213, opened 2016-09-05
listLearners output as S3 class with print

Coorsaa:
First try of making the listLearners output a bit "nicer".


berndbischl:
1) classname: you dont follow the style guide. and your class name is not good. 
use: "ListLearners"

2) of course, you still have to return a df! you have now changed the return type. just extend its class by using BBmisc::addClasses

3) you never call print in this case at the end of listLearners! thats why you are supposed to write the printer! simply return the normal object, the printer is then called automatically from the console!

4) why.... are you writing code which now outputs a gazillion lines per single learner?
just simply call printHead on the data.frame? and exlude the "note" column?
and then ask if you think something else needs to be improved?

Coorsaa:
referring to issue #1201:
I added the S3 class ListLearners and removed the note column from the output which shows the head of the resulting data frame. and looks now like that:

````
             class                                               name short.name      package    type installed numerics factors ordered missings weights prob oneclass twoclass multiclass class.weights
1   classif.avNNet                                     Neural Network     avNNet         nnet classif      TRUE     TRUE    TRUE   FALSE    FALSE    TRUE TRUE    FALSE     TRUE       TRUE         FALSE
2      classif.bdk                         Bi-Directional Kohonen map        bdk      kohonen classif      TRUE     TRUE   FALSE   FALSE    FALSE   FALSE TRUE    FALSE     TRUE       TRUE         FALSE
3 classif.boosting                                    Adabag Boosting     adabag adabag,rpart classif      TRUE     TRUE    TRUE   FALSE     TRUE   FALSE TRUE    FALSE     TRUE       TRUE         FALSE
4      classif.C50                                                C50        C50          C50 classif      TRUE     TRUE    TRUE   FALSE     TRUE    TRUE TRUE    FALSE     TRUE       TRUE         FALSE
5  classif.cforest Random forest based on conditional inference trees    cforest        party classif      TRUE     TRUE    TRUE    TRUE     TRUE    TRUE TRUE    FALSE     TRUE       TRUE         FALSE
6    classif.ctree                        Conditional Inference Trees      ctree        party classif      TRUE     TRUE    TRUE    TRUE     TRUE    TRUE TRUE    FALSE     TRUE       TRUE         FALSE
  featimp    se lcens rcens icens
   FALSE FALSE FALSE FALSE FALSE
   FALSE FALSE FALSE FALSE FALSE
    TRUE FALSE FALSE FALSE FALSE
   FALSE FALSE FALSE FALSE FALSE
    TRUE FALSE FALSE FALSE FALSE
   FALSE FALSE FALSE FALSE FALSE
````

mllg:
Thanks!

PR #1214, opened 2016-09-05
gamboost implementation + clean up code in other mboost learners

MariaErdmann:
I have two questions:

1) Implementing the learner like i just pushed it, leads to following "problem". If no family in par.vals is specified, then I get the following error:
 `Error in trainLearner.classif.gamboost(.learner = list(id = "classif.gamboost",  : 
  argument "family" is missing, with no default`

Mini-example

```
lrn = makeLearner("classif.gamboost")
```

There are thee  solutions I tried, I tend to do c) but I am not sure if this is the right way to do it.

a) `set par.vals = list(family = "Binomial", Binomial.link = "logit")` in makeLearner (I am pretty sure that this is not the right place to solve the 'problem' but this is they way it is implemented in classif.glmboost)

b) using following code snippet in trainLearner

```
defaults = getDefaults(getParamSet(.learner))
if (missing(family)) family = defaults$family
if (missing(Binomial.link)) Binomial.link = defaults$Binomial.link
```

c) Doing the following in the signature of trainLearner would also solve the 'problem'

```
trainLearner.classif.gamboost = function(.learner, .task, .subset, .weights = NULL, offset = NULL, mstop, nu, risk, stopintern, trace, family = "Binomial", Binomial.link = "logit", ...) {
...
}
```

2) predict.type = "prob" is only working for the binomial family, not for AUC or AdaBoost. How shall I handle this?
Note, in mboost's `predict()`, `type` can be either `"class"`, `"response"` or `"link"`. `"link"` is the default and is on the scale of the predictors. `"response"` is on the scale of the response and gives the predicted probabilities (thus corresponds to mlr's "prob"). `"class"` returns the predicted classes (thus, corresponds to mlr's "response").

Mini-example (you need mlr and mboost to run this).

``` library(mboost)
mod = gamboost(Class ~., data = binaryclass.df, family = mboost::AUC())
p = predict(mod, newdata = binaryclass.df, type = "response")
p
# [1] NA
```

Thank you!


ja-thomas:
> a) set par.vals = list(family = "Binomial", Binomial.link = "logit") in makeLearner (I am pretty sure that this > is not the right place to solve the 'problem' but this is they way it is implemented in classif.glmboost)

Well we change the defaults from the original gamboost, so we should put it in the par.vals. We also do that in glmboost.

MariaErdmann:
Ok, finally all three gamboost learners with tests. Unfortunatelly, the tests throw quite a lot of warnings since gamboost doesn't like linear extrpolation so much.
Going through the code of the other mboost learner to help me with gamboost, I saw some "beauty" issues   and improved the code at some places.

larskotthoff:
Thanks. Why did you comment some of the parameters for the existing learners?

MariaErdmann:
I outcommented those that come with family "ProppOdds" which is a family for tasks where the target must be ordered. I was told that ordered are to come soon and therefore just outcommented everything that has to do with this class. Is it better to remove all the code (also in the tests a quite big part is outcommented because of ProppOdds) or shall I write a comment?

larskotthoff:
Sounds good as it is -- I'll merge this, but could you please open an issue to remind us to change it when ordered factors are supported?

PR #1220, opened 2016-09-07
batchmark: parallel Benchmarks with batchtools

ja-thomas:
This is an alternative to the benchmark function + parallelMap, where we create one job for every train/test split. The jobs can be submitted with `submitJobs()` and later collected with `reduceBatchmarkResult()` which creates an benchmark object identical to `benchmark()`. There was some refactoring in the benchmark and resample functions.

This is using batchtools, which is not yet on CRAN.

There is still some cleanup to do and I want to do something smarter in the tests so that it is not so much copy+paste code between `test_base_benchmark` and `test_batchmark`.

But still some reviews might be helpful. The basic functionality seems to work since I can run all benchmark tests with batchmark.


PhilippPro:
What is the advantage to parallelMap?

ja-thomas:
We don't need to keep the master process alive. Also we can restart failed jobs without rerunning the whole benchmark.

ja-thomas:
I would guess jenkins need to be reconfigered.

But we will wait until checkmate (and batchtools) is on CRAN anyways

berndbischl:
side remark: see what i did in PR #1224 and do it here too

ja-thomas:
This now works directly with OpenML datasets, but now we also suggest OpenML. I would guess we don't want to merge that into master in the near future, but the usecase is still quite cool I think.

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+#' learners, tasks and resamplings. The experiments are then stored in a registry and the 
+#' runs can be started via \code{\link[batchtools]{submitJobs}]}. A job is one train/test split
+#' of the outer resampling. In case of nested resampling (e.g. with \code{\link{makeTuneWrapper}}), 
+#' each job is a full run of inner resampling, which can be parallelized in a second step 
+#' with \pkg{ParallelMap}. For details on the usage and support backends have 
+#' a look at the batchtools tutorial page: 
+#' \url{https://github.com/mllg/batchtools}.
+#' 
+#' The general workflow with \code{batchmark} looks like that:
+#' \itemize{
```
mllg: I think you are looking for \enumerate{}

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+#' \item{1.}{\code{batchtools::createExperimentRegistry()}}
+#' \item{2.}{\code{batchmark(...)}}
+#' \item{3.}{\code{batchtools:submitJobs()}}
+#' \item{3.5}{wait until the jobs are finished...}
```
mllg: I don't get why this is only a half step

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+#' \item{4.}{\code{reduceBatchtoolsResult()}}
+#' }
+#' @inheritParams benchmark
+#' @param data.ids [\code{integer}]\cr
+#'   Dataset IDs to download from Open-ML. Default is none.
```
mllg: I'm not sure that this is the right place. Is there a good argument against just getting the tasks before calling batchmark? 
```{r}
f = function(id) convertOMLTaskToMlr(getOMLTask(id))
tasks = lapply(data.ids, f)
```

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+  names(resamplings) = task.ids
+  
+  measures = ensureBenchmarkMeasures(measures, tasks)
+  
+  assertFlag(models)
+
+  reg$packages = union(reg$packages, "mlr")
+  
+  #set name of learner list
+  lrn.ids = vapply(learners, getLearnerId, FUN.VALUE = character(1))
```
mllg: BBmisc::vcapply

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+
+getAlgoFun = function(lrn, measures, models) {
+  force(lrn)
+  force(measures)
+  force(models)
+  function(job, data, instance) {
+
+    extract.this = getExtractor(lrn)
+    if (is.numeric(data$task))
+      data$task = OpenML::convertOMLDataSetToMlr(getOMLDataSet(data$task))
```
mllg: We read this twice for each job?

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+    
+    if (nrow(exps) > 0) {
+      problem = batchtools::makeJob(id = exps[1, ])$problem
+      pname = p
+      if (is.numeric(problem$data$task)) {
+        problem$data$task = OpenML::convertOMLDataSetToMlr(getOMLDataSet(problem$data$task))
+        pname = getTaskId(problem$data$task)
+      }
+      
+      rin = makeResampleInstance(problem$data$rdesc, problem$data$task)
```
mllg: This guy is not seeded. You will get different indices than those generated for computation.

R/batchmark.R
```diff
@@ -0,0 +1,182 @@
...
+  
+  for (p in problems) {
+    
+    exps = batchtools::findExperiments(prob.name = p, ids = ids)
+    
+    if (nrow(exps) > 0) {
+      problem = batchtools::makeJob(id = exps[1, ])$problem
+      pname = p
+      if (is.numeric(problem$data$task)) {
+        problem$data$task = OpenML::convertOMLDataSetToMlr(getOMLDataSet(problem$data$task))
```
mllg: We read it again?

R/batchmark.R
ja-thomas: Yes, it should be in the OML cache anyways so we do not want/need to save it for every job.

R/batchmark.R
ja-thomas: Same argument as earlier. Also we cannot keep it in memory since the `reduceBatchmarkResults` may be called in a different/later R session. Again, we get the dataset from the OML cache instead of saving it in the registry.

R/batchmark.R
```diff
@@ -0,0 +1,175 @@
...
+#'   Resampling strategy for each tasks.
+#'   If only one is provided, it will be replicated to match the number of tasks.
+#'   If missing, a 10-fold cross validation is used.
+#' @param reg [\code{\link[batchtools]{Registry}}]\cr
+#'   Registry, created by \code{\link[batchtools]{makeExperimentRegistry}}. If not explicitly passed,
+#'   uses the last created registry.
+#' @return [\code{data.table}]. Generated job ids are stored in the column \dQuote{job.id}.
+#' @export
+#' @family benchmark
+batchmark = function(learners, tasks, data.ids, resamplings, measures, models = TRUE, reg = batchtools::getDefaultRegistry()) {
```
giuseppec: I would remove the `data.ids` arg completely:
1) Users can simply get OpenML datasets and pass them into `tasks` theirselves. I would argue that this is something the user should do, like in https://github.com/mlr-org/mlr/pull/1570 where the user is forced to transform the factors to numerics.
2) This will simplify the `batchmark` code and we could remove some "strange" if-else statements below.

R/batchmark.R
```diff
@@ -0,0 +1,175 @@
...
+
+  requirePackages("batchtools", why = "batchmark", default.method = "load")
+
+  learners = ensureBenchmarkLearners(learners)
+  learner.ids = vcapply(learners, getLearnerId)
+  names(learners) = learner.ids
+
+  if (!missing(tasks)) {
```
giuseppec: replace `missing` with `is.null`?  See also the other cases below.

R/batchmark.R
```diff
@@ -0,0 +1,175 @@
...
+    OpenML::populateOMLCache(data.ids = data.ids)
+    did.tasks = as.list(data.ids)
+    names(did.tasks) = data.ids
+    tasks = if (missing(tasks))
+      did.tasks
+    else
+      c(tasks, did.tasks)
+    task.ids = names(tasks)
+  }
+  if (missing(tasks))
```
giuseppec: use checkmate here?

R/batchmark.R
```diff
@@ -0,0 +1,175 @@
...
+  pdes = Map(function(id, task, rdesc, seed) {
+    batchtools::addProblem(id, data = list(rdesc = rdesc, task = task, measures = measures, learners = learners), fun = resample.fun, seed = seed, reg = reg)
+    data.table(i = seq_len(rdesc$iters))
+  }, id = task.ids, task = tasks, rdesc = resamplings, seed = reg$seed + seq_along(tasks))
+
+  # generate algos
+  ades = Map(function(id, learner) {
+    apply.fun = getAlgoFun(learner, measures, models)
+    batchtools::addAlgorithm(id, apply.fun, reg = reg)
+    data.table()
```
giuseppec: Do we need this line?

R/batchmark.R
```diff
@@ -0,0 +1,175 @@
...
+#' @return [\code{\link{BenchmarkResult}}].
+#' @export
+#' @family benchmark
+reduceBatchmarkResults = function(ids = NULL, keep.pred = TRUE, show.info = getMlrOption("show.info"), reg = batchtools::getDefaultRegistry()) {
+
+  #registry and ids are asserted later
+  requirePackages("batchtools", why = "batchmark", default.method = "load")
+  assertFlag(keep.pred)
+
+  result = list()
```
giuseppec: preallocate with `vector("list", length(problems))` after line 137?

R/resample.R
```diff
@@ -110,28 +110,40 @@ resample = function(learner, task, resampling, measures, weights = NULL, models
   time2 = Sys.time()
   runtime = as.numeric(difftime(time2, time1, units = "secs"))
   addClasses(
-    mergeResampleResult(learner, task, iter.results, measures, rin, models, extract, keep.pred, show.info, runtime),
+    mergeResampleResult(learner$id, task, iter.results, measures, rin, models, extract, keep.pred, show.info, runtime),
```
giuseppec: This seems like a base change. @berndbischl needs to look into this and see if this could cause any problems or backwards incompatibilities.

R/batchmark.R
mllg: 100% agree. This is just bloat. Keep it simple.

R/batchmark.R
mllg: Without data.ids, this is pretty straightforward. You need to pass tasks, thus you can use checkmate here and assert a list of tasks (or a single task). Probably there is already a helper in mlr to do this.

R/batchmark.R
mllg: Yep, `addExperiments` expects a named list of `data.frame`/`data.table`. This is an easy way to construct it here.

R/batchmark.R
mllg: Obsolete if we strip out `data.ids`. :)

R/batchmark.R
berndbischl: there is quite a lot of use for dataset ids. argument: we had to do that in MULTIPLE projects now. and users cannot so easily set up OML tasks. thats actually quite hard. if this does not fuck up the code totally, i would argue to keep that feature

R/batchmark.R
berndbischl: dont use missing, it has nothing do do with data set ids, that one can avoid missing and use NULL instead

R/batchmark.R
mllg: 👍 

Should be
```r
res = replicate(length(problems), namedList(algorithms), simplify = FALSE)
names(res) = problems
```

R/batchmark.R
ja-thomas: `tasks = lapply(1:4, function(x) convertOMLDataSetToMlr(getOMLDataSet(x)))`

ja-thomas: easy enough? @berndbischl

berndbischl: yes, remove bloat then and doc in help page how you can do that

mllg:
Did some minor cleanups and doc fixes. Looks good and can be merged as soon as the one remaining test which currently fails is repaired. @ja-thomas takes over.

👍

mllg:
Error seems to be on my system only. Should be save to merge.

PR #1223, opened 2016-09-08
set default of shw.missing.values to TRUE

### LINKED ISSUE #1212: getHyperParsString(... , show.missing.values) should have a default
this broke mlrMBO (but is fixed there)

ja-thomas:
This fixes #1212 

Same default as in ParamHelpers for `paramValueToString`

EDIT: After short discussion with @berndbischl we set the value to TRUE, but this shouldn't change anything.

### DIFF
--- a/R/getHyperPars.R
+++ b/R/getHyperPars.R
@@ -27,7 +27,7 @@ getHyperPars.Learner = function(learner, for.fun = c("train", "predict", "both")
   pv[ns]
 }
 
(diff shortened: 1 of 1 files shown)


larskotthoff:
You're saying that it should be FALSE in the title, but set it to TRUE in the code -- which one should it be?

ja-thomas:
see updated first comment

larskotthoff:
Thanks, merging.

Issue #1191, opened 2016-08-24
Error when default argument is passed to par.vals in learner xgboost


philippstats:
The default of argument  `missing` in xgboost is set to `0` in par.set and  `NA_real_`  in par.vals.

When I pass the default par.vals to par.vals:
`
makeLearner("classif.xgboost", par.vals = list(missing = NA_real_))
`
This error occurs:
`
Error in setHyperPars2.Learner(learner, insert(par.vals, args)) : 
  NA is not feasible for parameter 'missing'! 
`
It seems like `makeNumericLearnerParam` does not accept `NA_real_` as numeric value. 

Also not working:
`makeLearner("classif.xgboost", par.vals = list(missing = NA))`

larskotthoff:
`missing` needs to be a number and not NA. It's set to NA "as this is how mlr expects missing values to be encoded" (see the note).

Does that answer your question?

berndbischl:
this can now be solved, very fortunately !, with @jakob-r 's new special.vals option in PH.

thx for detecting this.

berndbischl:
> missing needs to be a number and not NA. It's set to NA "as this is how mlr expects missing values to be encoded" (see the note).
> 
> Does that answer your question?

it does not really answer the question. we define a default value. then this is not settable, pro-actively, by mlr. ergo:  this is stupid, and @philippstats has detected a real problem.

the problem is, that sometimes some param allow certain values, that dont "fit" into the formal param type of PH. very often, but not always, these are also the default values. examples:
a num param, but stuff like NA, NULL, "auto", and so on.
PH now has an options special.vals, for param definitions, that allows to list these extra values, and defines them as feasible as well. we should have done this years ago. thx to @jakob-r we have it now.

larskotthoff:
Why would you ever want to set these special values?

berndbischl:
> Why would you ever want to set these special values?

i dont get what you mean. because they are valid setting of a parameter? and you want to control that param from mlr? and not have mlr forbid you to set these values?
i tried to explain it above with some examples?

larskotthoff:
At least in this example it's not a valid value. if you set `NA` in a case where `missing` is used, it would cause an error from `xgboost`, wouldn't it? And if you set it when the parameter isn't used, there's no need to set it to that specific value. The default is just something internal as far as I can see.

Do you have an example where it would make sense to set a parameter to a value that's not allowed by its definition?

berndbischl:
here is another example: 
[...]

PR #1225, opened 2016-09-08
xgboost: better handling of arg 'missing'

berndbischl:
should adress issue #1191


berndbischl:
i have also removed outdated todo-files and fixed some indendation

PR #1226, opened 2016-09-08
Deactivate test small values

ja-thomas:
We should reactivate the test as soon as Param Helpers is updated to 1.9

but otherwise we can not release mlr 2.10

### DIFF
--- a/tests/testthat/test_base_tuning.R
+++ b/tests/testthat/test_base_tuning.R
@@ -153,6 +153,7 @@ test_that("Learner defined with expression in param requires, see #369 and PH #5
 })
 
 
+# %FIXME% we should reactivate the test when ParamHelpers 1.9 is on CRAN
 # test_that("tuning does not break with small discrete values, see bug in #1115", {
 #   ctrl  = makeTuneControlGrid()
 #   ps = makeParamSet(
--- a/tests/testthat/test_base_tuning.R
+++ b/tests/testthat/test_base_tuning.R
@@ -153,14 +153,14 @@ test_that("Learner defined with expression in param requires, see #369 and PH #5
 })
 
 
-test_that("tuning does not break with small discrete values, see bug in #1115", {
-  ctrl  = makeTuneControlGrid()
-  ps = makeParamSet(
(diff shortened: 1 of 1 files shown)


larskotthoff:
Can you give some more context please? What breaks and why? Why are you commenting the tests? We use the github version of PH for Travis and Jenkins.

ja-thomas:
Sure

was written a bit confusing, but @berndbischl was in the same room so he knew what this is about. Essentially the mentioned test does **only** work with PH 1.9, not with the current CRAN release. So our tests are working on travis and jenkins, but since we planned to release mlr 2.10 today or tomorrow we need to deactivate this test, otherwise rcheck will fail

larskotthoff:
Ugh, this is terrible. I've created a new Jenkins job that uses the CRAN version of PH now. But this test won't be run by CRAN, will it? Do we actually need to disable this?

larskotthoff:
I'm getting more errors with the CRAN version of PH:

```
1. Error: downsample wrapper works with xgboost, we had issue #492 (@test_base_downsample.R#36) 
2. Error: getHyperPars (@test_base_getHyperPars.R#26) 
3. Error: tuning does not break with small discrete values, see bug in #1115 (@test_base_tuning.R#162) 
```

ja-thomas:
arent all base tests run by CRAN? at least R CMD CHECK failed because of this.

3. is the error of the test I disable in this PR

The other two errors didn't occured when I tested with the CRAN PH version

larskotthoff:
I can reproduce all errors locally when running `R CMD check --as-cran`.

ja-thomas:
So the Plan is now to update PH on CRAN first, so this PR should be obselete, but I'll keep this open until the update is actually done

larskotthoff:
Ok, sounds good. Can we please please please avoid such situations in the future or at least make PH releases *immediately* after such a change?

berndbischl:
> Ok, sounds good. Can we please please please avoid such situations in the future or at least make PH releases immediately after such a change?

we are really trying our best here....
actually we invested lots of time to figure out the best option....

IMHO this is now: release PH first.

ja-thomas:
PH is now 1.9 on CRAN

PR #1228, opened 2016-09-09
Fix some warnings in base tests

ja-thomas:
So there seem to occur some warnings in the base tests when running them locally

```
Warnings -----------------------------------------------------------------------
1. getParamSet (@test_base_getParamSet.R#12) - 'setId' is deprecated.
Use 'setLearnerId' instead.
See help("Deprecated")

2. makeLearners (@test_base_makeLearners.R#9) - 'function (learner, id) 
{
    .Deprecated("setLearnerId")
    assertString(id)
    learner$id = id
    return(learner)
}' is deprecated.
Use 'setLearnerId' instead.
See help("Deprecated")

3. makeLearners (@test_base_makeLearners.R#9) - 'function (learner, id) 
{
    .Deprecated("setLearnerId")
    assertString(id)
    learner$id = id
    return(learner)
}' is deprecated.
Use 'setLearnerId' instead.
See help("Deprecated")

4. check measure calculations (@test_base_measures.R#258) - MAPE is undefined if any truth value is equal to 0.

5. check measure calculations (@test_base_measures.R#259) - MAPE is undefined if any truth value is equal to 0.

6. measures quickcheck (@test_base_measures.R#659) - NAs produced by integer overflow
```

1 - 3 are just deprecated calls to `getId`

4 - 5 added supress warning test to check the actual value (`NA`) as well as an test for the warning message.

6 I have no idea why this is failing, seems to be consistently failing when running rcheck  but works when running the chunk of code by hand.

### DIFF
--- a/tests/testthat/test_base_getParamSet.R
+++ b/tests/testthat/test_base_getParamSet.R
@@ -9,7 +9,7 @@ test_that("getParamSet", {
   ps = getParamSet(lrn)
   expect_true(all(c("method", "fw.method") %in% names(ps$pars)))
 
-  lrn = makeModelMultiplexer(list(setId(lrn, "x")))
+  lrn = makeModelMultiplexer(list(setLearnerId(lrn, "x")))
   ps = getParamSet(lrn)
   expect_true(all(c("x.method", "x.fw.method", "selected.learner") %in% names(ps$pars)))
 
--- a/tests/testthat/test_base_makeLearners.R
+++ b/tests/testthat/test_base_makeLearners.R
@@ -6,7 +6,7 @@ test_that("makeLearners", {
   lrns1 = setNames(lapply(cls1, makeLearner), cls1)
   ids = c("a", "b")
   lrns1 = setNames(lapply(cls1, makeLearner), cls1)
-  lrns2 = setNames(mapply(setId, lrns1, ids, SIMPLIFY = FALSE), ids)
+  lrns2 = setNames(mapply(setLearnerId, lrns1, ids, SIMPLIFY = FALSE), ids)
   lrns3 = lapply(lrns1, setPredictType, predict.type = "prob")
 
   res = makeLearners(cls1)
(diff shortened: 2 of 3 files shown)


catastrophic-failure:
About 4-5 please look at #1211 so we keep it unified.

larskotthoff:
Why don't we see this in Travis or Jenkins?

berndbischl:
> Why don't we see this in Travis or Jenkins?

i am pretty sure we DO see them. its just that we dont look at it.
this was the reason why i always wanted to escalate warning to errors. 
but this then sometimes produced false positives.....

i still think that such an escalation would be ideal.

larskotthoff:
There's nothing in the output (and I think @ja-thomas is saying that you see this in the output, even if the build doesn't fail).

berndbischl:
> There's nothing in the output (and I think @ja-thomas is saying that you see this in the output, even if the build doesn't fail).

its at least visible when you run the tests locally. 
it think the difference is this: 
locally we run with rtest, on travis it is run inside of R CMD check. there you dont see the warnings.

larskotthoff:
Ok, but what does that do differently, i.e. how can we make those warnings show up?

berndbischl:
> Ok, but what does that do differently,

well, rtest just call testthat. r cmd check calls r cmd check.

> Ok, but what does that do differently, i.e. how can we make those warnings show up?

i dont think you can teach r cmd check to output more. 
this is why i said i would like to escalate warning.

catastrophic-failure:
Using `suppressWarnings` has the undesirable side effect that if a warning happens due to other reasons than the one you were expecting you won't see it.

berndbischl:
> Using suppressWarnings has the undesirable side effect that if a warning happens due to other reasons than the one you were expecting you won't see it.

thats not what we are discussing here. there are quite a few places in our tests where we produce (undesired) warnings. which are not suppressed. they are thrown because our tests or code is imperfect.
rtest shows these. r cmd   doesnt

Issue #1232, opened 2016-09-09
Partial dependence target scale not right


MasonGallo:
When I run the examples from `generatePartialDependenceData` docs, the scale of the target doesn't look right. Any ideas?

Here is the first example in the documentation:

```
lrn = makeLearner("regr.svm")
fit = train(lrn, bh.task)
pd = generatePartialDependenceData(fit, bh.task, "lstat")
plotPartialDependence(pd, data = getTaskData(bh.task))
```
![image](https://cloud.githubusercontent.com/assets/4913951/18404532/0944e8de-76b9-11e6-8cb9-c269cdedbe1f.png)

The graph above doesn't look right. It appears the target variable is off by a few decimal places?

```
pd$data
         medv     lstat
1  0.05248820  1.730000
2  0.04980801  5.756667
3  0.04666540  9.783333
4  0.04337672 13.810000
5  0.04027198 17.836667
6  0.03764451 21.863333
7  0.03571056 25.890000
8  0.03458451 29.916667
9  0.03427200 33.943333
10 0.03468078 37.970000
```

zmjones:
the issue is with the weight function that i added recently. the default aggregation function is still the mean which means that the scale is off by a factor of 100. setting the aggregation function to sum will fix the issue in this case.

generally, i am not sure of the best way to solve this.

i could change the default aggregation function to sum and add a note in the documentation. however, if you wanted to use the variance this wouldn't work, since a weighted variance is 1 / n * sum(w * (x - mu)^2) and in the current implementation i use fun(x) * w / sum(w) which wouldn't work.

i am thinking maybe i should remove the weights function totally, and instead add the ability to pass additional arguments to the aggregation function, where the user can compute weights if they like. 

thoughts?

larskotthoff:
Hmm, it sounds like removing the weights and allowing the user to do this in the aggregation function is the best solution. It looks like at the moment it's easy to get misleading results if the two don't match and impossible to check if they do in the function.

zmjones:
yea ok. will do that now.

zmjones:
so the problem with this is that to compute weights you need the training data and you need the prediction grid which is generated internally. in the function were the weights would be computed this object is called newdata. i'd like to be able to pass newdata (along with the training data) into the aggregation function so that users can compute weights but i would like the default aggregation function to still be mean and to not bother users with this if they aren't computing weights.

[...]

PR #1233, opened 2016-09-11
Add support for visualizing tasks with 2 or more hyperparameters

MasonGallo:
This PR will add support for visualizing hyperparameter tuning when tuning 2 or more hyperparameters simultaneously.

Example usage tuning C, sigma, degree of SVM on `pid.task`:

```
ps = makeParamSet(
  makeNumericParam("C", lower = -5, upper = 5, trafo = function(x) 2^x),
  makeNumericParam("sigma", lower = -5, upper = 5, trafo = function(x) 2^x),
  makeDiscreteParam("degree", values = 2:5))
ctrl = makeTuneControlRandom(maxit = 100L)
rdesc = makeResampleDesc("Holdout", predict = "both")
learn = makeLearner("classif.ksvm", par.vals = list(kernel = "besseldot"))
res = tuneParams(learn, task = pid.task, control = ctrl, 
  measures = list(acc,setAggregation(acc, train.mean)), resampling = rdesc, 
  par.set = ps, show.info = F)
data = generateHyperParsEffectData(res, partial.dep = TRUE)
```

We can isolate just `C`:

```
plotHyperParsEffect(data, x = "C", y = "acc.test.mean", plot.type = "line", 
  partial.dep.learn = "regr.randomForest")
```

image

We can also look at bivariate relationships, such as `C` and `sigma`:

```
plotHyperParsEffect(data, x = "C", y = "sigma", z = "acc.test.mean", 
  plot.type = "heat map", partial.dep.learn = "regr.randomForest")
```

image

Note the scales in the plots are off due to #1232

### DIFF
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -22,12 +22,13 @@
 #'  path.
 #'  Default is \code{FALSE}.
 #' @param partial.dep [\code{logical(1)}]\cr
-#'  Should partial dependence be generated based on converting to reg task? This
+#'  Should partial dependence be requested based on converting to reg task? This
 #'  sets a flag so that we know to use partial dependence downstream. This
 #'  should most likely be set to \code{TRUE} if 2 or more hyperparameters were
-#'  tuned simultaneously. Setting to \code{TRUE} will cause
-#'  \code{\link{plotHyperParsEffect}} to automatically plot partial dependence
-#'  when called downstream.
+#'  tuned simultaneously. Partial dependence should always be requested when
+#'  more than 2 hyperparameters were tuned simultaneously. Setting to
+#'  \code{TRUE} will cause \code{\link{plotHyperParsEffect}} to automatically
+#'  plot partial dependence when called downstream.
 #'  Default is \code{FALSE}.
 #'
 #' @return [\code{HyperParsEffectData}]
@@ -60,17 +61,21 @@
 #' @export
 #' @importFrom utils type.convert
 generateHyperParsEffectData = function(tune.result, include.diagnostics = FALSE,
-                                       trafo = FALSE, partial.dep = FALSE) {
+  trafo = FALSE, partial.dep = FALSE) {
+
   assert(
     checkClass(tune.result, "ResampleResult"),
     checkClass(tune.result, classes = "TuneResult")
   )
   assertFlag(include.diagnostics)
+  assertFlag(partial.dep)
 
   # in case we have nested CV
   if (getClass1(tune.result) == "ResampleResult"){
     d = getNestedTuneResultsOptPathDf(tune.result, trafo = trafo)
     num_hypers = length(tune.result$extract[[1]]$x)
+    if ((num_hypers > 2) && !partial.dep)
+      stopf("Partial dependence must be requested with partial.dep when tuning more than 2 hyperparameters")
     for (hyp in 1:num_hypers) {
       if (!is.numeric(d[, hyp]))
         d[, hyp] = type.convert(as.character(d[, hyp]))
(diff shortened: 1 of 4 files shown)


MasonGallo:
This is ready for review. Travis failure seems unrelated - sourcetools package is failing to build

larskotthoff:
Thanks, looks good! A few points:

- Could you expose the learner for the partial dependence model (currently `regr.randomForest`) as a parameter please?
- Looking at it now I'm not sure about the `x`, `y` and `z` argument names, given that `y` can be a response variable and a parameter depending on the plot. Maybe `params = c("C", "sigma"), value = "acc.test.mean")` or something like that?
- When I run your second example without partial dependence I get an empty plot. I would either raise an error or do the necessary interpolation there.
- Could you please add plot testing that parses the SVG and counts the number of elements, similar to what we have for some of the other plots? I realise that this may not be possible for all types of plots. Let me know if you have any questions about this.

I'm wondering whether `partial.dep` should default to `TRUE`, as it probably gives a more accurate picture of the dependence in general. Thoughts? @berndbischl @zmjones

MasonGallo:
Thanks!

> Could you expose the learner for the partial dependence model (currently regr.randomForest) as a parameter please?

yes

> Looking at it now I'm not sure about the x, y and z argument names, given that y can be a response variable and a parameter depending on the plot. Maybe params = c("C", "sigma"), value = "acc.test.mean") or something like that?

Hmm, I think I understand what you mean and I'm not completely opposed, but this would tradeoff "freedom" for the user like we discussed. Ex: if user gives 1 param and 1 value, we "force" the param to be on the x axis. I think I would honestly prefer to "force" the user - if they want to play with the axes they can just grab the data.

> When I run your second example without partial dependence I get an empty plot. I would either raise an error or do the necessary interpolation there.

I actually think this shouldn't work since it would be misleading without partial dependence. Perhaps we should force partial dependence if k > 2?

> Could you please add plot testing that parses the SVG

I was planning on adding this for all of the usecases in a separate "clean up" PR if that's okay? I wanted to just focus on getting pd done

> I'm wondering whether partial.dep should default to TRUE

Maybe it should default to `TRUE` if k>=2?

larskotthoff:
Re argument names: I see your point. I think in general I prefer a more "semantic" way of specifying these things, but I guess in this case it makes more sense to have the current parameters as the user can also specify the type of plot.

I wouldn't force partial dependence. My main point about the plot is that nothing is being shown without an indication of what's wrong. From what you're saying it sounds like we should throw an error in this case. (General note: We should figure out a way to clean this up in general -- there's lots of this kind of checking in the plotting functions and it's getting ugly.)

Separate PR for testing is ok.

I would keep a static default for `partial.dep`, otherwise it's easy to forget and get unexpected results.

larskotthoff:
Oh and if you rebase on master the builds should work.

MasonGallo:
Updates made. If the user doesn't request partial dependence when tuning >2 hyper parameters simultaneously, an error is thrown. User must also specify a learner for partial dependence if necessary.

> General note: We should figure out a way to clean this up

Yes, I'm planning on doing a separate PR for cleaning up tests, docs, args, better default style for plots for all cases. This will be rather tedious / time-consuming, so I want to do that separately.

Looks like Travis is still failing when attempting to install sourcetools?

larskotthoff:
Thanks. I would keep `regr.randomForest` as a default for the learner. And it looks like Jenkins is failing because of unused arguments.

MasonGallo:
> I would keep regr.randomForest as a default for the learner

This is different behavior than the interpolation arg. The interpolation arg defaults to `NULL` to force the user to specify if that's what they want. I wanted to keep the arg behavior consistent in the API.

> looks like Jenkins is failing because of unused arguments

Yes oops I forgot to fix the argname in tests. That should be fixed now.

larskotthoff:
Ok, I see your point, thanks.

PR #1234, opened 2016-09-11
fix xgboost tests

larskotthoff:
xgboost tests failed after latest change of default for `missing`. Removed the test as it seems unnecessary now.

@berndbischl ?!

### DIFF
--- a/tests/testthat/test_base_getHyperPars.R
+++ b/tests/testthat/test_base_getHyperPars.R
@@ -21,11 +21,4 @@ test_that("getHyperPars", {
 
   lrn = makeMultilabelBinaryRelevanceWrapper("classif.rpart")  
   expect_true(setequal(getHyperPars(lrn), list(xval = 0)))
-  
-  #Missing values should not be ommited and printed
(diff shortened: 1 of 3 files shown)


larskotthoff:
Merging this now so master isn't broken anymore, but @berndbischl please check whether all of this was intentional!

MariaErdmann:
Is this intended?
`makeLearner("classif.xgboost")` throws an error
`Error in makeNumericLearnerParam(id = "missing", default = NULL, tunable = FALSE,  : 
  unused argument (special.vals = list(NA, NA, NULL))`
Therefore, the tests of `'test_learners'all_classif'´ do not run through atm.

berndbischl:
> Is this intended?
> makeLearner("classif.xgboost") throws an error
> Error in makeNumericLearnerParam(id = "missing", default = NULL, tunable = FALSE, : unused argument (special.vals = list(NA, NA, NULL))

you have to update paramhelpers. but the new version for this is even released on CRAN already....

MariaErdmann:
Ohhh, yes. Thank you very much and sorry, I just forgot to do that.

PR #1235, opened 2016-09-11
remove weight.fun in place of expanded fun in generatePartialDependence

zmjones:
- internal wrapper for fun arg to allow passing of internal
   newdata (prediction grid) and data (training data from input arg)
   which allows computation of weights in fun instead of via an extra
   step using another arg, weight.fun (now removed)

two issues related to this. first is how to document it. i added to the docs here but we don't have a dependency free way to show an example of it (a situation where the aggregation function would use the prediction grid and/or the training data). 

the second issue is writing a test that would have caught the bug. how to do this doesn't seem obvious. you could check to see if the predictions are within the range of the training data target but it would be easy to make this break without having any bugs in the code.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -57,14 +57,12 @@
 #'   a numeric matrix with the number of columns equal to the number of class levels of the target.
 #'   For classification with \code{predict.type = "response"} (the default) the function must accept
 #'   a character vector and output a numeric vector with length equal to the number of classes in the
-#'   target feature.
+#'   target feature. Two variables, \code{data} and \code{newdata} are made available to \code{fun} internally via a
+#'   wrapper. `data` is the training data from `input` and `newdata` contains a single point from the
+#'   prediction grid for \code{features} along with the training data for features not in \code{features}.
+#'   This allows the computation of weights based on comparisons of the prediction grid to the training data.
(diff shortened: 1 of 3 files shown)


larskotthoff:
Complex examples are fine, you can wrap them in `\dontrun{}` if necessary. For the test, I would take exactly the case where Mason noticed it.

zmjones:
i think i have an ok case for that now. any tree based model will only do constant extrapolation, so if a partial dependence prediction is ever outside the range of the training data it is a bug. will add that.

i can add examples in dontrun that depend on packages that aren't in suggests? or i should add it to suggests?

zmjones:
looks like an unrelated fail for travis on devel R.

larskotthoff:
Hmm, could you add an example that doesn't depend on any other packages?

zmjones:
did that. i think it is quick enough to run. 

the example is a draw from a correlated multivariate normal with a continuous target feature and weights computed via the joint distribution. i added a note about what you would do in practice. 

the only alternative i can think of is to use oneclass classification (we have this in mlr right?) to compute P(outlier) and use that as a weight.

larskotthoff:
Looks good. I'll wait for the builds to finish and merge.

zmjones:
looks like the same error

Issue #1144, opened 2016-08-14
Better documentation for measures


schiffner:
We need to improve the measures doc page.
Similar as for learners (see #1072) some notes get longer and longer.

MasonGallo:
What if we linked to the "notes" within the tutorial site for each measure / learner? That way, we would create a page for each measure to list the info.

berndbischl:
> What if we linked to the "notes" within the tutorial site for each measure / learner? That way, we would create a page for each measure to list the info.

how to do this should not be discussed here, but here: #1072 

this issue is just for adding details about the measures

berndbischl:
@florianfendt 
some measures already contain nice formulas and so on in the notes. 
can you please make that equally could for all measures? 
so eg assume i have no idea what "mmce" means. can you fix that with 1-2 simple sentences in the note?

do you understand what i mean?

PR #1236, opened 2016-09-12
#1144 better doc for measures

florianfendt:


berndbischl:
@florianfendt 
if i have no further comments can this merged from your side?

berndbischl:
medse and others with median: note does not do much, use simple formula, you can use "median" in the formula

acc, mmce: try to produce formula, like mean(response != truth)

ppv and other: why dont you use mlr's abbreviations in the formula to make it shorter. like not "number of false positives" but fp

florianfendt:
@berndbischl
Adapted PR. Should be ok to merge from my side.
Don't really get why Travis fails, though. Error seems to be related to xgboost learner.

larskotthoff:
@florianfendt Rebase on master and Travis should work again.

florianfendt:
@berndbischl 
ping you here. Should be ready to merge

PR #1238, opened 2016-09-13
irace v2

berndbischl:
irace version 2 is soon going on cran.
they changed their interface, i adapted to the new one here.

things left to do:

- we need to make a note in NEWS that there are also some algorithmic changes in irace that can affect performance. irace now uses its new "elitist" version. its supposed to be better, but of course there is no guarantee

- irace can now also use the runtime from the target eval. we have to give it to irace in addition to "cost". currently i am returning NA. this can now be used to tune with irace with time bugdet set by the user.


berndbischl:
the branch should basically be  done, but irace is not on CRAN. we can only merge when irace 2.0 is released.

we also need to update the required version nr in DESC then

PR #1243, opened 2016-09-14
fixes for issue #63 in the tutorial

zmjones:
- incorrect jacobian function in doPartialDerivativeIteratoin
 - improper fun/fun.wrapper (for weights use)
 - test added based on tutorial fail
 - simplified code a bit

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -499,52 +499,45 @@ print.FunctionalANOVAData = function(x, ...) {
 }
 
 doPartialDerivativeIteration = function(x, obj, data, features, fun, td, individual, ...) {
-  if (!individual) {
-    fun.wrapper = function(x, newdata, data, ...) {
-      args = formals(fun)
-      if (all(c("newdata", "data") %in% names(args))) {
-        fun(x, newdata = newdata, data = data, ...)
-      } else if ("newdata" %in% names(args)) {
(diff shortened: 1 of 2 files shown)


larskotthoff:
Sorry, looks like some other tests are failing now.

larskotthoff:
Thanks. I'm getting a bit worried about this -- these seemingly small bugs always incur quite significant code changes. Is there any way this could be tested better or split up?

zmjones:
yea fair point. really i just introduced a bunch of bugs when i hastily added all this weighting stuff. i don't think anything has gone wrong other than that. that and given the amount of arguments has meant that the test coverage wasn't/isn't great.

i think also the size of the changes is misleading. i tend to do a little refactoring where i can when debugging. for example here i noticed i'd defined two sets of two functions which were basically the same, so i deleted one and moved the others (so from 4 to 2). the actual bug was one line and so was the test.

berndbischl:
> Thanks. I'm getting a bit worried about this -- these seemingly small bugs always incur quite significant code changes. Is there any way this could be tested better or split up?

@zmjones 
I was getting the same feeling (not criticizing you, just saying)
also watch out for not introducing "feature creep".

abizt the refactoring when fixing: i can understand that, i do the same.

zmjones:
fair point. i'll think some about whether i can modularize some more.

PR #1245, opened 2016-09-15
added the colsample_bylevel parameter in the xgboost learners

PhilippPro:
See http://xgboost.readthedocs.io/en/latest/parameter.html (the parameter is not written in the R help)


larskotthoff:
Does this depend on a particular version of xgboost?

berndbischl:
> Does this depend on a particular version of xgboost?

i dont think this matters, i think we have to rely on the newest version of xgboost in DESC anyhow

Issue #1240, opened 2016-09-13
xgboost mini issue: some params should not be defined as "tunable"


berndbischl:
print.every.n
verbose

and so on. needs to be checked.

classif.xgboost and regr.xgbosst

mllg:
Also, check the default for `maximize` and fix it or drop a comment why `TRUE` or `FALSE` is correct for the respective learner.

larskotthoff:
This should be fixed; please reopen if there's anything else.

PR #1247, opened 2016-09-16
Fix some params of xgboost (relates to ##1240)

MariaErdmann:
PR for #1240.
Changed some ranges according to [https://github.com/dmlc/xgboost/blob/master/doc/parameter.md] (which is a link to an explanation on the xgboost parameter. it comes from the xgboost manual).
Changed tunable to FALSE for `print.every.n`, `verbose`
Some questions remain:

- `nthread` is the number of parallel threads. default shall be the maximum number of threads available (in mlr it is 16). As I do not exactly know what this is, can somebody tells me if this is correct? Is this param tunable?

- `verbose`: `1` is the default which prinkts messages about performance. Shall mlr's default be `0` (silent)?

- `early.stop.round`  and `maximize`default is NULL. How handle this?


larskotthoff:
`nthread` shouldn't be tunable. I would set `verbose` to 0 by default. I don't understand your last question -- if the default is NULL, let's set it to NULL.

berndbischl:
i have rebased this on master and fixed the conflicts.
can you pls finish this ASAP?



> early.stop.round and maximizedefault is NULL. How handle this?

use special.vals for this. i think i specifically explained this, that we have this in PH now? it allows to set "special values" which are still feasible.

> nthread is the number of parallel threads. default shall be the maximum number of threads available (in mlr it is 16).

why is it 16 in mlr? i dont get this.

berndbischl:
i made a mistake when i resolved the conflict. fixed now.

i also set verbose = 0, as an mlr default. and set verbose = 0 in the unit tests

MariaErdmann:
> use special.vals for this.´

Done. Sorry, did not realise that.  Couldn't attend the last mlr-hiwi-meetings since the workshop and looking forward to meet up again!

> why is it 16 in mlr? i dont get this.

I didn't understand it either. I deleted the default now, since the documentation says: 
"default to maximum number of threads available if not set"

> i also set verbose = 0, as an mlr default

I added verbose = 0L to par.vals

berndbischl:
can somebody pls go through the params again and set a few to tuneable = FALSE?
then we are done here and can merge

MariaErdmann:
I reviewed my code again and set tunable to FALSE for `silent` (print running messages or stay silent), `eval-metric` (evaluation metrics) and `feval` (customized evaluation function).
I am not sure about setting tunable to FALSE for :

- `objective` (specifies the learning task and the corresponding learning objective, users can also pass a function to it)

- `base_score` (initial prediction score of all instances, global bias)

-  `maximize` (maximize = TRUE means the larger the evaulation score the better)

larskotthoff:
None of these should be tunable.

berndbischl:
> None of these should be tunable.

for some it is slightly debateable, but i agree with lars

larskotthoff:
If we want to make them tunable, we would need to add complex constraints between them, e.g. it doesn't make sense to change the maximization without changing the objective function.

larskotthoff:
But this is machine-readable information. As a user, a reasonable approach would be to tune over all parameters that have `tunable = TRUE` (without looking at them). This will not work as intended in this case.

larskotthoff:
I will merge this now, but we should discuss at some point what exactly "tunable" means.

PR #1249, opened 2016-09-17
ksvm mini tunable fix for hyper par settings

berndbischl:
pls merge ASAP

NB: a param like "cache" can influence runtime related meaures and is set it therefore to tunable

### DIFF
--- a/R/RLearner_classif_ksvm.R
+++ b/R/RLearner_classif_ksvm.R
@@ -28,8 +28,8 @@ makeRLearner.classif.ksvm = function() {
       makeNumericLearnerParam(id = "tol", default = 0.001, lower = 0),
       makeLogicalLearnerParam(id = "shrinking", default = TRUE),
       makeNumericVectorLearnerParam(id = "class.weights", len = NA_integer_, lower = 0),
-      makeLogicalLearnerParam(id = "fit", default = TRUE),
-      makeIntegerLearnerParam(id = "cache", default = 40L, lower = 1L, tunable = FALSE)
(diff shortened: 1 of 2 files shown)


larskotthoff:
Wait, I don't quite follow this. Yes, a cache can affect the runtime, but so can debug options and parallelization. Does that mean we should make those tunable as well?

berndbischl:
> Yes, a cache can affect the runtime, but so can debug options and parallelization. Does that mean we should make those tunable as well?

no, because they dont affect runtime in non-trivial ways. 
debug option: makes learner simply (potentially) slower
parallelization: makes it always faster, but in an "unfair" way.

Lars, its pretty simple. Just the "tunable = TRUE" setting simply says whether you could imagine ANY senario where you would tune for that param. debug and parallel options are a clear "no" .

the "cache" here i feel unsure. you can convince me to set it it back to tunable = FALSE. 
but it *does* affect runtime in a nontrivial way? unsure.

larskotthoff:
I would say that parallelization is more "tunable" than a cache. Using a cache is almost always a good idea, but parallelization depends on the machine -- can I use as many threads as processors, or does that make things worse (memory contention etc). If I want to tune something for runtime performance, I would *definitely* include parallelization.

berndbischl:
> Using a cache is almost always a good idea

its not a boolean, but the size of the cache.

berndbischl:
and its not about "more tunable". if it is, in any valid scenario, tuneable, we should set it to true.

larskotthoff:
Ok, let's try to come up with some guidelines at the next devel meeting. Merging this now.

PR #1250, opened 2016-09-18
New measures: Cohen's Kappa and Mean Quadratic Weighted Kappa

florianfendt:
Not really happy with the naming yet. Any better ideas than ```mqwk```?
Maybe only ```wk``` ?

Update for NEWS:
```
## measures - new
* kappa
* wkappa
```


larskotthoff:
I think I like `wk` better as the name. Your note and the formula you're actually using don't agree, could you please have a look at that?

florianfendt:
@larskotthoff 
Done. Thanks for the quick review!

larskotthoff:
Thanks! I'm getting nitpicky now, but it feels like if we have the weighted kappa, we should also have the normal kappa. Could you add that please?

florianfendt:
Sure, you're right. I'll do it

florianfendt:
Done.
After talking to Bernd we ended up renaming the measures to:
```kappa``` and ```wkappa```

larskotthoff:
Thanks, merging.

Issue #1251, opened 2016-09-19
for all learners: make sure that we check the manually used learner against the mlr implementatin with EMPTY param sets


berndbischl:
so we check default behavior.

for this one has to go through the unit tests for the base learners and check that we have 
list()
in there for the par.vals when we check

berndbischl:
see this bug here #1244

jakob-r:
Can you explain in short (but not too short) what is going on?
Some learners need some arguments to be called, right!? So the `list()` won't work.

larskotthoff:
But mlr has defaults that are taken in this case. So in principle giving no arguments should always work.

jakob-r:
Ah my bad. I thought we want to initialize the original learner always with empty par set.
For the future it might be interesting to do that for mlr learners that do not have specific defaults.

larskotthoff:
This was fixed in #1253.

Issue #1034, opened 2016-07-22
Add unit tests to check for wrong default values


giuseppec:
In mlr we maybe have wrong default values, for example in the `?mboost::boost_control` documentation it says that `center = TRUE`  is the default. In our mlr param set, the default is `FALSE`, see:

``` r
getParamSet(lrn)$pars[["center"]]
#     Type len   Def Constr Req Tunable Trafo
#1 logical   - FALSE      -   -    TRUE     -
# Used: train.
```

This is the issue that is caused:

``` r
lrn = makeLearner("regr.glmboost")
set.seed(1);(bench <- benchmark(lrn, bh.task))
#                task.id    learner.id mse.test.mean
#1 BostonHousing-example regr.glmboost      25.95779

lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))
set.seed(1);(bench2 <- benchmark(lrn2, bh.task))
#                 task.id    learner.id mse.test.mean
#1 BostonHousing-example regr.glmboost      42.50329
```

Maybe we should add a test for all learners such as:

```
testthat::expect_equal(as.data.frame(bench), as.data.frame(bench2))
```

Or maybe something like this is sufficient:

```
testThatLearnerHasCorrectDefaults = function(lrn, task) {
  lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))

  rin = makeResampleInstance("Holdout", task = task)
  set.seed(1);m1 = train(lrn, task, subset = rin$train.inds[[1]])
  lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))
  set.seed(1);m2 = train(lrn2, task, subset = rin$train.inds[[1]])

  p1 = predict(m1, task, subset = rin$test.inds[[1]])
  p2 = predict(m2, task, subset = rin$test.inds[[1]])
  expect_equal(performance(p1), performance(p2))
}
```

giuseppec:
In general, making sure that our defaults are always correct is very important for reproducibility in openml because we decided not to upload default hyperparameter values to openml. Applying a learner `lrn = makeLearner("regr.glmboost")` and `lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))` on the same task with the same seed should therefore always return the same result, shouldn't it?

larskotthoff:
Agreed. We probably should also have tests that compare the default mlr learner with the unwrapped learner.

jakob-r:
Related to handling #1032

PR #1253, opened 2016-09-20
test default parameters (#1251)

larskotthoff:
See issue #1251. This will likely break all sorts of things.

There are lots of learners that don't use the parset tests; those will need to be checked and changed if necessary.

### DIFF
--- a/tests/testthat/test_classif_C50.R
+++ b/tests/testthat/test_classif_C50.R
@@ -4,6 +4,7 @@ test_that("classif_C50", {
   
 requirePackages("C50", default.method = "load")
   parset.list = list(
+    list(),
     list(control = C50::C5.0Control(seed = getOption("mlr.debug.seed"))),
     list(trials = 5L, control = C50::C5.0Control(minCases = 10L,
       seed = getOption("mlr.debug.seed"))),
@@ -32,6 +33,7 @@ requirePackages("C50", default.method = "load")
   }
   
   parset.list = list(
+    list(),
     list(seed = getOption("mlr.debug.seed")),
     list(trials = 5L, minCases = 10L, seed = getOption("mlr.debug.seed")),
     list(rules = TRUE, bands = 100L, seed = getOption("mlr.debug.seed")),
--- a/tests/testthat/test_classif_LibLineaRMultiClassSVC.R
+++ b/tests/testthat/test_classif_LibLineaRMultiClassSVC.R
@@ -4,6 +4,7 @@ test_that("classif_LiblineaRMultiClassSVC", {
   requirePackagesOrSkip("LiblineaR", default.method = "load")
 
   parset.list = list(
+    list(),
     list(cost = 5L, epsilon = 0.1),
     list(cost = 5L, epsilon = 0.5),
     list(cost = 2L, epsilon = 0.1),
--- a/tests/testthat/test_classif_LiblineaRL1L2SVC.R
+++ b/tests/testthat/test_classif_LiblineaRL1L2SVC.R
@@ -4,6 +4,7 @@ test_that("classif_LiblineaRL1L2SVC", {
   requirePackagesOrSkip("LiblineaR", default.method = "load")
 
   parset.list = list(
+    list(),
     list(cost = 5L, epsilon = 0.1),
     list(cost = 5L, epsilon = 0.5),
     list(cost = 2L, epsilon = 0.1),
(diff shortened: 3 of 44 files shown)


larskotthoff:
See also #1034.

MariaErdmann:
Some tests needed adaption because the learner had par.vals set. I solved this with using two parset lists (one for the original learner and one for the mlr wrapper).

I set back all non working tests for the mboost learners to the old status (the status were an empty list was not tested yet), since these tests are adapted already in PR #1023, in order to avoid merge conflicts.

The test_classif_boosting with an empty list now takes a very long time compared to all the other learner tests (before it took, about a few seconds and now about 30 seconds on my machine). This is because of param `mfinal`which has a default value of 100 which is quite high. Shall we test with an empty list here, or is this too time consuming?

Finally, there is one learner/test I did not find out why the test fails with an empty list:
surv.cv.CoxBoost
I set the test back to the old status (without empty list), in order to see if the rest is fine.
Will look at it tomorrow. Ideas are welcome :-)

larskotthoff:
Thanks, that sounds like great progress! Tests taking a long time is ok; it's more important that things work with the default parameters.

I would check if the default coxboost parameters in mlr are actually the defaults for coxboost.

MariaErdmann:
> I would check if the default coxboost parameters in mlr are actually the defaults for coxboost.

Thanks. I found the problem.


I changed all the tests throwing an error in the first commit by @larskotthoff.
Now, look into every single test file to check whether it tests the empty parameter set scenario.
With the classification learners I am finished. This is what I did:

- Some tests just checked one scenario were some params were tested with `testSimple`/ `testProb`. 
Now they test both scenarios (one with empty paramset and one with some parameters passed) using `testSimpleParsets`/`testProbParsets`

- There are some files that just test one scenario: the case were no params are passed. I left those untouched.

- There was a bug in `classif.lssvm` (and consequently the test): there is
no parameter `fitted` but `fit`. I changed the learner and adapted the test.

- And there are some learner were it is not possible to use testSimpleParsets() or testProbParsets()
(test_classif_mlp.R, test_classif_neuralnet.R, test_classif_nnet.R, test_classif_nnTrain.R,
test_classif_saeDNN.R)
And here I need some advise, as I have no explanation for the weird behaviour, which has probably to do with the learners themself (all the files also use this capture.output function):
I want to test the empty paramset scenario and one scenario where parameters are passed with a parset.list (like we do i.e. in test_classif_C50).

Adapting the test such that I can use `testSimpleParsets` does  not work with:
````
parset.list = list(
   list(),
   list(param1 = 1, param2 = 1)
)
`````

But if just have one list in the parset.list:
````
parset.list(
   list()
)
````
and then I exchange list() with list(param1 = 1, param2 = 1) then it works. So the `testSimpleParsets()` does only work with one scenario and this means that one can also use `testSimple` two times successively to test the two scenarios, but this is bloats the code (see for example test_classif_saeDNN.R).  Though it works. Hence, how shall I fix these tests?

Thanks!

larskotthoff:
Thanks, sounds like very good progress.

What errors are you getting with the learners that `testSimpleParsets` does not work with?

MariaErdmann:
> What errors are you getting with the learners that testSimpleParsets does not work with?

````
as.character(cp$data$response) not equal to as.character(old.predicts).
Lengths differ: 107 vs 0
````

I changed test_classif_mlp.R such that it shows what I mean: right now, the parset.list just contains  one list (so travis and jenkins do not fail). But if you outcomment the second list, then you will get the error. It doesn't matter which list in the parset.list is outcommented. If there is only one it works fine, if there are two not:

works:
````
parset.list = list(
    #list(),
    list(size = 7L, maxit = 150L)
  )
````

does not work:
````
parset.list = list(
    list(),
    list(size = 7L, maxit = 150L)
  )
````

larskotthoff:
Wait, are those the tests that don't use `testSimpleParsets` to start with? I'm not sure why that is (ping @berndbischl @mllg), but those have explicit calls to `testSimple` with parsets -- just add another call with an empty parset. Thanks!

MariaErdmann:
I went through all the test files and I think I changed all test files where necessary.
There is only one test that fails: test_multilabel_randomForestSRC.R but the local check does not produce an error and travis/jenkins neither. Are the multilabel test files not tested?
As we/ I didn't change anything in the learner/ test file, the test must have been broken before.
The latest CRAN release of randomForestSRC was in July. I tried several things but I  don't know yet
what is wrong with the test.
As this is a prio-high PR and as I touched quite a lot of files (which makes the PR prone to merge conflicts) and as I don't know how long it will take me to find the bug in that one, I think it is probably better to solve thisbug in another PR and close this one as soon it is possible.
Hence, in my opinion this is ready for review/merge :-)

Issue #958, opened 2016-06-20
multilabelBinaryRelevanceWrapper is not working properly for classif.glmnet


MariaErdmann:
I am not sure if this is a bug:
prediction for classif.glmnet wrapped to a multilabelBinaryRelevanceWrapper throws an error. 
In the example below I subset the yeast.task but using the whole task will cause the same error.

```
#load_all() 
yeast = getTaskData(yeast.task)
yeast = yeast[1:300,c(1:3, 113:117)]
labels = colnames(yeast)[1:3]
yeast.task = makeMultilabelTask(id = "multi", data = yeast, target = labels)
yeast.task

# Create multilavelBinaryRelevanceWrapper for classif.glmnet
lrn = makeLearner("classif.glmnet")
lrn2 = makeMultilabelBinaryRelevanceWrapper(lrn)
mod = train(lrn2, yeast.task)
pred = predict(mod, task2)
#Error in predict(mod, yeast.task) : 
#  Assertion on 'subset' failed: Must be of type 'integerish', not 'double'.

```

I tried to find out where it comes from but was not successful (yet).
All other binary (or multiclass) classification learners do work with the Wrapper though.

catastrophic-failure:
Happened to be on a ensemble in a real application. Perhaps it has something to do with the number of predictors, as I was trying attribute bagging.

```splus
lrn = makeLearner("regr.glmnet", alpha = 0.5)
lrn = makeBaggingWrapper(lrn, bw.iters = 500L, bw.replace = FALSE, bw.size = 1, bw.feats = 1/50)
res = benchmark(lrn, task, rdesc.outer, list(mse, expvar, timeboth))
#Warning in predict.WrappedModel(m, task, subset = train.i) :
#  Could not predict with learner regr.glmnet.bagged: Error in benchmark(lrn, IQ.tasks[[1]], #rdesc.outer, list(mse, expvar,  : 
#  Assertion on 'subset' failed: Must be of type 'integerish', not 'double'.
```


```

schiffner:
I only had time to look at this for a minute:

It turns out that the error occurs because `subset` has the value `0.01`.
My guess is that by partial matching `subset` gets confused with the parameter `s` of `glmnet` which is set to `0.01` in the mlr interface to `glmnet` (cp. `lrn$par.vals`).
So the core of the problem would be how to unambiguously pass the params in `par.vals` to `predict`.

berndbischl:
julia is correct. i figured out the same problem right now.
[...]

PR #1257, opened 2016-09-21
fix 958 classif.glmnet in multilabelBinaryRelevanceWrapper

Coorsaa:
fixed bug from issue #958.
set subset to NULL in predict function in order to prevent the parameter s of glmnet from getting partial matching problems.

### DIFF
--- a/R/MultilabelBinaryRelevanceWrapper.R
+++ b/R/MultilabelBinaryRelevanceWrapper.R
@@ -50,11 +50,11 @@ trainLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .task, .subse
 }
 
 #' @export
-predictLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .model, .newdata, ...) {
+predictLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .model, .newdata, .subset = NULL, ...) {
   models = getLearnerModel(.model, more.unwrap = FALSE)
   f = if (.learner$predict.type == "response")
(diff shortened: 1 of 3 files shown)


larskotthoff:
Could you also add a unit test for this please?

berndbischl:
*at least* the thing from the issue report

R/MultilabelBinaryRelevanceWrapper.R
```diff
@@ -50,11 +50,11 @@ trainLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .task, .subse
 }
 
 #' @export
-predictLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .model, .newdata, ...) {
+predictLearner.MultilabelBinaryRelevanceWrapper = function(.learner, .model, .newdata, .subset = NULL, ...) {
```
giuseppec: We don't use `.subset` in the function, why?

larskotthoff:
Is this ready to be merged? @Coorsaa  @giuseppec

Coorsaa:
@larskotthoff: I think so, however, @giuseppec might want to look into it again.

giuseppec:
Looks good to me.

larskotthoff:
Ok, thanks. Just one question -- in `predict.R` you're changing the default for `subset` to `NULL`. Why?

Coorsaa:
glmnet has a parameter "s" so we had a partial matching with the subset if the default was not set as discussed in issue #958

larskotthoff:
Oh yes and you actually said that before. Sorry, I missed that!

PR #1260, opened 2016-09-22
Make hyperparseffect tests faster

MasonGallo:
@berndbischl let me know that the tests here were slow - I lowered the iterations and length of params. Should be much faster now.

### DIFF
--- a/tests/testthat/test_base_generateHyperParsEffect.R
+++ b/tests/testthat/test_base_generateHyperParsEffect.R
@@ -5,7 +5,7 @@ test_that("generate data", {
   ps = makeParamSet(makeNumericParam("C", lower = -5, upper = 5,
     trafo = function(x) 2^x)
   )
-  ctrl = makeTuneControlRandom(maxit = 10L)
+  ctrl = makeTuneControlRandom(maxit = 5L)
   rdesc = makeResampleDesc("Holdout")
   lrn = makeTuneWrapper("classif.ksvm", control = ctrl,
     resampling = rdesc, par.set = ps,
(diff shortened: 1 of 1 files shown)


larskotthoff:
Why do we need this? It doesn't look like the overall build time decreased.

berndbischl:
> Why do we need this?

because thats a test in base which is often run manually. at it was by far one of the slowest tests in base
(was a tad annoying when i looked at some stuff locally....)
"base" is also run on CRAN

MasonGallo:
> It doesn't look like the overall build time decreased.

To be fair, the overall build time could be driven by a number of external factors. This should certainly reduce at least processor time for this particular test.

larskotthoff:
Ok, so by how much is the time for this test reduced? Are we talking minutes to seconds? I'm hesitant about anything that reduces the test coverage.

Maybe make this test really small (1 iteration) and add another one with more iterations in not base?

MasonGallo:
In my non-scientific experiment on my linux box, I checked the timing.

On master:

```
   user  system elapsed 
104.780   0.056 104.838 
   user  system elapsed 
106.096   0.048 106.138 
   user  system elapsed 
105.740   0.040 105.761  
   user  system elapsed 
106.284   0.188 106.469 
```

And on my new branch:

```
   user  system elapsed 
 22.412   0.036  22.461 
   user  system elapsed 
 22.844   0.016  22.862 
   user  system elapsed 
 22.880   0.028  22.903 
   user  system elapsed 
 22.664   0.020  22.693 
```

So estimated savings of ~5x.

I changed this at @berndbischl 's request, so I'll let him weigh in here. My opinion is that we should have some small # of iterations like 5 or 10 so that we can still check that lines get drawn and interpolation happens. Or we can just leave as is if we don't care about base test speed.

larskotthoff:
I'll merge this now, but please @berndbischl comment.

Issue #1262, opened 2016-09-25
friedmanTestBMR - gives strange error message


xhudik:
```{r}
lrns=list(
  lrn.logreg = makeLearner("classif.logreg",predict.type = "prob"),
  lrn.svm = makeLearner("classif.svm",predict.type = "prob")
)
bmr = benchmark(lrns, pid.task)
friedmanTestBMR(bmr)
```
gives confusing and meaningless error message
```{r}
Error in apply(y, 1L, rank) : dim(X) must have a positive length
```
This is especialy confusing for newbies. Not sure whether it can be changed, but the current behavior is not user-friendly

thanks

larskotthoff:
Where exactly is this error message coming from? Can you post a complete, reproducible example please?

xhudik:
@larskotthoff  - the code I posted is complete. (just add library(mlr)). pid.task is an integarl part of the mlr package

larskotthoff:
This looks like a bug in `friedman.test()`.

schiffner:
I agree with Lars that `freedom.test` doesn't handle this well.

But in mlr we could improve the arg checks and docs for `friedmanTestsBMR`. It does not make sense to use the Friedman test if there aren't at least results for 2 learners and 2 tasks in the `BenchmarkResult`.

larskotthoff:
This should be fixed now. Closing here; please reopen if the issue persists.

Issue #705, opened 2016-02-12
Update docs of crit differences plots


berndbischl:
In this is PR @schiffner said: 
https://github.com/mlr-org/mlr/pull/695

§The docs of friedman*TestBMR, generateCritDifferencesData, plotCritDifferences need some improvements. It's probably easiest if I just do a commit in this branch after you are done, Bernd."

berndbischl:
I would be great if you 
1) could update the docs
2) if while you do that you see some easy clean ups in the code, go for it please
3) can you transform the detected problem in the PR thread into a unit test?

schiffner:
> 3) can you transform the detected problem in the PR thread into a unit test?
I can try to come up with tests for the "changing the `baseline`" problem (#695) and the earlier bug (in #643).

A general question about any tests for this: I never fully understood why mlr was broken on CRAN.
Any advice so that I don't accidentally build in a similar problem again?

berndbischl:
2 answers:

a) mlr was broken on cran because ggplo2 was API-changed. and @pfistfl simply created a unit test which with 1% probability failed, beacause it was stochastic and lets say sub-optimally designed. 
So this does not really concern you.

b) look at what @zmjones and @larskotthoff did to test correctness of plots and talk to them. because this seems to be the problem here, right?

schiffner:
Can I rename the `p.value` arg?
I don't know why I missed that before, but you don't pass a p-value to a test, but a significance level.

berndbischl:
> Can I rename the p.value arg?

yes do that

schiffner:
Just noticed in the tutorial that the x-axis of the critical differences plot is now missing (?!).
This is just to remind me that I need to investigate this, too.

PR #1267, opened 2016-10-05
mini fix for Friedman tests (#1262)

schiffner:
Added checks for the number of learners and tasks in a BenchmarkResult before doing the Friedman test + unit test. See #1262.

friedmanTestBMR and related functions require some more cleanup (see also #705). I will issue a separate PR for that.


larskotthoff:
Thanks, merging.

PR #1268, opened 2016-10-05
New measures squared log error and root squared log error

florianfendt:
Update for NEWS:

````
## measures - new
* msle
* rmsle
````


larskotthoff:
Thanks. Where did you get the definition for this? Given that the mean is taken, that should be reflected somewhere in the name (otherwise it's very similar to `sse`, but a different quantity).

florianfendt:
@larskotthoff 
Here's where I found the definition for the measures:
https://www.kaggle.com/wiki/RootMeanSquaredLogarithmicError

I renamed measures to msle and rmsle and also adjusted the test according to your comment on Monday. Hope everything's ok now and thanks for your support!

larskotthoff:
Thanks, looks good. Could you please add the link to the definition to the note please? And it looks like your tests are failing.

florianfendt:
@larskotthoff 
I think we are done here now.
Thanks for your help!

larskotthoff:
Thanks, merging.

Issue #625, opened 2015-12-19
test.join aggregation does not join probabilities


berndbischl:
```
task = sonar.task
lrn = makeLearner("classif.rpart", predict.type = "prob")

rin = makeResampleDesc("Holdout")
mm = setAggregation(auc, test.join)
r = resample(lrn, task, rin, measures = mm)
```

larskotthoff:
Did the PR solve this completely or should we leave this open?

giuseppec:
The part 

> NB: we have potentially more columns we need to join like se and the stuff from survival...

was not addressed in the PR, therefore we could leave this open.

jakob-r:
This error still prevails
```r
> task = wpbc.task
> lrn = makeLearner("surv.rpart")
> rin = makeResampleDesc("Holdout")
> mm = setAggregation(cindex, test.join)
> r = resample(lrn, task, rin, measures = mm)
[Resample] holdout iter: 1
Error in FUN(X[[i]], ...) : 
  You need to have 'truth.time' and 'truth.event' columns in your pred object for measure cindex!
```

jakob-r:
And
```r
task = yeast.task
lrn = makeLearner("multilabel.cforest")
rin = makeResampleDesc("Holdout")
mm = setAggregation(multilabel.acc, test.join)
r = resample(lrn, task, rin, measures = mm)
#[Resample] holdout iter: 1
#Error in `row.names<-.data.frame`(`*tmp*`, value = value) : 
#  invalid 'row.names' length
```

jakob-r:
In other words: We should not write custom code for every task class in `test.join` but find a generalized solution.

pat-s:
``` r
library(mlr)
#> Loading required package: ParamHelpers

task = wpbc.task
lrn = makeLearner("surv.rpart")
rin = makeResampleDesc("Holdout")
mm = setAggregation(cindex, test.join)
r = resample(lrn, task, rin, measures = mm)
#> Resampling: holdout
#> Measures:             cindex
#> [Resample] iter 1:    0.6156373
#> Error in FUN(X[[i]], ...): You need to have 'truth.time' and 'truth.event' columns in your pred object for measure cindex!
```

<sup>Created on 2019-12-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1275, opened 2016-10-07
Issue #625: solve aggregation issue for test.join with probs

giuseppec:
Try to solve https://github.com/mlr-org/mlr/issues/625

### DIFF
--- a/R/aggregations.R
+++ b/R/aggregations.R
@@ -235,8 +235,13 @@ test.join = makeAggregation(
     df = as.data.frame(pred)
     f = if (length(group)) group[df$iter] else factor(rep(1L, nrow(df)))
     mean(vnapply(split(df, f), function(df) {
+      if (pred$predict.type == "response") y = df$response
+      if (pred$predict.type == "prob") {
+        y = df[,grepl("^prob[.]", colnames(df))]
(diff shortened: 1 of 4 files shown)


larskotthoff:
Could you add the example from the issue report as a test please?

giuseppec:
Done. 

Note: For repeated-CV `test.join` currently computes the measure for each repeated cv separately and then averages this through the mean. For example, a 3 times repeated 10-fold CV creates cross-validated prediction for each repetition, e.g. y1, y2, and y3. In this case, currently `test.join` computes the measure for y1 and the real y, y2 and the real y, y3 and the real y and finally aggregates the three values using the mean. (why aggregating here? especially why using the mean?).
An alternative to this (without the issue of aggregating) would be to average the cross-validated predictions accross all reps, i.e., use the mean ymean = (y1+y2+y3)/3 and compute the measure using ymean and the real y.

larskotthoff:
Hmm, sounds to me like this should be configurable by the user (in particular how the aggregation is done). This is probably a separate PR though?

giuseppec:
Should we first merge this PR that fixes a bug? And then open up a new issue for that as this is rather an enhancement and no bug.

larskotthoff:
Sounds good. I'll merge this now.

Issue #1279, opened 2016-10-11
Faster filter functions


studerus:
I just discovered the `Rfast` package and saw that it has some very fast implementions of various univariate tests. I was thinking that we could use them for making faster filter functions. It seems that we could improve the speed of the correlation as well as ANOVA filters at least ten fold. See the following benchmarks:

```
dat <- data.frame(matrix(rnorm(50000), ncol = 500))
regr.task <- makeRegrTask(dat = dat, target = 'X1')
class.task <- makeClassifTask(dat = data.frame(y = factor(rbinom(nrow(dat), 1, 0.5)), dat), target = 'y')

rank1 = function(task) {
    y = FSelector::rank.correlation(getTaskFormula(task), data = getTaskData(task))
    setNames(y[["attr_importance"]], getTaskFeatureNames(task))
}

rank2 = function(task) {
    d = getTaskData(task, target.extra = TRUE)
    setNames(abs(Rfast::correls(d$target, d$data, type = 'spearman')[,1]), getTaskFeatureNames(task))
}

lincor1 = function(task) {
    y = FSelector::linear.correlation(getTaskFormula(task), data = getTaskData(task))
    setNames(y[["attr_importance"]], getTaskFeatureNames(task))
}

lincor2 = function(task) {
    d = getTaskData(task, target.extra = TRUE)
    setNames(abs(Rfast::correls(d$target, d$data, type = 'pearson')[,1]), getTaskFeatureNames(task))
}

anov1 = function(task) {
    data = getTaskData(task)
    sapply(getTaskFeatureNames(task), function(feat.name) {
      f = as.formula(stringi::stri_paste(feat.name,"~",getTaskTargetNames(task)))
      aov.t = aov(f, data = data)
      summary(aov.t)[[1]][1,'F value']
    })
}
anov2 = function(task) {
    d = getTaskData(task, target.extra = TRUE)
    setNames(Rfast::ftests(as.matrix(d$data), as.numeric(d$target))[,1], getTaskFeatureNames(task))
}


rbenchmark::benchmark(rank1(regr.task), rank2(regr.task), replications = 10)
all.equal(rank1(regr.task), rank2(regr.task))
rbenchmark::benchmark(lincor1(regr.task), lincor2(regr.task), replications = 10)
all.equal(lincor1(regr.task), lincor2(regr.task))
rbenchmark::benchmark(anov1(class.task), anov2(class.task), replications = 10)
all.equal(anov1(class.task), anov2(class.task))
[...]

PR #1280, opened 2016-10-11
Add faster filter functions using Rfast #1279

studerus:


larskotthoff:
Hmm, looks like Travis is failing because gsl isn't installed and it won't allow it to be installed! @mllg do you have any ideas about this?

mllg:
According to https://github.com/r-hub/sysreqsdb/blob/master/sysreqs/libgsl.json, you need to install libgsl0-dev. Should be whitelisted in https://github.com/travis-ci/apt-package-whitelist.

larskotthoff:
Thanks @mllg. Now quickcheck doesn't install properly and fails the tests. I'll kick this off in a bit again to see if it's fixed.

larskotthoff:
All good now, merging, Thanks.

PR #1283, opened 2016-10-12
re-enable R devel tests

larskotthoff:
R devel tests on Travis are currently broken because ranger doesn't compile. I've disabled them in master so the builds succeed; re-enable here when fixed.


larskotthoff:
Still not working, closing. We have devel tests on Jenkins.

Issue #1281, opened 2016-10-11
{generate,plot}HyperParsEffect{Data} and more than 2 hyperparameters: no example?


eddelbuettel:
When trying this (much appreciated) feature, I get

```r
Error in checkLearner(learner, "regr", props) : 
  Learner 'classif.xgboost' must be of type 'regr', not: 'classif'
R> 
```

That was a basic XGBoost with three parameters and `partial.dep=TRUE`.  There are two tests in 
`generateHyperParsEffect.R` which have 

```r
  if (checkClass(interpolate, "Learner") == TRUE ||
      checkString(interpolate) == TRUE) {
    lrn = checkLearnerRegr(interpolate)
  }
```

meaning that classifiers won't work.  I tried relaxing this to `checkLearner(interpolate)` but no luck so far.

schiffner:
Hi,

I hope I understand your question correctly.
{generate,plot}HyperParsEffect{Data} is intended to work for hyperparameters of a classification learner.

Since you set `partial.dep = TRUE`: What learner did you choose as `partial.dep.learn` in `plotHyperParsEffect`? Maybe the problem lies there because this must always be a regression learner (and the same holds for the `interpolate` learner)?

To make sure that classification with 3 hyperpars really works I made a small example for myself.
```
ps = makeParamSet(
  makeIntegerParam("degree", lower = 1, upper = 5),
  makeNumericParam("offset", lower = 0, upper = 5),
  makeNumericParam("C", lower = -5, upper = 5, trafo = function(x) 2^x))
ctrl = makeTuneControlRandom(maxit = 100L)
rdesc = makeResampleDesc("Holdout")
learn = makeLearner("classif.ksvm", par.vals = list(kernel = "polydot"))
res = tuneParams(learn, task = pid.task, control = ctrl, measures = acc,
  resampling = rdesc, par.set = ps, show.info = FALSE)
data = generateHyperParsEffectData(res, partial.dep = TRUE)
plt = plotHyperParsEffect(data, x = "C", y = "degree", z = "acc.test.mean",
  plot.type = "heatmap", partial.dep.learn = "regr.earth")
```

Maybe this is helpful to you.
Otherwise we would need a more complete code example from you.

eddelbuettel:
Looking at this I think I specified the `partial.dep.learn` argument wrong.  If you could just add what we have here to the manual page as an example things would be much better.

Your example surely works just fine, even over remote ssh+x11. So a big thanks!

[...]

PR #1288, opened 2016-10-13
Fixes plotHyperParsEffects for #1281

MasonGallo:
See #1281

- [x] Correctly accepts any regression learner for partial dependence
- [x] Arg check when user incorrectly calls `interpolate` and `partial.dep.learn` simultaneously
- [x] Stronger language in docs for using regression


schiffner:
Thanks very much. Looks good to me.

larskotthoff:
Thanks, merging.

Issue #1287, opened 2016-10-13
Hyperparameter effects data and plot: only subsets shown


eddelbuettel:
I am getting weird results from `plotHyperParsEffect()`. Here is a variant.  The grid is 3 x 3, but the plot 
omits two `ntree` values and and one `mtry` value.  That may be on purpose for lack of variability in the outcome -- if so, could we signal more clearly?

```
    paramset <- makeParamSet(makeDiscreteParam("ntree", values=c(100,500,1000)),
                             makeDiscreteParam("mtry", values=c(1,3,5)))
    ctrl <- makeTuneControlGrid()
    rdesc <- makeResampleDesc("CV", iters = 5L)
    learn <- makeLearner("classif.randomForest", predict.type="prob")
    lrn <- makeTuneWrapper(learn,
                           control = ctrl,
                           measures = list(acc, mmce),
                           resampling = rdesc,
                           par.set = paramset,
                           show.info = TRUE)
    res <- resample(lrn,
                    task = pid.task,
                    resampling = rdesc,
                    extract = getTuneResult,
                    show.info = TRUE)
    data <- generateHyperParsEffectData(res, partial.dep=TRUE)
    plt <- plotHyperParsEffect(data,
                               x = "ntree", #"nrounds",
                               y = "mtry",
                               z = "acc.test.mean",
                               plot.type = "heatmap",
                               show.experiments = TRUE,
                               nested.agg = mean,
                               partial.dep.learn = "regr.earth")  
    print(plt)
```

MasonGallo:
Thanks for submitting this. 

I would argue that you don't have enough data to "do" partial dependence here: you're attempting to learn a regression function over 2 features with essentially 9 observations. Based on your previous discussion, I think I'm preachign to the choir here :)

What's happening is partial dependence under the hood range is getting tripped up with integers. Looks like the problem is with `generatePartialDependenceData` not `plotHyperParsEffect`. I have an idea for a quick fix which I should get to tonight, assuming it doesn't break anything else.

Thoughts anyone else?

schiffner:
Agree with your conclusion. Let me or Zach know if we can help.

schiffner:
Maybe one more thought.
It would be nice to have faceting also for heat maps, in cases were we have a small parameter grid or some parameters with few values, but in total > 2 hyper pars (i.e, the case where at the moment the only way is partial dependence, but fitting a regression model might not work that well). Does that make sense?

MasonGallo:
[...]

PR #1289, opened 2016-10-13
Fix partial dep range grid for integer features (#1287)

MasonGallo:
See #1287 

- [x] Integer features now correctly create a sequential scale from min to max in partial dependence
- [x] Warning if numerical feature has small num unique obs (I don't love this so lmk)

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -45,13 +45,14 @@ generateFeatureGrid = function(features, data, resample, gridsize, fmin, fmax) {
         factor(rep(levels(data[[feature]]), length.out = cutoff),
                levels = levels(data[[feature]]), ordered = is.ordered(data[[feature]]))
       } else {
+        if (nunique <= 5L) {
+          warningf("Number of unique observations for %s is %i. Are you sure you have enough data?",
+            feature, nunique)
+        }
         if (resample != "none") {
           sort(sample(data[[feature]], cutoff, resample == "bootstrap"))
         } else {
-          if (is.integer(data[[feature]]))
-            sort(rep(fmin[[feature]]:fmax[[feature]], length.out = cutoff))
-          else
-            seq(fmin[[feature]], fmax[[feature]], length.out = cutoff)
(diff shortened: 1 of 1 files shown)


MasonGallo:
@zmjones is this alright? I'm not sure of the wider impact from this change

larskotthoff:
Thanks, merging. In the medium term we should think about how to test this better (but AFAIK you're working on a PR for that?).

zmjones:
hmmm i am not sure why i did the whole `rep(low:high)` thing. the actual bug was truncation?

i don't think the warning is appropriate here. there is nothing wrong with doing partial dependence on an integer feature with only a few unique values. Take for example any ordered factor encoded as an integer. For learners which don't directly handle ordered factors you'd have to do this. And of course you can encode any binary feature this way without any consequences. 

I also don't think this has anything to do with the amount of data you have. In general, just because I observe that there are few unique values for feature x, does not mean that I cannot estimate f_x(x) from f(x, y). I may simply have many repeated values of x. The variability in this estimate comes from the amount of training data, the complexity of the underlying function, and the rate at which the learner you use converges to this function or an approximation thereof.

I can see how this would be less likely in your functions @MasonGallo, but it would occur if you had a higher resolution grid for some parameters and a smaller one for others, no?

MasonGallo:
> hmmm i am not sure why i did the whole rep(low:high) thing. the actual bug was truncation?

Yes it seemed that integer features were not correctly drawing a grid. Basically only the `fmin` to `gridsize`  were drawn.

> I can see how this would be less likely in your functions @MasonGallo, but it would occur if you had a higher resolution grid for some parameters and a smaller one for others, no?

I agree, which is why I said I didn't love it. I had left it in for us to discuss. I am a fan of warnings of a similar nature to protect users from "black boxing", but prob elsewhere is a better place.

zmjones:
I am rewriting the prediction grid function now and will issue a pr shortly. I think we can better handle ordered factors and some other things too.

PR #1291, opened 2016-10-14
rewrite of generateFeatureGrid and better tests

zmjones:
- better handling of factors and ordered factors
 - test resample = "none" and "subsample" (not bootstrap)
 - test range match (for resample = "none") and factor level preservation


larskotthoff:
Thanks.

The switch statement is weird though -- is this the best way to do it? Could you extract the classes and then use `%in%` for example?

zmjones:
Why is that weird? I added this to my own version of BBmisc and did it using a method, but I thought this solution was cleaner than a series of if/else

larskotthoff:
I find the coercion into strings a bit weird. But ok, it does look cleaner and makes it obvious what you're checking for. I'll merge this.

PR #914, opened 2016-05-26
#747 mergeBenchmarkResults

florianfendt:
Not totally ready to merge yet.
We need to agree on how we want to treat the following situation:
All bmrs have been optimized on the same measure but for some bmrs additional measures were calculated.
Right now all the measures would need to be equal and we stop with an error otherwise.

PR #1293, opened 2016-10-17
store task.desc instead only task.id in ResampleResult

giuseppec:
This should be fast to review @larskotthoff . I outsourced this from PR #914 as this seems to be needed for several other issues.


larskotthoff:
Thanks, merging. I changed it so that the description is stored in addition to the ID.

PR #1297, opened 2016-10-18
Import cleanup

mllg:
Moved shiny and ggvis to suggests.

@zmjones Can you please have a look? I don't know how well the plotting stuff is covered by the unit tests.


berndbischl:
i think we tried this before, but i would be very happy if we could remove these (heavy) imports and put them in suggests. 

@mllg @zmjones @schiffner 
maybe build the tutorial after this to check this? if that works and travis is green here, i am pretty sure that we are safe

schiffner:
I can check the tutorial (if it builds and if anything looks strange).

zmjones:
this is fine. shiny/ggvis aren't tested automatically.

berndbischl:
> I can check the tutorial (if it builds and if anything looks strange).

as travis seems to run: i think you could do this now then we merge?

schiffner:
> as travis seems to run: i think you could do this now then we merge?

I think you can just merge. As Zach said this can't break the tutorial since we don't call any ggvis functions in the tutorial. (There are some examples that show how you would do this, but none of them are evaluated.)

larskotthoff:
Thanks, merging.

berndbischl:
has somebody at least tried to run 1 ggvis function with an mlr install?

schiffner:
I tried the tutorial examples. They run. The learning curve example needed a small update which I did but haven't pushed yet.
One small thing I noticed was that in `plotFilterValuesGGVIS` changing the number of features selected has no effect, but wanted to have a closer look before I open an issue for this.

berndbischl:
ok thx!

PR #1300, opened 2016-10-20
reate a list of `Prediction` objects for each single resample iteration

giuseppec:
I outsourced this from PR #914 as this seems to be necessary also for other issues. This new function creates a list of `Prediction` objects for each single resample iteration of a `ResampleResult` object. 
Since mlr by default computes measures for each resample iteration separately and then aggregates the measures, we can use this function to simply add additonal measures to an already existing `ResampleResult` object (see also the `addMeasure` function from PR #914, which I will also outsource from PR #914 after this PR has been merged).

### DIFF
--- a/R/ResampleResult_operators.R
+++ b/R/ResampleResult_operators.R
@@ -28,3 +28,62 @@ getRRPredictions = function(res) {
 getRRTaskDescription = function(res) {
   res$task.desc
 }
+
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, looks good. Could you change the string operations to use stringi functions please?

giuseppec:
Done

PR #1303, opened 2016-10-24
Add new measures to existing ResampleResult

giuseppec:
I outsourced this from PR #914 because it is easier to review in an extra PR. 
We can use this function to simply add additonal measures to an already existing `ResampleResult` object. After this is merged, we can finally adapt PR #914 and hopefully also merge that PR.


larskotthoff:
Thanks, merging.

PR #1304, opened 2016-10-24
fix multilabel unit tests

berndbischl:
1) there was a trvial bug in file test_multilabel_randomForestSRC.R

2) worse: the multilabel unit tests were not even *run* on travis, so we did not see this. a "run" file was missing. i added that here now.


berndbischl:
pls merge ASAP when / if travis is green

giuseppec:
Maybe this https://github.com/mlr-org/mlr/pull/1019 can also be done at the same time, there are two "pings" from @smilesun

PR #1307, opened 2016-10-25
convert newdata data.table to data.frame with warning in predict

### LINKED ISSUE #1306: predict fails when using a data.table instead of a data.frame
This should work imo

```
library(mlr)
library(data.table)
irisdt = data.table(iris)
mod = train("classif.qda", iris.task)
predict(mod, newdata = irisdt)

> Error in predict(mod, newdata = irisdt) : 
  Assertion on '.newdata' failed: Must be of type 'data.frame', not 'integer'.
```

ja-thomas:
Should fix #1306 

Base change so review thoroughly.

This is basically the same check we do for tasks. We could also create a helper function for that, especially when this is required at more points

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -52,7 +52,7 @@ predict.WrappedModel = function(object, task, newdata, subset, ...) {
     size = getTaskSize(task)
   } else {
     assertDataFrame(newdata, min.rows = 1L)
-    if (!("data.frame" %in% class(newdata))) {
+    if (class(newdata)[1] != "data.frame") {
       warningf("Provided data for prediction is not a pure data.frame but from class %s, hence it will be converted.",  class(newdata)[1])
       newdata = as.data.frame(newdata)
     }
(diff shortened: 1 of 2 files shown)


larskotthoff:
I would change the check to include data tables instead of coercing any data into a data frame. This seems overly permissive to me.

ja-thomas:
Well right before the coercing we do an `assertDataFrame`, so most stuff that is not coerable to a data.frame should be catched before.

That way we also remove a warning if we have a `tbl_df` (that works currently but gives an warning).

Not sure what/if other data.frame like objects exist

larskotthoff:
Ok, I changed your check a bit so it works if `data.frame` isn't the first class. I'll merge once Travis is done.

larskotthoff:
Argh, I broke it. I'm not really comfortable with this very general check that catches a specific case. I would prefer a specific check and conversion for a data table. Does that sound ok?

ja-thomas:
so basically this is the exact same we are doing when converting a data.frame (or data.table, tbl_df) to a task.

But we can also do it like that

`"data.frame" %in% class(newdata) && class(newdata)[1] != "data.frame"`

Then we will only convert objects that have the class `data.frame` but also some additional.

larskotthoff:
Ok, fair enough. You've convinced me :)

Merging.

PR #1310, opened 2016-10-26
Backports

mllg:
* Import backports (strrep() called, maybe more functions
* Replaced `any(is.na(x))` with `anyMissing`
* Added some final newlines where missing


berndbischl:
thx!

PR #1312, opened 2016-10-27
Profiled bottlenecks

mllg:
I've profiled some tuning runs and time was wasted unnecessarily in getting options, checks for setting hyperpars and getting learner properties. 

The program flow is sometimes still inefficient, but at least the operations are now faster.

### DIFF
--- a/R/BaggingWrapper.R
+++ b/R/BaggingWrapper.R
@@ -42,7 +42,7 @@
 #' @family wrapper
 #' @export
 makeBaggingWrapper = function(learner, bw.iters = 10L, bw.replace = TRUE, bw.size, bw.feats = 1) {
-  learner = checkLearner(learner, type=c("classif", "regr"))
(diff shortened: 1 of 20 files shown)


catastrophic-failure:
Nice improvements. One possible bottleneck is the use of formulas in learners that support matrix input (I think someone reported formulas in randomForest were specially problematic).

berndbischl:
> One possible bottleneck is the use of formulas in learners that support matrix input (I think someone reported formulas in randomForest were specially problematic)

yes, but we cannot do this here. the i have discussed this quite often with various persons, but the problem is how to handle factors? not every algorithm computes the "normal" model.frame inside of it and we dont want to change behavior.

although thinking about it: we could, at least for tasks with PURELY numerical data then use the matrix interface? 
(results on more complex learner interface code, but possibly a large speedup)
shall we open up an issue for this?

mllg:
Open a new issue for the formula stuff, this is something different.

Issue #1284, opened 2016-10-13
bug in resample using predict = "train"


giuseppec:
``` r
lrn = makeLearner("classif.rpart", predict.type = "prob")
rdesc = makeResampleDesc("CV", iter = 3, predict = "train")
mmce.train = setAggregation(mmce, train.mean)
res = resample(lrn, binaryclass.task, rdesc, mmce.train)
```

Issue 1: printing `res$pred` yields

```
Resampled Prediction for:
Resample description: cross-validation with 3 iterations.
Predict: train
Stratification: FALSE

threshold: 
time (mean): NA
Error in as.data.frame.default(x) : 
  cannot coerce class "c("ResamplePrediction", "NULL")" to a data.frame
In addition: Warning message:
In mean.default(x$time) : argument is not numeric or logical: returning NA
```

Issue 2: `res$pred$predict.type` seems to be `NULL` but should still have the same value as `lrn$predict.type`!
For example, using `predict = "both"` seems to work:

``` r
lrn = makeLearner("classif.rpart", predict.type = "prob")
rdesc = makeResampleDesc("CV", iter = 3, predict = "both")
mmce.train = setAggregation(mmce, train.mean)
res = resample(lrn, binaryclass.task, rdesc, mmce.train)
res$pred$predict.type
# [1] "prob"
```

MariaErdmann:
The problem is in `makeResamplePrediction`. See my (unfortunately not so short example below):
````
learner = makeLearner("classif.rpart")
task = binaryclass.task
resampling = makeResampleDesc("CV", iters = 2, predict = "train")
resampling = makeResampleInstance(resampling, task = task)
rin = resampling

mmce.train = setAggregation(mmce, train.mean)
extract = function(model) {}
more.args = list(learner = learner, task = task, rin = rin, weights = NULL,
  measures = list(mmce.train), model = FALSE, extract = extract, show.info = getMlrOption("show.info"))

library(parallelMap)
parallelLibrary("mlr", master = FALSE, level = "mlr.resample", show.info = FALSE)
exportMlrOptions(level = "mlr.resample")
iter.results = parallelMap(doResampleIteration, seq_len(rin$desc$iters), level = "mlr.resample", more.args = more.args)

ms.train = as.data.frame(extractSubList(iter.results, "measures.train", simplify = "rows"))
ms.train
ms.test = extractSubList(iter.results, "measures.test", simplify = FALSE)
[...]

PR #1315, opened 2016-10-28
Fix bug in resampling when using predict = "train" (#1284)

MariaErdmann:
suggestion for NEWS:

```
## Bug fixes
* fixed bug issued in #1284, a bug in resample when using predict = "train"
```


larskotthoff:
Thanks. Do the tests cover the case that you reported in the issue?

MariaErdmann:
> Do the tests cover the case that you reported in the issue?

I checked again and recognized I missed checking for predict.type and threshold.

I thought I check whether the entries in \<resample.object\>$pred for threshold and predict.type are not NULL which caused this error 
````
Error in as.data.frame.default(x) : 
  cannot coerce class "c("ResamplePrediction", "NULL")" to a data.frame
````
and then I check whether there are entries in time which are not NA to catch this warning 
````
In mean.default(x$time) : argument is not numeric or logical: returning NA
````
Moreover, I check whether the task description in the \<resample.object\>$pred is the same as the passed one.

larskotthoff:
Thanks, merging.

giuseppec:
@MariaErdmann don't forget to open up a new PR for the things we discussed and shortly summarize them.

MariaErdmann:
@giuseppec : Yes, thank you, see #1324 .

PR #1320, opened 2016-11-03
intergration of the earth classifier

rwarnung:
Version of the earth classifier with a lot of tags - maybe we should remove them.


larskotthoff:
Thanks for the PR! Can you please add a test for the new classifier? And for converting the predictions to probabilities, could you use a transformation as in #549 please?

The documentation looks fantastic, but I'm worried that it'll get out of date when the learner package is updated. Is there any information in there that's not in the help for the earth learner? I would keep that, but remove the things that are duplicated.

rwarnung:
@larskotthoff you are right about the documentation. I removed all the tags. For the transform: There is a glm with logit link in the function - so probabilities should not move outside of [0,1] and classes are determined from these probabilities. Do you still think we need the transform? Sorry, if I am mistaken. I will try to add the tests next week. Thanks for the feedback.

larskotthoff:
Ok, sounds good!

rwarnung:
Hi, I fixed the learner file (comments and a piece of logic) and added 3 tests. Please tell me if anything is missing.

larskotthoff:
Thanks, looks good. The tests are failing on Travis though -- are they passing for you locally?

rwarnung:
Thanks for habing a look. The checks work for me locally - but I get warnings that the probabiliies are numerically 0 or 1 - which is correct. Is a warning a problem?

tests/testthat/test_classif_earth.R
```diff
@@ -0,0 +1,39 @@
...
+    pars = c(pars, parset)
+    set.seed(getOption("mlr.debug.seed"))
+    m = do.call(earth::earth, pars)
+    set.seed(getOption("mlr.debug.seed"))
+    p = predict(m, newdata = binaryclass.test, type = "response")[,1]
+    old.predicts.list[[i]] = as.factor(ifelse(p > 0.5, levs[1L], levs[2L]))
+  }
+  
+  testSimpleParsets("classif.earth", binaryclass.df, binaryclass.target, binaryclass.train.inds, old.predicts.list, parset.list)
+})
```
schiffner: Please add `testProbParsets`.

tests/testthat/test_classif_earth.R
```diff
@@ -0,0 +1,39 @@
...
+  set.seed(getOption("mlr.debug.seed"))
+  m = earth::earth(formula = binaryclass.formula, data = binaryclass.train, glm = list(family = binomial(link = "logit"), maxit = 50) )
+  set.seed(getOption("mlr.debug.seed"))
+  p = predict(m, newdata = binaryclass.test, type = "response")[,1]
+  levs = binaryclass.class.levs
+  p.class = as.factor(ifelse(p > 0.5, levs[1L], levs[2L]))
+
+  testSimple("classif.earth", binaryclass.df, binaryclass.target, binaryclass.train.inds, p.class,list(maxit = 50))
+  testProb("classif.earth",   binaryclass.df, binaryclass.target, binaryclass.train.inds, p,list(maxit = 50))
+
```
schiffner: Lines 7-16 are not needed.
They are testing classif.earth without any additional hyperparameters set.
But exactly the same is done below for the first element of `parset.list`, which is an empty list.

schiffner:
Hi @rwarnung ,

thanks very much for the updates.

Just to make sure. Did you run *all* the classification tests locally or just the tests you created? I'm asking because the Travis fail happens in https://github.com/mlr-org/mlr/blob/master/tests/testthat/test_learners_all_classif.R, which automatically goes through all classification learners.

rwarnung:
@schiffner you are right - I just checked the tests that I created. Sorry for the confusion - I am just getting used to all this. Thank you for your patience.

schiffner:
Not a problem at all and my apologies. I forgot to mention these other tests in the tutorial, but am correcting this now.

rwarnung:
@schiffner I dived deeper in github and so forth and tested it all locally. Using

```
devtools::load_all()
library(testthat)
library(BBmisc)
test_file("test_classif_earth.R")
test_file("test_learners_all_classif.R",reporter = "stop")
```

In test_classif_earth.R I added `testProbParsets` and removed the first lines as you suggested. Furthermore I added more interesting parsets. The tests are succesfull but I get the usual warnings of probabilities that equal 0 or 1.

In test_learners_all_classif.R I added some parameters for the learner. The test fails but I think this is not my learner's fault but due to xgboost. The test call gives:

```
> test_file("test_learners_all_classif.R",reporter = "stop")
 Show Traceback
 
 Rerun with Debug
 Error: Test failed: 'learners work: classif '
* unused argument (special.vals = list(NA, NA, NULL))
1: mylist(task, create = TRUE) at test_learners_all_classif.R:26
2: listLearners(..., create = create) at C:\Users\Richard\Google Drive\R\mlrdevel\mlr\tests\testthat\helper_helpers.R:219
3: listLearners.Task(..., create = create) at C:\Users\Richard\Google Drive\R\mlrdevel\mlr/R/listLearners.R:103
4: listLearners.character(td$type, union(props, properties), quiet, warn.missing.packages, check.packages, create) at C:\Users\Richard\Google Drive\R\mlrdevel\mlr/R/listLearners.R:158
5: lapply(tab$id[tab$installed], makeLearner) at C:\Users\Richard\Google Drive\R\mlrdevel\mlr/R/listLearners.R:127
6: FUN(X[[i]], ...)
7: do.call(constructor, list()) at C:\Users\Richard\Google Drive\R\mlrdevel\mlr/R/makeLearner.R:67
8: (function () 
   {
       makeRLearnerClassif(cl = "classif.xgboost", package = "xgboost", par.set = makeParamSet(makeDiscreteLearnerParam(id = "booster", 
           default = "g 
```
Is my learner ok so far or should I do some more work on it at the moment? Thanks you and @larskotthoff for your patience. Kind regards.

larskotthoff:
Could you rebase on mlr master please? That should fix the xgboost tests.

rwarnung:
@schiffner and @larskotthoff as it seems my learner passed the tests now. Please let me know if there is any additional work that should be done. Kind regards

larskotthoff:
Thanks. I've had a look through the parameters and fixed some small things (default values, lower bounds). It looks like you've missed quite a few parameters though -- could you add them please?

rwarnung:
@larskotthoff thanks for adding the defaults and bounds! Concerning 'missed a few parameters' you mean in the definition of the learner or in the test (should each parameter be addressed in the test?)? So I should add parameters to the parsets, right? Thank you!

larskotthoff:
In the learner. All the parameters that are documented should be available through mlr. Thanks!

rwarnung:
I have to check why this has failed on my home system.

rwarnung:
@larskotthoff sorry for the delay .. had a mix-up with files. Now I added the parameters and the checks were successful. So far this is a binary classifier. I can try to make it work for binary and multiclass ... I will see what I can do.

larskotthoff:
Sounds good. I'll hold off merging then until you've had a chance to do that. Thanks!

rwarnung:
@larskotthoff I implemented the multiclass functions. The checks fail. I think the reason is that I use `requirePackagesOrSkip("earth", default.method = "load") ` . The error is that the object `'contr.earth.response' of mode 'function' was not found`. If I load the package using `library `or `require` then it works. Do you have an idea how I can fix this? I call `earth:earth` in the tests. If this is fixed then the learner should be ready.

larskotthoff:
Could this be caused by different earth versions locally and on the build systems?

rwarnung:
@larskotthoff I changed the method from `load `to `attach` then it worked. I can not explain this behaviour. The earth package is not the problem as the tests failed locally too. Now I wait for the checks to complete.

larskotthoff:
Ah yes, "load" doesn't attach the objects. Where exactly is `contr.earth.response` used?

rwarnung:
It is internally used in the call of `earth` and it is only needed for the multiclass classifier. Help says: "Contrasts function for factors in the earth response. For internal use by earth."

rwarnung:
@larskotthoff the checks have passed ...maybe the learner is ready :)

larskotthoff:
So this prevents it from working when using it as a normal classifier:

```
lrn = makeLearner("classif.earth")
m = train(lrn, iris.task)
Error in get(ctr, mode = "function", envir = parent.frame()) : 
  object 'contr.earth.response' of mode 'function' was not found
```

They should either export that or prefix with `earth::`.

For now, could you please remove/comment the multiclass stuff so we can merge?

rwarnung:
@larskotthoff good point, that the learner itself doesn't work because of this problem. I didn't realize this during my local tests. Now I removed the multiclass part of the learner and commented the multiclass test. Let's wait for the build tests to complete then the binary version is (again ;)) ready. Thanks for your help!

larskotthoff:
Thanks, merging. Could you open an issue/bug upstream about the multiclass thing so we can get this working here as well eventually please?

rwarnung:
@larskotthoff I opend up this issue https://github.com/mlr-org/mlr/issues/1387 as you can see about the multiclass problem with earth.

rwarnung:
@larskotthoff Hi Lars I took Bernd's advice and the learner with the multiclass functinality passes the local tests. How can I reopen this pull request?

larskotthoff:
This one is merged and can't be reopened. Open a new one please.

Issue #1244, opened 2016-09-15
Predictions for (regr.)blackboost differ


MariaErdmann:
I just checked regr.blackboost so far, but I think this applies also for the other blackboost learners.
Tests do not fail because they work around the "bug" (they do not test with an empty list of params):
Mini-example:

```
set.seed(760)
d = getTaskData(regr.task)
orig.mod = mboost::blackboost(medv ~ ., data = d)
orig.pred = predict(orig.mod, newdata = d)

lrn = makeLearner("regr.blackboost")
mod = train(lrn, regr.task)
pred = predict(mod, regr.task)

# Predictions are different
all.equal(as.vector(orig.pred), pred$data$response)


orig.mod.b = mboost::blackboost(medv ~ ., data = d, tree_controls = party::ctree_control(maxdepth = 0))
orig.pred.b = predict(orig.mod.b, newdata = d)

# now they are equal
all.equal(as.vector(orig.pred.b), pred$data$response)

```

Mlr's default settings are not equal to blackboost's default settings, because I need to explicitely set the maxdepth argument. Looking at the blackboost code, I thought maxdepth = 2 is the default, but changing the default to 2 in mlr, did not change the fact that predictions are different.

So I have the following questions:
Is the difference that occurs here, really due to different default settings?
If yes, I don't know where to look for the "real" tree_controls default settings.
I have the same problem also with the other control parameters of blackboost, when I implemented more families (also for count data). Then the only way to get equal results (in the tests) is by specifiying all parameters in the par.vals of mlr and in tree_controls, as well as in boost_control.

Does someone have an idea what is wrong here?

MariaErdmann:
Talking to @berndbischl, the reason why this is happening is as follows:

If one calls (normal) blackboost without specifying any parameters then the, parameters for tree_controls are like that:
````
blackboost(formula, data = list(),
           weights = NULL, na.action = na.pass,
           tree_controls = party::ctree_control(
               teststat = "max",
               testtype = "Teststatistic",
               mincriterion = 0,
               maxdepth = 2, savesplitstats = FALSE),
           ...)
````

But these are not the default values of party::ctree_control, which are:
````
ctree_control(teststat = c("quad", "max"), 
              testtype = c("Bonferroni", "MonteCarlo", 
                           "Univariate", "Teststatistic"), 
              mincriterion = 0.95, minsplit = 20, minbucket = 7, 
              stump = FALSE, nresample = 9999, maxsurrogate = 0, 
[...]

Issue #1042, opened 2016-08-01
mboost has a crossvalidation function in the fitting process


ja-thomas:
So mboost (and consequently glmboost and gamboost) does have this `m` parameter to optimize the number of iterations via CV or AIC while training the model.

This is kind of a special case and I'm not sure if we really want that to be possible. 

Should we remove that possibility, so that we are consistent where we train and where we tune models?

ja-thomas:
@MariaErdmann can you fix this in #1023

PR #1322, opened 2016-11-04
Introduce new mboost families, fix bug #1244, solve #1042

MariaErdmann:
Since my old branch broke unrepairably while trying to rebase and resolve merge conflicts I made this new PR. This is probably also better since the old branch comprised thousands of commits. I shortly summarize what is done in this PR:

- new families  are introduced, i.e. Huber, GammaReg, etc.
- it is also possible to set the family specific family parameters since these are introduced as a learner parameter, i.e. for `family = Binomial` one can set `Binomial.link = "probit"
- custom.families are also possible
- family PropOdds() cannot be implemented yet since this family is for ordered targets
- the blackboost learners had a bug (see #1244) because of their parameter tree_controls
- the learners had a crossvalidation function in the fitting process that was removed (see #1042)


MariaErdmann:
This PR solves following issues:
#1042
#1244 
#662

larskotthoff:
Thanks, looks good. Except that Travis is currently broken (because of old cached package versions), but that should hopefully sort itself out soon. I'll merge as soon as Travis passes.

Issue #1327, opened 2016-11-07
Broken tests generateThreshVsPerf


mllg:
Tests are broken on winbuilder and locally on OSX. Cannot fix it, I don't understand the tests.

```
1. Failure: generateThreshVsPerfData (@test_base_generateThreshVsPerf.R#43) ----
length(XML::getNodeSet(doc, black.line.xpath, ns.svg)) not equal to length(unique(pvs$data$iter)).
1/1 mismatches
[1] 0 - 2 == -2
```

Additional warning:
```{r}
the XPath query has no namespace, but the target document has a default namespace.
This is often an error and may explain why you obtained no results
```

berndbischl:
confirmed also on my (ubuntu) laptop
we need to fix this now, this means master tests are broken
@larskotthoff @zmjones

berndbischl:
how is it possible that travis runs then?

mllg:
Tests are failing. Probably a third party package (ggplot2, XML, ...?) has introduced this one. The last successful test was several days ago.

PR #1328, opened 2016-11-07
fix travis, install survival manually

berndbischl:
should fix our survival / hmisc problems in travis


berndbischl:
works. we can now see the errors reported here. 
https://github.com/mlr-org/mlr/issues/1327

Issue #1319, opened 2016-11-02
Faster performance measures from package Modelmetrics


studerus:
I noticed that package `caret` uses in its newest version the auc measure from package ModelMetrics, because it's programmed with Rcpp. ModelMetrics has many more performance measures all implemented in C++.

I did a quick speed comparison between our implementation of auc and the one in ModelMetrics. The one in ModelMetrics seems to be more than 10 times as fast.

```
library(mlr)
library(ModelMetrics)
library(microbenchmark)

pred <- predict(train(makeLearner('classif.logreg', predict.type = 'prob'), pid.task), pid.task)
prob <- getPredictionProbabilities(pred)
truth <- getPredictionTruth(pred)

microbenchmark(
  ModelMetrics = ModelMetrics::auc(truth, prob),
  mlr = measureAUC(prob, truth, negative = 'neg', positive = 'pos'))
```

```
Unit: microseconds
         expr      min       lq     mean   median       uq      max neval cld
 ModelMetrics  206.438  216.555  233.762  223.446  238.987  376.809   100  a 
          mlr 2657.890 2737.211 3026.499 2899.223 3066.808 5915.154   100   b
```

catastrophic-failure:
`measureAUC` is based on `ROCR`, the rest of the AUC metrics are based on `caTools`'s `colAUC` which I suppose must be faster.

larskotthoff:
Good catch! Could you make a PR for the `ModelMetrics` measures please?

PhilippPro:
`colAUC` is only twice as slow as the `ModelMetrics` package.

```
library(mlr)
library(ModelMetrics)
library(microbenchmark)
library(caTools)

pred <- predict(train(makeLearner('classif.logreg', predict.type = 'prob'), pid.task), pid.task)
prob <- getPredictionProbabilities(pred)
truth <- getPredictionTruth(pred)

microbenchmark(
  ModelMetrics = ModelMetrics::auc(truth, prob),
  mlr = measureAUC(prob, truth, negative = 'neg', positive = 'pos'),
  mlr2 = colAUC(prob, truth))
```
```

Unit: microseconds
         expr      min        lq      mean    median       uq      max neval cld
 ModelMetrics  320.223  338.3465  375.2747  355.4225  413.216  485.817   100 a  
          mlr 3501.354 3589.9820 4306.6350 3743.3890 5121.008 7983.279   100   c
[...]

PR #1344, opened 2016-11-14
colAUC instead of ROCR for AUC calculation.

PhilippPro:
Changed the calculation of the auc from the ROCR packages to our internal colAUC. As outlined in #1319 it is faster and we also diminish our dependency on external packages. 

The code for the calculation is a bit counter intuitive, because the truth vector is set to truth = (truth != positive), so colAUC calculates the correct value, but it works.


larskotthoff:
Thanks. Could you add a test for the new argument to `colAUC` please?

PhilippPro:
I added a test.

larskotthoff:
Great, thanks!

PR #1347, opened 2016-11-18
Closes #1345 extends parameters of ranger learners 

### LINKED ISSUE #1345: ranger: new hyperpars available
I can see at least 3 new parameters:
- `splitrule`: Splitting rule, regression and survival only. For regression one of "variance" or "maxstat" with default "variance". For survival "logrank", "C" or "maxstat" with default "logrank".
- `alpha`: For "maxstat" splitrule: Significance threshold to allow splitting.
- `minprop`: For "maxstat" splitrule: Lower quantile of covariate distribtuion to be considered for splitting.
Maybe there are more, we should extend the param set.

florianfendt:
* I added the new parameters for the split rules
* ````write.forest = TRUE```` is now the default anyway so I adapted our code
* Added parameter ````keep.inbag````
* Extended tests to include new split rule parameters

### DIFF
--- a/R/RLearner_classif_ranger.R
+++ b/R/RLearner_classif_ranger.R
@@ -16,18 +16,19 @@ makeRLearner.classif.ranger = function() {
       makeUntypedLearnerParam(id = "always.split.variables"),
       makeLogicalLearnerParam(id = "respect.unordered.factors", default = FALSE),
       makeDiscreteLearnerParam(id = "importance", values = c("none", "impurity", "permutation"), default = "none", tunable = FALSE),
-      makeLogicalLearnerParam(id = "write.forest", default = FALSE, tunable = FALSE),
+      makeLogicalLearnerParam(id = "write.forest", default = TRUE, tunable = FALSE),
       makeLogicalLearnerParam(id = "scale.permutation.importance", default = FALSE, requires = quote(importance == "permutation"), tunable = FALSE),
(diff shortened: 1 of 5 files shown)


R/RLearner_classif_ranger.R
```diff
@@ -16,18 +16,19 @@ makeRLearner.classif.ranger = function() {
       makeUntypedLearnerParam(id = "always.split.variables"),
       makeLogicalLearnerParam(id = "respect.unordered.factors", default = FALSE),
       makeDiscreteLearnerParam(id = "importance", values = c("none", "impurity", "permutation"), default = "none", tunable = FALSE),
-      makeLogicalLearnerParam(id = "write.forest", default = FALSE, tunable = FALSE),
+      makeLogicalLearnerParam(id = "write.forest", default = TRUE, tunable = FALSE),
```
jakob-r: the default changed indeed

tests/testthat/test_surv_ranger.R
```diff
@@ -1,33 +1,36 @@
 context("surv_ranger")
 
 ## TODO: Proper test required when predictions working
+## Suppress 'experimental' warning for splirule = maxstat
```
jakob-r: IMHO we should leave out the test for splitrule = "maxstat" as this is experimental.

jakob-r:
Argh. I still don't get this review system completely. Can you remove the tests for splitrule == "maxstat"? If we want to test other parameter settings then just ntree there are a lot of other options.

larskotthoff:
Thanks, merged.

Issue #1299, opened 2016-10-20
BenchmarkResult plots broken!


giuseppec:
``` r
set.seed(1)
lrn = list(
  makeLearner("classif.rpart"), 
  makeBaggingWrapper(makeLearner("classif.rpart", cp = 0), bw.iter = 100), 
  makeLearner("classif.randomForest", ntree = 100)
)
bmr = benchmark(lrn, binaryclass.task)

plotBMRRanksAsBarChart(bmr)
plotBMRBoxplots(bmr)
plotBMRSummary(bmr)
```
1. The legend does not show the names for wrapped learners in all three plots (however, it seems to work when `pretty.names = FALSE`)
2. The colours/ranks for `plotBMRRanksAsBarChart` seem to be wrong, i.e., randomForest seems to be rank 3 although it has the lowest mmce value.

giuseppec:
Another strange observation:

The `pretty.names` arg for `plotBMRBoxplots` does something completely different than the `pretty.names` arg for the other BMR plots (i.e., it prettifies the measure and not the learner names). Although it is correctly documented I asked myself why we don't simply also "prettify" the learner names here?

berndbischl:
@giuseppec 
can you try to fix this in a PR pls?
@larskotthoff @zmjones 
can we somehow unit test this now? please say yes....

larskotthoff:
Yes, we can test this by checking that the learner names (or IDs, or whatever) appear in the generated SVG. We can't easily check whether they appear in the correct place (which shouldn't be an issue here).

giuseppec:
The problem with the first issue is that we don't have a `short.name` slot for wrapped learners. Should we define a `getLearnerShortName` to solve this?

schiffner:
What would this `getLearnerShortName` do exactly?

giuseppec:
For a simple learner it would just return the `learner$short.name` . For a (multiple) wrapped learner, it would successively attach the additional string of the wrapper (which is also attached to the learner id, e.g., `.bagged` for the bagging wrapper and `.imputed` for the impute wrapper etc.).

Another alternative would be to modify each single mlr wrapper and add a `$short.name` slot to all wrappers. Due to laziness, I would prefer the `getLearnerShortName` solution.

giuseppec:
[...]

PR #1354, opened 2016-11-23
Fixes pretty.names argument for benchmark plots. Addresses #1299

florianfendt:
This should fix the issue we had with level reordering when the pretty.names argument was used
(specifically when used together with order.lrns)

Side question:
Why don't we have the order.lrns argument for plotBMRSummary? It would have the same effect as
in plotBMRRanksAsBarCharts. E.g: Only change ordering of legend labels.
Should I go ahead and add it? Maybe in a different PR though I guess.

### DIFF
--- a/R/plotBMRBoxplots.R
+++ b/R/plotBMRBoxplots.R
@@ -33,7 +33,13 @@ plotBMRBoxplots = function(bmr, measure = NULL, style = "box", order.lrns = NULL
   df = orderBMRTasks(bmr, df, order.tsks)
 
   if (pretty.names) {
-    levels(df$learner.id) = getBMRLearnerShortNames(bmr)
+    learner.short.names = getBMRLearnerShortNames(bmr)
+    if (!is.null(order.lrns)) {
+      learner.ids = getBMRLearnerIds(bmr)
+      names(learner.short.names) = learner.ids
+      learner.short.names = learner.short.names[order.lrns]
+    }
+    levels(df$learner.id) = learner.short.names
   }
 
   p = ggplot(df, aes_string("learner.id", measure$id))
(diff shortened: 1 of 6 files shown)


larskotthoff:
Thanks. Could you add a test that checks whether the issue is fixed please? Re the `order.lrns` argument: Sounds good, I would do this in a separate PR.

florianfendt:
How would I test this? Can we test this with our XML tests?
Right now I could only think of looking in the data of the ggplot object. E.g, do sth like this:
````
lrns = c("classif.randomForest", "classif.rpart")
bmr = benchmark(lrns, iris.task)
new.order = c("classif.rpart", "classif.randomForest")
p = plotBMRRanksAsBarChart(bmr, order.lrns = new.order)
new.learners = lapply(new.order, makeLearner)
new.order = extractSubList(new.learners, "short.name")
expect_equal(levels(p$data$learner.id), new.order)
````

larskotthoff:
You can check the generated plot directly, see the `testDocForStrings` function.

florianfendt:
Yes, I wrote that function ;)
But can we be sure that the short.names appear in the right order in the XMLNodeSets?

larskotthoff:
Ah, sorry :)

Yes, they should appear in the order in which they appear in the legend (top to bottom, left to right).

florianfendt:
Ok perfect. Thanks!
I will do this then now.

florianfendt:
@berndbischl 
I ping you here since you wanted this asap. I think we are done here and I'll add the order.lrns argument to plotBMRSummary in a different PR.

larskotthoff:
Thanks, merging.

PR #1356, opened 2016-11-24
use ModelMetrics to speed up auc calc

JackStat:
Using https://github.com/JackStat/ModelMetrics for AUC calculation.  If data is large it can cut the time in half. Although it is pretty quick already

    N = 2000000
    actual = as.numeric(runif(N) > .5)
    predicted = as.numeric(runif(N))
    
    (s1 <- system.time(a1 <- auc(actual, predicted)))
    (s2 <- system.time(a2 <- measureAUC(predicted, actual, 0, 1)))

    #> s1
    #   user  system elapsed 
    #  0.927   0.014   0.947 
    #> s2
    #   user  system elapsed 
    #  1.762   0.117   1.885


larskotthoff:
Thanks for proposing this! We actually had a look ourselves in #1319. We concluded that for very large datasets our current implementation is actually faster (and doesn't require another package). We're happy to be convinced, but could you demonstrate on some more benchmarks please?

JackStat:
That's hilarious.  I didn't even know.  Let me dig in a little more

PR #1362, opened 2016-11-27
check for sampled training task in resampling (fixes #1357)

### LINKED ISSUE #1357: generateLearningCurve evaluates on full training set instead of subsample
Original source: https://stackoverflow.com/questions/40765180/r-mlr-creating-learning-curve-from-subset-of-training-data-and-whole-test-data

It looks like during the generation of the learning curve the model is evaluated on the entire training data instead of the subsample taken during that iteration.

larskotthoff:
Turns out that this affected all wrappers that change the training task. I've fixed this in resampling, but I'm not sure if this needs to be addressed elsewhere.

@berndbischl @mllg


larskotthoff:
@berndbischl  @mllg Please have a look and merge -- this is a rather serious bug.

larskotthoff:
Merging...

PR #1364, opened 2016-11-28
cindex.uno for survival

mllg:


mllg:
Would be great if someone familiar with survival analysis could have a look.

studerus:
Thanks. Looks ok to me.

PhilippPro:
Maybe add a reference, where it also can be read what are advantages and disadvantages (compared to the normal c-index).

studerus:
Here's an interesting paper in which they compared the performance of different cindexes:
[A comparison of estimators to evaluate the discriminatory power of time-to-event models](dx.doi.org/10.1002/sim.5464)
The Uno-CIndex performed better than Harrell's CIndex, but the best one seems to be the one proposed by Heagerty and Zheng in this paper.
[Survival Model Predictive Accuracy and ROC Curves](dx.doi.org/10.1111/j.0006-341X.2005.030814.x).
I'm not sure if this is implented in a R package already.

adibender:
The `survAUC` package implements various methods (including the approach proposed by Heagerty and Zheng): https://cran.r-project.org/web/packages/survAUC/

larskotthoff:
Thanks. Could you add a test that checks that the computed value is correct on a small example please?

mllg:
```r
library(survC1)
library(survAUC)

set.seed(1)
data = getTaskData(wpbc.task)
train = sample(nrow(data), 2/3 * nrow(data))
data = list(train = data[train, ], test = data[-train, ])
S = list(train = Surv(data$train[["time"]], data$train[["status"]]), test = Surv(data$test[["time"]], data$test[["status"]]))

mod = coxph(Surv(time, status) ~ ., data = data$train)
y = unname(predict(mod, newdata = data$test))

tmp = cbind(S$test, y)
survC1::Est.Cval(mydata = tmp, tau = max(tmp[, 1L]), nofit = TRUE)$Dhat
survAUC::UnoC(S$train, S$test, y)
```
I get `0.4763045` and `0.2985769`. This isn't even close?

studerus:
Really strange. I would write to the authors of the packages.

mllg:
We agreed on using some of the implementations in `survAUC`.

PR #1368, opened 2016-11-29
Fix problem with integer overflow in measure mcc

studerus:
Fixes the following problem:
```
holdout('classif.logreg', oversample(pid.task, 5), measures = mcc)
````

```
[Resample] holdout iter 1: mcc.test.mean=  NA
[Resample] Aggr. Result: mcc.test.mean=  NA
Resample Result
Task: PimaIndiansDiabetes-example
Learner: classif.logreg
Aggr perf: mcc.test.mean=  NA
Runtime: 0.0150108
Warning message:
In (tp + fp) * (tp + fn) * (tn + fp) * (tn + fn) :
  NAs produced by integer overflow
```

### DIFF
--- a/R/measures.R
+++ b/R/measures.R
@@ -701,7 +701,7 @@ measureWKAPPA = function(truth, response) {
   expected.mat = rowsum %*% t(colsum) 
 
   # get weights
-  class.values = as.numeric(levels(truth))
+  class.values = seq_along(levels(truth)) - 1L
   weights = outer(class.values, class.values, FUN = function(x, y) (x - y)^2)
   
   # calculate weighted kappa
--- a/R/measures.R
+++ b/R/measures.R
@@ -1041,10 +1041,10 @@ mcc = makeMeasure(id = "mcc", minimize = FALSE,
 #' @rdname measures
 #' @format none
 measureMCC = function(truth, response, negative, positive) {
-  tn = measureTN(truth, response, negative)
-  tp = measureTP(truth, response, positive)
-  fn = measureFN(truth, response, negative)
-  fp = measureFP(truth, response, positive)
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks. Could you add a test for this please?

studerus:
Should the test be only for this specific measure or all measures?

larskotthoff:
For all measures would be great.

studerus:
@larskotthoff 
I added a test for all classification measures and thereby found another bug in wkappa, which I have also repaired. I don't know why one check was not successfull. I ran through locally. Can you help me here?

larskotthoff:
Looks like the check failed on the manual -- nothing related to this PR, ignore that.

It looks like you've changed the definition of weighted kappa. This is probably what it should have been to start with given the description, but could you please add a test that checks the value in a case where the two definitions would have given different results?

studerus:
I don't think I changed the definition of wkappa. In the wkappa test, the factor levels were "0" and "1", which were converted by `as numeric` to 0 and 1 and therefore still gave the correct result. However, if the factor levels are called differently, like `pos` and `neg` as in my new test, it is converted to `NA` and that's why the test failed. I therefore just changed it to seq_along(levels(truth)) - 1 to get 0 and 1 regardless of the names of the factor levels. I also compared the results of the updated wkappaMeasure with the result of the wkappa function of package psych and is was the same.

larskotthoff:
If the factor levels are "0", "10", and "100", the results will be different. I don't think this was intended in the old implementation and you've fixed that bug, but it would be great to have a test for it to make sure that it doesn't happen again.

larskotthoff:
Thanks, merging.

PR #1371, opened 2016-12-04
Allow Rfast filters to deal with missing data

studerus:
This should restore the behavior of the previous implementation with FSelector functions.


larskotthoff:
Thanks. Could you write a test for this as well please?

larskotthoff:
Thanks, merging.

PR #1391, opened 2016-12-18
closes #1147 createdummyfeatures

### LINKED ISSUE #1147: Refactor createDummyFeatures / add createDummyFeatures for single vectors
We could use a `createDummyFeature` function internally that just works on a vector (factor or character). 

There are for example a some measure that do something like `model.matrix( ~ . -1, data = as.data.frame(truth))`

If we want to do that correctly we should use the same function for the `createDummyFeature` on data.frames and tasks.

Coorsaa:
createDummyFeatures works now for vectors (numerics and characters). Hence this method is now used within the methods for data.frames and tasks.

Suggestion for NEWS
````
createDummyFeatures supports vectors now
````


larskotthoff:
Thanks. There are some warnings for S3 method consistency, could you have a look please?

larskotthoff:
Thanks, merging.

PR #1396, opened 2016-12-19
fix 1355, smote with integers

### LINKED ISSUE #1355: smote throws error if all variables are integers
It seems that the smote function cannot deal with data sets in which all predictor variables are of type integer:
```
dat <- getTaskData(pid.task)
i <- sapply(dat, is.numeric)
dat[,i] <- lapply(dat[,i], as.integer)
tsk <- makeClassifTask(data = dat, target = "diabetes")
smote(tsk, 2)
```
```
Error in smote(tsk, 2) : 
REAL() can only be applied to a 'numeric', not a 'integer'
```

ja-thomas:
Fixes #1355

Problem is that if all variables in a task are integer, `x.min.matrix` will become an integer matrix but `c_smote` expects a double matrix.

### DIFF
--- a/R/smote.R
+++ b/R/smote.R
@@ -76,6 +76,8 @@ smote = function(task, rate, nn = 5L, standardize = TRUE, alt.logic = FALSE) {
     }
   }
   x.min.matrix = as.matrix(x.min.matrix)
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, merging.

Issue #1403, opened 2016-12-21
Problem with plotHyperParsEffect


studerus:
The following minimal example does not work:

```
ps = makeParamSet(makeNumericParam("C", lower = 0.01, upper = 0.1))
gr = makeTuneControlGrid(resolution = 2)
res = tuneParams('regr.ksvm', bh.task, cv2, rmse, ps, gr)
dat = generateHyperParsEffectData(res)
plotHyperParsEffect(dat, x = "C", y = "rmse.test.rmse", plot.type = 'line')

# Error in eval(expr, envir, enclos) : object 'rmse.test.rmse' not found 
```

larskotthoff:
@MasonGallo Removing the `eval` in the pretty names stuff fixes this. Any reason the `eval` is needed (and it's probably not a good idea anyway...)?

MasonGallo:
The `eval` was used to convert the string input into an object so we can pull out the "pretty name". The bug could be fixed by changing the splitting criteria in pretty names or removing eval if there is a better way to convert the string input.

I am starting to wonder if pretty.names is useful API design in general for this context rather than just having the user edit it to what they would like. Maybe it should default to False.

larskotthoff:
Another option would be to remove this argument altogether and tell users in the documentation how they can set this themselves on the returned ggplot object. It would certainly simplify the code a lot.

MasonGallo:
Yes, that's where I was heading. I'm always happy to reduce code complexity. I think there could be a note in the documentation and a reference in the tutorial as a reminder to users. Should be quick I can prob get to this later today or tmw.

PR #1408, opened 2016-12-24
#1403 remove pretty.names

MasonGallo:
Per discussion in #1403 this removes `pretty.names` argument from `plotHyperParsEffect`. This reduces API complexity for something should probably be edited by the user anyway. I added a note in the documentation to remind users.

This should be ready when green.


larskotthoff:
Thanks, looks good. Merging.

PR #1409, opened 2016-12-25
Allow ranger to use observation weights

### LINKED ISSUE #1406: case.weights in ranger
Hi there,

It seems like the case.weights param isn't implemented yet. Would it be really hard to add this? It is super useful if you have unbalanced classes. You can set the class weights to the inverse of their frequencies, which ensures that during the bootstrap sampling the minority class(es) are upsampled. If I'm not mistaken this is what scikit-learn's RF is doing with its class_weight parameter.

Thanks a lot!
d

studerus:
Fixes #1406

### DIFF
--- a/R/RLearner_classif_ranger.R
+++ b/R/RLearner_classif_ranger.R
@@ -25,7 +25,7 @@ makeRLearner.classif.ranger = function() {
       makeLogicalLearnerParam(id = "keep.inbag", default = FALSE, tunable = FALSE)
     ),
     par.vals = list(num.threads = 1L, verbose = FALSE, respect.unordered.factors = TRUE),
-    properties = c("twoclass", "multiclass", "prob", "numerics", "factors", "ordered", "featimp"),
+    properties = c("twoclass", "multiclass", "prob", "numerics", "factors", "ordered", "featimp", "weights"),
     name = "Random Forests",
     short.name = "ranger",
     note = "By default, internal parallelization is switched off (`num.threads = 1`), `verbose` output is disabled, `respect.unordered.factors` is set to `TRUE`. All settings are changeable."
@@ -33,10 +33,10 @@ makeRLearner.classif.ranger = function() {
 }
 
 #' @export
-trainLearner.classif.ranger = function(.learner, .task, .subset, .weights, ...) {
+trainLearner.classif.ranger = function(.learner, .task, .subset, .weights = NULL, ...) {
   tn = getTaskTargetNames(.task)
   ranger::ranger(formula = NULL, dependent.variable = tn, data = getTaskData(.task, .subset),
-    probability = (.learner$predict.type == "prob"), ...)
+    probability = (.learner$predict.type == "prob"), case.weights = .weights, ...)
 }
 
 #' @export
--- a/tests/testthat/test_classif_ranger.R
+++ b/tests/testthat/test_classif_ranger.R
@@ -22,4 +22,11 @@ test_that("classif_ranger", {
   }
 
   testProbParsets ("classif.ranger", binaryclass.df, binaryclass.target, binaryclass.train.inds, old.probs.list, parset.list)
+  
+  #test that weights work as expected
+  lrn1 = makeLearner("classif.ranger", num.trees = 200L)
+  lrn2 = makeWeightedClassesWrapper(lrn1, wcw.weight = 2)
+  tpr1 = holdout(lrn1, pid.task, measures = tpr)$aggr
+  tpr2 = holdout(lrn2, pid.task, measures = tpr)$aggr
+  expect_true(tpr1 < tpr2)
 })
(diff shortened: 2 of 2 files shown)


larskotthoff:
Thanks, looks good. Could you add a test for the weights please?

studerus:
Isn't that covered by the general weights test already?

larskotthoff:
Yes, but it would be good to have a test that checks that weights make a difference here. I.e. test that it does what's expected, not only that it works technically.

studerus:
Where should this test be added? I think it would be inconsistent to only test this for classif.ranger but not for other learners that support weights.

larskotthoff:
Just add it in the test for the learner. Pull requests that test this for other learners welcome :) Thanks!

jakob-r:
To be fair, I could not find any learner that has been tested on weights specifically. So we could add the
>test that it does what's expected, not only that it works technically.
part globally and not only for `classif.ranger`.

Apart from that I thin it can be merged.

berndbischl:
hi,
thx for the fix, but why is nobody reading the docs in tests/helpers_learners_all

```
# test that a given learner respects its weights tag. we do this:
# train without weights, with weights = 1, and with changed weights
# then we check that changed weights actually change the output
# note: it is not always easy to find a good data and weights combo to reliably achieve this,
# as for some learners we need very extreme weights and for others exteme weights cause error.
#
# further args: train and test.inds, and weight vec for train inds
# we can also set pred.type and the getter for the output col from the preds.
testThatLearnerRespectsWeights = function(lrn, task, train.inds, test.inds, weights, hyperpars,

```

@studerus is IMHO completely right that this should be tested in general, and @larskotthoff is correct that we should test whether weights change the results - but i already coded that ...?

jakob-r:
So in conclusion(?): We can remove the weights specific test again?

berndbischl:
> So in conclusion(?): We can remove the weights specific test again?

yes. and RTFM ;-)

larskotthoff:
Thanks, merging (although Travis timed out for some reason...).

@berndbischl Exactly what part of the manual was I supposed to have read?

berndbischl:
> @berndbischl Exactly what part of the manual was I supposed to have read?

>>>> thx for the fix, but why is nobody reading the docs in tests/helpers_learners_all

berndbischl:
thats why we (i think including you?) were eager to have better documentation headers in the helper_learner_all file?

if you still think this is not clear enough, the docs, I mean, make a proposal to make this better.

larskotthoff:
:)

My point is that you need to read the source itself to figure this out. It's not in any of the other documentation.

berndbischl:
> , I mean, make a proposal to make this better.

larskotthoff:
Document all of these functions that you implemented at some point in the dev wiki.

berndbischl:
I dont like that, that directly results in copy paste. 
It would be much easier to simply summarize in the dev wiki that such tests exists, what their goal is and that one should look into the mentioned R test file for specifics?

danielhomola:
Thanks a lot for adding this! I installed the development version of mlr today, but when I try to do 
`lrn <- setHyperPars(lrn, case.weights=target_iw)`
I still get:
_classif.ranger: Setting parameter case.weights without available description object!_

`devtools::install_github("mlr-org/mlr")` will install from the `master` by default and #1409 was merged 5 days ago. How come I still get an error? Am I missing something? Thanks a lot for your help, and I apologize in advance if I'm missing something really obvious.. I'm still getting used to the R package landscape..

schiffner:
Hi Daniel,

In order to train ranger with observation weights, just do the following:
```
train("classif.ranger", iris.task, weights = 1:150)
```
Internally, the `weights` will be passed on to the `case.weights` argument of `ranger`.

Analogously, `resample` has a `weights` argument. Or alternatively, you can set weights when creating the learning task.

Best,
Julia

danielhomola:
Thanks a lot Julia for the quick answer!

Just out of curiosity could I use the weights argument of my task to ensure stratified resampling? Let's say I have a binary outcome variable and a classification task with n=100 samples, where the frequency of class 0 is 0.7 while the frequency of the positive class is 0.3. Let's also presume that the first 70 rows hold the class 0 and the last 30 the class 1 data. Could I set the weights with c(rep(0.7, 70), rep(0.3, 30)) to ensure stratified resampling? 

Many thanks,
d

larskotthoff:
How exactly the weights are taken into account depends on the learner. If you want stratified resampling, do that and don't try to emulate it through weights.

berndbischl:
@danielhomola 
Lars is basically rights. 

a) The "weights" arg does the following. The numerical vector you specify here will simply be passed down to the basic learner, (maybe indexed a bit during resampling, if you specify it in the task). What happens in the learning algorithm itself is the outside of mlr, but it simply refers to that single learner fit and is nearly always simply a reweighted  model fit / reweighted loss minimization.
(Well actually if it would be something else we would give the learner the property "weights" and pass the weights down, because thats the intended semantics of the mlr "weights" handling....)

b) Stratification in resampling is something orthogonal. Simply set it up independently. So I guess in your case simple say "stratify = TRUE" in makeResampleDesc.

HTH

PR #1415, opened 2017-01-05
Added regr and classif evtree and tests

catastrophic-failure:
I saw `evtree` in the potential learners in the Wiki, so I decided to implement it. The tests were mostly copied from `cforest`, so expect some changes.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -75,6 +75,7 @@ Suggests:
     elasticnet,
     elmNN,
     emoa,
+    evtree,
     extraTrees,
     flare,
     fields,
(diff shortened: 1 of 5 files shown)


catastrophic-failure:
I've tracked the error locally, have to delve into `testSimple` to understand what's going on. The last commit won't solve it yet.

larskotthoff:
Thanks. Where did you get the parameters from? They don't seem to be documented.

catastrophic-failure:
> Thanks. Where did you get the parameters from? They don't seem to be documented.

Yes, the documentation of this library is lacking several important bits. I'm using the defaults defined in the functions `evtree` and `evtree.control`.

larskotthoff:
Ok, thanks. Merging.

PR #1419, opened 2017-01-09
adds test for print.ParamSet. Closes #1352

### LINKED ISSUE #1352: Unit test: we need a print for each ParamSet for a learner
We had a bug here https://github.com/mlr-org/mlr/issues/1292

Such a test will catch this.

we can do this in testThatLearnerCanTrainPredict, maybe the name of the function could be generalized then.

florianfendt:
Added a test for the ParamSet printer in ``testThatLearnerCanTrainPredict`` and generalized the function name to ``testBasicLearnerProperties``.
Other name suggestions for the function?
We do the following in this function:
* check printer for learner and ParamSet
* check train/predict
* check predict types se and prob


larskotthoff:
Thanks. Could you update the documentation for that function as well please?

berndbischl:
otherwise looks good

larskotthoff:
Thanks, merging.

PR #1084, opened 2016-08-08
getOutOfBagPredictions

PhilippPro:
This is a new PR for #902, created a branch, so all can work on it. 

I just transformed it, so that a standardized "predict" object is provided, so that e.g. performance can be applied on the resulting object.

Further development could be changing it, so it is usable for Tuning and Bagging Wrapper.

Code example:

```

 set.seed(123)
 training.set = sample(1:150, 50)
 lrn = makeLearner("classif.ranger", predict.type = "prob", predict.threshold = 0.6)
 mod = train(lrn, sonar.task, subset = training.set)
 oob = getOutOfBag(mod, sonar.task)
 oob
 performance(oob, measures = list(auc, mmce))

```

Renamed some of the functions. 

Still not completely done, e.g.:

> This generates NAs if there are NAs in t.col.

PR #1420, opened 2017-01-09
Update SE estimation of random forest

jakob-r:


jakob-r:
We will first merge this with [getOutOfBagPredictions](https://github.com/mlr-org/mlr/pull/1084/) and then into master.

Issue #1246, opened 2016-09-16
mlbench pkg should be escaped in examples, eventually also in tests


jangorecki:
Hi,
Just came to inform that suggested packages used in examples should be escaped, so not mandatory to run examples. Best way is to use `if (requireNamespace(.)) {_actual_code_}` function. Otherwise you should move `mlbench` (and potentially others, `mlbench` was first to fail) to imports.

```
* checking examples ... ERROR
Running examples in ‘mlr-Ex.R’ failed
The error most likely occurred in:

> ### Name: makeClassifTask
> ### Title: Create a classification, regression, survival, cluster,
> ###   cost-sensitive classification or multilabel task.
> ### Aliases: ClassifTask ClusterTask CostSensTask MultilabelTask RegrTask
> ###   SurvTask Task makeClassifTask makeClusterTask makeCostSensTask
> ###   makeMultilabelTask makeRegrTask makeSurvTask
> 
> ### ** Examples
> 
> library(mlbench)
Error in library(mlbench) : there is no package called ‘mlbench’
Execution halted
```

The best way to check is remove all suggested deps and try R CMD check. More details in [R-exts](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Package-Dependencies).

berndbischl:
Hi Jan,

1) As this is simply a recommendation, and not required by checks with "--as-cran", and would create considerable work on our side: 
Could you please provide some context why you want us to change this?

2) Please note that nothing what you have suggested, as a "workflow solution", really does work so nicely. We need to ensure this then for all future versions of mlr. 

a) Importing such packages (needed in examples) is inappropriate. I am not enlarging the dependency chain of the package for *everybody* so some persons can run some examples in an slightly easier fashion.

b) > the best way to check is remove all suggested deps and try R CMD check

Sure. To do this once. Which is still annoying. In order to do this properly, and to ensure that this does not happen again in the future, we basically have to create a custom Jenkins job now, just for this thing.

3) I will not be able (whatever the result of this discussion) change this for the upcoming release. So the first possible potential fix for this would be in 4-6 weeks.

@larskotthoff  @mllg

jangorecki:
1.
[...]

PR #1422, opened 2017-01-09
Escape library(.) calls in examples. Closes 1246

florianfendt:


jakob-r:
Can you write shortly why this is necessary? In my eyes the examples look a bit cumbersome. If the library() call gives troubles I would try 
```r
#'@examples
#'\dontrun{
#'library(mlbench)
#'}
#'makeTask(...)
```

florianfendt:
@jakob-r 
Summary/Example of/for #1246 
- we import package* in examples
- Since the package is only listed in Suggests it is not verified during CRAN check that package* is available during submission to CRAN and we end up with an error.
- Someone wants to run examples and hasn't got package* or it isn't even available on the users platform.
- The ``requireNamespace`` call is the recommended way of handling such situations. 
[R-exts](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Package-Dependencies) (See 1.1.3.1 Suggested packages)

jakob-r:
@florianfendt Thanks for clearing me up.

PR #1432, opened 2017-01-16
fix regr.laGP: set g = 1/10000 (see learner bug list)

Coorsaa:
moreover removed unnecessary white spaces


larskotthoff:
Thanks, merging.

Issue #1370, opened 2016-12-02
Learner bug list: Test learners with defaults against no params


MariaErdmann:
Wrt to #1034 and #711 I implemented the following test and looked at all the errors. Since the test throws quite a lot of errors I will use this issue to make a list with all the errors that I get and which need to be fixed in order to get the test run through. We thought it would be better to fix the problems in seperate PRs.  What do you think?

The test is not finished yet since I do not know yet how to handle default parameters that have a require statement. But I considered this case when starting debugging the first errors.

````
testThatLearnerHasCorrectDefaults = function(lrn, task) {
  
  rin = makeResampleInstance("Holdout", task = task)
  set.seed(1); m1 = train(lrn, task, subset = rin$train.inds[[1]])
  lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))
  set.seed(1); m2 = train(lrn2, task, subset = rin$train.inds[[1]])
  
  p1 = predict(m1, task, subset = rin$test.inds[[1]])
  p2 = predict(m2, task, subset = rin$test.inds[[1]])
  expect_equal(performance(p1), performance(p2))
}
````

MariaErdmann:
Classification learners
- [ ] classif.avNNet: 

````
lrn = makeLearner("classif.avNNet", par.vals = list(repeats = 5))
mod = train(lrn, binaryclass.task)
lrn = makeLearner("classif.avNNet", par.vals = list(bag = FALSE))
mod = train(lrn, binaryclass.task)

# Problem is this part in the learner's file line 36 - 40
# ind = stri_detect_regex(nms, "repeats") --> produces a vector with one TRUE and rest is FALSE
# if (sum(ind)>0) --> is TRUE
# repeats = .learner$par.vals[[ind]] --> one [] is sufficient
# ind = stri_detect_regex(nms, "bag")
# if (sum(ind)>0)
#  bag = .learner$par.vals[[ind]]
```` 

- [ ] classif.glm

````
lrn = makeLearner("classif.binomial")
lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))
mod2 = train(lrn2, binaryclass.task)

# Problem is here: we need to add model into the signature of the train function
````

- [ ] classif.blackboost

````
lrn = makeLearner("classif.blackboost")
mod = train(lrn, binaryclass.task)
lrn2 = setHyperPars2(lrn, getDefaults(lrn$par.set))
mod2 = train(lrn2, task)

[...]

PR #1440, opened 2017-01-20
#1370 Learner bug list: fix defaults for multilabel.randomForestSRC

MariaErdmann:
Changed default of `mtry`, `nodesize`, `nodedepth`, `splitrule`, `sampsize`, `seed`, `membership`.
Probably there is a reason why the following 2 params were missing in the learner. In this case let me know. But in my opinion both can be added:
Added parameter `samp` (here one can specify a user specific bootstrap matrix) and in this context added value `by.user` to param `bootstrap`. Added also param `coerce.factor`. (names of variables to be analyzed as fators, useful for a large number of factors).


larskotthoff:
Thanks, merging.

PR #1443, opened 2017-01-20
Create featureless dummy models

giuseppec:
This is an old branch from @berndbischl .


berndbischl:
review:
1) pls reread the docs for typos and language, i see a bit
2) pls dont put documentation in the note. 
3) explicitly doc that we dont support weights at the moment
4) reference ZeroR and say that this is similar
5) shouldnt some fixmes be removed? what about the ties stuff?
6) i am unsure whether method "se" makes sense for regr. lets remove it for now

giuseppec:
> 1) pls reread the docs for typos and language, i see a bit

Done.

> 2) pls dont put documentation in the note. 

Done, I explained what the defaults are, not sure if that is ok (if not, what should I write here?)

> 3) explicitly doc that we dont support weights at the moment

Done.

> 4) reference ZeroR and say that this is similar

Done.

> 5) shouldnt some fixmes be removed? what about the ties stuff?

Since ZeroR from WEKA always predicts the **first** factor level in case of ties (see http://weka.8497.n7.nabble.com/ZeroR-on-dataset-with-equal-modes-td33355.html), I decided to let the user choose what to do in case of ties, i.e. I added ties.method to the param set.

> 6) i am unsure whether method "se" makes sense for regr. lets remove it for now

Done.

berndbischl:
> pls dont put documentation in the note.
> Done, I explained what the defaults are, not sure if that is ok (if not, what should I write here?)

also put that into the normal doc page. the "note" is simply for changes that we do for learners that are external to mlr. or "watch-out-warnings". (and we dont want to create doc pages for all of those).

just remove note completely then it will be = "", hence empty. that's totally normal and good.

berndbischl:
> Since ZeroR from WEKA always predicts the first factor level in case of ties (see http://weka.8497.n7.nabble.com/ZeroR-on-dataset-with-equal-modes-td33355.html), I decided to let the user choose what to do in case of ties, i.e. I added ties.method to the param set.

lets not overengineer this. is that "ties = first" option really good? I would say no, just have the random thing and remove it. but you can doc the difference to ZeroR of course, that would be good!

giuseppec:
Everything done. In case of ties only one class is sampled (only once) and always predicted.

berndbischl:
two comments:
1) for the regr case, can you store the median / mean in the model, not y? thats much faster.

2) i dont really like the tests. can you not, rather, handconstruct very small data sets? and compare what happens there?

giuseppec:
@berndbischl ping

PR #1444, opened 2017-01-20
#1370 Learner bug list: fix cluster.cmeans

MariaErdmann:
Adds param `reltol`, which is a control parameter (+"test") and deletes param `nstart`, which is not a param of cmeans.


larskotthoff:
Thanks, merging.

PR #1447, opened 2017-01-23
nicer code in measures.R: exchanged model.matrix by createDummyFeatures

Coorsaa:
Since createDummyFeatures works for vectors now, we can use it for our measures instead of model.matrix


larskotthoff:
Thanks, merging.

PR #1479, opened 2017-02-05
fixes classif.binomial (Learner bug list)

Coorsaa:
Maria was suggesting for add the model par to the train function signature in #1370. However, I was wondering why we set model = FALSE hardcoded inside the glm function call within the train function in the first place, since it's first - against the default value, and second - would make the param obsolete at all or did I get anything wrong here?
However I removed it and it works fine now as far as I can see.

### DIFF
--- a/R/RLearner_classif_binomial.R
+++ b/R/RLearner_classif_binomial.R
@@ -6,12 +6,12 @@ makeRLearner.classif.binomial = function() {
     par.set = makeParamSet(
       makeDiscreteLearnerParam("link", values = c("logit", "probit", "cloglog", "cauchit", "log"),
         default = "logit"),
-      makeLogicalLearnerParam("model", default = TRUE, tunable = FALSE)
+      makeLogicalLearnerParam("model", default = FALSE, tunable = FALSE)
     ),
     properties = c("twoclass", "numerics", "factors", "prob", "weights"),
     name = "Binomial Regression",
     short.name = "binomial",
-    note = "Delegates to `glm` with freely choosable binomial link function via learner parameter `link`."
+    note = "Delegates to `glm` with freely choosable binomial link function via learner parameter `link`. Against the original default in the stats package, model is set to FALSE, due to memory reasons."
   )
 }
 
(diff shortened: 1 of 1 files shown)


larskotthoff:
This increases the amount of memory used and was probably set to FALSE for this reason. It looks to me like it's fine to expose it, but then we should change the default. Or am I missing something @berndbischl ?

berndbischl:
> This increases the amount of memory used and was probably set to FALSE for this reason.

yes.


> It looks to me like it's fine to expose it, but then we should change the default.

i would agree. and mention this in the note.

larskotthoff:
Thanks, merging.

berndbischl:
this code change seems incorrect? 

1) the default in ?glm is "model =TRUE"
so this is the info that needs to be in the param set

2) if you want to override this, set 
par.values =list(model = FALSE)

let us PLEASE not get this policy muddled up.

now @Coorsaa did this wrong and @larskotthoff did not see this in review. this is a problem
we need to eliminate that general problem, otherwise i have to go back to fine-review every merge, which we know will not scale....

berndbischl:
actually even the note is incorrect

>> The default for 'model' in the original learner is FALSE, due to memory reasons."

Coorsaa:
@berndbischl @larskotthoff 
I'm on it right now. Sorry for doing it wrong in the first place. However my note was changed by Lars in the last commit before merging. What would be a sufficient note then?

berndbischl:
ok, comments:

- i see this more of a "review" problem. you should know what follows in my post here, too, but if you make mistake we need to catch it in review.

- it is even more important that we document CRYSTAL CLEAR what the meaning of these things is. please help me by making this better. ask yourself: is this perfectly documented? by checking, reading, and if not documented perfectly, doing that.

i thought it would be written down here, but seems not to be the case
https://mlr-org.github.io/mlr-tutorial/devel/html/create_learner/index.html

- what is the default in the ParamSet: it DOCUMENTS what is the default in the UNDERLYING function.
in this case, that is model = TRUE. it is not used in ANY way in code. so if you change it, it has no effect. 
what is in there, does not refer to mlr in any way. just for glm.

- if you want to change default behavior in a learner, you have to set par.vals. 
(we should really not do this very often)

- if you change par.vals, say in the note why.
here i would say: "model = FALSE was set by default to save memory."







- i

berndbischl:
@schiffner 
do you have something to add here?

larskotthoff:
Sorry, I missed the incorrect note. I wasn't aware that default learner parameters need to be changed this way.

PR #1481, opened 2017-02-06
'fix' cforest learners (Learner bug list)

Coorsaa:
There wasn't really a bug for all learners, but a missing requirement for param 'fraction' (replace == FALSE). Moreover some code cleanup (style guide).
Learner multilable.cforest however showed an error for the default value test which said the params ````p-value````, ````minprob```` and ````randomsplits```` are unused. Since they are actually present within ````party::cforest_control()````, as a first step, I added (only for multilabel.cforest) the ````special.val```` ````NULL```` for all three params which I directly set as default. 
Is that the the correct approach for this or should I completely remove those params for the multilabel learner instead?


larskotthoff:
Thanks. Do those parameters do anything at all for the multilabel learner? If not, remove them.

Coorsaa:
It seems that I can't remove them completely since I get errors then. Instead I removed the params and set them equal to NULL within the train function.

giuseppec:
Hi, I talked to @HeidiSeibold . 
She suggested not to use `party::cforest` in the future and to switch to `partykit:::cforest` (not now, but soon as there will be a bigger CRAN update). She already started to do an mlr implementation for the new cforest. 
I therefore would suggest to skip this whole `cforest_unbiased` and `cforest_classical` stuff we talked about @Coorsaa.
@HeidiSeibold would you mind looking at this PR if we are using the current `party::cforest` properly?

berndbischl:
> Thanks. Do those parameters do anything at all for the multilabel learner? If not, remove them.

@Coorsaa you did not reply here.

and what is the current status?

berndbischl:
> @HeidiSeibold would you mind looking at this PR if we are using the current party::cforest properly?

@HeidiSeibold ping

HeidiSeibold:
To be honest I have never used `party::cforest`. Is there anything in particular you are worried about?

Coorsaa:
@berndbischl Unfortunatelly, those three parameters are not documented, neither in  the ````cforest_control()```` nor in the ````ctree_control()```` doncumentations. Our other cforest learners, i.e. surv, classif, and regr, those three params simply do not appear. After the removal for multilabel here, they would be all the same at least.

berndbischl:
> @berndbischl Unfortunatelly, those three parameters are not documented, neither in the cforest_control() nor in the ctree_control() doncumentations. In our other cforest learners, i.e. surv, classif, and regr, those three params simply do not appear. After the removal for multilabel here, they would be all the same at least.

if they are not even documented i dont see why we need to "support" them.
also we should really make people aware of the fact that you can use mlr learners EVEN with parameters where mlr does NOT have param description objects present. 

see ?configureMlr on.param.without.desc

I hope everybody knows this?

and stefan: what do we now do to finish this here? are we done?

Coorsaa:
@berndbischl yes, we're done here

larskotthoff:
Thanks, merging.

HeidiSeibold:
Do you want to move from `party` to `partykit`? If so, I could help (preferably this week). 

Regarding backwards compatibility this would be a bit annoying. How would you deal with that?

berndbischl:
@HeidiSeibold can you please open a fresh issue / PR for this? here it will be lost.
you can work on this, and i will help you.
either change the old learner or provide a new one.

PR #1485, opened 2017-02-06
Out of bag and se estimators

jakob-r:
Unfortunately I merged the getOutOfBag and the se estimators (mbo_rf) branch into this one.

Now I think we should merge these two actually not so related things into the master, before it gets even more mixed up.


R/RLearner_regr_randomForest.R
```diff
@@ -24,9 +24,10 @@ makeRLearner.regr.randomForest = function() {
     par.set = makeParamSet(
       makeIntegerLearnerParam(id = "ntree", default = 500L, lower = 1L),
       makeIntegerLearnerParam(id = "ntree.for.se", default = 100L, lower = 1L),
-      makeDiscreteLearnerParam(id = "se.method", default = "jackknife",
-                               values = c("bootstrap", "jackknife",  "sd"),
-                               requires = quote(se.method %in% c("jackknife") && keep.inbag == TRUE)),
+      makeLogicalLearnerParam(id = "calibrate", default = TRUE, tunable = FALSE),
+      makeDiscreteLearnerParam(id = "se.method", default = "infjackknife",
```
ja-thomas: I thought we wanted to use jackknife as default, now it is `infijackknife`

R/RLearner_regr_randomForest.R
```diff
@@ -66,6 +67,18 @@ trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = N
   } else {
     data = getTaskData(.task, .subset, target.extra = TRUE)
     m = randomForest::randomForest(x = data[["data"]], y = data[["target"]], ...)
+    ## ugly hack
```
ja-thomas: I really dislike the way we are using `se.method = "bootstrap"`, we are creating a bagging wrapper internally. Why do we not drop `se.method = "bootstrap"` completely as it would still be possible to use a bagging wrapper on a random forest. This should result in the exact same model, or am I overlooking something?

R/RLearner_regr_randomForest.R
```diff
@@ -24,9 +24,10 @@ makeRLearner.regr.randomForest = function() {
     par.set = makeParamSet(
       makeIntegerLearnerParam(id = "ntree", default = 500L, lower = 1L),
       makeIntegerLearnerParam(id = "ntree.for.se", default = 100L, lower = 1L),
-      makeDiscreteLearnerParam(id = "se.method", default = "jackknife",
-                               values = c("bootstrap", "jackknife",  "sd"),
-                               requires = quote(se.method %in% c("jackknife") && keep.inbag == TRUE)),
+      makeLogicalLearnerParam(id = "calibrate", default = TRUE, tunable = FALSE),
```
ja-thomas: Isn't calibrate only required for `se.method = "infijackknife"` ?

R/RLearner_regr_randomForest.R
```diff
@@ -66,6 +67,18 @@ trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = N
...
+    ## the training data is used when predict.type = "se" and calibration is not FALSE
+    if (.learner$predict.type == "se") {
+      if (is.null(.learner$par.vals$calibrate)) {
+        m$data = data[["data"]]
+      } else {
+        if (.learner$par.vals$calibrate) {
+          m$data = data[["data"]]
+        }
+      }
+    }
```
ja-thomas: This looks pretty terrible and again, `calibrate` should only affect `se.method = "infjackknife"` right?

R/getOOBPreds.R
```diff
@@ -0,0 +1,64 @@
...
+getOOBPredsLearner = function(.learner, .model) {
+  UseMethod("getOOBPredsLearner")
+}
+
+# checks if the model was trained on the corresponding task by comparing 
+# the descriptions
+checkModelCorrespondsTask = function(model, task) {
+  if(!identical(task$task.desc, model$task.desc))
+    stopf("Description of the model does not correspond to the task")
+}
```
ja-thomas: This should be somewhere else, and don't we have something like that already?

PR #1489, opened 2017-02-06
random forest se estimation

jakob-r:
reopening this falsely merged PR: https://github.com/mlr-org/mlr/pull/1420


jakob-r:
[These changes](https://github.com/mlr-org/mlr/pull/1485/commits/fdcfd3c24d29442c6eadd56c59684e883c758d26) have to be added again.

R/RLearner_regr_randomForest.R
```diff
@@ -104,7 +119,7 @@ bootstrapStandardError = function(.learner, .model, .newdata, ...) {
     pred.boot = matrix(pred.boot, nrow = nrow(.newdata), ncol = R, byrow = TRUE)
   }
   var.boot = apply(pred.boot, 1, var) - bias
-  var.boot[var.boot <= 0] = 0
+  var.boot = pmax(var.boot, 0)
```
larskotthoff: This changes `var.boot` -- before only the values < 0 were changed, now all but one of them are. Was this a bug in the previous version?

R/RLearner_regr_randomForest.R
```diff
@@ -114,18 +129,39 @@ jackknifeStandardError = function(.learner, .model, .newdata, ...) {
...
+  oob = model$inbag == 0
   jack_n = apply(oob, 1, function(x) rowMeans(pred$individual[, x, drop = FALSE]))
   if (is.vector(jack_n)) {
     jack_n = t(as.matrix(jack_n))
   }
   jack = (n - 1) / n * rowSums((jack_n - pred$aggregate)^2)
   bias = (exp(1) - 1) * n / ntree^2 * rowSums((pred$individual - pred$aggregate)^2)
-  jab = jack - bias
-  jab[jab < 0] = 0
+  jab = pmax(jack - bias, 0)
```
larskotthoff: See above.

R/RLearner_regr_randomForest.R
```diff
@@ -66,6 +68,18 @@ trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = N
...
     m = randomForest::randomForest(x = data[["data"]], y = data[["target"]], ...)
+    ## ugly hack
+    ## if the formula interface is used the training data is saved, but not with the default interface
+    ## the training data is used when predict.type = "se" and calibration is not FALSE
+    if (.learner$predict.type == "se") {
+      if (is.null(.learner$par.vals$calibrate)) {
+        m$data = data[["data"]]
+      } else {
+        if (.learner$par.vals$calibrate) {
+          m$data = data[["data"]]
```
larskotthoff: Where is this data being used? I don't see a reference to it. Is this working around a bug in the learner?

R/RLearner_regr_randomForest.R
jakob-r: Maybe I am confused and don't get your point or you mixed up the behavior of `pmax` and `max`?

R/RLearner_regr_randomForest.R
jakob-r: Maybe @zmjones can explain better what happens but as far as I can see it it's because we need that data to be stored in the model `m` later in at least one `*StandardError` function.

zmjones: that is from a calibration procedure used in the infinitesimal jackknife which ensures that all variance estimates are positive, among other things. this requires the training data. `randomForestCI` never went to cran though, so are we including it now? i don't think the author has any intention of doing so. i could however put his code in mlr directly.

R/RLearner_regr_randomForest.R
zmjones: my reading of the docs for pmax is that this should do the same thing.

R/RLearner_regr_randomForest.R
zmjones: this is just a more efficient replacement no? you should get a logical matrix in both cases?

R/RLearner_regr_randomForest.R
larskotthoff: If we don't need it, let's remove it.

R/RLearner_regr_randomForest.R
larskotthoff: Ah sorry, I didn't realise that pmax changed the semantics of max.

R/RLearner_regr_randomForest.R
zmjones: most of this pr is adding this functionality, which we were using in that other project (so sensible it was in that branch). i'd suggest a new pr or a rebase to just include the minor changes. also why is this high priority?

berndbischl:
my 2cents on this:

- we need this URGENTLY in mlr because we need this in MBO, and other people are already using the less-than-optimal se estimation we already have on CRAN

- i really dont care whether we put every se-estim method for the RF under the sun into this PR and mlr.
actaully i would highly advised against this. we need to support what is reasonable, simple and useful.

berndbischl:
before we continue here:
lets us please define what we want to support, as methods, now

my opinion: 
sd: its simple, already there, a good baseline comparison. we should warn in docs that this is not the best method ever.
the jackknife-after-boostrap: it can out quite good in MBO. it seems to be a much more reasonable thing-

remove: infjackknife. it is in randomForestCI, which is not an CRAN. copying that code here sucks. if somebody wants to have this later, he can do another clean PR and needs to support and maintain this.

remove: bootstrap. is that really needed? it is also very expensive. maybe the noisy-boostrap, but does info exist that this is better than the jackknife? feel unsure here. 

please comment.

zmjones:
i don't see any reason to remove code that works. people can read the sexton and laake paper if they want comparisons. of course the inf jackknife can go for the reasons you mention.

berndbischl:
> i don't see any reason to remove code that works. people can read the sexton and laake paper if they want comparisons.

2 questions: 
- if code is lengthy and terrible and not many people use it, it creates a "burden" right? isnt that grounds for removal?
- how do you know it works? can i see that "proof"?

berndbischl:
> lengthy and terrible

i am NOT saying anybodies code in this PR is terrible! was trying to make a general point!

zmjones:
haha it is fine. i don't think it is that long and if people find bugs i am happy to look into them and so on. that said i don't have any burning desire to use any of this regularly, i just spent some effort in reimplementing all of this and can imagine it would be used.

no i don't have any proof that it works other than my own review of the code and that it worked reliably while we were using it in that project. that said i guess we could do some simulations, though i do not have the time at the moment. most of the procedures are fairly simple.

zmjones:
i am also confused. i looked at the commit list and thought this was adding the rewrite of the functionality i did a while ago, but this is really just adding the inf jackknife and a few very minor changes. i think we can just close this and have someone issue a pr with just the pmax and the one line of simplification i did on the inbag matrix. i don't see why this pr was started at all now.

berndbischl:
summary of what i discussed with zach:
- zach will take over now and finish quickly.
- docs will be checked and improved
- infjack goes
- bootstrap stays
- he will interate over the code to see if all is good

R/RLearner_regr_randomForest.R
jakob-r: High prio because we want to merge that for mlrMBO release.

jakob-r:
And what is about the `rf_se` branch and the PR #1497 ?
This is really confusing here.

zmjones:
that was supposed to replace this pr since most of the content of this one was adding the inf jackknife which we aren't going to merge.

jakob-r:
The for the sake of less confusion I will close this PR here.

PR #1493, opened 2017-02-08
closes #1492 fix classif.binomial

### LINKED ISSUE #1492: merged PR for classif.binomial was wrong!
see my comments here

https://github.com/mlr-org/mlr/pull/1479

```
this code change seems incorrect?

the default in ?glm is "model =TRUE"
so this is the info that needs to be in the param set

if you want to override this, set
par.values =list(model = FALSE)

let us PLEASE not get this policy muddled up.

now @Coorsaa did this wrong and @larskotthoff did not see this in review. this is a problem
we need to eliminate that general problem, otherwise i have to go back to fine-review every merge, which we know will not scale....

```

Coorsaa:
I fixed the default of param 'model' and set it to FALSE in the par.vals.
See #1492 and #1479

### DIFF
--- a/R/RLearner_classif_binomial.R
+++ b/R/RLearner_classif_binomial.R
@@ -14,7 +14,7 @@ makeRLearner.classif.binomial = function() {
     properties = c("twoclass", "numerics", "factors", "prob", "weights"),
     name = "Binomial Regression",
     short.name = "binomial",
-    note = "Delegates to `glm` with freely choosable binomial link function via learner parameter `link`. By default, 'model' was set to FALSE to save memory."
+    note = "Delegates to `glm` with freely choosable binomial link function via learner parameter `link`. We set 'model' to FALSE by default to save memory."
   )
 }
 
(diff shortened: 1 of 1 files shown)


berndbischl:
looks good but i would like issues not to be closed by "programming" in the copmmit msg

berndbischl:
can be merged if travis green

larskotthoff:
Re closing issues through commit messages, I'm in favour of that and have been asking people to do it. Otherwise we end up with "orphaned" issues.

berndbischl:
> Re closing issues through commit messages, I'm in favour of that and have been asking people to do it. Otherwise we end up with "orphaned" issues.

ok, noted. and i can understand. but then the "merger" needs to take responsibility by checking that there is nothing of "value" anymore in the issue, and that the PR deals really with 100% of what was discussed there

berndbischl:
(which you kinda need to do anyhow in the review)

larskotthoff:
Yep.

PR #1497, opened 2017-02-08
fix defaults for rf se estimation

zmjones:
- add sd method to docs
 - add JMLR paper to references
 - some code simplifications from @jakob-r
 - check sd method in tests

should close #1489.

### DIFF
--- a/R/RLearner_regr_randomForest.R
+++ b/R/RLearner_regr_randomForest.R
@@ -9,9 +9,11 @@
 #'
 #' If \code{se.method = "jackknife"}, the default, the standard error of a prediction is estimated by computing the jackknife-after-bootstrap, the mean-squared difference between the prediction made by only using trees which did not contain said observation and the ensemble prediction.
 #'
+#' If \code{se.method = "sd"}, the standard deviation of the predictions across trees is returned as the variance estimate.
+#'
 #' For both \dQuote{jackknife} and \dQuote{bootstrap}, a Monte-Carlo bias correction is applied and, in the case that this results in a negative variance estimate, the values are truncated at 0.
 #'
-#' @references [Joseph Sexton] and [Petter Laake],; [Standard errors for bagged and random forest estimators], Computational Statistics and Data Analysis Volume 53, 2009, [801-811].
+#' @references [Joseph Sexton] and [Petter Laake],; [Standard errors for bagged and random forest estimators], Computational Statistics and Data Analysis Volume 53, 2009, [801-811]. Also see: [Stefan Wager], [Trevor Hastie], and [Bradley Efron]; [Confidence Intervals for Random Forests: The Jackknife and the Infinitesimal Jackknife], Journal of Machine Learning Research Volume 15, 2014, [1625-1651].
 #'
 #' @name regr.randomForest
 NULL
@@ -44,19 +46,20 @@ makeRLearner.regr.randomForest = function() {
       makeLogicalLearnerParam(id = "keep.inbag", default = FALSE, tunable = FALSE)
     ),
     par.vals = list(
-      se.method = "bootstrap",
+      se.method = "jackknife",
       se.boot = 50L,
       ntree.for.se = 100L
     ),
     properties = c("numerics", "factors", "ordered", "se", "oobpreds", "featimp"),
     name = "Random Forest",
     short.name = "rf",
-    note = "See `?regr.randomForest` for information about se estimation. Note that the rf can freeze the R process if trained on a task with 1 feature which is constant. This can happen in feature forward selection, also due to resampling, and you need to remove such features with removeConstantFeatures."
+    note = "See `?regr.randomForest` for information about se estimation. Note that the rf can freeze the R process if trained on a task with 1 feature which is constant. This can happen in feature forward selection, also due to resampling, and you need to remove such features with removeConstantFeatures. keep.inbag is NULL by default but if predict.type = 'se' and se.method = 'jackknife' (the default) then it is automatically set to TRUE."
   )
 }
 
 #' @export
-trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = NULL, ...) {
+trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = NULL,
+  se.method = "jackknife", keep.inbag = NULL, ...) {
   if (.learner$predict.type == "se" &
         .learner$par.vals$se.method == "bootstrap") {
     base.lrn = setPredictType(.learner, "response")
@@ -65,7 +68,8 @@ trainLearner.regr.randomForest = function(.learner, .task, .subset, .weights = N
     m = train(bag.rf, .task, .subset, .weights)
   } else {
     data = getTaskData(.task, .subset, target.extra = TRUE)
-    m = randomForest::randomForest(x = data[["data"]], y = data[["target"]], ...)
+    m = randomForest::randomForest(x = data[["data"]], y = data[["target"]],
+      keep.inbag = if (is.null(keep.inbag)) TRUE else keep.inbag, ...)
   }
   return(m)
 }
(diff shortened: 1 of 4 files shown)


berndbischl:
@zmjones 
you seem to have a bug in there, travis fails.
looks like you are dropping a dim somewhere. 
i would test predicting on only one observation.

as this only fails now: do we have unit tests for all 3 se methods?

zmjones:
yes we do have tests for all 3 now. already only predicting one obs. in the test for the se estimation which is passing locally. not sure what is up at the moment.

larskotthoff:
Maybe package versions? Have you tried updating to the latest versions locally?

zmjones:
yea i think that is all up to date. still passing locally.

berndbischl:
are you really running the learners_all tests.....? sure?

zmjones:
no i just meant that regr_randomForest is passing. i don't have all the suggests installed on my lab computer though i am doing that now.

berndbischl:
well, have look at the travis output. really read the logs

```
1. Error: learners work: regr  (@test_learners_all_regr.R#36) ------------------
dim(X) must have a positive length
1: lapply(lrns, testBasicLearnerProperties, task = task, hyperpars = hyperpars, pred.type = "se") at testthat/test_learners_all_regr.R:36
2: FUN(X[[i]], ...)
3: predict(m, task) at /home/travis/build/mlr-org/mlr/mlr.Rcheck/tests/testthat/helper_learners_all.R:74
4: predict.WrappedModel(m, task)
5: measureTime(fun1(p <- fun2(do.call(predictLearner2, pars))))
6: force(expr)
7: fun1(p <- fun2(do.call(predictLearner2, pars)))
8: evalVis(expr)
9: withVisible(eval(expr, pf))
10: eval(expr, pf)
11: eval(expr, envir, enclos)
12: fun2(do.call(predictLearner2, pars))
13: do.call(predictLearner2, pars)
14: (function (.learner, .model, .newdata, ...) 

 SNIP

15: predictLearner(.learner, .model, .newdata, ...)
16: predictLearner.regr.randomForest(.learner, .model, .newdata, ...)
17: se.fun(.learner, .model, .newdata, ...)
18: apply(oob, 1, function(x) rowMeans(pred$individual[, x, drop = FALSE]))
19: stop("dim(X) must have a positive length")
```

so you need to run that files. 
just comment out stuff in there, and restrict the learners to just the RF to debug fast.

or, even better, copy the relevant code to an external file and run there

zmjones:
i still can't replicate this locally. i can pass r check and can see the error comes from this

```
  # regr with se
  lrns = mylist(task, properties = "se", create = TRUE)
  lapply(lrns, testBasicLearnerProperties, task = task, hyperpars = hyperpars,
    pred.type = "se")
```

but can't get that to work locally. something about the scope doesn't allow necessary functions like expect_output or extractSubList to be found despite being loaded in the global environment.

berndbischl:
i helped zach with debugging. we found the problem.

further stuff:
please make sure that indentation is correct, 2 spaces.
style guide should be followed. i see var names with underscores and some uppercase letters.

berndbischl:
we still wanted to get rid of par.vals, right?

berndbischl:
improving docs a little bit. give me 10 min

berndbischl:
@zmjones 
made docs a bit better, but: currently NONE of the extra parameters for the se.method are documented. 
please do that

jakob-r:
> NONE of the extra parameters for the se.method are documented.
please do that

@berndbischl what do you mean by this. The extra agruments are `se.boot` and `ntree.for.se`. Both are briefly descrobed in the section for `se.method = "bootstrap"`. What is missing?

berndbischl:
> what do you mean by this. The extra agruments are se.boot and ntree.for.se. Both are briefly descrobed in the section for se.method = "bootstrap". What is missing?

sorry, i was blind. can see it now, i guess it happened because it was late and the doc page just looks different here. 
anything else missing from doc? maybe we should say that "sd" is a very naive estimator?

jakob-r:
From my side nothing is missing. Added a sentence.

berndbischl:
- made docs better
- parvals needs to be removed
- zach, can you please document the noisy boostrap and the difference to the normal bs a bit better?
- are the defaults for the bs really good? they seem very expensive, what is in sexton and laake recommended?

zmjones:
i think the defaults for the bootstrap (which is the noisy bootstrap since the default ntree is 500 and ntree.for.se is 100 are fine. the bootstrap is going to be expensive any way you do it i think.

zmjones:
and i documented the differences a bit and also removed par.vals and moved the bootstrap defaults to that function

jakob-r:
why is it failing? i am on it.

zmjones:
so this looks like another error from removing the se.method default from par.vals. now the switch statement in the predict method doesn't work. 

i know we talked about removing all the se.method stuff from par.vals but i am not so sure this is a good idea now. this is also a problem for the additional se.method args: ntree.for.se and se.boot. 

thoughts?

jakob-r:
It took me a while but what fucks us over is the partial argument matching from `do.call(trainLearner, pars)`. We have to rename `ntree.for.se` to something else that does not get matched with `ntree`.

zmjones:
we could just put those two args back into par.vals. that would avoid this problem, no? this seems like a lot of pain for very little gain.

jakob-r:
The problem is that conceptually `par.vals` is the wrong place for that. We might want to improve the parameter settings behaviour in the future (there are already some old PRs that are just quite complicated and never got merged) and that would certainly be in conflict of the mechanisms we thought of.

jakob-r:
One question that arises: If we use `se.method = "bootstrap"` doesn't it influence the mean prediction heavily and do we want that or do we just want to use the bagged random forest to estimate the se and keep the normal random forest for the mean prediction?

jakob-r:
The following example let's me doubt about our understanding of the bootstrap se estimator:
```r
set.seed(1)
n = 1000
flat = data.frame(x = seq(0,1, length.out = n), y = rnorm(n, sd = 10))
plot(y~x, flat)
flat = makeRegrTask(data = flat, target = "y")
lrn = makeLearner("regr.randomForest", predict.type = "se", se.method = "bootstrap")
m = train(lrn, flat)
p = predict(m, flat)
mean(p$data$se)
```

jakob-r:
I want to note - also it looks like a lot changed it still is the same result. I just find it a bit more intuitive to read.

jakob-r:
And I now come to the same conclusion again I came some years ago. With the bootstrap we can just measure the variance of the estimators which is totally not the same because even if the data is totally noisy the estimators will all predict the mean and accordingly the variance of the estimation is really low.

zmjones:
well yes in practice it will be different. whether it will be better or worse will depend on the number of bootstrap iterations and the number of trees in the base random forests. since everything is independent just the product of those two i guess. i would guess that in general ntree.se * se.boot << ntree. for our defaults here this is the case anyways. in expectation the bootstrap of random forests and the random forest should give the same predictions though?

i don't understand your latest comment though. if the data is really noisy then the predictions of the base learners will also be very variable and the estimated variance will go up. can you give a concrete example where that wouldn't be the case?

jakob-r:
Well you can play around with the example I posted above and you will see that the variance will go down with a higher number of observations and a higher number of bootstraps. This makes sense in the way that of course with a higher number of observations - even if they are noisy - our estimation of the mean gets more reliable. However the estimate does not reflect the real distribution.

R/RLearner_regr_randomForest.R
```diff
@@ -90,43 +112,53 @@ getOOBPredsLearner.regr.randomForest = function(.learner, .model) {
...
+  # Bias is defined as
+  # (1/R - 1/M) / (BR*(R-1)) *
+  # (sum over all B:
+  #   (sum over all R:
+  #     (prediction for x of ensemble r in bootstrap b - average prediction for x over all ensambles in bootsrap b )^2
+  #   )
+  # )
+  bias = ((1 / se.ntree) - (1 / ntree)) / ( se.boot * se.ntree * (se.ntree - 1)) *
+    rowSums(matrix(
+      sapply(pred.boot.all, function(p)
```
ja-thomas: sapply :(

R/RLearner_regr_randomForest.R
```diff
@@ -1,19 +1,44 @@
...
+#'   the prediction made by only using trees which did not contain said observation and
+#'   the ensemble prediction.
+#' \item If \code{se.method = "bootstrap"} the standard error of a prediction is estimated by
+#'   bootstrapping the random forest, where the number of bootstrap replicates and the number of
+#'   trees in the ensemble are controlled by \code{se.boot} and \code{se.ntree} respectively,
+#'   and then taking the standard deviation of the bootstrap predictions. The "brute force" bootstrap
+#'   is executed when \code{ntree = se.ntree}, the latter of which controls the number of trees in the
+#'   individual random forests which are bootstrapped. The "noisy bootstrap" is executed when \code{se.ntree < ntree}
+#'   which is less computationally expensive. A Monte-Carlo bias correction may make the latter option
+#'   prefarable in many cases. Defaults are \code{se.boot = 50} and \code{se.ntree = 100}.
```
ja-thomas: Maybe emphasize that only the `sd` is calculated by the bootstrap and the `mean` is the prediction of the original forest

R/RLearner_regr_randomForest.R
```diff
@@ -90,43 +112,53 @@ getOOBPredsLearner.regr.randomForest = function(.learner, .model) {
...
-  var.boot = apply(pred.boot, 1, var) - bias
-  var.boot[var.boot <= 0] = 0
-  cbind(pred, sqrt(var.boot))
+# Set ntree = se.ntree for the brute force bootstrap
+# Set se.ntree << ntree for the noisy bootstrap (mc bias corrected)
+bootstrapStandardError = function(.learner, .model, .newdata,
+  se.ntree = 100L, se.boot = 50L, ...) {
+  single.model = getLearnerModel(.model)$single.model #get raw RF model
+  bagged.models = getLearnerModel(getLearnerModel(.model)$bagged.models) #get list of unbagged mlr models
+  pred.single = predict(single.model, newdata = .newdata, ...)
```
berndbischl: these lines should all be moved to predictLearner. the se.funs should only compute and return the se as a single vec

PR #1511, opened 2017-02-13
#1465: listTaskTypes and listLearnerProperties

### LINKED ISSUE #1465: have small function listTaskTypes
should return simply a char vector of all valid task types

unit test this against the types of all available tasks, possibly the tasks of all objects where we have examples in data?

MasonGallo:
Going through quick open issues. This should close #1465 when green.

I added tests and exported the methods.


larskotthoff:
Thanks, merging.

Issue #1470, opened 2017-02-02
pltBMRSummary custom shape


QuayAu:
Hi,
I am interested in changing the shape of the points plotted with plotBMRSummary.
I tried using scale_shape_manual(), which did not have any effect. 

```
library(mlr)
library(ggplot2)
lrns = list(makeLearner("classif.lda"), makeLearner("classif.rpart"))
rdesc = makeResampleDesc("Holdout")
bmr = benchmark(lrns, sonar.task, rdesc)
plt = plotBMRSummary(bmr)
plt + scale_shape_manual(values = c(1, 2))

```

jakob-r:
Hi,
the problem is, that we did not defined any variable to be the aesthetic for the shape so you don't affect something with `scale_shape_manual()` because the scale_shape is simply not present.
However there is a small hack that could help you in simple cases:

```r
plt$layers[[1]]$aes_params$shape = c(17, 15)
```

Another solution would be to write your own `plotBMRSummary2` function that you can simply modify, which should not be so hard.

Alternatively @everyone: Should we just change the line to
```r
p = ggplot(df, aes_string(x = meas.name, y = "task.id", col = "learner.id", shape = "learner.id"))
```
for plotBMRSummary() ?

berndbischl:
@QuayAu will likely add a fix here soon

Issue #1523, opened 2017-02-15
plotBMRBoxplots fails when the same learner is used multiple times (with different ids)


ja-thomas:
```
lrns = list(
  rf = makeLearner("classif.randomForest", id = "rf1"),
  rf2 = makeLearner("classif.randomForest", id = "rf2")
)


bm = benchmark(lrns, iris.task, hout)

plotBMRBoxplots(bm)
```

![bug](https://cloud.githubusercontent.com/assets/7561944/22991656/5298c3a8-f3bd-11e6-8ce6-ef8b3955714b.png)

This is because the `shortname` is the same for both learners, when I set `pretty.names=FALSE` it works

larskotthoff:
Is this actually a bug? I.e. do we want to detect such things and adjust the options accordingly? Or output a warning?

berndbischl:
> Is this actually a bug? I.e. do we want to detect such things and adjust the options accordingly? Or output a warning?

we want to output an error when the shortnames are not unique (and prettynames is on)
the resulting plot is otherwise extremely confusing.

and we want a better setter for names. like
setLearnerNames(id, short.name, name) i would suggest.

we can separate issues for this. not that for the plot functions we might have to check multiple functions for the same problem

PR #1525, opened 2017-02-16
add shape parameter for PlotBMRSummary 

QuayAu:
This is a PR for #1470 

This is still work in progress, as I need to add a unit test, that catches the problem that changing the shape parameter didn't have any effect. 
Is there any simple way to test if generated ggplots are the same?


R/plotBMRSummary.R
```diff
@@ -53,8 +53,10 @@ plotBMRSummary = function(bmr, measure = NULL, trafo = "none", order.tsks = NULL
     levels(df$learner.id) = getBMRLearnerShortNames(bmr)
   }
 
-  p = ggplot(df, aes_string(x = meas.name, y = "task.id", col = "learner.id"))
+  p = ggplot(df, aes_string(x = meas.name, y = "task.id", col = "learner.id", shape = "learner.id"))
```
ja-thomas: We don't really want to hardcode this right?

R/plotBMRSummary.R
```diff
@@ -53,8 +53,10 @@ plotBMRSummary = function(bmr, measure = NULL, trafo = "none", order.tsks = NULL
...
   p = p + geom_point(size = pointsize, position = position_jitter(width = 0, height = jitter))
+  # set shape to standard shape 19
+  p = p + scale_shape_manual(values = rep(19, length(getBMRLearnerShortNames(bmr))))
```
ja-thomas: Have a look at #1523, this might be problematic for wrapped learners

berndbischl:
what is the status here? work in progress? or ready for review?

QuayAu:
Work in progress, will have a look at it tomorrow

larskotthoff:
What's the status here @QuayAu ?

giuseppec:
Looks good to me. We internally set the shape to the default (using `scale_shape_manual`). This allows us to change the shape and replace the default from the outside using `scale_shape_manual`. This was not possible before this PR.

giuseppec:
@QuayAu can you add an example as requested by @larskotthoff ?

PR #1531, opened 2017-02-19
restrict k in rknn (closes #1530)

### LINKED ISSUE #1530: rknn crashes R when k > 98
... so it would make sense to constrain the `rknn` `k` parameter.

To reproduce:
```R
 > resample(makeLearner("classif.rknn", k=99), pid.task, cv5)
*** Error in `/usr/lib64/R/bin/exec/R': malloc(): memory corruption: 0x000055b296959d90 ***
======= Backtrace: =========
......
```

This happens because some rknn C code allocates some arrays of size 100 and then writes to the `[k+1]` index.

larskotthoff:


berndbischl:
thx to both of you

PR #1541, opened 2017-02-23
Kendalls tau and spearmans rho as ranking based regression measures

PhilippPro:
If one is mainly interested, if the order of the predicted values is correct one can use Kendalls tau and Spearmans rho as ranking based regression measures, see [Rosset et al.](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.95.1398&rep=rep1&type=pdf)

This is something similar to the AUC (I originally searched something like AUC for regression).

This can be especially interesting when you want to look if your trained model orders the observations correctly although e.g. mse is quite bad, because of outliers or wrong scale, that could be potentially solved by transforming the values before or after your training process. 
So this could be interesting as extra measure to e.g. the mse (which is strongly connected to pearson's correlation coefficient).

Implementation details:
- `use = "na.or.complete"`ensures, that something is calculated even if there are NA, if n-1 have at least one NA, it returns NA.
- The test is a bit trivial, but testing this is maybe trivial, too. ;)
- There is also a rescaled variant between 0 and 1, but I would prefer this

There are also some visual tools, see the paper for more information.

### DIFF
--- a/R/measures.R
+++ b/R/measures.R
@@ -402,7 +402,7 @@ tau = makeMeasure(id = "tau", minimize = FALSE, best = 1, worst = -1,
   }
 )
 
-#' @export measureTAU
+#' @export measureTau
 #' @rdname measures
 #' @format none
 measureTau = function(truth, response) {
@@ -422,7 +422,7 @@ rho = makeMeasure(id = "rho", minimize = FALSE, best = 1, worst = -1,
   }
 )
 
-#' @export measureRHO
+#' @export measureRho
 #' @rdname measures
 #' @format none
 measureRho = function(truth, response) {
--- a/R/measures.R
+++ b/R/measures.R
@@ -398,14 +398,14 @@ tau = makeMeasure(id = "tau", minimize = FALSE, best = 1, worst = -1,
   note = "Defined as: Kendall's tau correlation between truth and response. Only looks at the order.
   See Rosset et al.: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.95.1398&rep=rep1&type=pdf.",
   fun = function(task, model, pred, feats, extra.args) {
-    measureTAU(pred$data$truth, pred$data$response)
+    measureTau(pred$data$truth, pred$data$response)
   }
 )
 
 #' @export measureTAU
 #' @rdname measures
 #' @format none
-measureTAU = function(truth, response) {
+measureTau = function(truth, response) {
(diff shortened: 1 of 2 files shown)


catastrophic-failure:
Just for reference, the Gini Coefficient is usually called the AUC for regression.

PhilippPro:
Do you have a reference about this? I would like to know more about the relationship between these measures...

PhilippPro:
I found the relationship between Kendalls tau and the Mann–Whitney U test statistic (Wikipedia): The Mann–Whitney U test is related to a number of other non-parametric statistical procedures. For example, it is equivalent to Kendall's tau correlation coefficient if one of the variables is binary (that is, it can only take two values). 

Moreover AUC is directly proportional to the Mann Whitney U statistic. -> Kendall's tau is directly proportional to the AUC in case of classification. 

Gini coefficient is also very closely related to the AUC in case of classification: Gini+1=2×AUC. I did not find any formula for the Gini in case of regression...

catastrophic-failure:
See here: https://www.kaggle.com/wiki/Gini
The Gini coefficient is a measure of the conservation of ranks, and naturally extends to regression.

R/measures.R
```diff
@@ -389,6 +389,46 @@ rmsle = makeMeasure(id = "rmsle", minimize = TRUE, best = 0, worst = Inf,
...
+  fun = function(task, model, pred, feats, extra.args) {
+    measureTAU(pred$data$truth, pred$data$response)
+  }
+)
+
+#' @export measureTAU
+#' @rdname measures
+#' @format none
+measureTAU = function(truth, response) {
+  cor(truth, response, use = "na.or.complete", method = "kendall")
```
larskotthoff: Should be `measureTau`.

R/measures.R
```diff
@@ -389,6 +389,46 @@ rmsle = makeMeasure(id = "rmsle", minimize = TRUE, best = 0, worst = Inf,
...
+  See Rosset et al.: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.95.1398&rep=rep1&type=pdf.",
+  fun = function(task, model, pred, feats, extra.args) {
+    measureRHO(pred$data$truth, pred$data$response)
+  }
+)
+
+#' @export measureRHO
+#' @rdname measures
+#' @format none
+measureRHO = function(truth, response) {
```
larskotthoff: `measureRho`

PR #851, opened 2016-04-25
Adding debug traces to FailureModels and failed preds.

mb706:
Fix #750: This is my first draft implementing error dumps for mlr learners. It adds an `on.error.dump` mlr option and the functions `getFailureModelDump` and `getPredictionDump`. Usage is as follows:

``` R
> library("mlr")
> multiclass.task = makeClassifTask("multiclass", data = iris, target = "Species")
> configureMlr(on.error.dump = TRUE, on.learner.error = "warn")
>   lrn = makeLearner("classif.qda", predict.type = "prob")
> m = train(lrn, multiclass.task, subset = c(1,51,101))
Warning in train(lrn, multiclass.task, subset = c(1, 51, 101)) :
  Could not train learner classif.qda: Error in qda.default(x, grouping, ...) : 
  some group is too small for 'qda'

> debugger(getFailureModelDump(m))
Message:  some group is too small for 'qda'Available environments had calls:
1: train(lrn, multiclass.task, subset = c(1, 51, 101))
2: system.time(fun1(learner.model <- fun2(fun3(do.call(trainLearner, pars)))),
3: fun1(learner.model <- fun2(fun3(do.call(trainLearner, pars))))
4: fun2(fun3(do.call(trainLearner, pars)))
5: try(x, silent = TRUE)
6: tryCatch(expr, error = function(e) {
    call <- conditionCall(e)
    if (!
7: tryCatchList(expr, classes, parentenv, handlers)
8: tryCatchOne(expr, names, parentenv, handlers[[1]])
9: doTryCatch(return(expr), name, parentenv, handler)
10: fun3(do.call(trainLearner, pars))
11: withCallingHandlers(x, error = function(c) dump.frames())
12: do.call(trainLearner, pars)
13: (function (.learner, .task, .subset, .weights = NULL, ...) 
{
    UseMethod
14: trainLearner.classif.qda(.learner = list(id = "classif.qda", type = "classi
15: MASS::qda(f, data = getTaskData(.task, .subset, recode.target = "drop.level
16: qda.formula(f, data = getTaskData(.task, .subset, recode.target = "drop.lev
17: qda.default(x, grouping, ...)
18: stop("some group is too small for 'qda'")
19: .handleSimpleError(function (c) 
dump.frames(), "some group is too small fo
20: h(simpleError(msg, call))

Enter an environment number, or 0 to exit  Selection:
```

and also

``` R
> p = predict(m, iris.task)  # the failed model *OR* a normal model but the prediction fails
[...]

PR #1554, opened 2017-02-25
Fix #1553: tuneParams and tuneParamsMultiCrit have resample.fun option.

### LINKED ISSUE #1553: Tuning: Give user the option to specify his own resampling method
This would greatly enhance flexibility with respect to things like imputation or fallback learners. It also would allow the user to cache some intermediate results (e.g. filter values for feature selection--unless there already exists an option for that?).

I suggest a `resample.fun` argument that is expected to have the same interface as `resample` and also to return a `ResampleResult`.

mb706:
Fixes #1553.


mb706:
TODO: maybe write some documentation: `resample.fun` should return a list that contains `$aggr`, or, if `control$tune.threshold`, `$pred`, as well as maybe `$err.msgs` (and if #851 ever gets some love, `$err.dumps`).

berndbischl:
See discussion in issue.

PR #1555, opened 2017-02-25
New parameter 'method' for classif.plsdaCaret. This fixes #1544

### LINKED ISSUE #1544: classif.plsdaCaret sometimes does not terminate
... because it uses the `"oscorespls"` algorithm, which is prone to numerical instability. The [pls package paper](https://www.jstatsoft.org/article/view/v018i02/v18i02.pdf) mentions that the `"kernelpls"` algorithm gives the same result and is usually faster. I suggest changing it [here](https://github.com/mlr-org/mlr/blob/e8801bcb5ad1ac7c5fdfc7b862df7eb4c37a698a/R/RLearner_classif_plsdaCaret.R#L18).

```R
library(fastICA)
> dm <- fastICA(matrix(runif(6000), ncol=6), 6, method="C")$S
> mkt <- makeClassifTask("orthotask", data.frame(t=sample(factor(c('a', 'b')), 1000, replace=TRUE), dm), target='t')
> train(makeLearner('classif.plsdaCaret'), mkt)  # runs forever
```

mb706:
Make the `method` parameter and use the package's default of `kernelpls`. See #1544

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 * getOOBPreds: get out-of-bag predictions from trained models for learners that store them -- these learners have the new "oobpreds" property
 * listTaskTypes, listLearnerProperties
 
+## learners - general
+* classif.plsdaCaret: added parameter "method".
+
+
 ## learners - removed
 * {classif,regr}.avNNet: no longer necessary, mlr contains a bagging wrapper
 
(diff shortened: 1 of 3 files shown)


berndbischl:
thx for the quick and clean fix!

Issue #1535, opened 2017-02-20
DLL Error when calling classification tests in forecasting branch


Stevo15025:
In the forecasting branch, the code below will cause a maximum number of DLLs loaded error. The code is a simplified version of the beginning of [`test_learners_classif.R`](https://github.com/mlr-org/mlr/blob/master/tests/testthat/test_learners_all_classif.R)

```r
data(Sonar, package = "mlbench", envir = environment())
binaryclass.df = Sonar
binaryclass.target = "Class"
binaryclass.task = makeClassifTask("binary", data = binaryclass.df, target = binaryclass.target)
task = subsetTask(binaryclass.task, subset = c(10:20, 180:190),
                  features = getTaskFeatureNames(binaryclass.task)[12:15])
lrns = mylist(task, create = TRUE)
#  Error in dyn.load(file, DLLpath = DLLpath, ...) : 
#  unable to load shared object '/home/steve/R/x86_64-pc-linux-gnu-library/3.3/mclust/libs/mclust.so':
#  maximal number of DLLs reached...
#  Failed with error:  ‘package ‘mvoutlier’ could not be loaded’
#   Show Traceback
# 
# Rerun with Debug
# Error in requirePackages(package, why = stri_paste("learner", id, sep = " "),  : 
#  For learner classif.rrlda please install the following packages: rrlda 
```

In (R 3.4)[http://stat.ethz.ch/R-manual/R-devel/doc/html/NEWS.html] this won't be an issue as I requested that they allow for users to increase the maximum number of DLLs that R is able to load. 

> The maximum number of DLLs that can be loaded into R e.g. via dyn.load() can now be increased by setting the environment variable R_MAX_NUM_DLLS before starting R.)

But that be out until April and I was wondering if anyone has ideas on how to tackle this now. I plan to submit my masters thesis (the forecasting branch) to the R in Finance conference to present this April. While R 3.4 also is scheduled to come out in April, I think that is a little too close for comfort time-wise.

So,

1. This error only occurs in the tests for classifiers. For the purposes of my branch am I allowed to cut out those tests for now?

2. The error happens on the execution of `mylist()`, is there a way to limit the number of packages `mylist()` will load?

larskotthoff:
1. Yes, although you can also just split up the tests further.
2. `mylist` will load the packages necessary for the learners if you pass `create = TRUE`. You could restrict that number by restricting the number of eligible learners, but again the better solution here is to split up the tests further.

[...]

PR #1318, opened 2016-11-01
Forecasting

Stevo15025:
This is a pull request for forcasting in mlr. This branch is here to review the changes with @berndbischl

PR #1560, opened 2017-02-28
PR to fix MAX_NUM_DLL error

Stevo15025:
See #1318 and #1535 

@larskotthoff should this be broken up at the `test_that()` level?

### DIFF
--- a/tests/testthat/helper_learners_all.R
+++ b/tests/testthat/helper_learners_all.R
@@ -154,14 +154,14 @@ testThatLearnerHandlesMissings = function(lrn, task, hyperpars) {
   testBasicLearnerProperties(lrn = lrn, task = task, hyperpars = hyperpars)
 }
 
-# Test that the extraction of the out-of-bag predictions for the learner that supports 
+# Test that the extraction of the out-of-bag predictions for the learner that supports
 # this works correctly
 
 testThatGetOOBPredsWorks = function(lrn, task) {
   type = lrn$type
   mod = train(lrn, task)
   oob = getOOBPreds(mod, task)
-  
+
   if (type == "classif") {
     if(lrn$predict.type == "response") {
       expect_is(oob$data, "data.frame")
@@ -173,7 +173,7 @@ testThatGetOOBPredsWorks = function(lrn, task) {
   } else {
     if (type %in% c("regr", "surv")) {
       expect_is(oob$data$response, "numeric")
-    } 
+    }
   }
   expect_equal(nrow(oob$data), nrow(getTaskData(task)))
 }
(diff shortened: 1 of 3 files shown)


Stevo15025:
Also a new argument was added to `listLearners()` called subset which lets you select the `1:10` or `20:35` elements of the set of learners.

larskotthoff:
Thanks for looking at this. I'm not sure if the additional argument is needed though -- you can simply listLearners with `create = FALSE` and then subset. The general way of splitting up the tests is fine, but please pull out the common things into a function to reduce the code duplication. Please also have a look at the run-* files in tests/ to see if similar things need to be one there.

Stevo15025:
> Thanks for looking at this.

No problem!

> I'm not sure if the additional argument is needed though -- you can simply listLearners with create = FALSE and then subset.

Yes I did not see exactly what you meant by that before, but now mylist is called with `create = FALSE` and then a subset and lapply are used to create the learners.

> The general way of splitting up the tests is fine, but please pull out the common things into a function to reduce the code duplication. 

Added [here](https://github.com/mlr-org/mlr/pull/1560/files#diff-f6ed7c01d90bb7b026a3f16aa690a747R242)

> Please also have a look at the run-* files in tests/ to see if similar things need to be one there.

Sure, though I think we should try to do this as little as possible. In the end of the day this does make the tests a lot more convoluted. And when R 3.4 comes out we can set the MAX_NUM_DLLs so this won't be an error anymore.

Stevo15025:
Also testing mboost is causing my computer to crash? Very odd

larskotthoff:
Tests fail with

```
Running the tests in ‘tests/run-learners-classif.R’ failed.
Last 13 lines of output:
  2: lapply(lrns$class[subset], makeLearner) at /usr/local/jenkins/jobs/mlr-PRbuilder/workspace/mlr.Rcheck/tests/testthat/helper_learners_all.R:249
  3: FUN(X[[i]], ...)
  4: assertString(cl)
  5: makeAssertion(x, res, .var.name, add)
  6: mstop("Assertion on '%s' failed: %s.", var.name, res, attach = x)
  7: assertion_error(sprintf(msg, ...), call = sys.call(1L), attach = attach)
  
  testthat results ================================================================
  OK: 1186 SKIPPED: 0 FAILED: 1
  1. Error: learners work: classif 61:87 (@test_learners_all_classif.R#20) 
```

Stevo15025:
Can you check if 

```
 testThatDefaultClassifAllLearners(1:20)
```

causes your desktop version of R to crash? Everything is running besides that and the check for probability. both of those tests call mboost and after calling it crash.

larskotthoff:
Works for me. Have you tried updating to the latest version of mboost?

Stevo15025:
Just did and still failing. You are on mboost version 2.7? Maybe this is just my computer

larskotthoff:
Yep, just updated.

Stevo15025:
Yay! Success! I guess it is my machine that has a problem with mboost 

Does this all look okay?

larskotthoff:
Thanks, just a few style things:
- space after function parentheses and opening brace
- no need for `BBmisc::` in front of functions
- don't assign something to a variable and then not do anything with it (in the function to get the hyperparameters)

For the hyperparameters, you can even just declare them globally in the file; no need for a function.

Now remind me, this is only needed for the forecasting branch, isn't it? Is that branch ready to be merged? Otherwise I would hold off merging this.

Stevo15025:
Sounds good I'll make these fixes when I get home.

Has anyone reviewed the forecasting branch? I probably need to go and make these style changes in the forecasting branch. Also there are some places I would like people to look over. createLagDiff is pretty messy, but it was the only way I could find to perform one step ahead forecasts with machine learning models. Though I may rewrite it this weekend so that it uses data.table instead of xts. 

This is for the forecasting branch, however the MAX_NUM_DLL error is probably going to arise for the next branch to include a new numeric learner.

Stevo15025:
I also need to make the tutorial fit the mlr tutorial style

Stevo15025:
I'm not sure why, but placing the hyperpars in the test file does not let the hyperpars pass to the test. Which is odd because the global environment should pass to the test (for example `binaryclass.task` does pass through)

larskotthoff:
Ah, you may have to put them into one of the helper files which are always sourced.

Stevo15025:
> Ah, you may have to put them into one of the helper files which are always sourced.

I tried placing hyperpars outside of a function in 88e2d51 but it failed (?!). While not ideal, throwing the hyperpars into a function inside of `helper_learners_all` is not too bad

Stevo15025:
This is all good to go, after review. This is just to break up the classifer tests that were failing in the forecast package until R 3.4 comes out. Then we can set the maximum number of DLLs in jenkins and travis

larskotthoff:
Hmm, we talked about this a bit and are actually tempted to just wait until 3.4 comes out. This avoids merging something now and then reverting it later.

Do you need the forecasting stuff merged before then? That would be the only reason to pull this in.

Stevo15025:
Ehhhh

Would I _like_ forecasting to be merged before 3.4? Ya. 

Will I keel over if it doesn't happen in the next month? I think I'll live.

 that's reasonable, it can wait

larskotthoff:
Ok, let's see when we can review the PR first.

PR #1570, opened 2017-03-03
Removed factors from xgboost classif and reg properties

Stevo15025:


mb706:
As @giuseppec noted: The reasoning for removing the "factors" property should be noted in NEWS:
> Removed "factors" property from "regr.xgboost" and "classif.xgboost" since the xgboost learner does not handle categorial features. Previously they silently converted factors to integer features, which misinterprets the structure underlying the data.

tests/testthat/test_regr_xgboost.R
```diff
@@ -15,18 +15,18 @@ test_that("regr_xgboost", {
...
+    #model = xgboost::xgboost(data = data.matrix(regr.train[,1:3]), verbose = 0L,
     #label = as.numeric(regr.train[,14]),
     #nrounds = 20, objective = "reg:linear", missing = NULL)
-    old.predicts.list[[i]] = predict(model, data.matrix(regr.test[,1:13]))
+    old.predicts.list[[i]] = predict(model, data.matrix(regr.test[,1:3]))
   }
   
   #set.seed(getOption("mlr.debug.seed"))
-  testSimpleParsets("regr.xgboost", regr.df, regr.target, regr.train.inds,
+  testSimpleParsets("regr.xgboost", regr.df[, c(1:3, 14)], regr.target, regr.train.inds,
```
giuseppec: Maybe change this to `regr.df[, !vlapply(regr.df, is.factor)]`? So that it still works if we use a different data set for `regr.df`?

mb706: `regr.bgp`, `regr.blm`, `regr.bgpllm`, `regr.laGP`, `regr.km` work the same way. Though there seems to be a `regr.num.df` also. Is `regr.num.df` the canonical way? In that case the tests mentioned above also should be changed.

mb706: (as used in the test for e.g. `regr.xyf`)

giuseppec: We should then use `regr.num.df` here and open up an issue for the other learners (also maybe for the other classif learners).

tests/testthat/test_regr_xgboost.R
```diff
@@ -15,18 +15,18 @@ test_that("regr_xgboost", {
     if (is.null(parset$verbose)) parset$verbose = 0L
     if (is.null(parset$nrounds)) parset$nrounds = 1L
     if (is.null(parset$objective)) parset$objective = "reg:linear"
-    pars = list(data = data.matrix(regr.train[,1:13]), label = as.numeric(regr.train[,14]))
+    pars = list(data = data.matrix(regr.train[,1:3]), label = as.numeric(regr.train[,14]))
```
giuseppec: here `regr.num.train` and `regr.num.target` might be more appropriate

tests/testthat/test_regr_xgboost.R
```diff
@@ -15,18 +15,18 @@ test_that("regr_xgboost", {
...
     pars = c(pars, parset)
     set.seed(getOption("mlr.debug.seed"))
     model = do.call(xgboost::xgboost, pars)
-    #model = xgboost::xgboost(data = data.matrix(regr.train[,1:13]), verbose = 0L,
+    #model = xgboost::xgboost(data = data.matrix(regr.train[,1:3]), verbose = 0L,
     #label = as.numeric(regr.train[,14]),
     #nrounds = 20, objective = "reg:linear", missing = NULL)
-    old.predicts.list[[i]] = predict(model, data.matrix(regr.test[,1:13]))
+    old.predicts.list[[i]] = predict(model, data.matrix(regr.test[,1:3]))
```
giuseppec: `regr.num.test`

PR #1576, opened 2017-03-06
Expression handling in mlr

jakob-r:
Reopening #1126 after a reverted merge

Still open issues:

* `n` should not refer to the task size but maybe to the train set size.


R/Task_operators.R
```diff
@@ -470,14 +471,17 @@ getTaskWeights = function(task) {
 #' @export
 #' @examples
 #' task = makeClassifTask(data = iris, target = "Species")
-#' getTaskDictionary(task)
-getTaskDictionary = function(task) {
+#' getTaskDictionary(task, subset = NULL)
+getTaskDictionary = function(task, subset = NULL) {
+  getSubsetSize = function(subset) if (is.logical(subset)) sum(subset) else length(subset)
```
jakob-r: There was a PR and we now have `checkTaskSubset`. This might want to be used here.

larskotthoff:
What's the status here @jakob-r ?

jakob-r:
The status was that it is quite hard to make sure that `p` and `n` are the actual values for number of features and number of observations because of all the wrappers and resampling going on. Some heavy lifting would be required here. So we will do it in mlrNG?

larskotthoff:
Maybe something for the next workshop?

larskotthoff:
Looks like this isn't going to happen. Unless somebody objects, I'll close this soon.

jakob-r:
I guess you are right. But please do not delete the branch as it contains pretty much code that might be valuable to deal with expressions outside of mlr (in PH they are allowed at least)

larskotthoff:
Ok, closing and leaving branch.

PR #1580, opened 2017-03-06
Closes #1397 Drop option for BenchmarkResult operators

### LINKED ISSUE #1397: for the getBMR operators there should be a drop option
so you dont always get lists of lists of lists.
probably the default for drop is "off"

florianfendt:
For our BenchmarkResult operators, that used `getBMRObjects` we always got a list of lists like this:
`res[task.ids][learner.ids]`
The drop argument checks if only one task and/or learner was passed to benchmark. If that's the case
we unnest the list accordingly and drop the learner's/task's id from the lists names


berndbischl:
did you push MAN files? please dont do. you dont have to adapt the PR now, but please dont do that

berndbischl:
told @florianfendt what to do, dont merge

florianfendt:
@berndbischl 
Please review again. Does what you wanted now

berndbischl:
@florianfendt where is your partner review?

PR #1581, opened 2017-03-06
mlr class and entity overview

mb706:
Just to make you aware that this file exists, feel free to build some documentation (e.g. a graph of relationships between mlr objects) based on it.


mb706:
P.S. Apparently github renders .org files but removes newlines; The file is supposed to be viewed in raw text format.

PR #1583, opened 2017-03-06
tuneParams now works with ParamSet that exceeds messagef char limit.

### LINKED ISSUE #1528: tuneParams crashes when par.set is too big and show.info == TRUE
This may be an artificial example, but it gets annoying when using big `modelMultiplexer` models with many parameters. 
```R
> lrn = makeLearner("classif.rknn")
> ctrl = makeTuneControlRandom(maxit=3)
> # create long list of learner params
> psLength = 200
> longLearnerParams = do.call(c, lapply(seq_len(psLength), function(x) {
+   makeParamSet(makeIntegerLearnerParam(paste0('some.parameter', x), 1, 10))
+ }))
> lrn$par.set = c(lrn$par.set, longLearnerParams)
> tuneParams(lrn, pid.task, cv5, par.set = longLearnerParams, control = ctrl, show.info=TRUE)
[Tune] Started tuning learner classif.rknn for parameter set:
Error in sprintf(...) (from tuneParams.R#108) : 'fmt' length exceeds maximal format length 8192
```
I suggest not unnecessarily calling `sprintf` (via `messagef`) [here](https://github.com/mlr-org/mlr/blob/02b581f85304a210f77273de07243914ad2602ee/R/tuneParams.R#L107) (and just use `message` instead?).

mb706:
Fixes #1528

### DIFF
--- a/R/tuneParams.R
+++ b/R/tuneParams.R
@@ -104,7 +104,7 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
   opt.path = makeOptPathDFFromMeasures(par.set, measures, include.extra = (control$tune.threshold))
   if (show.info) {
     messagef("[Tune] Started tuning learner %s for parameter set:", learner$id)
-    message(printToChar(par.set))  # using message() since this can go over the char limit of messagef().
+    message(printToChar(par.set))  # using message() since this can go over the char limit of messagef(), see issue #1528
     messagef("With control class: %s", cl)
     messagef("Imputation value: %g", control$impute.val)
   }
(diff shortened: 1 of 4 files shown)


R/tuneParams.R
```diff
@@ -104,7 +104,7 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
   opt.path = makeOptPathDFFromMeasures(par.set, measures, include.extra = (control$tune.threshold))
   if (show.info) {
     messagef("[Tune] Started tuning learner %s for parameter set:", learner$id)
-    messagef(printToChar(par.set))
+    message(printToChar(par.set))
```
berndbischl: Please add a comment here, otherwise people might think of changing it back.

Otherwise this looks good. If this problem occours mutiple times we can also fix this in BBmisc upstream

berndbischl:
will add the comment myself now and then merge

mb706:
Ninja'd

berndbischl:
keep your dirty paws of it now

PR #1587, opened 2017-03-06
Log fun of tuneParams and selectFeatures now without memory stats.

### LINKED ISSUE #624: tuning: Export a lightweight version of logFunTune
When doing experiments with tuneParams() and fast learners / small datasets, the `gc()` calls made through `logFunTune` to report memory usage tend to make up >80% [1] of the runtime. Since I want to follow the optimization process but don't need the memory output, I ended up copying the logFunXX functions with the `gc()` calls removed. I can imagine this to be a common case, however, so I think giving an easier method for skipping memory reporting would be useful.

I propose
- exporting a lightweight version of `logFunTune` that does not report memory usage, or
- creating an `tune.report.memory` option by which the default log function behaviour can be changed

[1] According to `Rprof()`. Another call of `gc()` is implied by the `system.time()` call inside `evalOptimizationState()` (since it defaults to gcFist=TRUE). adding `gcFist=FALSE` here also improves runtime (seemingly with no downside besides inaccurate time reporting?)

mb706:
Displaying memory stats slows the tuning process down if showing output (sometimes significantly, for fast learners). Optionally, giving log.fun = "memory" now gives the old output. Fixes #624

Note: In #624 @berndbischl mentions a different idea how to implement this option; however while trying to implement it I found this way to be more elegant (also it works without adding another argument to `makeTuneControl`).


R/FeatSelControl.R
```diff
@@ -60,6 +60,8 @@
 #'   Default is none.
 #' @param log.fun [\code{function} | \code{NULL}]\cr
 #'   Function used for logging. If set to \code{NULL}, the internal default will be used.
+#'   If set to "memory", a slightly extended version of the default will be used that also displays
```
berndbischl: please use dQuote

R/FeatSelControl.R
```diff
@@ -60,6 +60,8 @@
 #'   Default is none.
 #' @param log.fun [\code{function} | \code{NULL}]\cr
```
berndbischl: apparently "log.fun" can now also be a string. thats not documented in the type

R/TuneControl.R
```diff
@@ -47,8 +47,10 @@
 #' @param tune.threshold.args [\code{list}]\cr
 #'   Further arguments for threshold tuning that are passed down to \code{\link{tuneThreshold}}.
 #'   Default is none.
-#' @param log.fun [\code{function} | \code{NULL}]\cr
-#'   Function used for logging. If set to \code{NULL}, the internal default will be used.
+#' @param log.fun [\code{function} | \code{character(1)}]\cr
+#'   Function used for logging. If set to \dQuote{default}, the internal default will be used.
+#'   If set to \dQuote{memory}, a slightly extended version of the default will be used that also displays
+#'   memory usage.
```
jakob-r: You use it twice. Probably use a roxygen template than.

R/FeatSelControl.R
```diff
@@ -109,12 +111,14 @@
 NULL
 
 makeFeatSelControl = function(same.resampling.instance, impute.val = NULL, maxit, max.features,
-  tune.threshold = FALSE, tune.threshold.args = list(), log.fun = NULL, ..., cl) {
+  tune.threshold = FALSE, tune.threshold.args = list(), log.fun = "default", ..., cl) {
```
jakob-r: Why the detour with the string. We could use the function names directly and put the defaults in the signature. But then we have to export these functions. :thinking:

R/FeatSelControl.R
```diff
@@ -109,12 +111,14 @@
...
 
   maxit = asCount(maxit, na.ok = TRUE, positive = TRUE)
   max.features = asCount(max.features, na.ok = TRUE, positive = TRUE)
-  if (is.null(log.fun))
+  if (identical(log.fun, "default"))
```
jakob-r: why not `log.fun == "default"`. somehow easier to read and more consistent with the rest of the code.

R/FeatSelControl.R
mb706: Would be (imho unnecessary) complexity for the user. There would be two more functions on the export list which have only the specific purpose of being fed to `log.fun`.

R/FeatSelControl.R
mb706: Leads to error if log.fun is a function (since `==` doesn't overload to functions)

R/TuneControl.R
mb706: It wasn't a template before (when it already was a duplicate). Is there a reason it was not?

jakob-r: lazyness

R/FeatSelControl.R
jakob-r: I guess you are right

R/FeatSelControl.R
jakob-r: Ah, i see.

R/FeatSelControl.R
```diff
@@ -58,8 +58,10 @@
 #' @param tune.threshold.args [\code{list}]\cr
 #'   Further arguments for threshold tuning that are passed down to \code{\link{tuneThreshold}}.
 #'   Default is none.
-#' @param log.fun [\code{function} | \code{NULL}]\cr
-#'   Function used for logging. If set to \code{NULL}, the internal default will be used.
+#' @param log.fun [\code{function} | \code{character(1)}]\cr
+#'   Function used for logging. If set to \dQuote{default}, the internal default will be used.
+#'   If set to \dQuote{memory}, a slightly extended version of the default will be used that also displays
+#'   memory usage.
```
larskotthoff: Can you mention here what the internal default is please?

R/TuneControl.R
larskotthoff: See above -- mention what the default is.

larskotthoff:
@mb706 There are merge conflicts. Could you have a look please?

mb706:
News:
> removed memory stats from "tuneParams" and "selectFeatures" default output for performance reasons. Using a control object with "log.fun" = "memory" restores this.

PR #1589, opened 2017-03-06
Issue #1417: performance NA when measure not applicable

### LINKED ISSUE #1417: performance NA when measure not applicable
assume the following:

you are benchmarking a number of learner on a number of tasks, and measuring different measures.
now, for some learners it might happen that a certain measure is not applicable, eg there is one learner that cannot predict probs, but you also have AUC in your measures. 

what happens now is that we exit with a hard error and you get nothing.

example:

```
load_all()

tasks = list(sonar.task)

lrn1 = makeLearner("classif.rpart", predict.type = "prob")
lrn2 = makeLearner("classif.logreg")
lrns = list(lrn1, lrn2)
ms = list(mmce, auc)

b = benchmark(lrns, tasks, resampling = hout, measures = ms)
print(b)
```

results in  (assume lrn2 cannot predict probs, i was too lazy)
```
Error in FUN(X[[i]], ...) : 
  Measure auc requires predict type to be: 'prob'!
```

that seems too conservative. eg. if lrn2 would crash one can also get a full benchmark object but with NAs.
I think best would be if "performance" could also return an NA here? with an option for this?
maybe with a warning?

giuseppec:
Fixes #1417

### DIFF
--- a/tests/testthat/test_base_performance.R
+++ b/tests/testthat/test_base_performance.R
@@ -52,7 +52,7 @@ test_that("performance", {
 test_that("performance is NA if 'on.measure.not.applicable' is not 'stop'", {
   default = getMlrOption("on.measure.not.applicable")
   vals = c("quiet", "warn", "stop")
-  for(i in vals) {
+  for (i in vals) {
     configureMlr(on.measure.not.applicable = i)
     lrn = makeLearner("classif.qda", predict.type = "response")
     mod = train(lrn, sonar.task)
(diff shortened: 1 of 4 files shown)


tests/testthat/test_base_performance.R
```diff
@@ -49,6 +49,24 @@ test_that("performance", {
   expect_equal(as.numeric(r$aggr["mmce.test.mean"]), as.numeric(r$aggr["mym.train.mean"]))
 })
 
+test_that("performance is NA if 'on.measure.not.applicable' is not 'stop'", {
+  default = getMlrOption("on.measure.not.applicable")
+  vals = c("quiet", "warn", "stop")
+  for(i in vals) {
```
berndbischl: style guide: after for there is a SPACE

giuseppec:
@berndbischl I added a test for benchmark.

PR #1594, opened 2017-03-06
get, has and list measure Properties

### LINKED ISSUE #1508: hasMeasureProperties
We have `hasLearnerProperties` but not `hasMeasureProperties`

(same goes for `getLearnerProperties`)

ja-thomas:
Fixes #1508.

I don't think splitting measure properties by task type makes much sense as only `req.prob` is not possible for all task types.

Also `cluster` property was missing in the description of Measure

### DIFF
--- a/R/Measure_properties.R
+++ b/R/Measure_properties.R
@@ -1,0 +1,43 @@
+#' @title Query properties of measures.
+#'
+#' @description
+#' Properties can be accessed with \code{getMeasureProperties(measure)}, which returns a
+#' character vector.
+#'
+#' The measure properties are defined in \code{\link{Measure}}
+#' @template arg_measure
+#' @param props [\code{character}]\cr
+#'   Vector of properties to query.
+#' @return \code{getMeasureProperties} returns a character vector with measure properties.
+#'  \code{hasMeasureProperties} returns a logical vector of the same length as \code{props}.
+#' @name MeasureProperties
+#' @rdname MeasureProperties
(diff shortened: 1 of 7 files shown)


R/Measure_properties.R
```diff
@@ -0,0 +1,43 @@
...
+hasMeasureProperties = function(measure, props) {
+  assertClass(measure, classes = "Measure")
+  assertSubset(props, listMeasureProperties())
+  props %in% getMeasureProperties(measure)
+}
+
+#' @title List the supported measure properties.
+#'
+#' @description
+#'   This is useful for determining which measure properties are available.
```
berndbischl: no indentation here

tests/testthat/test_base_measures.R
```diff
@@ -786,6 +786,16 @@ test_that("getDefaultMeasure", {
...
 })
 
+test_that("measure properties", {
+  #hasMeasureProps yields correct properties
+  expect_true(all(vlapply(listMeasures(create = TRUE),
+    function(m) all(hasMeasureProperties(m, m$properties)))))
+  props = listMeasureProperties()
+  #all props exist in mlr$measure.properties
+  expect_true(all(vlapply(listMeasures(create = TRUE),
+    function(m) all(getMeasureProperties(m) %in% props))))
```
larskotthoff: Could you check here whether the returned list of properties is not empty please? Same above for the other test.

PR #1595, opened 2017-03-06
Fixes pretty.names argument for several plotting functions. Closes #1523

### LINKED ISSUE #1523: plotBMRBoxplots fails when the same learner is used multiple times (with different ids)

florianfendt:
Output of plots were confusing when pretty.names was on and learners had the same id.
Now we stop with an error message which identifies the (first) problematic learner.

Affected functions:
* plotResiduals
* plotBMRBoxplots
* plotBMRSummary
* plotBMRRanksAsBarCharts

Connected issue: #1523


R/helpers.R
```diff
@@ -85,3 +85,15 @@ replaceDupeMeasureNames = function(measures, x = "id") {
...
 }
+
+# checks for duplicated entries in learner.names and stops
+# with error message containing the learner name that appeared more
+# than once
+checkDuplicatedLearnerNames = function(learner.names) {
+  dupl = duplicated(learner.names)
+  if (any(dupl)) {
+    dupl.learner = learner.names[which.first(dupl)]
+    stop(sprintf("Learner short names are not unique for: %s ",
```
ja-thomas: `stopf()`

ja-thomas: Also add a sentence "set `pretty.names=FALSE` to solve this" or something like that

R/helpers.R
```diff
@@ -85,3 +85,15 @@ replaceDupeMeasureNames = function(measures, x = "id") {
...
   unlist(meas.names)
 }
+
+# checks for duplicated entries in learner.names and stops
+# with error message containing the learner name that appeared more
+# than once
+checkDuplicatedLearnerNames = function(learner.names) {
+  dupl = duplicated(learner.names)
+  if (any(dupl)) {
+    dupl.learner = learner.names[which.first(dupl)]
```
jakob-r: what if multiple names are duplicated? the error will just return one of the pairs. what about `unique(learner.names[dupl]))`.

R/helpers.R
```diff
@@ -85,3 +85,15 @@ replaceDupeMeasureNames = function(measures, x = "id") {
...
+
+# checks for duplicated entries in learner.names and stops
+# with error message containing the learner name that appeared more
+# than once
+checkDuplicatedLearnerNames = function(learner.names) {
+  dupl = duplicated(learner.names)
+  if (any(dupl)) {
+    dupl.learners = unique(learner.names[dupl])
+    stopf("Learner short names are not unique for: %s. \n  Set 'pretty.names = FALSE' to resolve this.",
+      collapse(dupl.learners))
```
larskotthoff: The error message is very specific. I would either make it more generic, or move this function to a different file (plotUtils.R?).

florianfendt:
@larskotthoff 
I put the checkDuplicated... function into a new file now: utils_plot.R

PR #1600, opened 2017-03-07
fix printer for listFeatureFilter

### LINKED ISSUE #1592: listFilterMethods looks not good in printer
because we simply print the table, and the "desc" slot is too long

give result and S3 class and call clipString on the "desc"

ja-thomas:
fixes #1592 and #1600 

Magic number :(


R/Filter.R
```diff
@@ -75,7 +75,7 @@ listFilterMethods = function(desc = TRUE, tasks = FALSE, features = FALSE) {
     package = vcapply(extractSubList(filters, "pkg"), collapse)
   )
   if (desc)
-    df$desc = extractSubList(filters, "desc")
+    df$desc = clipString(extractSubList(filters, "desc"), len = 40)
```
jakob-r: Are you sure you want to clip the string here? I think @berndbischl pointed to a separate S3 class for the listFilterMethods result and clip the desc just in the printer

ja-thomas: yup, you're totally correct...

R/Filter.R
```diff
@@ -55,9 +55,11 @@ makeFilter = function(name, desc, pkg, supported.tasks, supported.features, fun)
 #'  Provide information on supported tasks.
 #' @param features [\code{logical(1)}]\cr
 #'  Provide information on supported features.
+#'  @param include.deprecated [\code{logical(1)}]\cr
```
berndbischl: indent looks wrong and please always mention default as last sentence in docs

berndbischl:
do we really need that option to include deprecated? just dont show them at all?

ja-thomas:
> do we really need that option to include deprecated? just dont show them at all?

that's why the default is set to `FALSE`. But an option to list them might be useful.

PR #1605, opened 2017-03-07
use checkTaskSubset

### LINKED ISSUE #1472: some function have a "subset" arg which must be consistently defined
- we have arg_subset for roxygen

- we should have a checkSubset function to encapsulate the arg check

- example where its not perfect "getTaskCosts" just says "subset" should be integer (but normally we allow "logical" too)

- search for other instances of subset being an argument

ja-thomas:
This fixes #1472 

The more important point was to introduce a `checkTaskSubset` function.

This touches some base code so review carefully.


berndbischl:
@ja-thomas 
in general prefer 
x = asInt(x) over assertInt

but here all seems ok

PR #1607, opened 2017-03-07
enable test for sum of predicted probabilities (closes #1017)

### LINKED ISSUE #1017: RLearners / tests: not all probs sum to 1
Here we have an now disabled check

https://github.com/mlr-org/mlr/blob/master/tests/testthat/helper_learners_all.R

```
# FIXME: the "sum to 1" apparently does not work for all learners?
# I can see differences up to 0.1 for some cases, reported in issue #
# expect_equal(info = info, unname(rowSums(probdf)), rep(1, NROW(probdf)), use.names = FALSE)
```

we normally are not that for off from 1, but like i said, im some cases 1e-1. 
i do think that this is too much.

do we want to repair this in mlr?

larskotthoff:


berndbischl:
i have nothing againts using the checkmate function, but in this case please really read (and possiby use?) expect_equal. read ?expect_equal, and ?all_equal and see that both support "tolerance"!

jakob-r:
Good to know but in this case we would lose the possibility to check for only smaller than 1.

berndbischl:
Why so you think the deviation should only go in one direction?  That's a bit weird. Expect-equal with tolerance is made EXACTLY for this type of hopefully numerical problems as here.

Using something else is just non-standard and somewhat confusing.

larskotthoff:
Ok, why is the tolerance argument for expect_equal not documented? It's only in the examples. not in the list of arguments.

PR #1615, opened 2017-03-08
Refactor make task desc

### LINKED ISSUE #1613: makeTaskDesc call in make*Class is bad code, refactor
look at this, from makeClassifTask

  task$task.desc = makeTaskDesc.ClassifTask(task, id, target, positive)

this calls makeTaskDesc on a *partially* constructed task. that is *really* bad code. probably we want a cleaner API for makeTaskDesc, where we simply pass in data and all extra information

berndbischl:
fixes #1613 

what i did:

- makeTaskDesc was really bad, it was constructed based an a partially constructed s3 task
- i now simply construct it from data and further info, much cleaner
- also made the name of the constructor more consistent with makeTask


berndbischl:
@jakob-r what I did is purely internal refactoring. You are now requesting to change exported api. That's is also muddling up issues which we agreed we don't want anymore.

If you think that's is better open an orthogonal issue. It still will potentially only create problems for users without too many benefits. But: different issue.

larskotthoff:
I'm fine with the changes. Agreed that for something purely internal we shouldn't worry too much about these things.

larskotthoff:
@jakob-r looks like you broke it.

jakob-r:
I am on it.

jakob-r:
# Warning
This change will break compatibility with Tasks generated with old mlr versions because the internal class names of the task descriptions are changed.

mllg:
Looks good. Can be merged as soon as travis agrees that I did not break anything during my small cleanup/speedup.

PR #1631, opened 2017-03-08
getLearnerProperties for ModelMultiplexer (fixes #791)

### LINKED ISSUE #791: ModelMultiplexer should have the 'properties' of the currently active learner
When creating a `ModelMultiplexer` from learners with different `$properties`, the resulting object has only the `$properties` that result from the set intersection of all the involved learner's properties. This results in ModelMultiplexer object being overly restrictive IMHO:

``` R
> lrn = makeModelMultiplexer(list(makeLearner("regr.bdk"), makeLearner("regr.lm")))                                                                                                                                                         
> lrn = setHyperPars(lrn, selected.learner = "regr.lm")                                                                                                                                                                                     
> train(lrn, bh.task)                                                                                                                                                                                                                       
Error in checkLearnerBeforeTrain(task, learner, weights) : 
  Task 'BostonHousing-example' has factor inputs in 'chas', but learner 'ModelMultiplexer' does not support that!
```

(even though the involved `"regr.lm"` learner is able to handle the `bh.task`:

``` R
> train(makeLearner("regr.lm"), bh.task)                                                                                                                                                                                                    
Model for learner.id=regr.lm; learner.class=regr.lm
Trained on: task.id = BostonHousing-example; obs = 506; features = 13
Hyperparameters: 
```

)
This behaviour gets in the way when using a `PreprocWrapper` that removes missings depending on which learner is active.

mllg:


berndbischl:
looks good. one very little thing: i usually like it if a very similar bug as reported in the issue is added as a unit test. but you could argue that yo did that here.

maybe check 2 min yourself, then merge

Issue #1522, opened 2017-02-15
Imputation on logical values


ja-thomas:
```
> x = data.frame(a = c(TRUE, FALSE, NA))
> 
> str(x)
'data.frame':	3 obs. of  1 variable:
 $ a: logi  TRUE FALSE NA
> 
> y = impute(x, cols = list(a = imputeConstant("__miss__")))
> 
> str(y$data)
'data.frame':	3 obs. of  1 variable:
 $ a: chr  "TRUE" "FALSE" "__miss__"
```

When we impute a logical column with a constant value we get back an character, wouldn't a factor make more sense?

berndbischl:
yeah, looks like a bug

berndbischl:
agreeing with that this should result in a factor

PR #1635, opened 2017-03-09
Fix for 1522 (Impute types)

mllg:


mllg:
Also: Detect if storage type has changed and throw an exception if this is the case (e.g., was numeric, now is logical for re-imputation).

jakob-r:
#1522

jakob-r:
> Running examples in ‘mlr-Ex.R’ failed

jakob-r:
The changes made a change in the example necessary. 
If a column of a data.frame only contains NAs they are not of any specific type so the imputation will change it's type and create an error message - the one that we introduce here.
Is that a behavior that we want?

jakob-r:
We talked and @mllg convinced us that this is a corner case that only happens if the user manually imputes/predicts on a data.frame with all-NA-columns. But thanks to the example code a solution is already given.

PR #1638, opened 2017-03-09
add fd.scalars

smilesun:
add another field for FDATask


berndbischl:
request other persons before me please

PR #1644, opened 2017-03-09
move costs to task.desc

### LINKED ISSUE #1614: costsenstask: costs should be stored in taskdesc
currently they are stored in task

also take a look at changeData, must be adapted then

ja-thomas:
Fixes #1614 

Base tests and `filter=cost` are running. We seem to consistently use the getter function so not much had to be changed.

Not sure what about `changeData` still works without a change.

lets see if all tests are running

### DIFF
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -402,7 +402,7 @@ changeData = function(task, data, costs, weights) {
   if (missing(data))
     data = getTaskData(task)
   if (missing(costs))
-  costs = getTaskCosts(task)
+    costs = getTaskCosts(task)
   if (missing(weights))
     weights = task$weights
   task$env = new.env(parent = emptyenv())
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -407,9 +407,6 @@ changeData = function(task, data, costs, weights) {
     weights = task$weights
   task$env = new.env(parent = emptyenv())
   task$env$data = data
-  # FIXME: I hate R, this is all bad
-  if (!is.null(costs))
-    task$env$costs = costs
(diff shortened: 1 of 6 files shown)


R/Task_operators.R
```diff
@@ -367,7 +367,7 @@ getTaskCosts = function(task, subset = NULL) {
   if (task$task.desc$type != "costsens")
     return(NULL)
   subset = checkTaskSubset(subset, size = task$task.desc$size)
-  d = task$env$costs[subset, , drop = FALSE]
+  d = task$task.desc$costs[subset, , drop = FALSE]
```
larskotthoff: Use getter `getTaskDescription` here.

berndbischl:
changeData MUST be touched, you still have env$costs in there?

if this now does NOT break tests, we need to ADD a test

i would also remove (try to?) "costs" from the signature of changeData

ja-thomas:
> changeData MUST be touched, you still have env$costs in there?

I don't think we need these lines anymore at all, as we have the costs in the Taskdescription and not in the environment anymore. This line is not triggered anymore as either

a) `cost` in `changeData` is missing, then we call `getTaskCosts` and either get `NULL` or the complete costs.

b) `cost` is not missing then we already got possibly subsetted costs and pass it to `makeCostSensTaskDesc` unchanged.


> if this now does NOT break tests, we need to ADD a test

I'm not sure if I can create a test to break this at all.

> i would also remove (try to?) "costs" from the signature of changeData

that might be difficult as we need to subset `costs` which is done in `getTaskCosts` and we do not pass the subset information to `changeData`.

PR #1651, opened 2017-03-10
added helper to suppress specific warnings

mllg:


mllg:
Can be used to suppress **specific** warnings, i.e. for survival ranger.

Issue #751, opened 2016-02-19
Make information about learners easily accessible.


mb706:
It would be nice if there were more information about the learners easily accessible, regarding
(**1**) the thing they actually do (e.g. `help(the_function_actually_called_inside_trainLearner)`). Right now one would need to use the appendix of the tutorial to get this information.
(**2**) maybe specific information about their workings within mlr or user experience.

**To (1):** It would be useful to add an `mlrHelp()` function that accepts a learner (or learner string) and calls help() for the actual function being called inside `trainLearner` or `predictLearner` (whichever is the more important one?). If the function being called depends on the parameters, or if both `trainLearner` and `predictLearner` call interesting functions (a function being 'interesting' if something in the ParamSet influences its behaviour), it should prompt the user to choose between different help pages. I would imagine the user interface to be something like:

```
> mlrHelp("classif.pamr")
pamr: ** Nearest shrunken centroid **
Choose help page:
1 : pamr::pamr.train
2 : pamr::pamr.predict
...:[user input]
```

Internally this would not be more than listing all the 'interesting' functions in a learner-object $property and having the `mlrHelp` function do the mentioned prompt and calling `help()`.

**To (2):** This seems to already exist in the form of `learner$notes`, but apparently, if someone has something to add to that, they would have to fork the repo and make a pull request? One could use the Github mlr Wiki (or something else that is publicly editable) to collect information about different learners; things that may be of interest to the user, maybe even have something about sensible parameter settings if a contributor has some expertise. During the build process, this information could be curl'd from the github wiki and included as data in the R package and made available also via `mlrHelp`.

jakob-r:
This is a quite extensive suggestion. I also note that the different learners could be documented more detailed.
I don't like the idea of a command line help interface so much as it is really unhandy. I prefer a more detailed and automatically generated documentation in the mlr tutorial. Obviously this should be mentioned in the internal documentation to not leave the user clueless.
Information about what parameters are sensible might be included in the future from other sides. 
What might be a good idea is to have a link to the internal documentation of the desired learner, so that `mlrHelp("classif.lqa")` would link you to `?lqa::lqa.default`  But this is also a good amount of not so demanding work.

mb706:
Something in the direction of
```R
# assuming learner$callees is a character vector containing the functions of
# package learner$package that are 'interesting'.
mlrHelp = function(learner, ...) {
  learner = checkLearner(learner, ...)
  if (is.null(learner$callees)) {
[...]

PR #1657, opened 2017-03-10
Add function learnerHelp

mb706:
Re #751: Give the user an easy way to look at help pages associated with a learner. This is especially useful to find out what parameters do.

For this to be nicely implemented, every learner should have a '`callees`' slot. An example how the callees is supposed to work can be seen in [RLearner_classif_blackboost.R](https://github.com/mlr-org/mlr/blob/78299468264ad26a38e67ea6c766378d6ccd1ae4/R/RLearner_classif_blackboost.R#L37): Mention all the functions called in `trainLearner.xxx` and `predictLearner.xxx` that are potentially interesting, or explain parameters in the learner's `ParamSet`.

The only ambiguity arising sometimes is whether to mention the `predict.LEARNEROBJECT` function that many packages supply. I suggest adding it to the `callees` only if it gives further insight into the learner, or if it takes parameters from the `ParamSet` (usually when a parameter has `when == "both"` or `when == "predict'`).


ja-thomas:
That is pretty awesome!

ja-thomas:
We need to handle the special characters in the pkg name. e.g. `"!earth"`

MariaErdmann:
These learners had the `package =!packagename`:
kknn, mda, penalized learners, dcSVM
And there were two things I stumbled over:
Why does the `fnn` learner is called `fnn`learner when it actually calls function `knn`?
Why did the learner `gaterSVM` had the package `e1071` in `packages = ...`. I deleted it. Tests were still working (but probabl I overlooked something.)

mb706:
- added function `learnerParamHelp`. Usage:
```R
> learnerParamHelp("classif.knn", "k")
  *k*:
number of neighbours considered. 
```
- added (not exported) function `listUndocumentedPars` to list the learners / parameters where the documentation can not be easily parsed:
```R
> ll = listLearners()
> for (l in ll$class[ll$type == "classif"]) if (length(u <- mlr:::listUndocumentedPars(l)) > 0) catf("%s: %s", l, collapse(u))
classif.bdk: xdim,ydim
classif.binomial: link
classif.blackboost: family,custom.family.definition,Binomial.link
classif.bst: minsplit,minbucket,cp,maxcompete,maxsurrogate,usesurrogate,xval,surrogatestyle,maxdepth,Learner
classif.cvglmnet: alpha,exact,nlambda,lambda.min.ratio,standardize,intercept,thresh,dfmax,pmax,exclude,penalty.factor,lower.limits,upper.limits,maxit,type.logistic,type.multinomial
classif.earth: thres,link,maxit,Adjust.endspan,Force.weights,Use.beta.cache,Force.xtx.prune,Get.leverages,Exhaustive.tol
classif.evtree: pmutatemajor,pmutateminor,pcrossover,psplit,pprune
classif.featureless: method
classif.gamboost: family,custom.family.definition,Binomial.link
classif.gausspr: sigma,degree,scale,offset,order
classif.glmboost: family,custom.family.definition,Binomial.link
classif.glmnet: s,exact
classif.h2o.deeplearning: use_all_factor_level
classif.IBk: I,F,K,E,W,X,A,output-debug-info
classif.J48: U,O,C,M,R,N,B,S,L,A,J,Q,output-debug-info
classif.JRip: F,N,O,D,S,E,P,output-debug-info
classif.ksvm: sigma,degree,scale,offset,order
classif.lda: predict.method
classif.lqa: lambda,gamma,alpha,c,a,lambda1,lambda2
classif.lssvm: sigma,degree,scale,offset,order
classif.mda: start.method,tries,criterion
classif.multinom: maxit,rang,decay,trace,abstoll,reltoll
classif.nnet: abstoll,reltoll
classif.nnTrain: max.number.of.layers
classif.OneR: B,output-debug-info
classif.pamr: threshold.predict
classif.PART: C,M,R,N,B,U,J,Q,output-debug-info
classif.penalized.fusedlasso: lambda1,lambda2
classif.penalized.lasso: lambda1
classif.penalized.ridge: lambda2
classif.plr: cp.type
classif.plsdaCaret: method
classif.qda: predict.method
classif.rda: lambda,gamma,kernel,SimAnn
classif.xgboost: booster,silent,eta,gamma,max_depth,min_child_weight,subsample,colsample_bytree,colsample_bylevel,num_parallel_tree,lambda,lambda_bias,alpha,objective,eval_metric,base_score,max_delta_step,nthread,sample_type,normalize_type,rate_drop,skip_drop
classif.xyf: xdim,ydim
```

mb706:
Some thoughts:
- The `callees` of the `classif` learners are done, the other learners still need to be done.
- The learner constructor writes all the information it manages to extract into a `help.list` slot; this gives the learner constructor the option to write additional info into this list. Some of the parameters are constructs of the learner itself (e.g. the `se.method` of the `randomForest`), some are renamed parameters (e.g. calling an argument `Learner` instead of `learner` to prevent a name collision), some of the learners don't provide good information about their parameters (mostly RWeka learners)
- if the second arg of `learnerParamHelp` is missing, all parameter help is displayed (which can turn out to be a lot), not sure if this is good.
- If anyone has better ideas how to display the parameter help, feel free to implement. It would for example be possible to extract the html help, write it into a temporary file, and open it with `browseURL` (just as normal help files are opened; I guess this is more comfortable if a long text is displayed). Another possibility is to somehow attach the help to the ParamSet, so that a part of the help is displayed when displaying parameters (probably not such a good idea though)
- wrappers and preprocessing is not covered, though definitely possible as a next step.

mb706:
@MariaErdmann:
- the `!packagename` thing is solved.
- the `fnn` learner uses the "`FNN`" ('fast nearest neighbour') package.
- gaterSVM doesn't call e1071 directly (SwarmSVM imports e1071 though), so I assume it is ok to remove it from the `package` list

mb706:
The remaining undocumented parameters (of the classifiers), I believe this is very close to what magic can do for us:
```R
> ll = listLearners()
> for (l in ll$class[ll$type == "classif"]) if (length(u <- mlr:::listUndocumentedPars(l)) > 0) catf("%s: %s", l, collapse(u))
classif.blackboost: custom.family.definition,Binomial.link
classif.bst: Learner
classif.cvglmnet: exact
classif.earth: maxit,Adjust.endspan,Force.weights,Use.beta.cache,Force.xtx.prune,Get.leverages,Exhaustive.tol
classif.evtree: pmutatemajor,pmutateminor,pcrossover,psplit,pprune
classif.featureless: method
classif.gamboost: custom.family.definition,Binomial.link
classif.gausspr: sigma,degree,scale,offset,order
classif.glmboost: custom.family.definition,Binomial.link
classif.h2o.deeplearning: use_all_factor_level
classif.IBk: I,F,K,E,W,X,A,output-debug-info
classif.J48: U,O,C,M,R,N,B,S,L,A,J,Q,output-debug-info
classif.JRip: F,N,O,D,S,E,P,output-debug-info
classif.ksvm: sigma,degree,scale,offset,order
classif.lda: predict.method
classif.lqa: gamma,alpha,c,a,lambda1,lambda2
classif.lssvm: sigma,degree,scale,offset,order
classif.multinom: abstoll,reltoll
classif.nnet: abstoll,reltoll
classif.nnTrain: max.number.of.layers
classif.OneR: B,output-debug-info
classif.pamr: threshold.predict
classif.PART: C,M,R,N,B,U,J,Q,output-debug-info
classif.plr: cp.type
classif.qda: predict.method
classif.rda: kernel,SimAnn
classif.xgboost: booster,silent,eta,gamma,max_depth,min_child_weight,subsample,colsample_bytree,colsample_bylevel,num_parallel_tree,lambda,lambda_bias,alpha,objective,eval_metric,base_score,max_delta_step,nthread,sample_type,normalize_ty
pe,rate_drop,skip_drop
```
The rest are
1. bugs (abstoll, reltoll for multinom, nnet)
2. things that I believe are bugs (kernel, SimAnn for rda)
3. just not documented (all the RWeka stuff)
4. renamed parameters (e.g. predict.method for lda, Learner for bst)
5. documented in a table within a table, apparently (xgboost, the missing from ksvm, lssvm, gausspr)

No 4 can be solved if there is some kind of table of renamed parameters. Someone who knows the XML package (maybe @giuseppec?) could look at 5. I guess for 3. we'd have to copy the docu from the web, or just leave it like this.

mb706:
Note: renamed the 'earth' parameter `thres` to `thresh` (as that is the proper name).

mb706:
I'd say from my side the branch is feature-complete. It introduces `learnerHelp` and `learnerParamHelp`.

### Examples
```R
> learnerHelp("classif.qda")
Choose help page:
1 : qda
2 : predict.qda
0 : cancel
...:
```

```R
> learnerParamHelp("classif.qda")
Documentation for the following parameters is present:
- method (MASS::qda)
- nu (MASS::qda)
The following parameters are missing documentation:
  predict.method
```

```R
> learnerParamHelp(makeLearner("classif.qda", nu = 10), "nu")
  *nu*:
     Type len Def   Constr Req Tunable Trafo
1 numeric   -   5 2 to Inf   Y    TRUE     -
Used: train.
Requires: method == "t"
Value: 10
Argument of: MASS::qda

degrees of freedom for method = "t". 

```

```R
> learnerParamHelp("classif.qda", "predict.method")
  *predict.method*:
      Type len     Def                      Constr Req Tunable Trafo
1 discrete   - plug-in plug-in,predictive,debiased   -    TRUE     -
Used: predict.
No documentation found. Try to consult learnerHelp("classif.qda").
```

### Missing Callees 
The `callees` for the learners that are not regression learners are still missing. Someone should collect these, but we could also merge this PR and collect the `callees` in a new branch.

### Renamed Parameters
Some learners have parameters that have been renamed to avoid name collisions -- the lookup functions can't deal with that, of course. I think a nice solution would be to add an `extras` parameter to `makeXXXLearnerParam` (which would be an mlr-agnostic extension to ParamHelpers). Then we could add the information about the function that a LearnerParam belongs to; e.g. in RLearner_classif_bst.R:
```R
# the 'learner' parameter of bst has been renamed 'Learner' to avoid a name collision
makeDiscreteLearnerParam(id = "Learner", default = "ls", values = c("ls", "sm", "tree"),
  extras = list(destination = "bst::bst(learner)")
```

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+    # we can give some more help
+    callees = c("mlr specifics", callees)
+  }
+  if (!is.null(callees)) {
+    n = 1
+    if (length(callees) > 1) {
+      repeat {
+        cat("Choose help page:\n")
+        cat(paste0(seq_along(callees), " : ", callees, "\n", collapse=""))
+        n <- readline("0 : cancel\n...: ")
```
giuseppec: styleguide `=` instead of `<-`

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+        n <- ifelse(grepl("\\D",n),-1,as.integer(n))
```
giuseppec: use stringi functions instead of grepl

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+#' @template arg_learner
+#' @param param [\code{character}]\cr
+#'   Parameter(s) to describe. If no argument is given, information on the documentation status
+#'   of all parameters is printed.
+#' @export
+#' @family learner
+#' @family help
+learnerParamHelp = function(learner, param) {
+  learner = checkLearner(learner)
+  if (missing(param)) {
```
giuseppec: maybe use `is.null` instead of `missing` (you have to set `param = NULL` before).

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+    }
+    missingdoc = setdiff(getParamIds(learner$par.set), names(learner$help.list))
+    if (length(missingdoc) > 0) {
+      catf("The following parameters are missing documentation:\n%s",
+        collapse(strwrap(collapse(missingdoc, sep = ", "), indent = 2), "\n"))
+    }
+    return(invisible(NULL))
+  }
+  assertCharacter(param, any.missing = FALSE)
+  all.param = getParamIds(learner$par.set)
```
giuseppec: Maybe use `getParamIds(getLearnerParamSet(learner))`  here? 
Otherwise this might not work if `learner` is a wrapped learner?

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+    } else {
+      catf("No documentation found. Try to consult learnerHelp(\"%s\").\n", getLearnerId(learner))
+    }
+  }
+}
+
+# remove nesting levels of XML tags
+simplifyNode = function(node) {
+  children = XML::xmlChildren(node)
+  lens = nchar(stri_trim(sapply(children, XML::xmlValue)))
```
giuseppec: vcapply instead of sapply?

R/learnerHelp.R
```diff
@@ -0,0 +1,214 @@
...
+    tbl = rbind(codeListToTable(html), tbl)
+    if (is.null(tbl)) {
+      next
+    }
+    for (row in seq_len(nrow(tbl))) {
+      # the following loop handles occasions where multiple parameters are mentioned in
+      # one row, separated by commas.
+      for (par.name in stri_split(tbl[row, 1], regex=", *")[[1]]) {
+        if (par.name %in% par.ids) {
+          help.list[[par.name]] = paste0("Argument of: ",
```
giuseppec: stri_paste instead of paste0

R/learnerHelp.R
mb706: Wrapped learners currently have no documentation whatsoever. I'm adding a check for wrapped learners instead.

R/learnerHelp.R
mb706: Wrapped learners currently have no documentation whatsoever. I'm adding a check for wrapped learners instead.

ja-thomas:
@mb706  Your done with the features? 

We still need the `callees` for `regr`, `surv`, `multilabel` and `cluster`.
@MariaErdmann can you continue?

mb706:
Yes, pretty much done from my side, see [my last comment](https://github.com/mlr-org/mlr/pull/1657#issuecomment-285937224) about stuff that could be improved long-term.

R/learnerHelp.R
```diff
@@ -0,0 +1,230 @@
...
+  lislis = lapply(lis, function(li) {
+      lichi = simplifyNode(li)
+      if (length(lichi) < 2 || names(lichi)[1] != "code") {
+        return(NULL)
+      }
+      parname = XML::xmlValue(lichi[[1]])
+      pardesc = stri_join_list(lapply(lichi[-1], XML::xmlValue), collapse = " ")
+      stri_trim(c(parname, pardesc))
+    })
+  as.data.frame(do.call(rbind, lislis), stringsAsFactors = FALSE)
```
giuseppec: instead of do.call, we could use one of: rbindlist, setDF(transpose(lislis)), setDF(lislis)

MariaErdmann:
If you can wait, I will add the callees for `regr`, `surv`, `multilabel` and `cluster` until friday. Starting now.

MariaErdmann:
Thanks for being patient :-)
There were some things I did not know how to handle and some things I saw:

- `cluster.kkmeans` uses `centers`, `kernelf`, `kernelMatrix` and BBmisc's `getMinIndex`. How relevant are those for the learnerHelp (right now I have not added them)
- Probably there is a parameter in `rFerns` missing: `saveForest`
- all mboost learners: I added `Family` to the callees for surv and regr. Do we want to have this one? Then we need to add in classif as well.
- in the `bst` learner there are a couple of parameters missing and my test (see issue #1370) also fails with this kind of learner. I started in the workshop to fix the learner, but it was more than I thought, so it will need some more time. But I am not mad, if somebody else loves to do it ;-)  Moreover, I am not sure, if we propbably need to add "rpart" in the `packages` section (but this will only be needed for the `learnerHelp`, the learner itself does not need to get the package
- regr.btgp: param `gamma` missing
- regr.mob:  Should `glinearModel` and `linearModel` added to the `learnerHelp()`? Where do parameters `part.feats` and `term.feats` come from?

mb706:
* I think you were right not to add these functions for `kkmeans`; In general I'd say to only add functions with parameters that come up in the learner's `param.set` (or that come up *as parameters*, like `family` in mboost learners). If mlr uses any other functions, it is usually either uninteresting detail (e.g. changing the data format for the learner) or should be documented within mlr (e.g. `regr.randomForest` does this).
* the `multilabel.rFerns` parameters seem to be missing quite a lot, compared with `classif.rFerns`
* Re `Family` in mboost, `linearModel` in `regr.mob`: Not easy. I think one could argue to include `"Family"` to the mboost `callees` and exclude the individual distributions (`"Gaussian"`, `"Weibull"`, ...) since all these point to the same help file. Similarly, for `regr.mob`, one could include `"Statmodel-class"`, which explains both `glinearModel` and `linearModel`. However, I find it cleaner to actually mention the functions that get used. The package might change in the future and have help pages for the individual distributions / models instead of a collective help page. Since listing all the individual distributions in mboost makes `"Family"` redundant (and since `"Family"` does not get called) I excluded it.
* I generally included packages that are "only" needed for `learnerHelp`, since they are imported by the learners anyways. In this case, `bst` imports `rpart`, so mentioning it in the packages doesn't hurt.
* Re `part.feats`, `term.feats`: These are among the parameters that mlr uses itself (to build the formula, in this case), and that are documented nowhere. In the long term, one could think about writing documentation for the learners that do things beyond just calling a different package; `regr.randomForest` is a good example of this. A different possibility is to describe the extra parameters in the notes, as `regr.km` does it. A third possibility would be to add extra parameter info in the parameters themselves, as I suggest in an earlier comment, with LearnerParam `extras`.

R/learnerHelp.R
mb706: do.call ∘ rbind does exactly the right thing here, including handling all the edge cases (empty list, list of NULLs) correctly. The data.table stuff makes it more complicated.

larskotthoff:
Thanks, this looks good! I'm in favour of merging this and then improving later, as it is a very large PR already and functionality that is completely orthogonal to the rest of the package and won't break anything. I'll merge this soon unless somebody objects.

larskotthoff:
Merging. Thanks @mb706

PR #1660, opened 2017-03-10
speedup task creation

mllg:


berndbischl:
breaks a test, but hopefully easy to fix?

PR #1671, opened 2017-03-14
ignore xgboost.model

ja-thomas:
Since with some update of xgboost it is always creating these files on disk.

(happens when running tests etc.)


berndbischl:
thank god!

PR #1699, opened 2017-03-16
ignore man/*.Rd but force by travis

jakob-r:
Is that a good idea? Hard to see the aftermath that PR would entail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ ehthumbs.db
 Icon?
 Thumbs.db
 *xgboost.model
+man/*.Rd
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ after_success:
   - git config credential.helper "store --file=.git/credentials"
   - echo "https://$GT:@github.com" >> .git/credentials
   - git config push.default matching
-  - git add man/*
+  - git add --force man/*
   - git commit man DESCRIPTION NAMESPACE -m "update auto-generated documentation [ci skip]" || true
   - git push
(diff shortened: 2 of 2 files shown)


larskotthoff:
Can you explain please how what we have at the moment is a problem?

smilesun:
The problem occurs when we need to merge a conflict. One way is to use the online merging tools by github but most people seem to do this locally by using tools like meld?  When we do this, a lot of RD files will be automatically staged.

berndbischl:
i think it would help a lot, if we dont have the RD on github anymore.
they just create hassle, they are complete "temporary" files. 

i think we should try out @jakob-r's proposal. it would be a big improvement.

they only question is whether github_install works - and i am not even sure how much i care about that. people should install from CRAN

@mllg

mllg:
Only travis/jenkins pushing man pages sounds reasonable. I don't see any downside.

berndbischl:
> Only travis/jenkins pushing man pages sounds reasonable. I don't see any downside.

some time ago you mentioned that we can get away WITHOUT having the man-dir on CRAN. 
i mean that seems to be the best option. any news / experience about this?

i can try this out with our new focussearch package, to test it. but maybe you know more?

larskotthoff:
In principle, we don't need them, but it makes making a release easier.

jakob-r:
@berndbischl: With this PR the Rd files would be still part of the git but ignored unless force added (as it is done by travis)

berndbischl:
> but it makes making a release easier.

not really. for a release i run "rbuild" with rt to create the targz. that should autobuild all man files. 
i also upload to winbuilder before. 

so it is exactly the same command(s) i run now, with or without the man files on github.

and it was always like this. the ONLY thing i worried about before - i think @mllg suggested some time ago to remove the man files from GH - was that install_github from devtools would not work.

suggestion: 
lets test the whole think, maybe with a smaller package than mlr. if it works reasonably, we exclude the man files.

PR #1710, opened 2017-03-19
Add missings property for h2o learners

### LINKED ISSUE #1690: H2O randomForest and GBM models DO support missing values
Hello,

I wanted to recommend adding the "missings" property to H2O's randomForest and GBM models as both of these models internally deal with missing values without imputation. The other H2O models will skip over missing values and correctly do not have the "missings" property. However, as you can see from the documentation below, H2O RF and GBM do support NAs. I can confirm from changing the missings property for H2O's RF on my machine that everything works properly and the response does not contain NAs.

Thanks!

## H2O documentation page for random forest: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/drf.html

**How does the algorithm handle missing values during training?**
Missing values are interpreted as containing information (i.e., missing for a reason), rather than missing at random. During tree building, split decisions for every node are found by minimizing the loss function and treating missing values as a separate category that can go either left or right.

**How does the algorithm handle missing values during testing?**
During scoring, missing values follow the optimal path that was determined for them during training (minimized loss function).

## H2O documentation for GBM: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/gbm-faq/missing_values.html

**How does the algorithm handle missing values during training?**
Missing values are interpreted as containing information (i.e., missing for a reason), rather than missing at random. During tree building, split decisions for every node are found by minimizing the loss function and treating missing values as a separate category that can go either left or right.

So how does the algorithm determine what goes left or right? The algo decides which feature and level/number value to split on its decision to make things go left or right is purely based on value (i.e. everything less than the split point value goes left, and everything greater and equal goes right). This means that missing values can go left or right in the same way as any other categorical because NAs are now seen as a new categorical.

A note on nbins_cats: This parameter specifies the number of bins to use for non-Na categoricals, it does not bin missing values. Regardless of the number of bins a user chooses for nbins_cats, an additional bin will be created specifically for the missing values.

**How does the algorithm handle missing values during testing?**
During scoring, missing values follow the optimal path that was determined for them during training (minimized loss function).

mllg:
Fixes #1690.

### DIFF
--- a/R/RLearner_classif_h2ogbm.R
+++ b/R/RLearner_classif_h2ogbm.R
@@ -14,7 +14,7 @@ makeRLearner.classif.h2o.gbm = function() {
       makeIntegerLearnerParam("max_after_balance_size", lower = 0L, default = 1L),
       makeIntegerLearnerParam("seed", tunable = FALSE)
     ),
-    properties = c("twoclass", "multiclass", "numerics", "factors", "prob"),
+    properties = c("twoclass", "multiclass", "numerics", "factors", "prob", "missings"),
     name = "h2o.gbm",
     short.name = "h2o.gbm",
     note = "'distribution' is set automatically to 'gaussian'."
--- a/R/RLearner_classif_h2oglm.R
+++ b/R/RLearner_classif_h2oglm.R
@@ -21,7 +21,7 @@ makeRLearner.classif.h2o.glm = function() {
       makeUntypedLearnerParam("beta_constraints"),
       makeLogicalLearnerParam("intercept", default = TRUE)
     ),
-    properties = c("twoclass", "numerics", "factors", "prob", "weights"),
+    properties = c("twoclass", "numerics", "factors", "prob", "weights", "missings"),
     name = "h2o.glm",
     short.name = "h2o.glm",
     note = "'family' is always set to 'binomial' to get a binary classifier."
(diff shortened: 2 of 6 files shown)


kdpsingh:
Thanks for merging. However, as per my original note, only h2o GBM and h2o randomForest support missings. The other h2o models (including GLM) currently don't support missing values -- there is an interface to perform mean imputation via h2o but I wouldn't count that as supporting missing values.

I would suggest removing the missing values property for the h2o GLM and leaving it in for GBM and RF only. Thanks!

ledell:
@kdpsingh All the h2o models support missing values by default -- no extra interface is required to do this.  The tree algos support missing values natively; GLM and DL support it through mean imputation.  

@kdpsingh @mllg I'll make a PR for adding "missings" to GLM and DL.

kdpsingh:
Thanks @ledell. Agree with the PR to add support to h2o glm and deeplearning for missing values. I think it would be nice info to have on the mlr tutorial page that glm and deeplearning perform mean imputation for missing values in the "Note" column on the integrated learners page.

berndbischl:
@ledell do you want to quickly sweep over the notes? before i do that and write partially untrue stuff?

ledell:
@berndbischl PR for updates to notes here: https://github.com/mlr-org/mlr/pull/2031

PR #1715, opened 2017-03-21
Fix classif lqa

### LINKED ISSUE #1714: classif.lqa w/ penalty genet and oscar with custom penalty parameters don't work
`oscar` doesn't work because the name `c` collides with some other name somewhere
`genet` doesn't work because the parameters `alpha` and `gamma` are taken in the wrong order.
Bonus: `alpha` is restricted to be [0, 1], not [0, &infin;).

mb706:
Fix  #1714


berndbischl:
i looks like some param have been renamed. at least oscar.c, but apparently also lambda1 and lambda2?

in these cases please ALWAYS explain this in the note.

and: thanks.

mb706:
Done. lambda1, lambda2 were not renamed, though.

berndbischl:
> ambda1, lambda2 were not renamed, though.

they were. but not by you. i can see this in the code, and by looking at the docs.  we should add a note about this too. i know, not your fault.

mb706:
(Many of the penalty parameters for `classif.lqa` are "renamed" in some sense, since the parameter of the penalty functions is always just a one, two or three dimensional vector named "`lambda`".)

mb706:
@berndbischl I think this should be merged? classif.lqa is really broken otherwise.

PR #1716, opened 2017-03-21
Put classif.penalized.XXX together into classif.penalized.

mb706:
Having three learners with slightly different names that all expose the same function is unnecessary.

Also, so far, it was not possible to have `fusedl == FALSE` and give L1 as well as L2 penalization at the same time.


berndbischl:
i have stupidly merged this, then reverted
issue is here #1718

PR #1720, opened 2017-03-21
Join penalized

### LINKED ISSUE #1718: join the penalized learners
@mb706 

1) sorry that i merged your PR too quickly then had to revert
old branch is here
https://github.com/mlr-org/mlr/tree/fix_penalized

2) in general i think your idea is good

3) but if we do this, we HAVE to do this consistently at least for the surv and classif learners for penalized, too

mb706:
Fix #1718 
Please someone have a closer look at this, I never use regr or surv learners, so can't say anything about the quality of this.


berndbischl:
looks good, while you are at it, you could set unpenalized to tuneable = FALSE? that really is something where we cannot even tune the data structure. and is usually dictated by the application

mb706:
Done

larskotthoff:
Thanks, merging.

PR #1723, opened 2017-03-21
Oneclass: Add new makeOneClassTask() and rlearner svm for one class

MinhAnhL:
Create new Task type: oneclass


R/OneClassTask.R
```diff
@@ -0,0 +1,30 @@
...
+
+makeOneClassTaskDesc = function(id, data, weights, blocking) {
+  target = character(0L)
+  td = makeTaskDescInternal("oneclass", id, data, target, weights, blocking)
+  return(addClasses(td, c("OneClassTaskDesc", "UnsupervisedTaskDesc")))
+}
+
+#' @export
+print.OneClassTask = function(x, ...) {
+  print.UnsupervisedTask(x)
```
mb706: missing ellipsis, use `print.UnsupervisedTask(x, ...)`. And should probably be `NextMethod()`.

`print.OneClassTask` could be omitted entirely, since a `OneClassTask` will always be an `UnsupervisedTask` (?), but I guess this makes sense if the plan is to print some more info or uncouple UnsupervisedTask later.

P.S. I know `ClusterTask` does it like this, and `ClusterTask` is just as wrong.

MinhAnhL: Thanks for the review :) 
I will add the ellipsis, and I guess Classiftask.R and Clustertask.R (and probably all other tasks) also need to add ellipsis, when calling `print.Supervised()`/`print.Unsupervised()` in their print-fct.

Can you explain what you mean by `NextMethod()`?

mb706: [`NextMethod()`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/UseMethod.html) automatically jumps to the same function one step up in the class hierarchy. If `x` is a `OneClassTask` with `class(x)` something like `c("OneClassTask", "UnsupervisedTask", "Task")` (I think?) and you evaluate `print(x)`, then R automatically calls `print.OneClassTask` (since that is the first class); if you then call `NextMethod()` within this function (instead of `print.UnsupervisedTask`), it goes to the next step (which would be `print.UnsupervisedTask`, so pretty much what you want). You should try it!

This may not work so well if you call `print.OneClassTask(x)` *directly*, but you probably shouldn't do that anyways.

ja-thomas:
Ok, so some thoughts:

- It might be better to let `oneClassTask` inherit from `supervisedTask` instead of `unsupervisedTask`.

- For that task it should be optional to add a target column. If it is missing we define everything as "normal".  If a target column is added it should be binary.

MinhAnhL:
I made a a pull request for my makeOneClassTask() few weeks ago. 
I worked on it in a subbranch  of the master called 

> oneclass_base

.
In a further subbranch called 

> oneclass_rlearner_svm

 (so basically subsubbranch from master) I worked on my first learner, which is the svm for one class. For that I made a pull request 1-2 weeks ago, but it didn't appear in my pull request list. I was wondering what I did wrong. Now I realised it is merged into the onceclass_base-branch and I don't know how to reverse it. But I think it is fine to have the makeOneClassTask() and its first learner in one PR.

pat-s:

Issue #1730, opened 2017-03-24
new learnerHelp functions: rename and API


berndbischl:
1) this functions are totally cool

2) but not 100% style guide and i hate to rename later.

they should always be in "verb style", so doThatThing.

so here:
- getLearnerHelper, or even better: helpLearner? showLearnerHelp?
i think helpLearner is closest to normal R help?

3) for learnerParamHelp:
we should really return the string instead of printing it directly? that seems much more useful

then rename getParamHelpString

4) i would like to display the param help string later in the print-out of the paramset for the learner. but this is a new issue. just mentioning here,

@mllg @larskotthoff @jakob-r @mb706

mb706:
Some considerations:
* `learnerParamHelp` does some pretty-printing of the help string with `strwrap`, to wrap it to the `width` option of the R session; the text is hard to read otherwise. If it is to only return the string, one should ideally give it a class and write a `print` function that does this wrapping.
* `learnerParamHelp` also prints `mlr` related info about the parameter: The parameter itself (i.e. the `print`-out of the `LearnerParam` object), the requirement (if any), and the current value (if set). I found this to be a useful set of answers to the question "what can I know about this parameter?". The beauty of showing the current value is that it it shows the "default" value when `learnerParamHelp` is given a learner ID string, but it shows the value currently in effect when given a learner object with changed hyperparameters.
* Originally `learnerParamHelp` printed the info about *all* parameters when no `param` argument was given. I found this made output very long, complex and hard to read, so I opted for the current state (where `learnerParamHelp` with `param == NULL` just prints an overview of available params, and which params have docu). I therefore suggest re your (4) that, while it may be useful to display the help string when printing a *single* param, I would not show it when printing the param *set*. (In fact, `learnerParamhelp`'s output comes close to what I wished a single parameter's `print` or `summary` output would look like.)

berndbischl:
@mb706 is this fixed? forgot to close?
or something still open? 

careful as i marked this as blocking
@larskotthoff

larskotthoff:
This should be fixed and can be closed.

mb706:
I didn't mark #1734 as closing this issue, since it really only changes the name; it does not address the other issues / ideas you raised. If you are happy with the state now you can close. If the name was the 'blocking' aspect but you still want the API to change with a lower priority, you can change the priority tag.

mb706:
[...]

PR #1734, opened 2017-03-24
Rename help learner

mb706:


mb706:
Rename functions as suggested in #1730, though does not touch the issue of changing the internals or API.

larskotthoff:
Thanks.

PR #1742, opened 2017-03-28
Allow passing of train indices to measures

mllg:
This is required for some survival measures and would greatly simplify #1372.

### DIFF
--- a/R/Measure.R
+++ b/R/Measure.R
@@ -43,7 +43,8 @@
 #'     \item{req.task}{Is task object required in calculation? Usually not the case}
 #'     \item{req.model}{Is model object required in calculation? Usually not the case.}
 #'     \item{req.feats}{Are feature values required in calculation? Usually not the case.}
-#'     \item{req.prob}{Are predicted probabilites required in calculation? Usually not the case, example would be AUC.}
+#'     \item{req.prob}{Are predicted probabilities required in calculation? Usually not the case, example would be AUC.}
+#'     \item{req.train}{Are training indices required in calculation? Usually not the case, some survival measures need them.}
 #'   }
 #'   Default is \code{character(0)}.
 #' @param fun [\code{function(task, model, pred, feats, extra.args)}]\cr
--- a/R/Measure_custom_resampled.R
+++ b/R/Measure_custom_resampled.R
@@ -65,7 +65,7 @@ makeCustomResampledMeasure = function(measure.id, aggregation.id, minimize = TRU
   assertString(note)
 
   force(fun)
-  fun1 = function(task, model, pred, feats, extra.args) NA_real_
+  fun1 = function(task, model, pred, feats, train, extra.args) NA_real_
   # args are checked here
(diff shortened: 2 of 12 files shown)


larskotthoff:
Looks like only the lintr failed. I guess that means it works :)

mllg:
```r
pred$data = as.data.frame(ss[!is.train, , drop = FALSE])
```
Is this really better?
@berndbischl @mb706

mb706:
You mean as in "Will it pass lintr?" - Yes. If you mean "Is this the kind of world we want to live in?" - well (1) before lintr the codebase was inconsistent between `, , ` and `,, `, both occurring about equally often; (2) no R style guide I found had any opinion on that matter(*); (3) the "space after comma" rule already beats the "no space before closing brackets" rule; (4) its what `lintr` enforces by default.

(*) actually, in a strict reading, the style guide forbids two consecutive commas, be it with or without space

berndbischl:
1) First of all, I dont want the name to be "train". Thats not expressive. Use at least train.inds, also consistent with the rest of the package.

2) Is this really necessary. I mean, the Prediction object contains the indices? So why do we need to blow up the API here?

berndbischl:
> Is this really necessary. I mean, the Prediction object contains the indices? So why do we need to blow up the API here?

ok, that is incorrect, i got confused. the prediction object contains the, of course, the indices of the TEST set. but the model contains the train.indices. so why not require simply that?

berndbischl:
> ok, that is incorrect, i got confused. the prediction object contains the, of course, the indices of the TEST set. but the model contains the train.indices. so why not require simply that?

also: for a ResamplePrediction we also have the indices. of course this does not help, as you need to deal with a normal "Prediction" object. I guess I would argue to put the train.inds simply into that one? That can then be used in more places and is cleaner?

mllg:
> First of all, I dont want the name to be "train". Thats not expressive. Use at least train.inds, also consistent with the rest of the package.

Agreed.

> Is this really necessary. I mean, the Prediction object contains the indices? So why do we need to blow up the API here?
> [...]

Blow up the interface, or blow up the objects. So I take you prefer to blow up the objects?

berndbischl:
> Blow up the interface, or blow up the objects. So I take you prefer to blow up the objects?

i am not 100% sure. lets go through the options.


a) can we require the model? does that blow up memory? during resampling it is automatically there. during manual evaluation you have to just add it to the call. thats  not worse than adding the indices, what you suggest. actually: it is easier?
the only downside is, that if at some point later, you only have the Prediction object, you cannot evaluate, when you have not stored the model as well (which is large). but again: same problem for your approach?
you would need to extra-store the training indices.

b) put the indices in Prediction object. i dont know.
what mlr is really missing is a two-way conversion list of Preds <--> ResamplePrediction.
but the indices do not really help there. it would be (also very much ok)
list of Preds + ResampleInstance <--> ResamplePrediction.

tell me your thoughts, maybe a) is really best? wouldnt that work?

mllg:
Found an easy way to solve this.

PR #1744, opened 2017-03-28
test for #1467 (closes #1467)

### LINKED ISSUE #1467: getTaskFormula should check if x is actually a task object
Otherwise you get funny errors

```
> getTaskFormula(convertOMLTaskToMlr(tsk))
Error in if (type == "surv") { : argument is of length zero
```

This is just a small detail, but could improve user experience ;)

Just FYI: I figured out now that it should be
`getTaskFormula(convertOMLTaskToMlr(tsk)$mlr.task)`
instead :lollipop:

mllg:


mllg:
Fixed via #1711, this is just a unit test.

PR #1754, opened 2017-03-31
fix getLearnerModel for nested preproc wrapper

mllg:
`getLearnerModel(..., more.unwrap = TRUE)` did not give the underlying R model, only the next learner.
`getLeafModel()` should in principle do this, but (a) is not exported and (b) also fails for nested chain models


larskotthoff:
Thanks, merging.

PR #1761, opened 2017-04-04
Oneclass r learner svm

MinhAnhL:
Add oneclas_svm learner for anomalydetection/ one class classification.
Note: there is no property prediction for svm in e1071


MinhAnhL:
I checked locally with `devtools::check()` and `devtools::test(filter = "base")` and didn't get an error. Travis check seems to throw an error in run-lint.R for the line `seq_along(parset.list)` in the file "test_oneclass_svm.R".
Where   
```
parset.list = list(
    list(),
    list(gamma = 20),
    list(kernel = "sigmoid", gamma = 10),
    list(kernel = "polynomial", degree = 3, coef0 = 2, gamma = 1.5)
  )
```
(Before I used `1: length(parset.list)` and got an error, therefore I use `seq_along(parset.list)` instead)
Can I ignore travis here? 
Or does anyone has a alternative suggestion for `seq_along(parset.list)`

PR #1763, opened 2017-04-05
Allow specification of formulas in tasks

giuseppec:
This PR enables one to do stuff like:

```r
task = makeRegrTask(formula = Petal.Length ~ Petal.Width + I(Petal.Width^2), data = iris, target = "Petal.Length")
lrn = makeLearner("regr.lm")
mod = train(lrn, task)

pd = generatePartialDependenceData(mod, task, features = "Petal.Width")
plotPartialDependence(pd)
```

index

TODOs:

- if formula is specified, the target can be generated from the formula
- currently, it works only for learners that internally use `getTaskFormula` and support the formula interface (maybe this is also sufficient). 
- add "formula" property for learners?
- change behaviour of `getTaskData` if task contains formula?


larskotthoff:
Thanks, looks interesting. Should we have a new learner property "formula" for this?

berndbischl:
hard to review this. there seems to be zero documentation and unit tests for this?


also, yes, if this only works for some learners we need to introduce a property?
and what about the other tasks, you only did classif and regr?

giuseppec:
That's why this PR has a work in progress label. I just had this too long in my local branch and did not want to loose this.

berndbischl:
@giuseppec ok. can you comment on the problem is just noted in the other PR?
is there a way to "untangle" both of these things?

mb706:
> currently, it works only for learners that internally use getTaskFormula and support the formula interface (maybe this is also sufficient).

A solution here would be to have `getTaskData` to actually get the data in the form that the formula demands. (The function could have a parameter that determines whether that happens). `e1071:::svm.formula` is an example of how this could work. This particular example appears to only work for numerical features (since categorical features are dummy-encoded).

One would have to look into whether R  has an automated way of turning `data.frame` + `formula` into a new `data.frame` with general data types, or if one might want to implement this for `mlr`. I think it would be a worthwhile feature, because it would give a simple interface for specifying interactions, even for learners that don't use `formula`.

be-marc:
I would like to start the discussion again about the integration of formulas into MLR. I am working on the implementation of `mgcv::gam` and `MASS::glmmPQL` into MLR. It is not possible to integrate them without formulas. 

I discussed this issue with @pat-s and @giuseppec, who already proposed solutions in #1763 and #1899 (Further information in #1740). 

To sum up the drawbacks of both PR:

1. Integration of the formula into the learner like `classif.lrn = makeLearner("classif.gam", formula = "target ~ s(p1) + s(p2)"`
   - A benchmark with one or multiple learners and multiple tasks would fail because the data set in one task might be not compatible with the formula    specified in the learner
   
2. Integration of the formula into the task like `classif.task = makeClassifTask(data = df, target = "target", formula = "target ~ s(p1) + s(p2)")`
   - The task could be used just with one specific learner
   - Benchmark would also fail if the formula specified in the task is not compatible with the classifier
   

In our discussion @pat-s came up with a new solution. A task could store a named vector with the classifier name and the associated formula. 

```
# Create named vector
formula_vector = structure(c("target ~ s(p1) + s(p2)", "target ~ p1 + (1|p2)"), names= c("classif.gam", "classif.glmmPQL"))

# Set in task
classif.task = makeClassifTask(data = df, target = "target", formula = formula_vector)
```

`getTaskFormula()` could be used to return a formula if there is a matching one for the specified learner.

This solution has the advantage that

+ It is possible to use different learners with one task
+ Benchmarking would still be possible
+ We keep somehow the spirit of MLR that the task defines the formula. It would be even possible to force the user that the formula contains all columns of the data frame like it was always done with the automatic generated formula. 

If we can agree on this solution, I would start with the implementation and add `gam` and `glmmPQL` later.

pat-s:
Since we decided to implement any new major changes into _mlr3_, I'll close here and open an issue there.

PR #1766, opened 2017-04-07
pass subset argument through explicitly (fixes #958)

### LINKED ISSUE #958: multilabelBinaryRelevanceWrapper is not working properly for classif.glmnet

larskotthoff:


berndbischl:
ok this looks great. but we need to extend the unit tests with the problem raised in the issue, and maybe even 2-3 more wrappers

berndbischl:
so at least lets test the baggingwarpper with the glmnet learner, and like i said maybe 2 more

mllg:
Thanks!

PR #1774, opened 2017-04-14
merge Fda_pull1_task_feat_extract into fda_pull1_task

smilesun:


smilesun:
not relevant anymore, since we changed the datastructure for FDA.

PR #1775, opened 2017-04-14
merge Fda_pull1_task_featextract into  old FDA branch

smilesun:


smilesun:
This is totally out of date since we made so many changes afterwards, all features have been transmitted to the other branches.

PR #1776, opened 2017-04-14
Added monotone parameter to xgboost

vinaywv:
Xgboost 0.6 version has monotone_constraints property added to the package. I added monotone_constraints as a parameter into the RLearner_regr_xgboost.R


larskotthoff:
Thanks. Is this only for the regression version, or does it apply to the classifier as well?

vinaywv:
I have added this to the classifier as well

vinaywv:
@larskotthoff , I am not sure why the checks were not successful. I have added the same line of code in both regression and classification. While the Travis-CI check was successful earlier for regression, the same is not for classification.

larskotthoff:
Thanks, the failure seems to be unrelated. The documentation says that the values for the vector can only be integers, but you've declared them as numeric -- are real values allowed?

vinaywv:
@larskotthoff, I think I made the parameter to be a integer vector with lower limit of -1 and upper limit of 1

larskotthoff:
Perfect, thanks!

PR #1781, opened 2017-04-25
Add rng as parameter to the generatePartialDependence function

vinaywv:
partialdependeance creates its own range of values based on the minimum and maximum values of a features in the train data. For instance in the iris data, the petal width is automatically divided into 6 buckets 0.1000000,0.3666667,0.6333333,0.9000000,1.1666667,1.4333333. But the user cannot specify if they want to get the output in buckets such as 0.1,0.2,0.5,0.8,1.0,1.2, 1.5. I am just passing the 'rng' as a parameter for the user to be able to provide these values. If the parameter is not provided, it will give the default values as above.

### DIFF
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -232,7 +232,7 @@ generatePartialDependenceData = function(obj, input, features,
     stop("fmax must be a named list with an NA or value corresponding to each feature.")
   assertCount(gridsize, positive = TRUE)
 
-  if (!is.null(rng))
+  if (is.null(rng))
     rng = generateFeatureGrid(features, data, resample, gridsize, fmin, fmax)
   if (length(features) > 1L & interaction)
     rng = expand.grid(rng)
--- a/R/generatePartialDependence.R
+++ b/R/generatePartialDependence.R
@@ -168,7 +168,7 @@
 generatePartialDependenceData = function(obj, input, features,
   interaction = FALSE, derivative = FALSE, individual = FALSE, center = NULL,
   fun = mean, bounds = c(qnorm(.025), qnorm(.975)),
-  resample = "none", fmin, fmax, gridsize = 10L, rng = NULL, ...) {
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, this looks useful. A few things though:
- Please correct the roxygen argument type to list.
- Instead of checking for `missing()`, can you check for the default `NULL` please?
- Could you add a unit test for the new functionality please?

Thanks!

vinaywv:
Thanks @larskotthoff, I have made the first two changes you had suggested. I am not familiar on how to create the unit tests, but I tried to look at the previous unit tests and tired to add one myself. If the unit test is not correct, can you please let me know of any document which will point out on how to write the unit tests?

larskotthoff:
You simply missed a closing parenthesis at the end of the test :)

Could you after calling the function please also ideally check whether the returned object respects the range that you've passed? I.e. check whether you have data for those points. Have a look at the other `expect*`s in the file to see how to do that.

You can run the tests locally with `library(testthat); test(filter = "generatePartialDependence")`.

larskotthoff:
Hmm, do the tests pass for you locally?

vinaywv:
Hi Lars, I was getting an error on the local machine saying 'Cannot find package root' when I try to run the test on the local machine. I tried debugging it but was not able to find the cause of that problem.

larskotthoff:
Thanks, the tests pass for me locally. Just one last thing: could you rename the parameter to "range" please? The abbreviation saves only 2 letters and makes it less obvious what it is.

vinaywv:
This variable rng was internally created in the function and I had just let the user pass the range if required. It is used elsewhere in the code - do you want me to change it elsewhere especially in the partial plot for interactions

larskotthoff:
No, just the name of the argument. It doesn't really matter what it is called internally. Thanks!

larskotthoff:
Thanks for the update; it looks like the tests are failing though. Could you fix please?

larskotthoff:
Thanks, merging.

PR #1782, opened 2017-04-25
new learner from Maierhofer knn and kernel

smilesun:
used this pr to keep track of changes, working now , do not merge


smilesun:
1. the branch name is called fda_pull1_task as well in the Maiehofer implementation
2. the original fda.usc.classif.knn interface to mlr is extended to allow probability prediction
3. FDA. classifKnn(no dot) is from the Fuchs paper (Nearest neighbor ensembles for function dta)
4. FDA.classifKernal is an extension to the Fuchs paper by integrating the kernel method.
5. proxy package is used to compute the proximity between two points
6. fdasrvf: Elastic Functional Data Analysis is used to create a new semi-metrics for the learner FDA.classifKnn
7. stakedLeaners, method = brierscore or others ,  one  of them is from fuchs.
----
7. the expression does not work and will be removed
8. Code review needed

smilesun:
Thomas has created a new pr for his stuff since we have changed dramastically the data structure, so this pull request is not relevant anymore.

PR #1784, opened 2017-04-26
Update variance estimation methods for RF learners

ja-thomas:
- Extratrees is now an available splitting method for classification and regression

- Add se estimation method for `regr.ranger`. (This is mostly copy paste code from `regr.randomForest`) and only one of the three alternatives there.


berndbischl:
isnt it a bit inconsitent to add only 1 se method from rf...?

yes, it is a clear improvement. can you easily convince me that we should not copy the other 2 methods?

ja-thomas:
yup you're right, all three methods would make sense. 

But I just needed one for now to test some stuff for MBO. But in general it would make sense if the se estimations for random forest would work for all supported RFs, but thats more work.

So the short answer is I was too lazy to copy all three methods... And this is mostly a PR so that the code/branch does not get lost

jakob-r:
We should really have extra methods for se calculation for bagged learners...

Stevo15025:
What else needs done here? May have time next week to help if you need any

jakob-r:
@berndbischl We should merge and be okay with this double code because calculating the SE for bagged learners in general would be a seperate issue. This should not hold back such important progress for the usage of ranger.

berndbischl:
i really dont want the package to become inconsistent here. 
if people want to have this so urgently, please help making this better.

it is really not that hard here.

jakob-r:
Okay. I added generalized functions for sd and jackknife. We might want to put them in a separate file. For a better diff i will do that later (after suggestions for names).

jakob-r:
@berndbischl your opinion would be much appreciated :wink:

larskotthoff:
I've tried to sort out the merge conflicts. Could someone please have a look whether this still makes sense? @jakob-r

jakob-r:
I solved the conflict. If travis passes this can be merged.

larskotthoff:
Ok, I've fixed some obvious stuff, but Travis is still failing and have no idea why. Can't reproduce locally.

pat-s:
@ja-thomas 	just FYI, I merged the master branch. Let's see if the build passes. Do you still think that this is ready to merge then? 

(+ an entry to NEWS.md)

pat-s:
@jakob-r 
For some reason the `par.vals` list in 

https://github.com/mlr-org/mlr/blob/c89a3c2a61786ae86b93cd13ccdd4f9590680add/tests/testthat/test_regr_ranger.R#L31

is not present during the `train()` call in

https://github.com/mlr-org/mlr/blob/c89a3c2a61786ae86b93cd13ccdd4f9590680add/R/RLearner_regr_ranger.R#L41

That is what causes Travis to fail. 

Do you know if this is a general problem of `par.vals` not being forwarded?

jakob-r:
ranger now uses only its own se methods (jacknife and infjacknife (default)). 
Normal rf should work again.
all necessary parameters are set automatically if predict.type = "se" is demanded.

pat-s:
Cool :+1:
The doc still needs to be transformed to markdown and we a NEWS entry, then we're good to go here.

PR #1792, opened 2017-05-08
nnet: MaxNWts is not tunable

jakob-r:
According to the manual: 

>The maximum allowable number of weights. There is no intrinsic limit in the code, but increasing MaxNWts will probably allow fits that are very slow and time-consuming.


larskotthoff:
Thanks, merging.

PR #1795, opened 2017-05-08
Fix lint issues that apparently weren't caught before.

mb706:
This should stop lintr bot from complaining all the time.

I need to go now, if there are still some minor issues, please fix these and commit.

Everyone else: In your pull request, just do 'git merge master' after *this* pr was merged and you pulled it, then lintr should also stop complaining about your branch.


mb706:
I think I'm done here.

larskotthoff:
Thanks, merging.

PR #1801, opened 2017-05-19
bracket was wrong

MinhAnhL:
instead of 

> length(threshold > 1L)

it should be

> length(threshold) > 1L


larskotthoff:
Thanks for spotting this! Could you also add a unit test that checks for this bug please?

berndbischl:
> Thanks for spotting this! Could you also add a unit test that checks for this bug please?

no she cannot, as you should see yourself after thinking about this....
i mean the code is equivalent right?
although better now.

PR #1803, opened 2017-05-21
Small changes of the texts in the note section

MinhAnhL:
…it make sense to also introduce the term sensitivity

fdr: wrong definition of the denominator (wikipedia roc, englisch version)
mcc: missing definition of numerator


berndbischl:
@giuseppec can you quickly read this then i will merge

berndbischl:
@MinhAnhL are you done here?

MinhAnhL:
Yes I'm done

PR #1804, opened 2017-05-23
port to new backports

mllg:


mllg:
We should wait for the binaries to hit the mirrors (1-2 days) before merging.

mllg:
Something went wrong in this branch...

PR #1819, opened 2017-06-01
Oneclass: Rlearner ksvm

MinhAnhL:
_RLearner_oneclass_ksvm.R_ is build analogue to _RLearner_oneclass_svm.R_
Only in `trainlearner.oneclass.ksvm` there is a slight difference in the structure.
It is more analog to _RLearner_classif_ksvm.R_, because ksvm require the parameters regarding the kernel to be in `kpar`,  mlr allows the user to pass them directly.

_test_oneclass_ksvm_ is analog to _test_oneclass_svm_ and _test_classif_ksvm_.
`parset.list1` and `parset.list2` are identical, except that `parset.list1` list the param regarding the kernel in `kpar`


pat-s:

Issue #1697, opened 2017-03-16
treat feature filters as functions


mb706:
Some of the provided feature filtering methods have parameters that influence their execution / result. These parameters are given to the `filterFeatures` or related functions directly, which is a bit inelegant. Furthermore, the *actual parameters* are documented in `makeFilter`; I had to grep through the mlr source to find that out.

I think it would be much nicer to have feature filters as functions, in the same way `glm` treats its `family` argument. One could still give the user the option to use strings naming the filter methods as an alternative.

An illustration:
```R 
> filterFeatures(task, "chi.squared", abs = 3)  # old way for compatibility
> filterFeatures(task, chi.squared(), abs = 3)  # new way
> filterFeatures(task, chi.squared, abs = 3)  # possibly allow this, too
> help(chi.squared)  # the relevant text currently found in ?makeFilter
> # model predictiveness, using defaults: rpart and holdout:
> filterFeatures(task, "univariate.model.score", abs = 3)
> filterFeatures(task, univariate.model.score(), abs = 3)
> # giving an argument to the method:
> filterFeatures(task, univariate.model.score(perf.learner = "classif.logreg"), abs = 3)
```

To save relevant meta-information about the filter methods as, one could use the function's `attributes`, or have these functions return objects that, besides a `fun` that does filtering, also contains meta-information. I like the first of these possibilities more, as it would make it easier to tabulate the filter methods, and also to print them nicely.

PR #1827, opened 2017-06-05
CPO: Composable Preprocessing Operators

mb706:
This is my GSoC project. See the [preliminary vignette](https://rawgit.com/mlr-org/mlr/ComposablePreprocOperators/todo-files/CPO%20Vignette.html) for a quick overview ([more compact version with the R output removed](https://rawgit.com/mlr-org/mlr/ComposablePreprocOperators/todo-files/CPO%20Vignette_compact.html)).
# Description for a General Audience
@everyone. If you have questions, ideas or feedback, please don't hesitate to write me, here or in other places!
## What is this?
Functions for data manipulation and pre-processing **&#x2795;** a replacement for `makePreprocWrapper` **&#x2795;** lots of syntactic sugar.

## Description
CPOs are called like functions and create an object that has Hyperparameters that can be manipulated using `getHyperPars`, `setHyperPars` etc.
```R
> cpoPca()
pca(center = TRUE, scale = TRUE)
> cpoPca(scale = FALSE)
pca(center = TRUE, scale = FALSE)
> cpo = cpoPca(scale = FALSE)
> getHyperPars(cpo)
$center
[1] TRUE

$scale
[1] FALSE

> setHyperPars(cpo, center = FALSE)
pca(center = FALSE, scale = FALSE)
```
These objects can be applied to `Task`s or `data.frames` to manipulate data, or can be attached to a `Learner` to create a wrapped learner similar to `makePreprocWrapper`.
```R
> # PCA-rotate pid.task
> rotated.pid.task = pid.task %>>% cpoPca()
>
> # rotate a data.frame
> rotated.attitude = attitude %>>% cpoPca()
> # use the same rotation matrix on a
> # shorter version of the DF
> short.rotated.attitude = head(attitude) %>>% retrafo(rotated.attitude)
> all.equal(head(rotated.attitude), short.rotated.attitude,
+   check.attributes = FALSE)
[1] TRUE
> 
> # Centering / Scaling *after* PCA
> neoPCA = cpoPca(center = FALSE, scale = FALSE, id = "pca") %>>% cpoScale()
> neoPCA
(pca.pca >> scale)(pca.center = FALSE, pca.scale = FALSE, center = TRUE, scale = TRUE)
>
> # Attach the above to learner
> pcaLogreg = neoPCA %>>% makeLearner("classif.logreg")
> getHyperPars(pcaLogreg)
$model
[1] FALSE

$pca.center
[1] FALSE

$pca.scale
[1] FALSE

$center
[1] TRUE

$scale
[1] TRUE
```

Custom CPO constructors can be created using `makeCPOObject` or `makeCPOFunctional`. Note it is possible to write the (re)transformation operations with curly braces, with the function header getting added automatically. 
```R
> # bogus example, multiply first column
> cpomultiplier = makeCPOFunctional("multiplierF", factor = 1: numeric[~., ~.],
+   cpo.trafo = {  # implicit 'function(data, target, factor = 1)' here
+     data[[1]] = data[[1]] * factor
+     attr(data, "retrafo") = function(data) {
+       data[[1]] = data[[1]] / factor
+       data
+     }
+     data
+   })
> head(getTaskData(pid.task %>>% cpomultiplier(10000)))  # note first column
  pregnant glucose pressure triceps insulin mass pedigree age diabetes
1    60000     148       72      35       0 33.6    0.627  50      pos
2    10000      85       66      29       0 26.6    0.351  31      neg
3    80000     183       64       0       0 23.3    0.672  32      pos
4    10000      89       66      23      94 28.1    0.167  21      neg
5        0     137       40      35     168 43.1    2.288  33      pos
6    50000     116       74       0       0 25.6    0.201  30      neg
```

# Implementation details
@berndbischl, @mllg 
## "CPO" (the name)
Would you want me to use "TaskTransform" instead of CPO (or something entirely different)?
## makeParamSet Syntactic Sugar
I wrote a function `paramSetSugar` that makes creating `ParamSet`s much less painful. Example
```R
> paramSetSugar(a: logical, b: integer[0, 10], c: numeric[, ]^2)
           Type len Def      Constr Req Tunable Trafo
a       logical   -   -           -   -    TRUE     -
b       integer   -   -     0 to 10   -    TRUE     -
c numericvector   2   - -Inf to Inf   -    TRUE     -
```
Do you like this idea in general (maybe you want to incorporate it into `ParamHelpers`?) or would you rather not like me to use this in my project?
## Object based vs. Functional CPOs
I implemented both, one in `R/CPOObjectBased.R`, the other in `R/CPOFunctional.R`; the code shared by both is mostly in `R/CPOAuxiliary.R`. Both have some advantages and disadvantages. The object based could use less memory in theory, since it does not carry around an `environment` in its model that usually contains the training data. It is also easier to debug if you like to use `debugonce`. In turn, the functional implementation can be applied directly to `Task` objects (since the CPO objects are just functions in this case) and could probably quite easily be coerced into collaborating with the `magrittr` package.
Maybe have a look at the concrete implementations of mine to see which one you like more.
### Note About Object Based Implementation
The `makePreprocWrapper` implementation in `mlr` relies on the transformation function returning an object `list(data = [data], control = [control])`. I had the idea of just having it return the resulting data, and using R magic to inspect the function's environment to get at the control. See e.g. the implementation of `cpoScale`:
```R
cpoScale = makeCPOObject("scale", center = TRUE: logical, scale = TRUE: logical,
  cpo.trafo = {
    ## boilerplate :-( Hope to get rid of this at some point
    targetdata = data[target]
    data[target] = NULL
    ## here we go
    result = scale(as.matrix(data), center = center, scale = scale)
    data[] = result
    data[target] = targetdata
    ## the 'control' object will be retrieved by the CPO machinery
    ## and given to cpo.retrafo
    control = list(center = attr(result, "scaled:center"),
      scale = attr(result, "scaled:scale"))
    data
  }, cpo.retrafo = {
    ## here we have the 'control' object
    as.data.frame(scale(as.matrix(data),
      center = control$center, scale = control$scale))
  })
```
What is your opinion about this? Alternatives are: Copying the entire `cpo.trafo` namespace to `cpo.retrafo`, so the user wouldn't need to worry about which variables are available and which are not. The downside to this: This would take the entire training data and save it inside the model, might be memory intensive. I could also stop being fancy and just return the `list(data, control)` as in `makePreprocWrapper`. There is a way to inspect `cpo.retrafo` and copy only the objects that are used by it, but this inspection is bound to be incomplete (the problem is halting problem equivalent) and could copy more data than the `retrafo` part needs.
## Composition operator
I choose `%>>%`, since it is similar, but not used by, `magrittr`. It applies to CPO in conjunction with Learners on the right and Tasks on the left, but does not do `Task %>>% CPO %>>% Learner` because of the associativity problem.
## State of implementation
The current roadmap, as I see it; comments?
* [X] makeParamSet syntactic sugar (`R/ParamSetSugar.R`)
* [X] Object based CPO (`R/CPOObjectBased.R`)
* [X] Functional CPO (`R/CPOFunctional.R`)
* [X] CPO composition
* [X] CPO attachment to Learner
* [X] CPO application to Task / data.frame
* [X] Easily do re-trafo on data after training on other data (i.e. perform the `predict` step without being attached to a Learner) 
* [X] Basic documentation (needs polish)
* [X] Base functionality tests
* [X] Small example CPOs (`R/CPO_concrete.R`)
* [X] nicer API to avoid boilerplate code that is the same in most CPOs
* [X] `properties` handling
* [X] CPO multiplexer
* [X] cpoCbind
* [x] metaCPO
* [x] list all built-in CPOs
* [x] select features to affect / ignore
* [x] Actually useful CPOs
* [x] Testing concrete CPOs
* [ ] export / fix hyperparameters
* [ ] cpoCbind avoid copying redundant columns
* [ ] target CPO

Maybe some day...
* [ ] Task weights
* [ ] Task blocking


larskotthoff:
Thanks, this looks great! A few comments:

- Would it be possible to have a `cpo()` "wrapper" method that takes an existing method like `pca()` as an argument?
- What's the difference between CPO objects and CPO functionals? Isn't the former just the 0 arity version of the latter?
- What happens when a preprocessing method and a learner have parameters with the same name?

Regarding the name: I would go with just "po", or probably "preprocOp" to give the user a better idea of what it is.

Composition operator: Why not overload `>>`?

Roadmap: Looks good to me.

mb706:
* The wrapper CPO operator is in planning.
* From the user's perspective, the added functionality of functional CPOs is that he can do `cpo1(cpo2(task))`, and use e.g. `magrittr` to do these kind of things. Otherwise, the API is identical (by design).
* when parameter names clash, an error is reported with a hint how to solve this: Using the `id` parameter.
```R
> cpoclash = makeCPOObject("clash", model: logical,
+   cpo.trafo = {
+     control = 0
+     data
+   }, cpo.retrafo = {
+     data
+   })
> cpoclash() %>>% makeLearner("classif.logreg")  # both have a 'model' param
Error in parameterClashAssert(cpo, learner, cpo$name, learner$name) : 
  Parameter "model" occurs in both clash and Logistic Regression
Use the id parameter when constructing, or setCPOId, to prevent name collisions.
> lrn = cpoclash(id = "cpo") %>>% makeLearner("classif.logreg")
> # id is prefixed to cpoclash's parameter names
> getParamSet(lrn)
             Type len  Def Constr Req Tunable Trafo
cpo.model logical   -    -      -   -    TRUE     -
model     logical   - TRUE      -   -   FALSE     -
```
* I think in R one can not overload arbitrary operators, only things delimited by `%`, and actually present operators. (`ggplot` does this to `+`, but I think overloading a normal operator would be a bit too bold for this project at this stage)

R/CPOAuxiliary.R
```diff
@@ -0,0 +1,373 @@
...
+}
+
+#' @export
+`%>>%.default` = function(cpo1, cpo2) {
+  stopf("%%>>%% not defined for objects of class c(%s)", paste0('"', class(cpo1), '"', collapse = ", "))
+}
+
+#' @export
+`%>>%.data.frame` = function(cpo1, cpo2) {
+  name = deparse(substitute(cpo1), 20)[1]
```
jakob-r: Why 20? :thinking:

R/getHyperPars.R
```diff
@@ -1,19 +1,19 @@
...
 #' @param for.fun [\code{character(1)}]\cr
 #'   Restrict the returned settings to hyperparameters corresponding to \code{when}
 #'   the are used (see \code{\link[ParamHelpers]{LearnerParam}}).
 #'   Must be a subset of: \dQuote{train}, \dQuote{predict} or \dQuote{both}.
 #'   Default is \code{c("train", "predict", "both")}.
 #' @return [\code{list}]. A named list of values.
 #' @family learner
+#' @family CPO
 #' @export
 getHyperPars = function(learner,  for.fun = c("train", "predict", "both")) {
```
jakob-r: Why don't we make this directly S3 then?

R/setHyperPars.R
```diff
@@ -21,9 +22,8 @@
 #' print(cl2)
 setHyperPars = function(learner, ..., par.vals = list()) {
```
jakob-r: and this as well.

jakob-r:
It looks pretty neat. I just had a fast look and wonder if this could by a separate package. You didn't change so much in mlr itself but I don't know how much you rely on non-exported functions. 
Also many functions should be in ParamHelpers. 

The idea of `paramSetSugar` is nice for us lazy developers but it's really hard to get other people to learn a new syntax (as simple as it might be). For example without your example I would assume that `numeric(, )^2` creates a numeric param with the trafo `f(x) x^2` :wink: .

R/CPOAuxiliary.R
mb706: I just want a short identifying "name" for the task. The task will be destroyed after application of `%>>%` anyways, so the user only sees this if there is some error message.

mb706: 20 was the shortest `deparse` was willing to give me.

R/getHyperPars.R
mb706: Don't get your comment. `getHyperPars` is S3?

mllg:
I'll have a closer look later, but here are my $.02 on `paramSetSugar`:

* It will be harder to debug.
* Silencing R CMD check will be tricky. Have you tried `codetools::checkUsagePackage()`?

mb706:
`paramSetSugar` isn't that much new syntax, is actually really easy:
```
NAME [= DEFAULT]: TYPE RANGE[^DIMENSION] [OPTIONS]
```
(of which only the stuff after the colon is really 'new', and if you need to use `DIMENSION` you have much bigger problems than confusing syntax). It's a shame that it is apparently so unpopular, to me it is the `ParamHelpers` that I always wanted.

It's also not a problem for debugging (unless you mean bugs in `ParamHelpers` itself?) since the resulting object is a vanilla `ParamSet`.

mb706:
The `R CMD check` complaints are not my fault, they are caused by the `"mxff"` learners.

mllg:
> The R CMD check complaints are not my fault, they are caused by the "mxff" learners.

If I'm not mistaken `R CMD check --as-cran` does a static code analysis. You should get errors similar to those in this example:
```r
foo = function() {
  paramSetSugar(a: logical, b: integer(0, 10), c: numeric(, )^2)
}
> checkUsage(foo)
<anonymous>: no visible binding for global variable 'a' (:2)
<anonymous>: no visible binding for global variable 'b' (:2)
<anonymous>: possible error in integer(0, 10): unused argument (10) (:2)
<anonymous>: possible error in numeric(, ): unused argument () (:2)
```
`dyplr` has a vignette on NSE (<https://cran.r-project.org/web/packages/dplyr/vignettes/nse.html>). Summed up, it is required to use either formulas, `quote()`, strings or `lazyeval` in packages to avoid these messages. 



> It's also not a problem for debugging (unless you mean bugs in ParamHelpers itself?) since the resulting object is a vanilla ParamSet.

That's true, but what if you mess up things up during initialization, e.g. if you have a typo in your default. Do you get good error messages?

Don't get me wrong, I do not want to discourage you. But I've been down the NSE road before and I had to make so many compromises to overcome such technical issues that I didn't like my own mini language anymore and finally rolled everything back.

I would suggest to support only strings and make the format more flexible to cover everything (discrete params, expressions, different bounds per dimension, ...). This way, we would also have a short ASCII representation of the object and would not need to fiddle with the R language.
But be warned: This is usually quite tedious (c.f. <https://github.com/mllg/checkmate/blob/master/src/qassert.c>).

R/getHyperPars.R
jakob-r: My bad. I got confused because it was kind of wrong before with `assertClass(learner, classes = "Learner")` and you corrected it.

mb706:
Wow, thanks, I hadn't expected to see these kinds of problems. If there is no way to convince the checker to accept this, the solution I can think of is to make the `=` mandatory with `NA` meaning no default (so all param names look like named parameter passing) and use square brackets for the range indication.
```R
> foo = function() {
+   paramSetSugar(a = NA: logical, b = NA: integer[0, 10], c = NA: numeric[, ]^2)
+ }
> codetools::checkUsage(foo)
>
```
I really wouldn't want to use strings for this if avoidable. In the current form, one can use actual expressions for most of the values, have syntax highlighting, wrap around lines, etc. To prove a point, and to respond to your suggestion to make extensions:
```R
> paramSetSugar(a = expression(10 + 10): integer(c(1, 2, 3), c(4, 5, 6))^3)
           Type len     Def         Constr Req Tunable Trafo
a integervector   3 10 + 10 1,2,3 to 4,5,6   -    TRUE     -
```
I didn't even *know* this would work. (Though imho the `^3` should be made redundant if vectorial bounds are given.)

Re debugging: Internally I call the R constructors, so I get no more and no fewer errors than I would usually get:
```R
> paramSetSugar(a = 11: integer(0, 10))
Error in makeParam(id = id, type = "integer", learner.param = TRUE, lower = lower,  : 
  a : 'default' must be a feasible parameter setting.
```
If you mess up the syntax (and you happen to have something that is valid R so the parser doesn't complain), you get things like
```R
> paramSetSugar(integer(, ))
Error in formerr(pstring, "`:` was missing or at unexpected position.") : 
  Parameter 'integer(, )' must be of the form
NAME [= DEFAULT]: TYPE [RANGE] [^ DIMENSION] [SETTINGS]
`:` was missing or at unexpected position.
```

mb706:
I have now added `retrafo()` which lets one retrieve an object to perform the trained transformation on a new object.
```R
> # rotate a data.frame
> rotated.attitude = attitude %>>% cpoPca()
> # use the same rotation matrix on a
> # shorter version of the DF
> short.rotated.attitude = retrafo(rotated.attitude)(head(attitude))
> all.equal(head(rotated.attitude), short.rotated.attitude,
+   check.attributes = FALSE)
[1] TRUE
```
The result of `retrafo` is currently a function which can be applied to new data. Alternatively, it could be an object that gets put into `predict` together with the new data. Would you rather want me to use this second approach?

Also, I have now changed `paramSetSugar` to use square brackets instead of round parentheses for range indication, as well as use `value = NA` to give it no default value but still avoid the static code analysis to nag about undefined variable access. (To give the actual default `"NA"`, use `(NA)`)

mb706:
```
LaTeX errors when creating PDF version.
This typically indicates Rd problems.
LaTeX errors found:
! LaTeX Error: File `xkeyval.sty' not found.
```
Anyone have any idea why this happens?

mb706:
News:
* `predict`, `%>>%` now work with retrafo
* decompose a list of preprocessors OR retrafos by using `as.list`; re-compose them with `chainCPO`
* retrafo state can be inspected
```R
> ppchain = cpoPca(id = "a", center=F, scale=F) %>>% cpoScale()
> as.list(ppchain)
[[1]]
pca.a(a.center = FALSE, a.scale = FALSE)

[[2]]
scale(center = TRUE, scale = TRUE)

> rtchain = retrafo(pid.task %>>% ppchain)
> as.list(rtchain)
[[1]]
[RETRAFO pca(center = FALSE, scale = FALSE)]

[[2]]
[RETRAFO scale(center = TRUE, scale = TRUE)]

> getRetrafoState(as.list(rtchain)[[2]])
$center
[1] TRUE

$scale
[1] TRUE

$control
$control$center
          PC1           PC2           PC3           PC4           PC5 
-1.638530e+02  3.839000e+01 -1.497887e+00 -1.360065e-01  3.442811e-01 
          PC6           PC7           PC8 
-4.998194e-01 -4.381292e-02 -5.983028e-03 

$control$scale
       PC1        PC2        PC3        PC4        PC5        PC6        PC7 
87.1191830 81.7970328 22.0721032 14.1024163 10.7807413  7.3319184  2.7978175 
       PC8 
 0.3226096 
> predict(rtchain, pid.task)
Supervised task: PimaIndiansDiabetes-example
Type: classif
[...]
```

larskotthoff:
I'm still not wild about the name. It's a bit cryptic. Could we make it a bit more intuitive?

mb706:
News:
* check that prediction data conforms to training data (same column names, similar enough column types)
* properties handling (not tested yet): Add `.properties`, `.properties.needed`, `.properties.adding` parameters to indicate the kind of data a CPO can handle, to indicate the kind of data it needs the data receiver (e.g. attached learner) to have, and the properties it adds to a given learner. An is a CPO that converts factors to numerics: The receiving learner needs to handle numerics, so `.properties.needed = "numerics"`, but it *adds* the ability to handle factors (since they are converted), so `.properties.adding = c("factors", "ordered")`.
* option to split up data into types (numeric, factor, ordered, other) to easily write CPOs that only operate on a certain type. E.g. centering only the numeric columns:
```R
> cpo = makeCPOObject("scale", center = TRUE: logical, scale = TRUE: logical,
+ .datasplit = "most", cpo.trafo = {
+ # `data` is a list with three entries: $numeric, $factor, $other
+   result = scale(as.matrix(data$numeric),
+     center = center, scale = scale)
+   control = list(center = attr(result, "scaled:center"),
+     scale = attr(result, "scaled:scale"))
+   data$numeric = as.data.frame(result)
+   data }, cpo.retrafo = {
+   data$numeric = as.data.frame(scale(as.matrix(data$numeric),
+     center = control$center, scale = control$scale))
+   data })
> scale = cpo()
> data.frame(a = 1:3, b = 3:5, c = c("a", "b", "c"), d = c("a", "a", "a"))
  a b c d
1 1 3 a a
2 2 4 b a
3 3 5 c a
> df = data.frame(a = 1:3, b = 3:5, c = c("a", "b", "c"), d = c("a", "a", "a"))
> df
  a b c d
1 1 3 a a
2 2 4 b a
3 3 5 c a
> df %>>% scale
   a  b c d
1 -1 -1 a a
2  0  0 b a
3  1  1 c a
```

ja-thomas:
You might want to have a look at the recipes package: https://topepo.github.io/recipes/

This tries to do something similar, maybe there is some interesting stuff in there.

mb706:
Thanks, this really looks kind of like what I am doing! As I understand it, their 
```R
recipe(formula, data) %>% operation(...) %>% operation(...)
```
would be quite similar to
```R
makeClassifTask(id, data, target) %>>% operation(...) %>>% operation(...)
```
(obviously CPO is greater, it even has one more 'greater than' sign in its operator)

They can also handle weights (which I am ignoring for now), and they seem to be able to specify subsets of columns on which to operate for each operator (which went straight to my to-do list). CPO, on the other hand, has the advantage of `setHyperPars`, and that it can attach operations to learners and tune them.

mb706:
### News:
* `%<<%` as an alternative for `%>>%`, if the user happens to prefer to write CPOs in the other direction
* extended option to split up data into subsets by type: it is now possible to specify that the `cpo.trafo` only gets handed numerics / factors / etc. instead of the whole data. This reduces boilerplate code further.
* getting CPO from a learner (not tested)
* removed "object based" and "functional" CPO, the new CPO backend handles both. Depending on whether `cpo.retrafo` is specified, the CPO expects `cpo.trafo` to either have a `control` variable that will be passed to `cpo.retrafo`, or to define the `cpo.retrafo` inside the function.
* `NULLCPO`: neutral element of `%>>%` ( `x %>>% NULLCPO == x`)

### Outlook:
I am currently working on CPOs that affect the target column(s) of a task. This would be useful for e.g. log-transforming a target variable, but also to convert between task types. Examples (not yet implemented):
```R
train.task = subsetTask(bh.task, 1:100)
test.task = subsetTask(bh.task, 101:200)

log.trans.learner = logtransform() %>>% makeLearner("regr.lm")

model.trans = train(log.trans.learner, train.task)
pred.trans = predict(model.trans, test.task)

# alternatively, if we don't want to attach
# to a learner for some reason
train.task.trans = train.task %>>% logtransform()
model = train(makeLearner("regr.lm"), train.task.trans)
# we need to re-transform "test.task" using the retrafo
rt = retrafo(train.task.trans)
test.task.trans = test.task %>>% rt
pred.trans.logdomain = predict(model, test.task.trans)
# now this prediction is in the log-domain.

# if we want to have the predictions de-log-transformed
# we would need to get the target inverter
# using `inverter()`
# First we need to tell mlr to build the inverter using `tagInverse`
test.task.trans = tagInverse(test.task) %>>% rt
pred.trans.logdomain = predict(model, test.task.trans)
# now we get the inverter
inv = inverter(test.task.trans)
pred.trans.inverted = invert(inv, pred.trans.logdomain)
# --> should be similar to pred.trans
```
Please write if you have any comments, suggestions, opinions!

larskotthoff:
I'm not sure about allowing `%<<%` as well. What's the semantics of `preproc %>>% learner %<<% data`? Or even weirder `data %>>% learner %<<% preproc`.

mb706:
All operators are evaluated left to right, so `a %>>% b %<<% c` is the same as `c %>>% a %>>% b`. If the user writes programs that confuse him, its his own fault ;)
Also `data %>>% learner` currently doesn't work since this could (without NSE tricks I'd try to avoid here) not be told apart from `data %>>% preproc() %>>% learner`. If the user writes this, he probably means `data %>>% (preproc() %>>% learner)`, but that is not the order of evaluation (its `(data %>>% preproc()) %>>% learner`).
Incidentally, there wouldn't be this problem when using `%<<%`.

mb706:
News:
* "stateless" CPOs which don't depend on training data when predicting.

mb706:
### News:
* **CPO Multiplexer**: Combine many CPOs into one, with an extra `selected.cpo` parameter that chooses between them.
```R
> df = data.frame(a = 1:3, b = 6:8, c = 1:3 * 2)
> cp = cpoMultiplex(list(cpoScale, cpoPca))
> getParamSet(cp)
                 Type len   Def    Constr Req Tunable Trafo
selected.cpo discrete   - scale scale,pca   -    TRUE     -
scale.center  logical   -  TRUE         -   Y    TRUE     -
scale.scale   logical   -  TRUE         -   Y    TRUE     -
pca.center    logical   -  TRUE         -   Y    TRUE     -
pca.scale     logical   -  TRUE         -   Y    TRUE     -
> df %>>% cp
   a  b  c
1 -1 -1 -1
2  0  0  0
3  1  1  1
> df %>>% setHyperPars(cp, selected.cpo = "pca")
        PC1           PC2 PC3
1 -1.732051 -1.110223e-16   0
2  0.000000  0.000000e+00   0
3  1.732051  1.110223e-16   0
> df %>>% setHyperPars(cp, selected.cpo = "pca", pca.center = FALSE)
         PC1         PC2           PC3
1 -0.8362205  0.26395263 -5.551115e-17
2 -1.3408104  0.05294043 -1.110223e-16
3 -1.8454003 -0.15807177 -1.110223e-16
```
* **CPO Applicator**: A simple CPO with one parameter which gets applied to the data as CPO. This is different from the multiplexer in that its parameter is free and can take any value that behaves like a CPO. On the downside, this does not expose the argument's parameters to the outside.
```R
> cp = cpoApply()
> cp.scale = setHyperPars(cp, cpo = cpoScale())
> getParamSet(cp.scale)
       Type len Def Constr Req Tunable Trafo
cpo untyped   -   -      -   -    TRUE     -
> df %>>% cp.scale
   a  b  c
1 -1 -1 -1
2  0  0  0
3  1  1  1
```
### Outlook:
The task-affecting CPOs described [earlier](https://github.com/mlr-org/mlr/pull/1827#issuecomment-307820959) is almost done, in that all the code is there, it just doesn't work yet. As soon as that is done, I will focus on tests, stability, and documentation.

larskotthoff:
What's a use case for the multiplexer?

larskotthoff:
What's a use case for the multiplexer?

berndbischl:
> What's a use case for the multiplexer?

use ICA *or* PCA.

berndbischl:
regarding the ParamSetSugar thing: martin and I had a longer discussion today. we designed 3 systems
1) his one (ok that existed before)
2) a non-NSE  (so SE....) one, with just shorter function names
3) a string based one

we should discuss which one should be supported by ParamHelpers because this belongs there and has nothing to do with this thread.

berndbischl:
@mb706 can you please maintain the first post here so it explains the current state with all current options correctly? or maybe soon already right a few tutorial style pages (short ones, that simply outline the API to intelligent readers)

larskotthoff:
Re use case: Ok, so this would be mostly for tuning? What I'm getting at is that there seems to be no mechanism to make the switch based on e.g. the kind of data being passed in. That is, a "flexible" CPO that adapts to the data you put in it.

berndbischl:
> Re use case: Ok, so this would be mostly for tuning?

yes, sure. at least the my ICA PCA use case that i came up with now. 


> What I'm getting at is that there seems to be no mechanism to make the switch based on e.g. the kind of data being passed in. That is, a "flexible" CPO that adapts to the data you put in it.

well now it is your turn to define a concrete use case for that.

larskotthoff:
If all features numeric do PCA else do nothing.

berndbischl:
> If all features numeric do PCA else do nothing.

that seems unreasonable and is not a concrete use case. if only some features are numeric, you would only do the PCA on those. and martin already supports that. and if you want more flexibility here, you would add more options to the PCA op i guess.

larskotthoff:
Ok, how about if missing values run expensive imputation x, else don't (and I really don't want to run the method because it is expensive even when there's nothing to do). Or scale features only if the range is more than y. Or don't run PCA to reduce to 3 dimensions if the data has fewer than 3 features.

And here's a really tricky one (we probably can't do this): scale and center values only if we do clustering.

berndbischl:
> Ok, how about if missing values run expensive imputation x, else don't (and I really don't want to run the method because it is expensive even when there's nothing to do)

i dont get this: x = impute %>>% rpart
if i run x on a dataset that has no NAs, nothing happens. of course the imputation op will do nothing?

> Or scale features only if the range is more than y. 
what is y? a selected threshold? that would be hyper par of the scale OP?

> Or don't run PCA to reduce to 3 dimensions if the data has fewer than 3 features.
that would be a hyperpar of the PCA op? 

ok the last 2 i somehow get, in the sense of what you technically might want. are you suggesting data dependent pre-conditions so that the CPO is actually applied?

mb706:
News:
* **cpoSelect**: drop all excet a few selected columns from a `data.set`/`Task`.
```R
> head(iris)
  Sepal.Length Sepal.Width Petal.Length Petal.Width Species
1          5.1         3.5          1.4         0.2  setosa
2          4.9         3.0          1.4         0.2  setosa
3          4.7         3.2          1.3         0.2  setosa
4          4.6         3.1          1.5         0.2  setosa
5          5.0         3.6          1.4         0.2  setosa
6          5.4         3.9          1.7         0.4  setosa
> head(iris) %>>% cpoSelect("factor")
  Species
1  setosa
2  setosa
3  setosa
4  setosa
5  setosa
6  setosa
> head(iris) %>>% cpoSelect(pattern = "Width")
  Sepal.Width Petal.Width
1         3.5         0.2
2         3.0         0.2
3         3.2         0.2
4         3.1         0.2
5         3.6         0.2
6         3.9         0.4
```
* **cpoCbind**: `cbind` of other CPOs as operation.
```R
> df = data.frame(a = 1:3, b = 3:1)
> df
  a b
1 2 6
2 4 4
3 6 2
> operation = cpoCbind(
+   centered = cpoScale(scale = FALSE, id = "s1"),
+   scaled = cpoScale(center = FALSE, id = "s2"),
+   copied = NULLCPO)
> df %>>% operation
  centered.a centered.b  scaled.a  scaled.b copied.a copied.b
1         -2          2 0.3779645 1.1338934        2        6
2          0          0 0.7559289 0.7559289        4        4
3          2         -2 1.1338934 0.3779645        6        2
> # printing 'operation' gives the graph representation
> # as can be seen, the hyperparameters are exported
> operation
cbind(s2.center = TRUE, s2.scale = FALSE, s1.center = FALSE, s1.scale = TRUE, ...
O     scale.s2(s2.center = FALSE, s2.scale = TRUE)
|
+<O   scale.s1(s1.center = TRUE, s1.scale = FALSE)
|
O   CBIND[centered,scaled,copied]

> # what happens if we switch off the 'scaling' flag of the second cpoScale application
> # (s2.scale -> FALSE)?
> df %>>% setHyperPars(operation, s1.scale = FALSE)
  centered.a centered.b scaled.a scaled.b copied.a copied.b
1         -2          2        2        6        2        6
2          0          0        4        4        4        4
3          2         -2        6        2        6        2
```

mb706:
### News:
* **listCPO**: List built-in CPOs
* **cpoDummyEncode**
```R
> head(iris[5]) %>>% cpoDummyEncode()
  Speciessetosa Speciesversicolor Speciesvirginica
1             1                 0                0
2             1                 0                0
3             1                 0                0
4             1                 0                0
5             1                 0                0
6             1                 0                0
```
* **affect.\***: Every CPO gets parameters to possibly only affect a subset of columns. Select by index, type, name, or regex pattern.
```R
> head(iris) %>>% cpoPca(affect.pattern = "Width")
  Sepal.Length Petal.Length Species        PC1           PC2
1          5.1          1.4  setosa -0.1089088  0.0534891611
2          4.9          1.4  setosa  0.3831705 -0.0351560166
3          4.7          1.3  setosa  0.1863388  0.0003020545
4          4.6          1.5  setosa  0.2847546 -0.0174269810
5          5.0          1.4  setosa -0.2073247  0.0712181967
6          5.4          1.7  setosa -0.5380303 -0.0724264147
```

mb706:
### News:
* **cpoMeta**: A CPO that builds data-dependent CPO networks. This is a generalized CPO-Multiplexer that takes a function which decides (from the data, and from user-specified hyperparameters) what CPO operation to perform. This addresses the concerns expressed by @larskotthoff [here](https://github.com/mlr-org/mlr/pull/1827#issuecomment-308891052).
(Nonsensical) Example:
```R
s.and.p = cpoMeta(logical.param: logical,
.export = list(cpoScale(id = "scale"), 
  cpoPca(id = "pca", scale = FALSE, center = FALSE)),
cpo.build = function(data, target, logical.param, scale, pca) {
  if (logical.param || mean(data[[1]]) > 10) {
    scale %>>% pca
  } else {
    pca %>>% scale
  }
})
```
The resulting CPO `s.and.p` performs scaling and PCA, with the order depending on the parameter `logical.param` and on whether the mean of the data's first column exceeds 10. If either of those is true, the data will be first scaled, then PCA'd, otherwise the order is reversed.
The all CPOs listed in `.export` are passed to the `cpo.build`. It is also possible to use non-exported CPOs, but the ones in `.export` have their parameters exported:
```R
> s.and.p(logical.param = TRUE)
meta(logical.param = TRUE, scale.center = TRUE, scale.scale = TRUE,
  pca.center = FALSE, pca.scale = FALSE)
```

larskotthoff:
Great, thanks @mb706. What about the last example I mentioned (applying an operator depending on what else is in the CPO chain)? Do you think that that would be possible?

To be clear, I'm not saying that this must be supported, but I'm curious whether we could.

mb706:
### News:
There is a quick-and-dirty "vignette" showing off some CPO features [here](https://github.com/mlr-org/mlr/blob/ComposablePreprocOperators/todo-files/CPO%20Vignette.ipynb). It is based on jupyter, which github tries to render, but some of the output is a bit broken. It is, however, an ideal starting point if you want to play around a bit with CPOs.

mb706:
@larskotthoff I don't see a good way to have behaviour depending on the rest of the chain. I am trying to make
`(data %>>% CPO1) %>>% CPO2` behave the same way as `data %>>% (CPO1 %>>% CPO2)`. However, in the first version (without some NSE tricks that would easily be defeated if an intermediate variable was used), `CPO1` could not know what `CPO2` is doing, while in the second version it could.

I guess the reason why this behaviour *could* be useful is that if someone wants to do tuning with a multiplexed learner, and a CPO that should do something only when the multiplexed learner is in a certain state, then there would be two hyperparameters (one controlling the `ModelMultiplexer`, one controlling the CPO) that should be dependent on each other. I think the "right" solution here would be to have a `HyperParsTransformer` learner-wrapper that changes certain hyperparameters in dependence of others and hence only exports a subset of a learner's parameters.
```R
newlearner = makeHyperParsTrafoWrapper(learner,
  hide.params = c("param1", "param2"),  # hyperpars that will not be part of 'newlearner'
  add.params = paramSetSugar(combined12: numeric[0, 1]),  # new hyperpar to introduce
  hptrafo = function(params.in) {
    params.in$param1 = params.in$combined12 ^ 2
    params.in$param2 = ifelse(params.in$combined12 > 0.5, "yes", "no")
    return(params.in)
  # makeHyperParsTrafoWrapper calls this function whenever setHyperPars is called
  # and changes the underlying learner's parameters according to the returned object.
  # it is smart enough to remove 'combined12' from the returned list.
  })
```

larskotthoff:
Ok thanks, that makes sense.

mb706:
### News:
* Imputation: Basically a wrapper around `mlr::impute` and `mlr::reimpute`. There are two general CPOs that have arguments pretty much mirroring `impute`, and there are "specialised" `impute` CPOs that implement individual imputation methods.
```R
> df = data.frame(a = 1:3, b = c(NA, 2, 3))
> df
  a  b
1 1 NA
2 2  2
3 3  3
> df %>>% cpoImpute(cols = list(b = imputeMedian()))
  a   b
1 1 2.5
2 2 2.0
3 3 3.0
> df %>>% cpoImputeMedian()
  a   b b.dummy
1 1 2.5    TRUE
2 2 2.0   FALSE
3 3 3.0   FALSE
> df %>>% cpoImputeMedian(make.dummy.cols = FALSE)
  a   b
1 1 2.5
2 2 2.0
3 3 3.0
```
The specialised CPOs are:
```
name              description
cpoImputeConstant Imputation using a constant value.                                       
cpoImputeHist     Imputation using random values with probabilities approximating the data.
cpoImputeLearner  Imputation using the response of a classification or regression learner. 
cpoImputeMax      Imputation using constant values shifted above the maximum.              
cpoImputeMean     Imputation using the mean.                                               
cpoImputeMedian   Imputation using the median.                                             
cpoImputeMin      Imputation using constant values shifted below the minimum.              
cpoImputeMode     Imputation using the mode.                                               
cpoImputeNormal   Imputation using normally distributed random values.                     
cpoImputeUniform  Imputation using uniformly distributed random values.                    
```

mb706:
### News:
* Feature Filtering: Basically a wrapper around `mlr::featureFilter`. There is a general `cpoFilterFeatures` cpo that takes any method as an argument, or many specialised CPOs for each filtering method.
```R
> head(getTaskData(iris.task %>>% cpoFilterFeatures(method = "variance", perc = 0.5)))
  Sepal.Length Petal.Length Species
1          5.1          1.4  setosa
2          4.9          1.4  setosa
3          4.7          1.3  setosa
4          4.6          1.5  setosa
5          5.0          1.4  setosa
6          5.4          1.7  setosa
> head(getTaskData(iris.task %>>% cpoFilterVariance(perc = 0.5)))
  Sepal.Length Petal.Length Species
1          5.1          1.4  setosa
2          4.9          1.4  setosa
3          4.7          1.3  setosa
4          4.6          1.5  setosa
5          5.0          1.4  setosa
6          5.4          1.7  setosa
```
The specialised CPOs are:

| name | description |
| --- | --- |
| cpoFilterAnova | Filter features using analysis of variance. |
| cpoFilterCarscore | Filter features using correlation-adjusted marginal correlation. |
| cpoFilterChiSquared | Filter features using chi-squared test. |
| cpoFilterGainRatio | Filter features using entropy-based information gain ratio |
| cpoFilterInformationGain | Filter features using entropy-based information gain. |
| cpoFilterKruskal | Filter features using the Kruskal-Wallis rank sum test. |
| cpoFilterLinearCorrelation | Filter features using Pearson correlation. |
| cpoFilterMrmr | Filter features using 'minimum redundancy, maximum relevance'. |
| cpoFilterOneR | Filter features using the OneR learner. |
| cpoFilterPermutationImportance | Filter features using predictiveness loss upon permutation of a variable. |
| cpoFilterRankCorrelation | Filter features using Spearman correlation. |
| cpoFilterRelief | Filter features using the ReliefF algorithm. |
| cpoFilterRfCImportance | Filter features using party::cforest variable importance. |
| cpoFilterRfImportance | Filter features using randomForest variable importance. |
| cpoFilterRfSRCImportance | Filter features using randomForestSRC.rfsrc. |
| cpoFilterRfSRCMinDepth | Filter features using randomForestSRC minimal depth. |
| cpoFilterSymmetricalUncertainty | Filter features using entropy-based symmetrical uncertainty |
| cpoFilterUnivariate | Filter features using the predictiveness using a given learner. |
| cpoFilterVariance | Filter features using feature variance. |

berndbischl:
hi, 

a few things

1) can you please maintain the overview with the news on new stuff on a separate page? we will lose the overview here. maybe the wiki?

2) can you please add a CPO to remove constant features? 

3) i "dislike" a bit that you are adding so many CPOs for eg filtering and imputation. they all have the same API. would it not be better to have cpoImpute(method = "foo")?
thats would also make it much simpler to tune over the method

mb706:
1. Sure, sorry for the spam. I will keep the "preliminary vignette", linked to in the first line of the PR, up to date instead.
2. Sure.
3. The nice thing with having the specialised CPOs is that if they have different parameters, it is easier to document. I actually ran into this problem in production code: I tried to optimize over filtering methods and only noticed later that some of them have well hidden and poorly documented parameters. [I even opened an issue about that](https://github.com/mlr-org/mlr/issues/1697). If a user wants to optimize over different filtering or imputation methods, he now has the choice of using a CPO Multiplexer or using the "general" CPOs (`cpoFilterFeatures`, `cpoImputeAll`) which work pretty much the way you suggest.

mb706:
re "they all have the same API" -- I was thinking about that. I am taking many arguments from the original functions just because of "institutional inertia", i.e. people would complain if they weren't there any more. However, one could also try to make individual CPOs as simple as possible, remove parameter settings, and obligate the user to use further CPOs if he wants something that was previously controlled by parameters. For imputation, this could entail:
1. Removing the `make.dummy.cols` argument, always creating dummy cols, and having the user use `cpoSelect(pattern = "^dummy\\.", invert = TRUE)` to remove them if he doesn't want them
2. Alternatively, removing the `make.dummy.cols` argument and creating a separate CPO that solely creates dummy columns. If the user wants dummy columns, he has to prepend that other CPO.
3. Removing the `force.dummies` argument. Whichever 1. or 2. gets used, always create *all* dummy columns. If the user doesn't like constant features / unnecessary dummy columns, he can use a constant feature removing CPO.
4. Removing the `impute.new.levels` and `recode.factor.levels` argument. There should be a separate CPO that equalizes all data between train and test runs (akin to `fix.factor.levels` in learners).

Examples could look like this:
```R
data %>>% cpoImputeMin(make.dummy.cols = TRUE, force.dummies = FALSE,
  impute.new.levels = TRUE, recode.factor.levels = TRUE)
# these are all the default settings
# so the actual line would be as short as
data %>>% cpoImputeMin()

# this could be turned into
data %>>% cpoFixFactors() %>>%
  cpoDummyCreator(force.dummies = FALSE) %>>% cpoImputeMin()
# or possibly into
data %>>% cpoFixFactors() %>>%
  cpoDummyCreator() %>>% cpoConstFeatRemover() %>>% cpoImputeMin()
# or maybe, if cpoDummyCreator() just *replaces* columns
data %>>% cpoFixFactors() %>>%
  cpoCbind(cpoImputeMin(), dummy = cpoDummyCreator() %>>% cpoConstFeatRemover())
```
This is definitely very elegant, because it would break each CPO down into its essential elements. However, some of these things, for example 4., should pretty much always happen, and that would mean that the user would need to use two or three CPOs (he would probably end up using `cpoFixFactors()` most of the time) when in the past, he just used one function call (`impute` / `reimpute`, in this case) that handled all the sensible defaults for him.

One could export the basic operations as "CPO Atoms", and the operations users usually want to use (e.g. including `cpoFixFactors`) as "CPO Molecules" on top of that, but this is obviously going in the opposite direction of what you were suggesting you want this thing to go.

berndbischl:
> Sure, sorry for the spam. I

it really isnt spam, i just want a "complete" doc thats easier to read than this thread.

berndbischl:
regarding the "joining" of the CPOs: 
OK i get your argument, and suspected you would bring it, as I kinda thought about this too. and kinda agree :)
if the multiplexer is simple to use, you are right. lets discuss this on sunday. 

for now add more CPOs, especially simple but useful stuff like remove const.
then we can check use-ability through cases.

mllg:
My 2 cents:

* Would be great to be able to use `magrittr` pipes instead of the one you've implemented.
* Name changes to consider:
  * Chain -> Pipe/Pipeline
  * Meta-CPO -> Conditional CPO
  * CBind CPO -> Gather/Union CPO
* Seems like a one-to-many operator is missing (e.g., bootstrap). This could be called spread/burst, and followed by a gather/union operator.
* Maybe its worth to have a quick look at [flowr](https://github.com/sahilseth/flowr) where workflows are defined with the verbs serial/burst/gather.

mb706:
Thanks for your input!
* I will think about the `magrittr` thing. So far I see two options:
  1) Make CPOs just "functions", so that `CPO1 %>% CPO2` could also be written `CPO2(CPO1)`. I had that, at one point ("CPOFunctional"), but it turned out to be quite ugly internally. I have some ideas now how to make that nicer, so if there is demand for it I think I can do it.
  2) Overload the `%>%` operator, which I think is the worse option. The few looks I had at `magrittr` internals gave me the impression it isn't built to be extended that way.
* Thanks for the ideas! Please (everyone) write me all the ideas for names that you have. Many of the current names are suboptimal, and get worse whenever features get added and warp the actual use of some of the commands / CPOs involved. I will probably make a big renaming session when the project is more mature and this won't happen as much.
* I think the one-to-many thing is in a way already implemented, and in a different way it is not and I don't know how it should work.
  1) Right now, `cpoCbind` recognizes automatically when a common operation is performed for any of its inputs. E.g. when you use
      ```R
      selector = mlr:::cpoSelect(type = "numeric")
      scale = cpoScale(id = "scale")
      scale.pca = scale %>>% cpoPca(center = FALSE, scale = FALSE, id = "pca")
      cbinder.select = cpoCbind(
        scaled = selector %>>% scale,
        pcad = selector %>>% scale.pca,
        original = NULLCPO)
      ```
      `cpoCbind` recognizes that `selector` and `scale` happen both in the `scaled` and `pcad` lane, so in a way there is a "burst" (going to output directly, and going through `pca`) happening. When you print `cbinder.select`, it even shows the data flow network (flow is downward or right/left where indicated by "`>`" or "`<`"; "`O`" indicates a CPO doing something; inflow of the original data is not shown and implied in the parentless `O` and the `cpoCbind`):
      ```R
      O     select(type = numeric, index = integer(0), names = character(0), pattern
      |    = <NULL>, pattern.ignore.case = FALSE, pattern.perl = FALSE,
      |    pattern.fixed = FALSE, invert = FALSE)
      |    
      O>+   scale.scale(scale.center = TRUE, scale.scale = TRUE)
      | |  
      +<O   pca.pca(pca.center = FALSE, pca.scale = FALSE)
      |   
      O   CBIND[scaled,pcad,original]
      ```
  2) I don't really know how to implement anything more sophisticated, e.g. bootstrap. If different CPOs subset the data in different ways, the `cbind` metaphor falls apart and I don't know in what other ways the data should be combined. Currently the CPOs operate with the assumption that one line of input should remain one line of output -- I think under these constraints the functionality mentioned in "i." is all there could be (?). I plan to also implement CPOs that can subset input rows, or create new input rows from the present data, but so far was planning to just forbid `cpoCbind` to use any of these kinds of CPOs. Do you have any ideas how one would combine different data streams that were subsetted in different ways? Or how to "burst" or "union" data in ways not yet possible? (Example R input and what you would it expect to do is especially helpful ;-) )
* The `flowr` "burst" seems to mean launching a bunch of jobs in parallel, each with different "hyperparameter" settings (??), which I think would be quite hard to map to CPOs and their emphasis on "exporting" hyperpars so that a mlr "tuner" has access to them.

larskotthoff:
What's the status here? Does this still need to merged for mlrCPO to work?

mb706:
Yes please, `mlrCPO` relies on a few kinda-internal functions of `mlr`. Should be pretty stable as well (my last few commits were just running after merge conflicts).

larskotthoff:
Thanks, merging.

PR #1833, opened 2017-06-09
Improve Survival stuff

mllg:
Summary:

* New helper to set measure parameters (`setMeasurePars()`).
* Two new survival measures: `cindex.uno` and `iauc.uno`. We had more measures in the other PR (#1372), but they turned out to be numerical unstable. They can still be integrated in separate PRs in the future (if required).
* `surv.cforest` predicted median survival times of the respective terminal nodes (note that this is undocumented). But we expect the survival learners to predict a numerical value which is high if the individual is expected to die quickly. Thus this learner yielded reversed predictions. Fixed by multiplying with -1. Additionally, I've added a test to check if the performance on a really simple task is closer to the best possible value than to the worst possible.
* There is currently no way to let `surv.penalized` predict risks. The method suggested by the package authors seems to be broken, yielding constant predictions for very basic tasks. I've removed the learner, and glmnet seems to be a viable alternative for penalized survival regression.


R/measures.R
```diff
@@ -1337,19 +1340,57 @@ measureMultilabelTPR = function(truth, response) {
...
   }
 )
 
+#' @export cindex.uno
+#' @rdname measures
+#' @format none
+cindex.uno = makeMeasure(id = "cindex.uno", minimize = FALSE, best = 1, worst = 0,
+  properties = c("surv", "req.pred", "req.truth", "req.model"),
+  name = "Uno's Concordance index",
+  note = "Fraction of all pairs of subjects whose predicted survival times are correctly ordered among all subjects that can actually be ordered. In other words, it is the probability of concordance between the predicted and the observed survival. Corrected by weighting with IPCW as suggested by Uno.",
```
larskotthoff: Is there a paper reference for this?

R/measures.R
```diff
@@ -1337,19 +1340,57 @@ measureMultilabelTPR = function(truth, response) {
...
+    times = seq(from = 0, to = max.time, length.out = extra.args$resolution)
+    surv.train = getTaskTargets(task, recode.target = "rcens")[model$subset]
+    y = getPredictionResponse(pred)
+    if (anyMissing(y))
+      return(NA_real_)
+    survAUC::AUC.uno(Surv.rsp = surv.train, Surv.rsp.new = getPredictionTruth(pred), times = times, lpnew = y)$iauc
+  },
+  extra.args = list(max.time = NULL, resolution = 1000)
+)
+
```
larskotthoff: Could you add hand-constructed tests for the new measures please?

R/RLearner_surv_gamboost.R
```diff
@@ -52,8 +52,5 @@ trainLearner.surv.gamboost = function(.learner, .task, .subset, .weights = NULL,
 
 #' @export
 predictLearner.surv.gamboost = function(.learner, .model, .newdata, ...) {
-  if (.learner$predict.type == "response")
-    predict(.model$learner.model, newdata = .newdata, type = "link")
-  else
-    stop("Unknown predict type")
+  predict(.model$learner.model, newdata = .newdata, type = "link")
```
larskotthoff: Why is the if no longer necessary here (and below)?

R/RLearner_surv_cforest.R
```diff
@@ -53,7 +53,8 @@ trainLearner.surv.cforest = function(.learner, .task, .subset,
 
 #' @export
 predictLearner.surv.cforest = function(.learner, .model, .newdata, ...) {
-  predict(.model$learner.model, newdata = .newdata, ...)
+  # cforest returns median survival times; multiply by -1 so that high values correspond to high risk
+  -1 * predict(.model$learner.model, newdata = .newdata, type = "response", ...)
```
larskotthoff: Could you add a test for this please?

R/measures.R
mllg: I've added references.

R/measures.R
mllg: What do you mean by hand-constructed? Calculating these measures without a package would require a few hundred LOC.

R/RLearner_surv_cforest.R
mllg: There is a test in this PR which detects if the predictions are reversed/inverted.

R/RLearner_surv_gamboost.R
mllg: Survival learners do not support multiple predict types currently. There was an attempt to support survival probabilities, but this is not implemented. The calling function checks for predict type and matches against properties, so this is dead code.

tests/testthat/helper_objects.R
```diff
@@ -92,6 +92,13 @@ getSurvData = function(n = 100, p = 10) {
   cens.time = rexp(n, rate = 1 / 10)
   status = ifelse(real.time <= cens.time, TRUE, FALSE)
   obs.time = ifelse(real.time <= cens.time, real.time, cens.time) + 1
+
+  # mark large outliers in survival as censored
```
PhilippPro: Why?

tests/testthat/test_base_measures.R
```diff
@@ -898,3 +916,25 @@ test_that("measures MCC denominator 0 (#1736)", {
   res = measureMCC(c(TRUE, TRUE, TRUE), c(TRUE, TRUE, TRUE), TRUE, FALSE)
   expect_equal(res, 0)
 })
+
+test_that("setMeasurePars", {
+  mm = mmce
+  expect_list(mm$extra.args, len = 0L, names = "named")
+  mm = setMeasurePars(mm, foo = 1, bar = 2)
```
PhilippPro: Example with mmce is a bit senseless? (Maybe ok, it is just for testing) Are there other measures (except of cindex.uno and iauc.uno) where setting parameters could be useful at the moment?

tests/testthat/test_surv_measures.R
```diff
@@ -0,0 +1,43 @@
+context("survival measures")
+
```
PhilippPro: Why an extra file instead of putting it into test_base_measures.R?

tests/testthat/test_base_measures.R
mllg: Don't know if there are others. Parameters of cindex.uno and iauc.uno are tested in the other file.

tests/testthat/test_surv_measures.R
mllg: This is just personal preference (in order to let the base group of tests run quickly), we could also join the two files.

R/measures.R
larskotthoff: For most of the other measure tests along the lines of incorrect predictions 5, correct predictions 10, therefore error rate 33%. Check that implemented measure gets that number. The point is to check that the number is correct for specific cases (and these can be constructed, i.e. you know what the answer should be).

tests/testthat/test_surv_measures.R
larskotthoff: I prefer splitting into more files as well. We're already running into issues with tests being too large.

R/measures.R
mllg: It's complicated. I've added a small test to check if perfect predictions lead to (nearly) perfect performance if there is no censoring. For all other cases, I'd need an external package because you cannot compute this by hand (in a reasonable time frame). I guess we have to rely on the package authors of `survAUC` for correctness.

@PhilippPro Do you have any ideas how to test those measures?

PhilippPro: No, not really. Of course one can construct simple cases without censoring that can be calculated by hand, but with censoring we have to use the complicated formulas from Uno's Paper here (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3079915/), which do not look very simple at first glance.

larskotthoff:
Is this ready to be merged?

mllg:
Tests have been failing randomly (connection issues, timeouts). But now everything looks good, so yes, merging. Thanks for reviewing.

PR #1848, opened 2017-06-13
Reenable Lintr

### LINKED ISSUE #1844: reenable lintr
```

# FIXME: disabled because it took more than 10min without output and broke travis
# library(testthat)
# test_check("mlr", filter = "lint")

```

mb706:
This should fix #1844

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,8 @@ before_script:
 script:
   - |
     R CMD build .
-    travis_wait 100 R CMD check mlr*.tar.gz
+    travis_wait 100 R CMD check mlr*.tar.gz --as-cran --run-donttest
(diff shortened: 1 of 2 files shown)


mb706:
Note this disables output *during* testing, so if the tests hang we wouldn't know until timeout. Since the tests run forever anyways I don't think that's a problem, but users might be surprised by that at first.

mb706:
I think I'm done here.
```
Running ‘run-base.R’ [12m/11m]
```

berndbischl:
> Note this disables output during testing, so if the tests hang we wouldn't know until timeout.

can you please explain with more words what this means?

berndbischl:
that just refers to what happens on travis, of course?

mb706:
It only affects travis output during the build: It does not live-update any more. After completion, it looks the same again.

PR #1851, opened 2017-06-13
Fix dropped weights

### LINKED ISSUE #1850: changeData ignores weights for multiclass task
because of a bug in changeData. This probably has other ramifications as well.

mb706:
Should fix #1850

### DIFF
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -439,10 +439,7 @@ changeData = function(task, data, costs, weights) {
     weights = task$weights
   task$env = new.env(parent = emptyenv())
   task$env$data = data
-  if (is.null(weights))
-    task["weights"] = list(NULL)
-  else
-    task$weights = weights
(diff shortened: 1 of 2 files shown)


berndbischl:
why not add clustering in the tests?

mb706:
I remembered there was one kind of task without weights. It turned out to be costsens, not cluster, but for some reason I removed both.

mb706:
Can be merged as soon as tests pass imho.

larskotthoff:
Cluster task is still missing from the tests?

berndbischl:
> Cluster task is still missing from the tests?

yup pls add

mb706:
Isn't `noclass.task` a cluster task? What change exactly do you guys want me to make?

berndbischl:
> Isn't noclass.task a cluster task?

yes. and sorry!

larskotthoff:
Argh, sorry, should have given that task a better name...

PR #1856, opened 2017-06-14
set measure id = measure name for measure arsq

Coorsaa:
Measure ````arsq```` had "adjrsq" as id which was not consistent to all other measure/id combinations.


larskotthoff:
Thanks, merging.

PR #1858, opened 2017-06-19
Fix 1857 fixup.data

### LINKED ISSUE #1857: makeClusterTask, makeMultilabel tasks ignore fixup.data

mb706:
This fixes #1857 
I also made the `makeSupervisedTask` and `makeUnsupervisedTask` paramters non-optional. This will prevent this kind of error in the future: `fixup.data` should always be given to these functions, since they are always called from functions that have an opinion on this.

Maybe it will break some tests. let's see.

### DIFF
--- a/R/ClusterTask.R
+++ b/R/ClusterTask.R
@@ -6,7 +6,7 @@ makeClusterTask = function(id = deparse(substitute(data)), data, weights = NULL,
   assertChoice(fixup.data, choices = c("no", "quiet", "warn"))
   assertFlag(check.data)
 
-  task = makeUnsupervisedTask("cluster", data, weights, blocking)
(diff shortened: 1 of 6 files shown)


larskotthoff:
You're only changing things for unsupervised, not multilabel here?

mb706:
I corrected both `makeClusterTask` and `makeMultilabelTask` and changed `makeSupervisedTask` and `makeUnsupervisedTask` to both have no optional arguments (since this will prevent these kinds of bugs in the future).

larskotthoff:
I mean that `makeMultilabelTask` is not in the diff.

mb706:
https://github.com/mlr-org/mlr/pull/1858/files#diff-150271cb6a0cb656eb2994e8536fc93f ?

larskotthoff:
Argh, I was looking for a literal `makeMultilabelTask`.

Thanks, merging.

Issue #1264, opened 2016-10-01
classif.xgboost - error if early.stop.round is set?


DanielKuehn87:
Not sure what I am doing wrong here:

```
library(mlr)
mlr_task <- pid.task
mlr_learner <- makeLearner("classif.xgboost",
                           par.vals = list(nrounds = 250,
                                           early.stop.round = 25,
                                           maximize = FALSE))

mlr_train <- train(mlr_learner, mlr_task)
```

This results in an error (I tested the mlr CRAN & github version):

> Error in xgb.train(params, dtrain, nrounds, watchlist, verbose = verbose,  : 
> Please set maximize to note whether the model is maximizing the evaluation or not.

MariaErdmann:
Hi DanielKuehn87,
this error message is not related to mlr. It is related to xgboost package itself. A similar issue was reported to the R xgboost team on github: [https://github.com/dmlc/xgboost/issues/916](url), but it seems this is not fixed on CRAN yet.

DanielKuehn87:
Ok, thanks for the link. After using the xgboost version from the drat repo (and updating early.stop.round to early_stopping_rounds in mlr) it works now.

Still a bit weird, since this works with CRAN-xgboost:
```
library(xgboost)
library(mlbench)
data(PimaIndiansDiabetes)

xg_data <- data.matrix(PimaIndiansDiabetes[,1:8])
xg_label <- PimaIndiansDiabetes$diabetes
xg_matrix <- xgb.DMatrix(data = xg_data, 
                         label = xg_label)

xg_watchlist <- list(msr = xg_matrix)

xg_model <- xgb.train(data                = xg_matrix,
                      watchlist           = xg_watchlist,
                      nrounds             = 250, 
                      early.stop.round    = 25,
                      maximize            = FALSE
)

```

larskotthoff:
Hang on, so `early.stop.round` has been renamed in the repo version? We need to update that when it's released on CRAN.

Also, are you saying that you're getting an error with the current development mlr version and the xgboost version on CRAN?

DanielKuehn87:
After reading the xgboost repo docu, it seems like they even changed a bit more: 
1. Missing default is NA now (instead of 0) 
2. Early.stop.round is renamed to early_stopping_rounds
3. Additional parameters save_period, save_name, xgb_model and callbacks were added

Yes, I got the error message above with the latest mlr version on github and xgboost from CRAN. After installing the xgboost repo version it works now.

[...]

PR #1859, opened 2017-06-20
xgboost: expose watchlist and callbacks

khotilov:
- expose `watchlist` and `callbacks` through parameters #1264
- remove `silent` from the parameters - it's overridden by `verbose` in xgboost R interface and has no effect
- set default `lambda=1` as it is such for tree boosters
    - Question: is there any way to conditionally set default lambda=0 when booster=='gblinear'?
- add `tweedie_variance_power` parameter
- placeholders for some recently added parameters (commented out until the next CRAN release)

### DIFF
--- a/R/RLearner_classif_xgboost.R
+++ b/R/RLearner_classif_xgboost.R
@@ -7,7 +7,7 @@ makeRLearner.classif.xgboost = function() {
       # we pass all of what goes in 'params' directly to ... of xgboost
       # makeUntypedLearnerParam(id = "params", default = list()),
       makeDiscreteLearnerParam(id = "booster", default = "gbtree", values = c("gbtree", "gblinear", "dart")),
-      makeIntegerLearnerParam(id = "silent", default = 0L, tunable = FALSE),
+      makeUntypedLearnerParam(id = "watchlist", default = NULL, tunable = FALSE),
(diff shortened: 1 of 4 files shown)


larskotthoff:
@mb706 Can we have the lintr allow TODO comments please?

mb706:
that would be direct violation of the style guide <https://github.com/rdatsci/PackagesInfo/wiki/R-Style-Guide> point 17 ;-)
Anhway, to disable this, comment out the "todo.comment" linter here: <https://github.com/mlr-org/mlr/blob/master/tests/testthat/helper_lint.R#L269>

larskotthoff:
That style point doesn't apply here -- it's a todo for when a new package version is released. The style guide doesn't say anything about not allowing TODOs in general.

khotilov:
TODO codetags have fairly different semantics (informal tasks pending completion) to FIXME's (something bad or suspicious that needs cleanup, usually before a release). To me, the style guide's p17 simply emphasizes this distinction, even if in a potentially confusing way.

larskotthoff:
@khotilov Could you please disable the TODO lintr as suggested so we can see if the build succeeds then?

khotilov:
@larskotthoff The build has passed.

larskotthoff:
Thanks! Merging.

Issue #1861, opened 2017-06-21
Feature selection + hyperparameter tuning + model validation (nested-nested CV)?


tjpalanca:
There are some learners (e.g. Random Forests) where I have to do: (a) partition dataset into training and test, (b) select features within the training set via Wrapper approach, and (c) tune hyperparameters for each candidate feature set (Wrapper) approach.

I know how to do model validation + feature selection or model validation + tuning (or with filter approach), but is there any way I can perform the task above? Thank you.

larskotthoff:
I'm not sure if I understand entirely what you're trying to do. Have you seen the part of the [tutorial](https://mlr-org.github.io/mlr-tutorial/devel/html/feature_selection/index.html) that explains how to fuse a learner with feature selection and then tune it?

Note that random forests perform feature selection internally and it is usually not necessary to do it externally.

tjpalanca:
Apologies that I wasn't clear.

The nested-nested resampling I mentioned would ideally be as follows. (I'm using SVMs here instead of RFs as you're right about RF internal feature selection).

```
library(mlr)
library(magrittr)
makeLearner(cl = "classif.ksvm", predict.type = "prob") %>% 
  # Inner tuning loop
  makeTuneWrapper(
    par.set = makeParamSet(
      makeDiscreteParam("C", values = 2^(-2:2)),
      makeDiscreteParam("sigma", values = 2^(-2:2))
    ), 
    measures = list(auc, acc),
    control = makeTuneControlGrid(),
    resampling = makeResampleDesc(method = "CV", iters = 10)
  ) %>% 
  # Middle feature selection loop
  makeFeatSelWrapper(
    measures = list(auc, acc),
    control = makeFeatSelControlExhaustive(),
    resampling = makeResampleDesc(method = "CV", iters = 10)
  ) %>% 
  # Outer model performance estimation loop
  benchmark(
    resamplings = makeResampleDesc(method = "RepCV", reps = 10, folds = 10),
    measures = list(auc, acc)
  ) ->
  svm.model
```

So there would be two wrappers as there will be resampling -> feature selection -> tuning within each iteration. The way I'm doing this right now is to use `modelr` to do the outer model validation loop then training a one wrapper learner on each of those iterations, but it would be good to be able to do this within mlr.

Please let me know if I am approaching this correctly. Thank you!

mb706:
[...]

Issue #1806, opened 2017-05-23
ImputationWrapper and FeatSelWrapper conflict when imputed column is missing from candidate feature set


tjpalanca:
Hi,

I'm trying to set up nested resampling by combining data-dependent imputation and feature selection. However, when specifying the `cols` argument in `makeImputeWrapper` instead of just a general class, and error is raised when that specific feature is not included in the candidate feature set in the feature selection process.

```{r}
# Libraries
library(mlr)
library(magrittr)

# Load iris dataset
data(iris)

# Add 5 artificial missing values to Sepal.Length
set.seed(7292)
iris[sample(1:nrow(iris), size = 5), c("Sepal.Length")] <- NA

# Create task
makeClassifTask(
  data = iris,
  target = "Species"
) -> iris.task

# Train rpart with imputation and feature selection
makeLearner(
  cl = "classif.rpart",
  predict.type = "response"
) %>% 
  # Imputation Wrapper
  makeImputeWrapper(
    cols = list(
      Sepal.Length = imputeMedian()
    )
  ) %>% 
  # Feature Selection Wrapper
  makeFeatSelWrapper(
    resampling = makeResampleDesc("CV", iters = 5),
    measures = list(mmce),
    control = makeFeatSelControlExhaustive()
  ) %>% 
  train(iris.task)
```

This seems perfectly reasonable to do, and in this case I think the desired behavior is to just ignore `Sepal.Length` imputation if it is not in the candidate set. However, it raises an error:

```
[FeatSel] Started selecting features for learner 'classif.rpart.imputed'
With control class: FeatSelControlExhaustive
Imputation value: 1
[FeatSel-x] 1: 0000 (0 bits)
[FeatSel-y] 1: mmce.test.mean= 0.7; time: 0.0 min
[FeatSel-x] 1: 1000 (1 bits)
[FeatSel-y] 1: mmce.test.mean=0.293; time: 0.0 min
[FeatSel-x] 1: 0100 (1 bits)
Error in impute.data.frame(obj = list(Sepal.Width = c(3.2, 3, 2.8, 3.7,  : 
  Column for imputation not present in data: Sepal.Length
```

Thanks in advance!

tjpalanca:
Closed by #1871

PR #1871, opened 2017-06-29
Modify `makeBaseWrapper` to unlock wrapped feature selection and tuning and to prevent errors in wrapped feature selection and imputation

tjpalanca:
This change relaxes the restrictions in `makeBaseWrapper` so that:

1. Only wrappers of class `TuneWrapper` (instead of any of class `OptWrapper`) must be placed at innermost loop and cannot wrap around anything else.

This unlocks two use cases related to issues #1861 and #1806:

1. #1861 Allows feature selection in an outer loop and tuning within each candidate feature set.
2. #1806 Allows imputation to take place before the feature selection loop so that it does not fail in the case when the candidate feature set in feature selection does not include specified column names via the `cols` argument.

Thank you!

### DIFF
--- a/R/BaseWrapper.R
+++ b/R/BaseWrapper.R
@@ -1,10 +1,7 @@
 makeBaseWrapper = function(id, type, next.learner, package = character(0L), par.set = makeParamSet(),
   par.vals = list(), learner.subclass, model.subclass) {
-
-  if (inherits(next.learner, "TuneWrapper"))
-    stop("Cannot wrap a tuning wrapper with something else!")
-  if (inherits(next.learner, "OptWrapper") && inherits(next.learner, learner.subclass))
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks. Could you add a test for the new checks please?

tjpalanca:
I've changed the incorrect logic in the first commit to reflect the required behavior.

I've also added two tests for: (1) the required use case mentioned in #1861 and (2) catching the error when a `TuneWrapper` is not in the innermost resampling loop.

larskotthoff:
Thanks, this looks good to me. I'd like to have somebody else to have a look though @berndbischl @mllg @jakob-r

larskotthoff:
Ping @berndbischl @mllg @jakob-r

PR #1872, opened 2017-06-29
mlrFDA new proposal by sub-classing data frame to "fda.data.frame"

smilesun:
@pfistfl @berndbischl @lbeggel .
1. The discussion with Bernd came out that we do not have much time left before the Reisenberg talk since we do not want to change the API after the talk, so we need to make a workable version.  
So I would propose we start working on it now.
2. @pfistfl  Bernd suggest I should assist you in implementing the new idea. So whether you could work on this pull request or create another one which is the list of matrix version. 
3. Maybe it is not a good idea for multiple to work on the same branch, so @pfistfl  please indicate how much you could contribute and I could assist you or we can create separate branches on top of the branch "fda_subclass_dataframe" and only merge the subbranch into this branch.


pfistfl:
I am a little confused. 
I thought I would implement the "functional feature" version? 

I think I can work on this during this weekend.
Should we not base the changes on top of the ```fda_pull1_task``` branch? 
 I will quickly create a sketch on what needs to be done.

pfistfl:
See makeFDAData.R and discuss

pfistfl:
@smilesun 
Should I continue on this? Can we maybe quickly talk before I start implementing in order to prevent unneccessary work?

pfistfl:
I would even suggest to keep the API for creating Tasks using fd.features and fd.grids... 
just pass the params on to ```makeTask``` and start from there.

```
#' Create a functional feature from a numeric matrix
#'
#' @param m [\code{matrix}] \cr Numeric matrix.
#' @param fname [\code{character}] \cr Name of the functional feature
#' @param fd.grids [\code{numeric}] \cr fd.grids
#' @return [\code{matrix}] of class c("functional", "matrix") containing the functional values
#' @export
makeFunctionalFeature = function(m, fname = deparse(substitute(m)), fd.grids = NULL) {
  if (is.data.frame(m))
    m = as.matrix(m)
  assertMatrix(m, mode = "numeric")
  if(is.null(fd.grids))
    fd.grids = seq_len(ncol(m))
  assertCharacter(fname, len = 1L)
  assertNumeric(fd.grids, len = ncol(m))
  attr(m, "fd.grids") = fd.grids
  attr(m, "fname") = fname
  addClasses(m, "functional")
}


library(BBmisc)

fd.features = list("fd1" = c(1:3),  "fd2" = c("X4"))
fd.grids = NULL

data = data.frame(matrix(rnorm(5*4), ncol = 5))

fd.features = lapply(fd.features, function(fdc) {
  # Can be either column name or index
  assertChoice(class(fdc), choices = c("integer", "character"))
  if (is.character(fdc))
    fdc = which(colnames(data) %in% fdc)
  return(fdc)
})

fdfns = names(fd.features)

if (is.null(fd.grids)) {
  fd.grids = setNames(lapply(X = names(fd.features), FUN = function(name) {
    fd.grids = lapply(fdfns, function(name) seq_along(fd.features[[name]]))
    seq_len(length(fd.features[[name]]))
  }), names(fd.features))
}

fdf = lapply(fdfns, function(fn) {
  makeFunctionalFeature(data[, fd.features[[fn]], drop = FALSE], fn, fd.grids[[fn]])
})


data = data[, - do.call("c", fd.features), drop = FALSE]
data[, fdfns] = fdf
```

smilesun:
`data[, 2:3]` actually returns more than 2 columns which is incompatible with the mlr core

smilesun:
```
lapply(data, class)
$X5
[1] "numeric"

$fd1
[1] "functional" "matrix"    

$fd2
[1] "functional" "matrix"    

$x6
[1] "character"
```

smilesun:
tem = data[1:2, ]
lapply(tem, class)

$X5
[1] "numeric"

$fd1
[1] "matrix"

$fd2
[1] "matrix"

$x6
[1] "character"

smilesun:
After discussing with Florian, we decide to abandon this approach.
Waiting for comments,  @berndbischl @lbeggel

pfistfl:
Basically the result of talking to Xudong is the following:
The "functional feature" approach by adding a matrix is hard for the following reasons:
- subsetting columns via [, 2:3] can only index **whole matricies** but not part of the functional feature.
- subsetting the rows of a functional feature via [1:5, ] drops the attributes and class of the functional. 
For this reasons, I think my approach is obsolete

Adding a class``` fda.data.frame``` and subsetting this does not seem to yield the same problems.

smilesun:
we will use this branch "fda_df_attr" instead

smilesun:
This is one of the proposal worked out together with Florian, but we do not need this anymore.

Issue #1669, opened 2017-03-13
[FDA] Stratification for tasks of type 'fdaclassif' not supported


maierhofert:
```
res = makeResampleDesc(method = "RepCV", predict = "test",
                       stratify = FALSE,
                       folds = 2L, reps = 2L)

# resampling instances
res_instances = lapply(tsks, makeResampleInstance, desc = res)
```
I see no theoretical problem with allowing stratification for FDA tasks. I tried out the branch fda_pull1_task.

pfistfl:
Issue is obsolete due to the way we finally decided to implement. Added a unittest nonetheless

Issue #1846, opened 2017-06-13
use chainofResponsibility to resolve the too many Task type problem


smilesun:
in change data, we have this long switch statement, and there is a forseeable possibility that it will increase length:
```
  task$task.desc = switch(td$type,                                                                                                                                                         
448     "classif" = makeClassifTaskDesc(td$id, data, td$target, task$weights, task$blocking, td$positive),                                                                                     
449     "regr" = makeRegrTaskDesc(td$id, data, td$target, task$weights, task$blocking),                                                                                                        
450     "cluster" = makeClusterTaskDesc(td$id, data, task$weights, task$blocking),                                                                                                             
451     "surv" = makeSurvTaskDesc(td$id, data, td$target, task$weights, task$blocking, td$censoring),                                                                                          
452     "costsens" = makeCostSensTaskDesc(td$id, data, td$target, task$blocking, costs),                                                                                                       
453     "multilabel" = makeMultilabelTaskDesc(td$id, data, td$target, td$weights, task$blocking),                                                                                              
454     "fdaclassif" = makeFDAClassifTaskDesc(td$id, data, td$target, td$positive, td$fd.features, td$fd.grids, td$weights, task$blocking),                                                    
455     "fdaregr" = makeFDARegrTaskDesc(td$id, data, td$target, td$fd.features, td$fd.grids, td$weights, task$blocking)                                                                        
456 )             
```      
**# Use the following code to optimize it.**
replace above with the following line:
`task$task.desc = makeTaskDescritption(getchain(),td)`

where

```
makeTaskDescription = function(node, td, task, data) {                                          
   UseMethod("makeTaskDescription")                                                                                                                                                              
}   

base.node = makeS3Obj("ChainNode",                                                                                                                                                              
exec = NULL,                                                                                                                                                                
next = NULL)    
``` 
and 

```
classifNode =  addclass(base.node, "classif")
classifNode$exec = function(td, data, task) {makeClassifTaskDesc(td$id, data, td$target, task$weights, task$blocking, td$positive)}

makeTaskDescription.classifNode = function(node, td, task, data)
{
if("classif" %in% class(node))  return(makeClassifTaskDesc(td$id, data, td$target, task$weights, task$blocking, td$positive)) 
# or if("classif" %in% class(node)) return(do.call(node$exec, td, task, data))
node$next$exec()
}

```
in function zzz.R
```
getchain = function() {
nodes= c(classif.node, regr.node,......)
[...]

Issue #1793, opened 2017-05-08
A general way to make mlr support all operation on FDA task


smilesun:
If I supply a FDATask to a feature selection method in mlr, I got the following error
```
Error in generateFilterValuesData(instance$task, method = filter.method) : 
  Filter(s) 'information.gain' not compatible with task of type 'fdaclassif'
Timing stopped at: 0.004 0 0.005
```
Of course, we could add some patch to this again and again, but is there is way to update all properties to a new Task ? Using Observer for example.

smilesun:
There are still problem on this if one try 
```
generateFilterValuesData(phoneme.task)       
```
And get 
                                                       
```
Error in generateFilterValuesData(phoneme.task) : 
  Filter(s) 'randomForestSRC.rfsrc' not compatible with features of type 'functionals' respectively
```

PR #1877, opened 2017-07-03
Add attributes to dataframe and get another class for dataframe 

smilesun:
@pfistfl @lbeggel


pfistfl:
So @smilesun 

We need the ```fda.data.frame``` for two reasons:
1) We want to use S3 method dispatch.
2) We want to somehow store additional info as attributes. 

Prerequisite: ```data[i, j, drop = FALSE]```  needs to work.

```
library(checkmate)
# Example Data
data = data.frame(matrix(rnorm(10*50), ncol = 10),
  "a" = as.factor(sample(letters[1:3], 50, replace = TRUE)), 
  "int" = sample(c(1L, 3L), 50, replace = TRUE),
  "b" = as.factor(rep(c(0,1), times = 25)))

fd.features = list("fd1" = 1:7, fd2 = 8:10)
fd.grids = list("fd1" = 1:7, fd2 = 1:3)

# Basic makeFDAData Function
makeFDAData = function(data, fd.features, fd.grids) {
  data = addClasses(data, "fda.data.frame")
  attr(data, "fd.grids") = fd.grids
  attr(data, "fd.features") = fd.features
  return(data)
}

df = makeFDAData(data, fd.features, fd.grids)

checkFDAData = function(df) {
  assertClass(df, classes = c("fda.data.frame", "data.frame"))
  assertList(attr(df, "fd.features"), len = length(attr(df, "fd.grids")))
  assertList(attr(df, "fd.grids"))
}


checkFDAData(df[1:5, , drop = FALSE]) # works
checkFDAData(df[, 1:13, drop = FALSE]) #breaks even if we actually do not subset anything because ncol(df) = 13

```

mb706:
When you overload ``` `[.data.frame` ```, don't forget to also handle the case `df[SUBSET]` which should be equivalent to `df[, SUBSET, drop = FALSE]`.

smilesun:
@pfistfl , so now you are with me for using the dispatch instead of the ifelse ?
@mb706  You are right, Florian and I tried out this approach on Monday but we only tested subsetting row
df = iris
attr(df,"fd.grids") = list(fd1 = 1:2, fd2 = 3:4)
df = addClasses(df, "fda.data.frame")
> attributes(df[1:3, ])
$names
[1] "Sepal.Length" "Sepal.Width"  "Petal.Length" "Petal.Width"  "Species"     

$fd.grids
$fd.grids$fd1
[1] 1 2 

$fd.grids$fd2
[1] 3 4


$row.names
[1] 1 2 3

$class
[1] "fda.data.frame" "data.frame"    
> attributes(df[3])
NULL
> attributes(df[, 3])
NULL

smilesun:
@berndbischl  
1. from my side,  the view is as before. It might  be against the idea of OO to merge the FDATask into ClassifTask and RegrTask, the idea of OO is basically to use child class to replace ifelse ( and what we want to do now is the opposite) and currently the number of child class of Task is not many in my point of view, it looks to be a lot since the code of organizing them need to be optimized. For example, both Martin and I have suggested a way to change the code in task_operators.R, which will not affect other parts of R, and Martin's solutions look more concise and easy to implement.
2. dataframe is an R object other people build, we might only want to wrap around it using the proxy pattern and adapter pattern instead of changing its behavior which will result unexpected senario that we currently could not imagine as Martin pointed out for instance. Suppose someone wrote a Java library, one might only want to use this as a service, instead of getting the source code for this Java library and re-engineer it. 
3. if we eliminate the FDATask,  we basically lose the multiple inheritance possibilities. Adding another layer of abstraction(Data) under TASK to me currently sounds not necessary, since we do not have yet a sparkData and imageData and videoData yet, and simply getting a new layer of abstraction seems to be too much for just FDA, and once we have the sparkData and ImageData and videoData , the code we wrote now will basically be gone once we redesign everything. 
4. Image what we are doing, we have the fd.grids and fd.features defined in our new "fda.data.frame" Class, and what we do in "makeClassifTask" is basically copying all those fields again to the TaskDesc, which seems to be trivial work since we did nothing except for copying ?
5. DO you still think it is a good idea to subclass the dataframe?

smilesun:
@berndbischl , if you are worrying that we have two many subclasses of Task which will grow exponentially, we could basically divide them orthogonally which turns the number of subclasses from exponential to polynomial. For example, use the "Decorator" Design pattern, which could remedy the disadvantage that s3 from the language level does not support multiple inheritances. 

For example,  in Java a typical Decorator pattern would be
```
BufferedReader in = new BufferedReader(new FileReader("foo.in"));
```
Which set the BuffereReader inheritance family and FileReader inheritance family orthogonal so they do not need to define a new class called BufferedReaderAndFileReader(which will explode exponentially)

in our case, what we currently have is FDAClassifTask inherited from both FDATask and Task. If we do not want to have more classes, we simply add some decorator on top of the ClassifTask.

@pfistfl @mb706  How would you think?

pfistfl:
@smilesun 
I have not understood everything you wrote above but the key message is clear I think. 

SO: Basically the Task with an fda.data.frame would be a **normal** ...Task (e.g. classifTask) with the following changes:
- ```fda.data.frame``` instead of ```data.frame``` as underlying data. 
a) Does this involve a copy / manipulation of the ```data.frame``` when creating the Task?
b) In this case, would creating a Task maybe become very slow for huge data.frames?
c) In the normal case, I think we basically do not touch the ```data.frame``` (?!)


- ```fd.features``` and ```fd.grids``` as extra list elements in the ```TascDesc``` 
d) I think we need this in any viable case.

- e) Why have we not decided to just have functionals in a normal Task and just store ```fd.features``` and ```fd.grids``` in case we have functionals, and have them empty in case there are none? 
This introduces two more parameters to make...Task()

Sorry for creating more questions than solving.

smilesun:
@pfistfl 
a,b,c   yes,  I would suggest we keep a data.frame as it is, do not touch it
d.   this is what we currently have
e,  I am not convinced by this idea as we have to change the code of makeCLassifTask and a thousand mlr developer need to rely on this and it will strongly couple the development team. Suppose if someone want to change something in makeClassifTask, he has to take care of what we wrote there and make sure his editing does not destroy our FDA module, but this is extra work for him.  I prefer we keep what we have now  since we still have some bugs in our intial implementation and need some time to fix this as well.

pfistfl:
> e) 

It depends on how much we change I think. If it is ONLY the two extra list items in ```task.desc``` I think this should not create problems.

Nonetheless, I think you are right. I think we should focus on the as-is state, and thus consider the API to be 
```fd.features``` and ```fd.grids``` in a ```(FDA)...Task```.

I think even if something changes down the line it should not be a big problem for users, as long as the fd.features and fd.grids stays the same.

smilesun:
Could someone maybe first fix these problems in the fda_df_attr branch then we could merge it into the fda_pull1_task branch?  To make sure they work well for the following (I currently am blocked by other teachings the annoying  lrz problem)

- setThreshold

- tuneThreshold

- calculateROCMeasures

- generateThreshVsPerfData

- generateCalibration

pfistfl:
I can do that.

1) I  will branch off fda_pull1_task, isn't this more intelligent?
2) Should I merge master as well?
3) The requirement is that those things work with fda learners? I did not know we had problems there.

smilesun:
@pfistfl  Ok, thanks, whatever you like, yes, they do not completely work with FDA, please add some tests in one separate file to test those features work with FDA task.  Then if some tests fails, fix it in the branch. 
Tell me the new branch you forked.

pfistfl:
The branch is called: ```fda_fix_tunestuff```

ToDo's: 
- [ ] Implement the  proposed by @mb706  in #1846 ? 
- [ ] Fix the stuff mentioned in #1793
- [ ] Fix the stuff mentioned by @smilesun above?
- [ ] Fix #1669
- [ ]  Add a warning for scalar features and drop those?

pfistfl:
@smilesun 
Questions: 
- In ```trainLearner.fdaclassif.knn```we currently hand ALL data to the algorithm. We should only provide the functional(s)!
Do we need a getFunctionalData(task, ...) that does something like  ``` intersect(unlist(fd.features), colnames(data)) ```?

smilesun:
@pfistfl  
1. Do not do a thousand things at one time, we could review it together step by step
2. Please write the test first before you write any code
3.  Do as you want and we will confirm that in the review

pfistfl:
I merely try to create a full list of stuff that needs to be done, I will not fix everything at once. 

I think I fixed the stuff you and @mb706 mentioned in #1846 
Care to review? See #1880

smilesun:
since we have agreed on a new structure for FDA which is already in the master branch, I will close this pr

Issue #1610, opened 2017-03-07
Scalar features in Functional data analysis


smilesun:
todo: we need to add scalar features to functional data analysis task, at the moment , the scalar features are ignored

berndbischl:
learners need a property whether they support scalar features.

*if* they support scalar features, learners should of course get and handle them

if they dont, you have 2 options:

a) throw an error, if scalar features are in the task, before training

b) drop them, but warn about it

berndbischl:
maybe b) is more convenient now.

but *document* that well

pat-s:
Eventually fixed in #1880? Either way, not adding more detailed support for FDA to {mlr}.

PR #1880, opened 2017-07-06
Fix some bugs / missing feature in FDA Branch

### LINKED ISSUE #1669: [FDA] Stratification for tasks of type 'fdaclassif' not supported

pfistfl:
This PR on top of fda_pull1_task reflects what needs to be fixed. 

ToDo's: 
- [ ] Implement the  proposed by @mb706  in #1846 ? 
- [ ] Fix the stuff mentioned in #1793
- [ ] Fix the stuff mentioned by @smilesun above?
- [ ] Fix #1669
- [ ]  Add a warning for scalar features and drop those (#1610) ?
- [ ] Do we need a getFunctionalData(task, ...) that does something like  ``` intersect(unlist(fd.features), colnames(data)) ```? All FDA-learners currently 1) Do not warn in case there are scalar fetures and 2) do not throw out scalar features.


smilesun:
```
Error in generateFilterValuesData(task = task, method = method, nselect = nselect,  : 
  Filter(s) 'randomForestSRC.rfsrc' not compatible with task of type 'classif'
```

pfistfl:
Yes, I know, I only fixed it for the information.gain case to showcase my idea. I do not know if it is okay to change the idea for all Filters.

Basically: 
- Are you okay with how I create the make...TaskDesc in changeData()? 
- What do you think about the solution for the "fdaclassif" in the Filter Case. Basically my idea is to instead of  checking for ```td$task.type = "classif"``` to check whether the Task inherits from ```ClassifTask```.

smilesun:
1. I am with you on this point to use inherits rather than use string match. Only if Bernd will support us on this idea. It seems that we have too much to do to really merge FDA into mlr and currently only u and me are available, but we definitely need people to get involved in making the change from string match to inherits
2.  I am working now to fix the above mentioned bug

smilesun:
@pfistfl   
I have a feeling that we might have too much to do now only for us two, how about the following approach?

1. We First create a test file covering the following, then I merge this into branch fda_pull1_task
- setThreshold

- tuneThreshold

- calculateROCMeasures

- generateThreshVsPerfData

- generateCalibration

2. Currently the above test will fail I guess ,  we fix those test without changing  the string match to inherits

3. we change the string match to inherits in the last step

Please tell your idea on this, open for discussion

smilesun:
Currently for the filter , we could not actually change all the filters since there are too much, maybe for filter itself, we need to change from string match to inherits
```
.FilterRegister[["rf.importance"]]
Filter: 'randomForestSRC.rfsrc'
Packages: 'randomForestSRC'
Supported tasks: classif,regr,surv
Supported features: numerics,factors,ordered

```

pfistfl:
We either have to add ```fdaregr``` and or ```fdaclassif``` to each Filter or we replace them... 
We have to touch all of them anyways. Touching everything twice .... I don't know.

Other Idea is to fix everything only for One Filter / One Case....

smilesun:
I just thought about this as well, in either case, we have to touch those things, so maybe we change from string match to inherits everywhere which makes things easier within one step.

I have time this afternoon for this, let us solve this filter first. If you find out where is the problem for the current example failure, ping me.

smilesun:
@pfistfl ,  I THINK the problem is here 
```
assert(checkClass(task, "ClassifTask"), checkClass(task, "RegrTas
    k"), checkClass(task, "SurvTask")) 
```

smilesun:
I will build a hash map from the task type to task itself

pfistfl:
@smilesun 
I can probably help you tomorrow.
I have an Idea, lets see what you think.

Instead of seeing which task we inherit from, we do something like 
```
getBasicTaskType(task.type) {
  task.type = switch(task.type, 
  "fdaregr" = "regr"
  "fdaclassif" = "classif", 
    .... )
  return(task.type)
}
```

Using this we do not have to change a lot in terms of what is checked. For every new task.type we just have to add it in the function.

smilesun:
@pfistfl , I will probably not work tomorrow.  I will merge this branch into pull1_task, the first thing I did is that we should leave the original code of changeData in the comments form so we could refer to that later

smilesun:
Currently this branch has the problem in failing the above test, have not find out what is the problem yet, maybe someone could look into this
```
if (identical(Sys.getenv("TRAVIS"), "true") || identical(Sys.getenv("R_EXPENSIVE_TEST_OK"), "true")) {
  test_check("mlr", "_learners_all_regr")
}

```

Issue #811, opened 2016-03-29
Error when predicting due to missing levels?


andreashandel:
Hi,
I'm trying to use 10-fold cross validation on a simple linear model ("regr.lm") with the resample() function. I'm getting the following error message:
[Resample] cross-validation iter: 1
[Resample] cross-validation iter: 2
Error in model.frame.default(Terms, newdata, na.action = na.action, xlev = object$xlevels) : 
  factor XXX has new levels NN

(where XXX and NN can change depending on setting different seeds)

I assume it's because during the CV split, some factor levels are not in the training data and are dropped during the train part and then can't be used when predicting the holdout set, i.e. [like described in this stackoverflow topic.](http://stackoverflow.com/questions/4285214/predict-lm-with-an-unknown-factor-level-in-test-data)

Question: Is there some way to have resample() automatically deal with the error during prediction, e.g. similar to the answer to the stackoverflow question referenced above? Or other suggestions on how to best handle this within the mlr framework? (I want to eventually do feature selection and model tuning using CV with a few different methods).
Thanks!

larskotthoff:
When creating the task, you can specify that empty levels shouldn't be dropped through `fixup.data`, e.g.

    task = makeClassifTask(id = "foo", data, fixup.data = "no")

andreashandel:
Thanks for the quick reply! Tried it, does unfortunately not work. 
I think the problem is at the level of lm() itself, it discards the empty factors (see discussion in link above) and mlr would need to 'fix' what lm() produces as in the stackoverflow example before using predict().

larskotthoff:
Ah, I see. That would mean removing data from the predictions though? I'm not sure if we want to do that in a predictor, as it would probably cause problems later on when something relies on predictions being there for all rows. I guess we could have a nicer error message though?

@berndbischl What's your take on this?

andreashandel:
Wouldn't need to remove observations/rows from the prediction set, just setting factors in the prediction set that are not in training to NA. Would make for a worse prediction (as it should) but otherwise I think it could be ok. Should probably not be the default, but maybe on option one could turn on? 

The problem right now is that if I did a big tuning/feature selection/whatnot optimization and in just one of my many train/predict runs the training part of the sample by chance includes a factor level that is not in the prediction set, the whole thing likely fails (haven't tested, but I assume). Would be nice if mlr could somehow 'deal with' this one failure but still produce results for everything else.

larskotthoff:
[...]

PR #1882, opened 2017-07-08
Account for missing factor levels in train introduced by partitioning for lm* models

pat-s:
Addresses #811 

# Concept & changes

* Factor levels present in test but missing in train will be set to `NA` (in test) so that `predict()` will work
* new helper function: `remove.missing.levels.lm()`
* This is achieved by modifying `test` in `calculateResampleIterationResult()`
* The resulting predicted values will contain `NA`s. These are removed then. Also, other objects returned from `predict()` and `makePrediction()` (i.e. `tune` and `newdata`) will be subsetted accordingly to the correct length of `p` to avoid 'length' errors
* new spatial task data added (`bc.task.spatial`).  Factor level 'lithology' contains 18 levels and is the major player here 
* Effectively, observations with missing factor levels will not be used / set to `NA` -> they will not be included during performance calculation 
* This fix not only accounts for the train/test factor level problem but also helps in cases where `stratify.cols` was used until now (to stratify factor levels across folds). `stratify.cols` only works if the number of factor levels is > number of folds, which is not always the case
* The fix is only applied when setting `fix.factors.prediction = TRUE` in the learner
* an informative info message is given on how many observations are affected

# Problems/Questions/To Do

- [x] Informative message stating the name of the factor level for which levels were dropped is not printed (called within `predict()` -> fixed
- [x] currently only implemented in `resample` -> also working for train + predict now. https://github.com/mlr-org/mlr/pull/1882/commits/3769df901f3bc7cf68aef52c6312769b4b849a15
- [x] more unit tests? Currently only for `classif.binomial` -> since only `lm` models are affected this should be sufficient
- [x] Style Guide compatibility? https://github.com/mlr-org/mlr/pull/1882/commits/e5d15f1bd7598f7e32e4bd8cecd46f21588cbdff https://github.com/mlr-org/mlr/pull/1882/commits/e906d0b3a58826889a3be53581ca72511648f486 https://github.com/mlr-org/mlr/pull/1882/commits/351c1fff9b669321096c57ea566925bee8c15cff
- [x] rename `remove_missing_levels()`? -> `remove.missing.levels.lm()`
- [x] fix R CMD Check

# Proof of concept

`devtools::install_github("pat-s/mlr@missing-levels")`

#### Pull request

```r
data("bc.task.spatial")
task = bc.task.spatial

rdesc = makeResampleDesc("RepCV", folds = 10, reps = 2)

lrn = makeLearner("classif.binomial",
                  link = "logit",
                  predict.type = "prob",
                  fix.factors.prediction = TRUE)

resa = resample(lrn, task, rdesc, measures = list(auc))

[Resample] repeated cross-validation iter 1: auc.test.mean=0.897
[Resample] repeated cross-validation iter 2: auc.test.mean=0.989
[Resample] repeated cross-validation iter 3: auc.test.mean=0.999
[Resample] repeated cross-validation iter 4: auc.test.mean=0.979
[Resample] repeated cross-validation iter 5: auc.test.mean=0.973
[Resample] repeated cross-validation iter 6: auc.test.mean=0.971
[Resample] repeated cross-validation iter 7: auc.test.mean=0.968
[Resample] repeated cross-validation iter 8: 
Setting missing levels in 'lithology', only present in test data but missing in train data, to 'NA'.
This affected 1 of 94 total observations in test data.
auc.test.mean=0.988
[Resample] repeated cross-validation iter 9: auc.test.mean=0.987
[Resample] repeated cross-validation iter 10: auc.test.mean=0.947
[Resample] repeated cross-validation iter 11: auc.test.mean=0.959
[Resample] repeated cross-validation iter 12: auc.test.mean=0.971
[Resample] repeated cross-validation iter 13: auc.test.mean=0.993
[Resample] repeated cross-validation iter 14: auc.test.mean=0.977
[Resample] repeated cross-validation iter 15: auc.test.mean=0.961
[Resample] repeated cross-validation iter 16: 
Setting missing levels in 'lithology', only present in test data but missing in train data, to 'NA'.
This affected 1 of 94 total observations in test data.
auc.test.mean=0.97
[Resample] repeated cross-validation iter 17: auc.test.mean=0.991
[Resample] repeated cross-validation iter 18: auc.test.mean=0.973
[Resample] repeated cross-validation iter 19: auc.test.mean=0.99
[Resample] repeated cross-validation iter 20: auc.test.mean=0.993
[Resample] Aggr. Result: auc.test.mean=0.998
```

#### mlr master

```r
data("basque") # not available in master, only in fork
basque$geometry <- NULL

                          target = "diplo01", positive = "1")

rdesc <- makeResampleDesc("RepCV", folds = 10, reps = 2)

lrn <- makeLearner("classif.binomial",
                    link = "logit",
                    predict.type = "prob")

resa <- resample(lrn, task, rdesc, measures = list(auc))

[Resample] repeated cross-validation iter 1: auc.test.mean=0.999
[Resample] repeated cross-validation iter 2: auc.test.mean=   1
[Resample] repeated cross-validation iter 3: auc.test.mean=0.999
[Resample] repeated cross-validation iter 4: auc.test.mean=0.975
[Resample] repeated cross-validation iter 5: auc.test.mean=0.964
[Resample] repeated cross-validation iter 6: auc.test.mean=0.987
[Resample] repeated cross-validation iter 7: auc.test.mean=0.986
[Resample] repeated cross-validation iter 8: auc.test.mean=0.983
[Resample] repeated cross-validation iter 9: auc.test.mean=0.936
[Resample] repeated cross-validation iter 10: Error in model.frame.default(Terms, newdata, na.action = na.action, xlev = object$xlevels) : 
  factor lithology has new levels Rocas volcnicas piroclsticas
```

#### train + predict

```r
data("bc.task.spatial")
task = bc.task.spatial

n = getTaskSize(task)
train.set = seq(1, n, by = 2)
test.set = seq(2, n, by = 2)
mod <- train(lrn, task, subset = train.set)

task.pred = predict(mod, task = task, subset = test.set)
Setting missing levels in 'soil', only present in test data but missing in train data, to 'NA'.
This affected 6 of 472 total observations in test data.

print(task.pred)
Prediction: 466 observations
predict.type: prob
threshold: 0=0.50,1=0.50
time: 0.01
```


larskotthoff:
Thanks, this looks like a good start. To answer your questions:

1. Use the function `warningf()`.
2. If somebody uses `train` and `predict` separately, I don't think that this can be done. In `resample` there should probably be a check and warning when the data is split already.

In general, please have a look at the [coding guidelines](https://github.com/mlr-org/mlr/wiki/mlr-Coding-Guidelines), especially the style conventions.

mb706:
Preamble: I don't have the authority to accept / reject things here, so this is just my advice and not obligatory.

I don't think the implementation is going in the right direction. IMHO:
1. The resample code should not contain special cases for two kinds of models in particular.
2. `predict` should not filter out `NA`.
    1. The user might use `predict` on unknown data and expect the prediction to have the same number of rows as `newdata`
    2. Different measures would probably want to handle `NA` differently. MMCE, for example, could count an `NA` as a misclassification by default. One could think about making the measures ignore `NA`, but that would mean that a model can get a higher score for predicting missings on data for which it is uncertain.
3. [As I said before](https://github.com/mlr-org/mlr/issues/811#issuecomment-313704053), it is much more general to use the `TaskDesc` to remove missing factor levels on prediction data.

My suggestions on how to solve this problem:
1. at the start of `train`, do something like
    ```R
    if (learner$fix.factors.prediction) {
      td = getTaskData(task)
      tt = getTaskTargetNames(task)
      newt = cbind(droplevels(dropNamed(td, tt)), td[tt])
      task = changeData(task, newt)
    }
    ```
    Didn't actually test this, but I think it is clear what I mean: The new `task` will only contain the factor levels that were present during training, and `mlr:::predictLearner2` will then drop the unseen factors.
2. see how different learners handle `NA` during prediction. In theory, every model should treat each line of input data as if it were standing on its own, but there might be learners that crash when one line of prediction data contains `NA`s. If many learners fail at that, maybe somehow route only the data without `NA` to the model on prediction, and predict `NA` for all `NA` containing rows by default otherwise. Of course, this should depend on the learner's `"missings"` property. If you only care about `lm` and `glmmPQL`, this step might be optional.
3. Consider how measures should handle `NA`. There could be (a) a global `configureMlr` option (I don't like global `mlr` options, but the easy way out at least is easy), (b) an option for `resample` and `tune` akin to `na.action`, or (c) (my favourite, but I can see why people might disagree) have specific measures that handle `NA` differently. As mentioned above, just dropping `NA` lines *will* distort the resampling score.

P.S., Maybe think twice about introducing these dependencies. `magrittr` seems to be syntactic sugar only, `stringr` sounds like it doesn't add anything above `stringi`.

pat-s:
> Use the function warningf().

Thanks!

> If somebody uses train and predict separately, I don't think that this can be done. In resample there should probably be a check and warning when the data is split already.

My implementation works in between the train and predict step because it needs the fitted model from `train` + `newdata` before going into `predict()`. I'll take a look :)

@mb706 Thanks for your extensive reply! Much appreciated! 

> The resample code should not contain special cases for two kinds of models in particular.

As far as I can see, this problem only applies to models of the lm* family. `glmmPQL` also has the problem although it uses repeated calls to lme4 (which might inherit `lm`?).  That's why I choose to implement it like this. Others, ML in particular, do not face the problem and just deal with it (although I do not know how they do it). **But there are of course a lot of models left which need testing on whether they need this fix or not!** 

> predict should not filter out NA.
> The user might use predict on unknown data and expect the prediction to have the same number of rows as newdata

> Different measures would probably want to handle NA differently. MMCE, for example, could count an NA as a misclassification by default, or 1/[number of target levels] misclassification. One could think about making the measures ignore NA, but that would mean that a model can get a higher score for predicting missings on data for which it is uncertain.

Of course, that's the main issue here. 

How does the NA / NA removal influence the error measures? You are right that higher performances could be achieved simply because we have less cases to check on. I do not have a solution for this / what's the best way for error measures to handle this situation. On the other hand, at least we get an error measure even if its biased rather than getting an error for the whole fold. Maybe that's the price to pay here. That's when the number of repetitions in CV comes into play to 'average out' such cases as good as possible (if that's even possible). 

It also highly depends on the factor level. If you have a factor level with 18 levels and 5-6 of them with n(levels) < 10 and (I think) 3 with levels with only 1 occurence (feel free to check the data), it might even be worth to drop the variable or clean it up beforehand. However, if they user decides to go with it, there need to be an option in `mlr` to handle such cases even if comes with a price of biased error measures. 

> As I said before, it is much more general to use the TaskDesc to remove missing factor levels on prediction data.

If I understand correctly, you do not have information about the learner in `taskDesc` (sorry, I am only into `mlr` since last week). Since it only applies for the lm* family, I choose to go with the option to do the altering of `test` conditionally on the learner family. Please correct me here! 

Regarding your approach:

I see your points. Dropping factor levels in test rather than in train is just another way to go here. In the end, both approaches do only have "valid" predictions on cases for which everything "works" i.e. we did not alter anything. 

I think the topic is more general and we should first agree on how to handle such cases in general, looking at the possible inferences for the error measures. 
Let's ping @berndbischl @giuseppec @jakob-r (feel free to ping more people who have the rep to decide here) and hear their opinion. This was meant as a proof of concept to tackle this problem and was already done before your last reply in #811. Maybe this PR is not the place to discuss this and we should use #811 or a complete new issue here for discussion so that this PR only discusses possible implementation changes of this approach. 

> P.S., Maybe think twice about introducing these dependencies. magrittr seems to be syntactic sugar only, stringr sounds like it doesn't add anything above stringi.

Sure, that's why I asked. I just used the implementation from my package `sperrorest`. `magrittr` can be easily dropped and for sure also `stringr` in favor of `stringi`.

larskotthoff:
Regarding the NAs for unknown levels: I think that this is a reasonable approach to encode a "don't know" for the model. Of course this needs to be documented (and probably shouldn't be the default behaviour), and it may even be a good idea to, if there are NAs, show explicitly how many there are and what a measure looks like with the NAs included and excluded.

One potential issue I see with just dropping the data with missing factor levels in test is that it becomes a bit harder to communicate to the user what's happening and how the results are affected. For example, the accuracy may be 100% when the data are dropped, but the performance of the model is obviously much worse than that because it doesn't know about some of the data. This fact may be hidden somewhere in a warning much further up, and even if it tells you how much data was dropped it's not necessarily straightforward to see how that affects the measure.

The advantage of having NAs is that the measure is affected directly (and yes, in different and sometimes non-intuitive ways), but at least the numbers for models with some levels missing and models that have all levels are directly comparable.

pat-s:
So the question is how do we continue here? This issue has potential for an infinite discussion. 

I agree that this is not an optimal solution but I also doubt that there is one in general. I do not know how other algorithms beides `lm` handle this case - maybe they also drop data in such cases but do not report it? 

I think this approach is a possible solution and should be made available as an option for users who accept the "consequences". These are hard to quantify and always depend on the data. The higher the relative number of dropped observations, the higher the potential bias in model performance estimation.

We could issue a detailed warning with how many observations have been dropped due to non-available factor levels. In the end it can just give a "feeling" of how this number of dropped observations might have affected the performance estimation. 

I see that there are different opinions/approaches to solve this issue (e.g. @mb706). I would like to see a "do it" or "don't do it" on whether this approach should be finalized.

R/remove_missing_levels.R
```diff
@@ -0,0 +1,84 @@
...
+    if (any(!found)) {
+      # track which variable
+      var <- predictors[i]
+      # set to NA
+      test_data[!found, predictors[i]] <- NA
+      # drop empty factor levels in test data
+      test_data %>%
+        droplevels() -> test_data
+      # issue warning to console
+      cat(sprintf(paste0("\n'sperrorest()': Setting missing levels in",
```
jakob-r: 1) Use `BBmisc:catf`
2) `paste0` is not realy necessary here, right?
3) This seems to be some code of your `sperrorest` package? Thans for sharing but you might want to edit the error message

jakob-r:
At a short first glimpse on the code:

> 3 new imports needed (magrittr, gdata, stringr) -> should I rewrite the code to avoid this?

We use `stringi` instead of `stringr`. That goes for all string operations like `gsub` also.
`magrittr` can easily be avoided.
`gdata::unmatrix` probably also. From my first impression `names(unmatrix(x))` is `outer(rownames(x), colnames(x), paste, sep=":" )` (See [here](https://github.com/cran/gdata/blob/master/R/unmatrix.R))

> Style Guide compatibility?
Looks quite good. Just replace `<-` with `=`.

R/remove_missing_levels.R
pat-s: > Use BBmisc:catf

Thanks

> paste0 is not realy necessary here, right?

No, just to stay within a margin of 80. 

> This seems to be some code of your sperrorest package? Thans for sharing but you might want to edit the error message

Yes, solved.

pat-s:
@larskotthoff @berndbischl please review 👍

R/datasets.R
```diff
@@ -125,6 +125,16 @@ NULL
...
+#' Basque pathogen infection spatial binary classification task.
+#'
+#' Contains the task (\code{bc.task.spatial}).
+#'
+#' @name bc.task.spatial
+#' @references Iturritxa, E., Mesanza, N., & Brenning, A. (2014). Spatial
+#' analysis of the risk of major forest diseases in Monterey pine plantations.
+#' Plant Pathology, 64(4), 880-889. doi:10.1111/ppa.12328
+#'
+NULL
```
larskotthoff: This doesn't have anything to do with missing factor levels, right? It should be a separate PR.

R/predict.R
```diff
@@ -72,6 +72,26 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
     newdata = newdata[subset, , drop = FALSE]
   }
 
+
+  # set factor levels, present in test but missing in train, to NA
+  if (model$learner$fix.factors.prediction == TRUE &&
+      any(vcapply(newdata, function(x) class(x)) == "factor") &&
```
larskotthoff: Why not `is.factor()`?

R/predict.R
```diff
@@ -72,6 +72,26 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
...
+      any(class(model$learner.model) == "lm" | class(model$learner.model) == "glmmPQL")) {
```
larskotthoff: Why only those particular models? If the fixing should be done only for models of a particular type, this should be a property of the learner that can be checked here.

R/predict.R
```diff
@@ -72,6 +72,26 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
...
+    # sometimes we have no task here, e.g. in test_tune_tuneMBO@34
+    if (!missing(task)) {
+      subset = task$env$data[subset, ]
+    } else {
+      subset = newdata
+    }
```
larskotthoff: Why does `subset` need to be set again here? And it looks like before `subset` was a set of indices and now it's the actual data.

R/remove.missing.levels.lm.R
```diff
@@ -0,0 +1,77 @@
...
+#' @importFrom BBmisc warningf
+#'
+#' @param fit fitted model on training data
+#'
+#' @param test.data data to make predictions for
+#'
+#' @return data.frame with matching factor levels to fitted model
+#'
+#' @keywords internal
+#'
```
larskotthoff: Please have a look at the other Roxygen comments -- types are missing, formatting.

R/remove.missing.levels.lm.R
```diff
@@ -0,0 +1,77 @@
+#' @title remove.missing.levels.lm
```
larskotthoff: The name is misleading. The function doesn't actually remove anything. Also, why is there an `.lm`?

R/remove.missing.levels.lm.R
```diff
@@ -0,0 +1,77 @@
...
+    # Obtain factor predictors in the model and their levels
+    factors = stri_replace_all_regex(names(unlist(fit$learner.model$xlevels)),
+                                     "[-^0-9]|as.factor|\\(|\\)", "")
+
+    if (length(factors) == 0) {
+      return(test.data)
+    }
+    factor.levels = unname(unlist(fit$learner.model$xlevels))
+    model.factors = as.data.frame(cbind(factors, factor.levels))
+  }
```
larskotthoff: It seems like the special handling for a particular type of model should be in the definition of the learner and not a completely unrelated file.

R/remove.missing.levels.lm.R
```diff
@@ -0,0 +1,77 @@
...
+  # For each factor predictor in your data, if the level is not in the model,
+  # set the value to NA
+
+  for (i in seq_len(length(predictors))) {
+    found = test.data[, predictors[i]] %in% model.factors[
+      model.factors$factors == predictors[i], ]$factor.levels
+    if (any(!found)) {
+      # track which variable
+      var = predictors[i]
+      # set to NA
```
larskotthoff: Trivial comment. We probably don't need it.

R/remove.missing.levels.lm.R
```diff
@@ -0,0 +1,77 @@
...
+      test.data[!found, predictors[i]] = NA
+      # drop empty factor levels in test data
+      test.data = droplevels(test.data)
+      # issue warning to console
+      catf("Setting missing levels in '%s', only present in test data but missing in train data, to 'NA'.",
+                      var)
+      catf("This affected %s of %s total observations in test data.",
+           table(is.na(test.data))[[2]], nrow(test.data))
```
larskotthoff: The first part looks incorrect -- what if previous iterations of the loop set something to NA already? Wouldn't that be counted several times?

R/resample.R
```diff
@@ -172,12 +172,27 @@ calculateResampleIterationResult = function(learner, task, i, train.i, test.i, m
...
+      if (is.null(test.i)) {
+        test.i = m$learner.model$data[test.i, ]
+      }
+
+      newdata = remove.missing.levels.lm(m, test.i)
+      pred.test = predict(m, newdata = newdata)
+    } else {
+      pred.test = predict(m, task, subset = test.i)
+    }
+      if (!is.na(pred.test$error)) err.msgs[2L] = pred.test$error
```
larskotthoff: Looks like the indentation is wrong here.

R/datasets.R
pat-s: Its the dataset required for the tests. Its also the same dataset that will be used in #1891. Still a separate PR then?

pat-s: changed

R/predict.R
pat-s: changed

R/predict.R
pat-s: Other models than those inheriting `lm` seem to work without any problems in cases addressed by this PR. I would prefer not to change anything in the data if everything is working smoothly as this may cause unwanted side-effects. 

Doing it globally here for any models that inherit `lm` is way cheaper than including/duplicating this in all `lm`* learners? I will remove `glmmPQL` again as it is not a supported learner atm

R/remove.missing.levels.lm.R
pat-s: Because so far it is only applied to lm* models. I changed the name to `missingLevelsTrain()`. Better?

R/resample.R
pat-s: solved

R/remove.missing.levels.lm.R
pat-s: The counting is done per predictor and also the warning is issued per predictor. Every `test.data` of a loop iteration is therefore unrelated to the previous/next iteration. Hence, no `NA` can be counted several times.

R/remove.missing.levels.lm.R
pat-s: addressed. better now?

R/predict.R
pat-s: > And it looks like before subset was a set of indices and now it's the actual data.

Yeah, confusing. Changed the object name to `newdata` so that `subset` is still the set of indices

pat-s:
ping @larskotthoff

R/datasets.R
larskotthoff: Is this dataset really required for this PR? Could you use another dataset? Just trying to keep things as contained as possible.

pat-s:
@larskotthoff changed tests to existing data set and cleaned up. mergeable now?

larskotthoff:
Looks good, but I'd like a second review for this.

pat-s:
ping @jakob-r @mllg

pat-s:
@larskotthoff How should we proceed if no one is able to review right now? I do not want to ping x times and rebase this PR once a week.. :/

larskotthoff:
Well, I'm afraid that's just the way it is for these things. If nobody has done it by then, there's a chance at the mlr workshop in about a month though (especially if you remind people immediately before).

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
...
+#' @description
+#' Accounts for missing factor levels only present in test data
+#' but not in train data by setting levels to NA
+#' @param fit \cr
+#'   fitted model on training data
+#' @param test.data [`data.frame`] \cr
+#'   data to make predictions for
+#' @return `data.frame` with matching factor levels to fitted model
+#' @keywords internal
+#' @export
```
jakob-r: Whz do we  export this?

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
...
+missingLevelsTrain = function(fit, test.data) {
+
+  # https://stackoverflow.com/a/39495480/4185785
+
+  # drop empty factor levels in test data
+  test.data = as.data.frame(droplevels(test.data))
+
+  # 'fit' object structure of 'lm' and 'glmmPQL' is different so we need to
+  # account for it
+  if (any(class(fit) == "glmmPQL")) {
```
jakob-r: There are many `if(class(x))` statements. Maybe you can increase use of S3?

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
...
+    model.factors = as.data.frame(cbind(factors, factor.levels))
+  } else {
+    # Obtain factor predictors in the model and their levels
+    factors = stri_replace_all_regex(names(unlist(fit$learner.model$xlevels)),
+                                     "[-^0-9]|as.factor|\\(|\\)", "")
+
+    if (length(factors) == 0) {
+      return(test.data)
+    }
+    factor.levels = unname(unlist(fit$learner.model$xlevels))
```
jakob-r: Where do the xlevels come from?

R/predict.R
```diff
@@ -72,6 +72,26 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
     newdata = newdata[subset, , drop = FALSE]
   }
 
+
+  # set factor levels, present in test but missing in train, to NA
+  if (model$learner$fix.factors.prediction == TRUE &&
+      any(vlapply(newdata, function(x) is.factor(x))) &&
+      any(class(model$learner.model) == "lm")) {
```
jakob-r: This code should be part of `predictLearner.regr.lm`

R/resample.R
```diff
@@ -174,12 +174,27 @@ calculateResampleIterationResult = function(learner, task, i, train.i, test.i, m
...
-    pred.test = predict(m, task, subset = test.i)
-    if (!is.na(pred.test$error)) err.msgs[2L] = pred.test$error
-    ms.test = performance(task = task, model = m, pred = pred.test, measures = measures)
-    names(ms.test) = vcapply(measures, measureAggrName)
-    err.dumps$predict.test = getPredictionDump(pred.test)
-  } else { # "both"
+    # set factor levels, present in test but missing in train, to NA
+    if (m$learner$fix.factors.prediction == TRUE &&
+      any(class(m$learner.model) == "lm")) {
+      test.i = task$env$data[test.i, ]
```
jakob-r: use getTaskData

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
+#' @title missingLevelsTrain
```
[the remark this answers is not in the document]
pat-s: 1cccf53

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
...
+#' @description
+#' Accounts for missing factor levels only present in test data
+#' but not in train data by setting levels to NA
+#' @param fit \cr
```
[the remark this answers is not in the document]
pat-s: 2d3ebb3

R/missingLevelsTrain.R
pat-s: removed 2d3ebb3

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,70 @@
...
+#'   fitted model on training data
```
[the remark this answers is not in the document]
pat-s: updated a515733

R/missingLevelsTrain.R
pat-s: I'll try 👍

R/missingLevelsTrain.R
pat-s: Its a sub-object of a fitted lm* model and contains the factor levels of the trained model

R/resample.R
pat-s: done 1f86782 👍

R/predict.R
pat-s: moved from `predict.WrappedModel()` to `predictLearner.regr.lm`  15b14a4

R/missingLevelsTrain.R
pat-s: see c4d3030. Is this correct?

R/predict.R
```diff
@@ -120,6 +120,19 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
     }
     time.predict = measureTime(fun1({p = fun2(fun3(do.call(predictLearner2, pars)))}))
 
+    # remove NAs in p (occurs if model inherits "lm" and misses factor levels in train)
+    if (model$learner$fix.factors.prediction == TRUE &&
+        any(is.na(p))) {
+      index.na = which(p %in% NA)
```
jakob-r: `isNA(p)` ?

jakob-r:
```
ind.not.na = !is.na(p)
...
p = p[ind.not.na]
```

pat-s: ef9494ed4814188ba834199f61bed235857f2068

R/missingLevelsTrain.R
pat-s: and this bac6533? I replaced the ´class(x)` call and instead created a .default S3 function which basically does nothing for all models except lm*

pat-s:
@jakob-r Changes addressed, Travis passing -> please give feedback :+1:

pat-s:
@jakob-r @larskotthoff ping

pat-s:
ping @larskotthoff @jakob-r @berndbischl

pat-s:
ping @larskotthoff @jakob-r @berndbischl

R/RLearner_regr_lm.R
```diff
@@ -28,6 +28,18 @@ trainLearner.regr.lm = function(.learner, .task, .subset, .weights = NULL,  ...)
...
+  if (.model$learner$fix.factors.prediction == TRUE &&
+      any(vlapply(.newdata, function(x) is.factor(x)))) {
+
+    # cheap error catching here
+    # in @test_base_generateFilterValuesData.R#93 data is not stored in m$learner.model ??
+    if (is.null(subset)) {
+      .newdata = .model$learner.model$data[subset, ]
+    }
+    .newdata = missingLevelsTrain(.model, .newdata)
+  }
```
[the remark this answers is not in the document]
pat-s: https://github.com/mlr-org/mlr/pull/1882/commits/b3ddc6111f76e01253df15d9562e6ccf2e5a0ae0

tests/testthat/test_missingLevelsTrain.R
```diff
@@ -0,0 +1,42 @@
...
+  rdesc = makeResampleDesc("RepCV", folds = 5, reps = 2)
+
+  lrn = makeLearner("classif.binomial",
+                    link = "logit",
+                    predict.type = "prob",
+                    fix.factors.prediction = TRUE)
+
+  resa = resample(lrn, task, rdesc, measures = list(auc))
+
+  expect_length(resa, 12)
```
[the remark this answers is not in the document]
pat-s: https://github.com/mlr-org/mlr/pull/1882/commits/55269be6c693edbd5fafd6fa3e54ecd6e3660908

tests/testthat/test_missingLevelsTrain.R
```diff
@@ -0,0 +1,42 @@
+context("missingLevelsTrain.R")
+
+test_that("missingLevelsTrain.R", {
```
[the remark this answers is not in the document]
pat-s: https://github.com/mlr-org/mlr/pull/1882/commits/55269be6c693edbd5fafd6fa3e54ecd6e3660908

R/missingLevelsTrain.R
```diff
@@ -0,0 +1,15 @@
...
+#' @param test.data [\code{data.frame}] \cr
+#'   Data set for prediction.
+#' @return A \code{data.frame} with matching factor levels of test data and fitted model
+missingLevelsTrain = function(model, test.data) {
+  lmod = getLearnerModel(model)
+  UseMethod("missingLevelsTrain", lmod)
+}
+
+missingLevelsTrain.default = function(model, test.data) {
+  return(test.data)
```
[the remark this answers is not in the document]
pat-s: Hi @larskotthoff, could you be a bit more precise here? What could go possibly wrong here? I see no case since this S3 method only applies for models of the lm* family and `.default` will apply for all others. 

Do you mean I should issue a message that it does nothing for other learners? I thinks that is not needed since the aim here was to only implement it for lm* models.

R/missingLevelsTrain_lm.R
```diff
@@ -0,0 +1,42 @@
+missingLevelsTrain.lm = function(model, test.data) {
```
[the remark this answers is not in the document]
pat-s: Maybe there are more gentle ways to write this but it works and is quite robust (at least I hope so :smile:). 
I do not really want to change anything here because (1) it works, (2) is well documented and (3) traceable. Do you really see an important need to change the complete approach here?

jakob-r:
@pat-s Can you fetch the upstream master and merge it into your branch to resolve the conflicts?

R/RLearner_classif_binomial.R
```diff
@@ -27,6 +27,11 @@ trainLearner.classif.binomial = function(.learner, .task, .subset, .weights = NU
 
 #' @export
 predictLearner.classif.binomial = function(.learner, .model, .newdata, ...) {
+
+  # checks factor levels, and fixes eventually occuring problems due to additional
```
mb706: I think you mean the German "eventuell" here, which would translate to "possibly" in English? The English word "eventually" means "in the end, after some time", and I feel this is not what you want to say here. Sorry to be nitpicking, but this might actually confuse people who don't know the German word.

tests/testthat/test_missingLevelsTrain.R
```diff
@@ -0,0 +1,38 @@
+context("missingLevelsTrain.R")
+
+test_that("additional factor levels in test data are removed and resample() does not error", {
+
+  task = sonar.task
+  # introduce factor level with only one observation
+  task$env$data$test_fac = factor(c(rep("1", 1), c(rep("2", 3),
+    rep("3", length(task$env$data$V1) - 4))))
```
mb706: Note that the data in `task$env` has reference semantics: You are actually *changing* the `sonar.task` here, which might introduce hard to fix bugs in other (future?) tests. I would do `task = subsetTask(sonar.task)`, which creates a copy.

R/RLearner_classif_binomial.R
pat-s: no worries, every comment is appreciated :smile: 
You're right, I changed it :+1:

tests/testthat/test_missingLevelsTrain.R
pat-s: Ok, interesting. Did not know that because I thought that I just change the `task` object which got created from `sonar.task`. Nevertheless, I changed it to your suggestions if that is more robust :+1:

pat-s:
Closing as superseeded in many ways by #2100.

PR #1883, opened 2017-07-09
Tune threshold multiclass fix

Coorsaa:
for multiclass classification threshold tuning uses now ````constrOptim```` instead of ````cmaes::cma_es```` since it should be a lot faster.
However, this needs to be checked by some benchmarks.
Moreover, possible alternatives for ````constrOptim```` could be ````solnp```` since it supports inequality- and equality-constraints, or ````bobyqa````.

Furthermore its not clear how many restarts should be made.


Coorsaa:
right now, number of restarts is set to ````1L````, but can be set via control, e.g.:
````tuneThreshold(pred, control = list(restarts = 5L))````
which results here in 6 iterations. The initial one is set to ````1/k````, where ````k```` is the number of classes, while the restarts are generated by ````runif(k)```` (sum up to 1).
The "best" iteration is finally chosen.

**Comment by larskotthoff** [comment]
Thanks. I assume you're going to post some benchmarks here as well before this is merged?

**Comment by Coorsaa** [comment]
you're assuming totally right @larskotthoff

**Comment by ja-thomas** [comment]
Thanks for your work. 

What is the plan for the benchmarks?

- cmaes (current version)
- Nelder-Mead 
- solnp
- bobyqa

There is some work using particle swarm optimization for multiclass threshold tuning: https://rd.springer.com/chapter/10.1007/978-3-642-24031-7_54

That might be worth a look, but I think we should stick to simple stuff first :)

Regarding the benchmarks, please benchmark for a different number of classes, including some more 'extreme' cases with more than 10 classes.

**Comment by Coorsaa** [comment]
Finally, things are ready.
For improving multiclass threshold tuning, we did two benchmarks. First was to compare different methods. Results can be seen below:

tunethreshold_plot

Since GenSA outperformed the rest, the next benchmark was about to find the best hyperparameters for GenSA. We tested 480 different param settings on different data/learner/measure combinations and got the following results:

gensa_perc_all
 
Or more detailed in the interesting area:

gensa_perc_all_cut

We chose the dark red curve as the best setting for our puposes whose param values I've just implemented.

**Comment by ja-thomas** [comment]
@larskotthoff @jakob-r @mllg @berndbischl
can a second person have a look and approve/suggest changes, otherwise I'd really like to merge this now.

**Comment by ja-thomas** [comment]
Speedup is quite impressive. 
Threshold Tuning on MNIST runs in under 1 minute. 
With the current implementation it was at least(!) over >15 minutes, was too lazy to wait any longer

**Comment by berndbischl** [comment]
ping me for a review in person, then merge?

File: R/performance.R [hunk 1]
```diff
@@ -12,6 +12,8 @@
 #'   Features of predicted data, usually not needed except for clustering.
 #'   If the prediction was generated from a \code{task}, you can also pass this instead and the features
 #'   are extracted from it.
+#' @param simpleaggr [\code{logical}]\cr
+#'   If TRUE, aggregation of \code{ResamplePrediction} objects is skipped. This is used internally for threshold tuning.
```
larskotthoff: Mention default in docs.

R/performance.R
```diff
@@ -110,6 +112,9 @@ doPerformanceIteration = function(measure, pred = NULL, task = NULL, model = NUL
   }
 
   # if it's a ResamplePrediction, aggregate
+  if (simpleaggr) {
+    measure$fun(task, model, pred, feats, m$extra.args)
+  } else {
   if (inherits(pred, "ResamplePrediction")) {
```
larskotthoff: Indentation of code in this block is wrong.

R/performance.R
Coorsaa: done

larskotthoff:
@berndbischl did you want to have a look before merging?

jakob-r:
Should we merge [this branch here](https://github.com/mlr-org/mlr/compare/tuneThreshold_multiclass_fix...fix2184_tuneMBO_parallel_threashold?expand=1) and then merge to master? It's somewhat related.

ja-thomas:
good idea

pat-s:
Now that Travis is stable and green again, can we merge this? Or should we still wait for a review from Bernd?

pat-s:
@mllg Thanks for merging. Could you add an entry to NEWS.md? You may know better how to phrase the changes :)

PR #1884, opened 2017-07-10
Fix probabilities for fdaclassif tasks

pfistfl:
fdaclassif.knn can now do probabilities

Introduced getBasicTaskType function, I think we should use it (or something equal) throughout our code.


pfistfl:
Somehow my gui bugged. Closing and reopening

PR #1885, opened 2017-07-10
Fix probabilities for fdaclassif tasks 

pfistfl:
fdaclassif.knn can now do probabilities

Introduced getBasicTaskType function, I think we should use it (or something equal) throughout our code.

@lbeggel @smilesun


lbeggel:
Looks good. 
You also added a small gunpoint task? We should then replace the "large" gunpoint.task in each unit test.

pfistfl:
I think we should. This just decreases runtime for unittests

smilesun:
@pfistfl  Good job, but what happened  to the travis?  Why the build is stalled?  Did someone stopped that ?

pfistfl:
@smilesun 
Should we use my function (getBasicTaskType() ) or the HashMap you created?
I think we need to settle with one.

I think we can close here, continued in #1886

smilesun:
@pfistfl , to use which one  is not important now we could keep both, first of all we finish the lists I wrote for unit test, but we should not put two many files but should rather put more tests in one file.

I would like to merge this PR first to our main branch for FDA but the travis did not finish I have no idea what happened there

smilesun:
@pfistfl , we should keep this branch before I merge it.  So things get changed in small step each time.

pfistfl:
I think there are unittests now for all of the following. They are in #1886 
So I think we have everything?

setThreshold
tuneThreshold
calculateROCMeasures
generateThreshVsPerfData
generateCalibration

pfistfl:
Since fdaclassif basically behaves exactly like classif, with the exception that there is a different ```task.type``` involved don't you think it is overkill to add even more unittests?

smilesun:
@pfistfl , fdaclassif does not behave exactly like classif, only similar, to do so is because we want to keep it consistent with all the functionalities of mlr, let it work or throw an error.

lbeggel:
@pfistfl Couly you create the test_fda_base_generateThreshvsPerf.R where you extract the fda part from the test_base_generate... ? Thanks!

lbeggel:
everything integrated in #1890

smilesun:
This is an old pr which is not consistent with our current Datastructure agreement, the major point here is to let the fda learner compute probabilities.  In the master, probablities are already available, after discussing with Florian, we decide to close this

PR #1886, opened 2017-07-10
Add unittests for fdaclassif 

lbeggel:
tests for response type "prob" and ROC measures


smilesun:
@lbeggel @pfistfl  I think it is a better idea to put the unit test in one file which is called fda_base or some similar names, since we could 1 easily keep track of what we have to do to make FDA work like general ML and also keep track of what we have done   2. FDA is a seperate module, so we should keep it as more apart from others parts of mlr as possible, i.e. do not touch other files too much.  WHat do you two think?

Please add more unit test and finish the list I wrote once you have time.

lbeggel:
The list you wrote is already complete. 
Ok, i will remove the tests in the individual files and create one base fda test.

lbeggel:
Everything integrated in #1890

smilesun:
since everything is integrated in #1890, this pr is not relevant anymore

PR #1891, opened 2017-07-12
Add task attribute "is.spatial" and implement spatial partitioning method

pat-s:
# Introduction

Spatial partitioning is essential in ecological modeling as spatial autocorrelation within the data yields to overoptimistic model performances (-> violated independence assumption of observation & residuals). Spatial partitioning in cross-validation is able to reduce this bias and yields more accurate model performances. 

rplot
_Comparison of spatial & non-spatial cross-validation partitioning (this is just an arbitrary example of an intermediate result of a study)_

## How does this work?

Spatial coordinates of observations (`x` and `y`) are used to set up a spatial clustering using `kmeans`. These variables are not used during training/prediction, only for partitioning. 

The resulting spatial clusters of observations are less affected by spatial autocorrelation due to their spatial separation. See [Brenning 2005](https://doi.org/10.5194/nhess-5-853-2005) or [Brenning 2012](https://doi.org/10.1109/IGARSS.2012.6352393) for more details.

rplot
_Red = test, black  = training_

Currently, this is implemented in [sperrorest](https://github.com/pat-s/sperrorest). At useR!2017, @berndbischl and I agreed on implementing spatial partitioning into `mlr`, making `sperrorest` somewhat obsolete. Somewhat because `mlr` currently lacks some models which are frequently used within spatial modeling (`mgcv::gam`, `MASS::glmmPQL`, `dismo::maxent`). Having the option to do spatial modeling im `mlr` opens the door to a new community! 
Also, we hope to make the spatial community more aware of the importance of spatial partitioning by implementing these methods into `mlr`. Having this reproducible workflow and the good documentation, `mlr` has the potential to become an easy-to-use (re-use) tool for scientists doing spatial modeling. They often lack programming skills / motivation diving into many models due to model complexity, tuning problems and/or different syntax. 

# Technical implementation

I implemented a logical task attribute whether the task is spatial (`is.spatial`) or not. Referring to this attribute, coordinates are used during resampling and dropped for training/prediction. 
Also, resampling methods `SpCV` and `RepSpCV` are now available. 

## Features

- [x] Add global Task description attribute `is.spatial`
- [x] Implement spatial partitioning method based on `kmeans` clustering (`SpCV` & `RepSpCV`)

## Data and tests

- [x] New task data "bc.task.spatial" (binary)
- [x] add test "classif_binomial_spatial" 
- [x] ~~add more tests, at least covering all response types to see if task attribute `is.spatial` is correctly implemented~~ -> datasets needed for this!
- [x] update all other task datasets to have the `is.spatial` attribute -> commit https://github.com/mlr-org/mlr/pull/1891/commits/b54379ad7b08bf8f6b0e23e142c7ca64693d1ad2

## Documentation & NEWS

- [x] ~~Write a "basic" documentation section how to do spatial modeling with `mlr`~~ -> I guess that can be postponed and put into a new PR as it has to go to mlr-tutorial repo anyways
- [x] Update NEWS

## Issues of spatial partitioning

Training/test splits are more heterogenous than for random partitioning. Hence, we more often face problems with missing factor levels either in train or test (#811) or non-converging models during tuning. 

## Proof of concept

`classif.binomial` 

```r
# make task
task <- bc.task.spatial
getTaskDesc(task)

$id
[1] "pathogen_data"

$type
[1] "classif"

$target
[1] "diplo01"

$size
[1] 944

$n.feat
numerics  factors  ordered 
      10        6        0 

$has.missings
[1] FALSE

$has.weights
[1] FALSE

$has.blocking
[1] FALSE

$is.spatial # new!
[1] TRUE

$class.levels
[1] "0" "1"

$positive
[1] "1"

$negative
[1] "0"

attr(,"class")
[1] "ClassifTaskDesc"    "SupervisedTaskDesc" "TaskDesc"     

####
# make resample object
rdesc_rep <- makeResampleDesc("SpRepCV", fold = 5, reps = 10)

# make learner
lrn <- makeLearner("classif.binomial", 
                   link = "logit",
                   predict.type = "prob")

# run
resa <- resample(learner = lrn, task = task, 
                 resampling = rdesc_rep, measures = list(auc))

[Resample] repeated spatial cross-validation iter 1: auc.test.mean=0.982
[Resample] repeated spatial cross-validation iter 2: auc.test.mean=0.997
[Resample] repeated spatial cross-validation iter 3: auc.test.mean=0.996
[Resample] repeated spatial cross-validation iter 4: auc.test.mean=0.986
[Resample] repeated spatial cross-validation iter 5: auc.test.mean=0.987
...
[Resample] repeated spatial cross-validation iter 46: auc.test.mean=0.986
[Resample] repeated spatial cross-validation iter 47: auc.test.mean=0.964
[Resample] repeated spatial cross-validation iter 48: auc.test.mean=0.996
[Resample] repeated spatial cross-validation iter 49: auc.test.mean=0.989
[Resample] repeated spatial cross-validation iter 50: auc.test.mean=0.99
[Resample] Aggr. Result: auc.test.mean=0.99
```


Quirouette:
Thank you for considering this!  

Not sure if this is useful or not, but make_folds {origami} has a 'cluster_ids' argument which allows groups of observations to be partitioned.
> 'cluster_ids' : a vector of cluster ids. Clusters are treated as a unit - that is, all observations within a cluster are placed in either the training or validation set.

[https://github.com/jeremyrcoyle/origami/blob/1e5ff6d7cf3c2ef1448fc1c3009404d1a312a272/R/make_folds.R](url)

Thanks again!

pat-s:
now that travis is finally passing -> please review @berndbischl @jakob-r @larskotthoff (and others I do not know yet who should review new implementations) :smiley:

.travis.yml
```diff
@@ -39,14 +39,12 @@ before_install:
   - Rscript -e 'message("INSTALLED PACKS:"); installed.packages()[,1:3]'  # we need a very fresh version of survival for Hmisc, needs be done early in the script
   # we need a very fresh version of survival for Hmisc, needs be done early in the script
   - Rscript -e 'install.packages("survival")'
-  - mkdir $HOME/bin
   - ln -s $(which gcc-4.9) $HOME/bin/gcc
```
larskotthoff: Please pull this out into a separate PR.

DESCRIPTION
```diff
@@ -156,7 +156,7 @@ Suggests:
     testthat,
     tgp,
     TH.data,
-    xgboost (>= 0.6-2),
```
larskotthoff: Same as above -- separate PR please.

NEWS.md
```diff
@@ -3,8 +3,10 @@
 ## general
 * relaxed the way wrappers can be nested -- the only explicitly forbidden
   combination is to wrap a tuning wrapper around another optimization wrapper
+* added logical task attribute 'is.spatial'
```
larskotthoff: It's great that you're doing this, but in the future please put the changes for NEWS in the PR thread -- this avoids merge conflicts with other PRs.

R/ResampleInstances.R
```diff
@@ -14,6 +14,19 @@ instantiateResampleInstance.CVDesc = function(desc, size, task = NULL) {
...
 
+instantiateResampleInstance.SpCVDesc = function(desc, size, task = NULL) {
+  coords = data.frame(task$env$data$x, task$env$data$y)
+  # perform kmeans clustering
+  inds = kmeans(coords, centers = desc$iters)
+  inds = factor(inds$cluster)
+
+  # get row indices of test set from clustering
+  test.inds = lapply(levels(inds), function(x, spl)
+    which(spl == x), spl = inds)
```
larskotthoff: I don't understand this code. Why does it give the test indices? Doesn't this simply assign everything to their clusters?

R/SupervisedTask.R
```diff
@@ -1,13 +1,13 @@
...
+  task = makeTask(type = type, data = data, weights = weights, blocking = blocking, fixup.data = fixup.data, check.data = check.data, spatial = spatial)
 
   if (check.data) {
     # costsens does not have a target col...
     # assertCharacter(target, any.missing = FALSE, min.len = 1L)
     w = which.first(target %nin% colnames(data))
     if (length(w) > 0L)
       stopf("Column names of data doesn't contain target var: %s", target[w])
-    checkTaskData(task$env$data, cols = setdiff(colnames(data), target))
+    checkTaskData(task$env$data, cols = setdiff(colnames(task$env$data), target))
```
larskotthoff: What is this change here? Was there a bug?

R/SupervisedTask.R
```diff
@@ -29,4 +29,6 @@ print.SupervisedTask = function(x, print.target = TRUE, print.weights = TRUE, ..
   if (print.weights)
     catf("Has weights: %s", td$has.weights)
   catf("Has blocking: %s", td$has.blocking)
+  if (print.target)
+    catf("Is spatial: %s", td$is.spatial)
```
larskotthoff: Why is this controlled by `print.target`? It's a property of the task, isn't it?

R/Task.R
```diff
@@ -70,6 +70,9 @@
 #'   Should sanity of data be checked initially at task creation?
 #'   You should have good reasons to turn this off (one might be speed).
 #'   Default is \code{TRUE}.
+#' @param spatial [\code{logical(1)}]\cr
+#'   Is the task spatial? I.e. does it contain coordinates ("x" and "y) which
+#'   shall be used for spatial partitioning?
```
larskotthoff: Can you expand on this a bit and explain how they are used for partitioning and how this partitioning is used please?

R/Task_operators.R
```diff
@@ -320,6 +320,18 @@ getTaskData = function(task, subset = NULL, features, target.extra = FALSE, reco
...
+    # from the GlobalEnv. If cond2 is FALSE, 'getTaskData' was called from
+    # 'subsetTask' in a nested resampling call. In this case we remove x and y
+    # later as we still need it for partitioning
+    if (!identical(parent.frame(n = 1), globalenv()) &&
+        !sys.call(-2) == "subsetTask(.task, .subset)" &&
+        task$task.desc$is.spatial == TRUE) {
+      res$x = NULL
+      res$y = NULL
+    }
   }
```
larskotthoff: Why do you need all the additional checks here, and why are you deleting x and y? Your comment doesn't make sense to me ("remove because we need it later"?).

R/predict.R
```diff
@@ -112,6 +112,10 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
       on.exit(options(warn = old.warn.opt))
       options(warn = -1L)
     }
+    if (td$is.spatial == TRUE) {
+      pars$.newdata$x = NULL
+      pars$.newdata$y = NULL
+    }
     time.predict = measureTime(fun1({p = fun2(fun3(do.call(predictLearner2, pars)))}))
```
larskotthoff: Why are you setting x and y here?

tests/testthat/test_classif_binomial_spatial.R
```diff
@@ -0,0 +1,35 @@
...
+    p.class = as.factor(binaryclass.spatial.class.levs[ifelse(p > 0.5, 1, 2)])
+    old.predicts.list[[i]] = p.class
+    old.probs.list[[i]] = p
+  }
+
+  testSimpleParsets("classif.binomial", binaryclass.spatial.df, binaryclass.spatial.target, binaryclass.spatial.train.inds,
+                    old.predicts.list, parset.list2)
+  testProbParsets("classif.binomial", binaryclass.spatial.df, binaryclass.spatial.target, binaryclass.spatial.train.inds,
+                  old.probs.list, parset.list2)
+})
```
larskotthoff: Could you write a test for the resampling as well please?

R/resample.R
```diff
@@ -75,7 +75,7 @@ resample = function(learner, task, resampling, measures, weights = NULL, models
   n = getTaskSize(task)
   # instantiate resampling
   if (inherits(resampling, "ResampleDesc"))
-    resampling = makeResampleInstance(resampling, task = task)
+    resampling = makeResampleInstance(resampling, task = task, coords = NULL)
```
larskotthoff: The `coords` argument doesn't seem to be used anywhere.

R/ResampleInstances.R
pat-s: In simple words: It correctly formats the desired indices for partitioning from the initial `kmeans` clustering. If you set a breakpoint there and run it line by line, you'll see in detail what it does 👍

larskotthoff: Can you explain in more detail in the comment please?

R/SupervisedTask.R
pat-s: In my case, sometimes when I arrived there no `data` object existed and I needed to extract the information for `checkTaskData` from `task$env$data`. 

I mean, `task$env$data` is always present anyways so this should be safe to use. However, I do not know if this may slow down the code since it first need to (lazy?) load `task$env$data`?

R/SupervisedTask.R
pat-s: Right, this should be changed. No idea why I wrote it that way. Possible copy/pasting issue.

R/SupervisedTask.R
larskotthoff: But data is an argument to the function? How can it not be there?

R/Task.R
pat-s: Okay. My suggestion would be to link to a tutorial site which I am going to write in the next days to keep the help file concise. Thoughts?

R/Task_operators.R
pat-s: Because I have to make the removal of `x` and `y` dependent on the function which called `getTaskData`.  There are differences between the `train` + `predict` combination and `resample`.  In the latter case, we arrive here (`getTaskData`) twice. The first time we need the complete data, as we need the coordinates for partitioning. The second time, we subset the data to not include `x` and `y` in the training and prediction. 

Here is the updated comment

> In this case we remove x and y later (later = we arrive in 'getTaskData'  twice in a 'resample' call) as we still need it for partitioning in upcoming function calls and only need to remove `x` and `y` before we proceed to the training step. 

More clear now? Happy to expand more if needed.

R/predict.R
pat-s: I remove them because the data was trained without `x` and `y`. Hence, the test data (= `newdata`) also needs a removal of `x` and `y`

tests/testthat/test_classif_binomial_spatial.R
pat-s: Okay! I'll check the already existing ones and figure something out

R/resample.R
pat-s: Right, seems to be a leftover of an superseded implementation approach 👍

R/SupervisedTask.R
pat-s: I'll try to reproduce my issue

pat-s: In some cases, `data` is not the full dataset and hence does not contain `x` and `y` but just data with subsetted columns.

Set a breakpoint at `if (check.data)` in `makeSupervisedTask()` in the following example. You will see how `data` changes. I need `x` and `y` because I use them to check if the the task `is.spatial`. If they are not present due to some subsetting, `makeSupervisedTask()` will report `is.spatial = FALSE` although it should be `TRUE`. 

```r
data(basque, package = "sperrorest")
basque$geometry <- NULL
                        target = "diplo01", positive = "1",
                        spatial = TRUE)
```

So there was no bug up to now - I just need always need the full dataset here for my further checks.

R/Task_operators.R
larskotthoff: This is quite ugly. I don't think it's a good idea for mlr to modify the underlying data like this. I would change the internal structure of the task to put this data somewhere else and adapt the constructors accordingly, e.g. by passing a list of column names through a `spatial` argument.

R/predict.R
larskotthoff: See comment above. I think this should be handled in a non-destructive way.

R/Task_operators.R
pat-s: To verify that I understand it correctly: You would prefer to 'remove' `x` and `y` before creating the task, i.e. not having them in a data frame initially but passing them (numeric vectors) via a list in an argument called `spatial` like so?:

```r
                        target = "diplo01", positive = "1",
                        spatial = list(x = randomvectorA, y = randomvectorB))
```

R/ResampleInstances.R
pat-s: New comment: 

```
  # uses resulting factor levels from kmeans clustering to set up a list of
  # length x (x = folds) with row indices of the data referring to which fold
  # each observations is assigned to
```

```
str(test.inds)
List of 5
 $ : int [1:248] 1 2 9 10 11 12 14 15 16 18 ...
 $ : int [1:169] 3 48 49 50 51 52 53 54 55 56 ...
 $ : int [1:249] 132 161 162 163 164 165 166 167 168 169 ...
 $ : int [1:119] 5 6 7 8 13 17 136 137 138 143 ...
 $ : int [1:159] 4 21 29 45 46 47 131 133 134 135 ...
```

R/SupervisedTask.R
larskotthoff: Right, so that makes another argument for storing the x and y information somewhere else in the task. It can't break if you subset, and your change changes the way the checks work for everything else as well.

pat-s: I thought about this. One of my first implementations indeed stored `x` and `y` within `task$task.desc` instead of `data`. 

One problem that occurred was that I need `x` and `y` to follow along the dimensions of `data` so that it always fits the data and partitioning (especially in a nested case) is performed correctly. 

Keeping `x` and `y` at `data` has the opportunity that it gets automatically subsetted with `data` and only needs to be removed once (before the training + predict step).

Placing `x` and `y` in `task$task.desc` would mean that I would need to do an additional subsetting step every time something changed in `data` to keep `x` and `y` up to date with `data`.  This would happen way more often than the removal of `x` and `y` before "train + predict". Therefore I think it is more error-prone and would affect runtime more than the other approach.

tests/testthat/test_classif_binomial_spatial.R
pat-s: see https://github.com/mlr-org/mlr/pull/1891/commits/0c090392891cbe51ecd4ce4efdf2cb7530d4ecab

R/SupervisedTask.R
larskotthoff: I think what you outline is the right approach. Yes, it's more work, but we have tests to check that it's implemented correctly, and I don't think that modifying the underlying data without telling the user is ever the right way.

For example, can you build models multiple times on the same task? Once x and y are deleted they aren't put back.

R/Task_operators.R
larskotthoff: That would be one way, yes. I was thinking of having the spatial argument name the columns in `data` that represent x and y.

R/SupervisedTask.R
pat-s: > modifying the underlying data without telling the user is ever the right way.

Well, we tell beforehand in the help file of `Task` and in the (not yet written) tutorial. 

Implementing so many manual subsetting steps would really cause a lot of new + duplicate code and is error prone. I would still vote against it atm. Maybe we can have a third opinion here?

I could try to display it more prominently that `x` and `y` will only be used for partitioning and not used during modeling? Maybe a info message during task creation? 

> For example, can you build models multiple times on the same task? Once x and y are deleted they aren't put back.

the task itself stays the same, its only modified during `resample` or `train`. Or do you mean 'stacking' when talking about multiple models?

larskotthoff: If you modify the task, the other stuff the user wants to do with it will potentially break. Yes, there's a bit more code, but I don't think it's actually that bad. Documenting this isn't enough, it's completely unexpected behaviour -- why should building a predictive model change the underlying data?

pat-s: > If you modify the task, the other stuff the user wants to do with it will potentially break. 

Could you expand on this a little? Right now I do not know what the user should want to do with the task which should not be possible anymore. My code only removes `x` and `y` before the training  + predict step in the respective objects created just before these steps. In all other steps, `x` and `y` are still present in `data`. 

If I would put `x` and `y` in `task$task.desc` I would always need to have two objects of `x` and `y` there: (1) which change with the data as it changes (with the additional duplicate subset steps) and (2) backup objects storing the unmodified `x` and `y` objects.

Leaving `x` & `y` at `data` (like now) has the opportunity that this would not be necessary. 

Also, its much more natural to leave `x` and `y` at `data` when doing any other spatial analysis stuff. Forcing the user to separate them from `data` and pass them as an extra (list) object would also cause some confusion I presume. In relation to this, if `x` and `y` are stored with the data, one can use OpenML more easily because the complete dataset is stored there.

R/Task_operators.R
pat-s: Apart from the discussion where to store `x` and `y`, this could be implemented additionally; i.e. passing custom names of spatial coordinates if they are not named `x` and `y` in `data` but e.g. `lat` & `lon`.

pat-s:
@berndbischl we need your review/opinion here 👍

R/SupervisedTask.R
pat-s: @larskotthoff @berndbischl  pls comment on this. I still favour the implementation as it is right now. Also changing it would probably cost me couple of days to get it running as smooth as it is now. However, I would like to see it merged into `mlr` within the next weeks. We have already couple of people here using my branched version right now for a paper and it would be great if we could reference to `mlr` master. So if you insist on changing it, I will do it. I will also write a tutorial page for it after this is done here.

pat-s: ping @larskotthoff @berndbischl

larskotthoff: See my question above -- does building multiple models with the same task work? It seems to me that as soon as you're removing `x` and `y` this will break.

pat-s: You mean e.g. benchmarking? works 👍 

Maybe there is still some confusion going on. I remove `x` and `y` from the task that arrives at the train/predict step - the original task is not touched and on every iteration it starts with the default. Hence, there should be no worries that the user will work with a modified task in other cases.

Also, if the user insists on having `x` and `y` as predictors in the modeling part, the user can just pass the coordinates as inputs which are named different than `x`/`y`, e.g. `x1` and `y1`. These variables will not be removed before the train/predict step. 
I will document all this once this PR is merged. 

```r
## Two learners to be compared
lrns = list(makeLearner("classif.lda"), makeLearner("classif.rpart"))

## Choose the resampling strategy
rdesc = makeResampleDesc("SpCV")

## Conduct the benchmark experiment
bmr = benchmark(lrns, bc.task.spatial, rdesc)

Task: basque, Learner: classif.lda
Resampling: spatial cross-validation
Measures:             mmce      
[Resample] iter 1:    0.0462963 
[Resample] iter 2:    0.2857143 
[Resample] iter 3:    0.1219512 
[Resample] iter 4:    0.2076923 
[Resample] iter 5:    0.0987654 
[Resample] iter 6:    0.1025641 
[Resample] iter 7:    0.0427350 
[Resample] iter 8:    0.3176471 
[Resample] iter 9:    0.2500000 
[Resample] iter 10:   0.2601626 


Aggregated Result: mmce.test.mean=0.1733528


Task: basque, Learner: classif.rpart
Resampling: spatial cross-validation
Measures:             mmce      
[Resample] iter 1:    0.1018519 
[Resample] iter 2:    0.2321429 
[Resample] iter 3:    0.0975610 
[Resample] iter 4:    0.1769231 
[Resample] iter 5:    0.2222222 
[Resample] iter 6:    0.0769231 
[Resample] iter 7:    0.1282051 
[Resample] iter 8:    0.3117647 
[Resample] iter 9:    0.2250000 
[Resample] iter 10:   0.2032520 
```

pat-s: ping @larskotthoff :)

larskotthoff: Ok, sounds like it should be fine then.

pat-s: Then I clean up and see that travis is passing?
Who else should review this?

larskotthoff: Either @berndbischl or @mllg should take a look.

pat-s:
@larskotthoff can you change your review status to accepted? travis passed all tests :)

pat-s:
ping @berndbischl @jakob-r

pat-s:
ping @berndbischl @jakob-r @larskotthoff

pat-s:
ping @larskotthoff @jakob-r @berndbischl

R/Task.R
```diff
@@ -127,6 +136,17 @@ makeTask = function(type, data, weights = NULL, blocking = NULL, fixup.data = "w
     }
   }
 
+  if (spatial == TRUE) {
+    # check if coords are named 'x' and 'y'
+    if (!any(colnames(data) == "x" | colnames(data) == "y")) {
+      stop("Please rename coordinates in data to 'x' and 'y'.")
```
larskotthoff: This is going to pass when there's only one of "x" and "y".

R/Task.R
```diff
@@ -127,6 +136,17 @@ makeTask = function(type, data, weights = NULL, blocking = NULL, fixup.data = "w
...
+    }
+  }
+  # check if setting 'spatial = TRUE' was forgotten by accident
+  if (spatial == FALSE && any(colnames(data) == "x" & any(colnames(data) == "y"))) {
```
larskotthoff: Your parentheses seem off here -- please add a test for this code and the check above.

R/Task.R
pat-s: https://github.com/mlr-org/mlr/pull/1891/commits/6be14bc85da9be2b1f9a5b06a109d5449109ebd2

R/Task.R
pat-s: https://github.com/mlr-org/mlr/pull/1891/commits/4728a2a16183bdebbe0375c774a957c1194e9335 and https://github.com/mlr-org/mlr/pull/1891/commits/550db62ded6c9020f3fe862b30650f2b4996e373

pat-s:
ping @larskotthoff

pat-s:
@larskotthoff Can we merge with only 1 review?

larskotthoff:
Ok, I'm going to merge this and endure the wrath of our overlord which is sure to follow.

pat-s:
@larskotthoff My ping button says "thanks" :smile: 
(Happy to make additional changes of course!)

PR #264, opened 2015-03-18
New Learner: added GAM(binary classification) learner

rishy:
Not sure, whether this is already working properly. I was getting some error on my local machine while running the tests. It's something related to mgcv::gam, which basically doesn't allows the usage of '.' in formula. Can anyone help me with the work around?

PR #1899, opened 2017-07-16
add `mgcv::gam` as RLearner_classif_gam

pat-s:
Picking up this previous [PR](https://github.com/mlr-org/mlr/pull/264) (which was never merged) and extending it. 

# To Do

- [x] make `link` Param depending on `family` similar as in e.g. `regr.glm` https://github.com/mlr-org/mlr/pull/1899/commits/83182fd70c0a345c9caefd64822b0105b736d311
- [x] write unit test https://github.com/mlr-org/mlr/pull/1899/commits/ae8540247d38d06559856f598ddae52e09ee9109

# Implementation Problems

- [x] How to deal with gam smoothing terms (e.g. `s()`) which are wrapped around predictor variables? -> currently passing it via "untypedParam" named "formula"

# Proof of concept

```r
# Info: 'bc.task' data is not suitable since it contains ordered factors
data(basque, package = "sperrorest")
basque$geometry = NULL
basque$lithology <- NULL
basque$soil <- NULL
basque$fus01 <- NULL
basque$x <- NULL
basque$y <- NULL
basque$hail <- NULL
                        target = "diplo01", positive = "1")

rdesc_ec <- makeResampleDesc("CV", iters = 5)

rdesc_ec_rep <- makeResampleDesc("RepCV", fold = 5, reps = 2)

lrn_ec <- makeLearner("classif.gam", 
                      family = "binomial",
                      binomial.link = "logit",
                      predict.type = "prob",
                      optimizer = "perf",
                      method = "REML",
                      formula = "diplo01 ~ s(slope) + s(ph) + 
                                      hail_new + s(age) + year + s(p_sum) + s(r_sum)")

[Resample] repeated cross-validation iter 1: auc.test.mean=0.847
[Resample] repeated cross-validation iter 2: auc.test.mean=0.812
[Resample] repeated cross-validation iter 3: auc.test.mean=0.878
[Resample] repeated cross-validation iter 4: auc.test.mean=0.865
[Resample] repeated cross-validation iter 5: auc.test.mean=0.902
[Resample] repeated cross-validation iter 6: auc.test.mean=0.828
[Resample] repeated cross-validation iter 7: Algorithm did not convergeauc.test.mean=0.82
[Resample] repeated cross-validation iter 8: auc.test.mean=0.882
[Resample] repeated cross-validation iter 9: auc.test.mean=0.881
[Resample] repeated cross-validation iter 10: Algorithm did not convergeauc.test.mean=0.886
[Resample] Aggr. Result: auc.test.mean=0.86
```


larskotthoff:
Meta-comment: please use the labels instead of changing the title.

pat-s:
@larskotthoff I have no rights to access the labels :)

larskotthoff:
You should now.

berndbischl:
please dont touch travis.yml in such PRs
also you dont need to commit NAMESPACE (or should not I guess)

berndbischl:
please use proper indentation: always 2 spaces

berndbischl:
you are also pushing and RMD and html file, i think those should also be removed

berndbischl:
"formula" should not be in the paramset

berndbischl:
dont set "tunable = TRUE", thats the default

berndbischl:
you have set many param as "tunable = FALSE" but they potenially affect model performance.
ONLY set tunable = F, is you are 100% sure that this param cannot be tuned, so e.g. logging params nd other stuff

pat-s:
> please dont touch travis.yml in such PRs

ok, after rebase with `master` this should be solved

> also you dont need to commit NAMESPACE (or should not I guess)

ok, that's new to me. It's done automatically when using `devtools::document()` during the build step but I can remove it. 

> you are also pushing and RMD and html file, i think those should also be removed

yet, leftovers 

> please use proper indentation: always 2 spaces

ok

> you have set many param as "tunable = FALSE" but they potenially affect model performance.
> ONLY set tunable = F, is you are 100% sure that this param cannot be tuned, so e.g. logging params nd other stuff

ok

> "formula" should not be in the paramset

That's on purpose, see the switch in `trainLearner.classif.gam` and also `?mgcv::gam`. Smooth terms need to be specified in the formula for `mgcv::gam`.  If not, response-predictor relationships are assumed to be linear and behave like a glm rather than a gam. So the user needs to define his smoothing structure using some argument and directly using `formula` here seems to be the most logic approach.

berndbischl:
> That's on purpose, see the switch in trainLearner.classif.gam and also ?mgcv::gam. Smooth terms need to be specified in the formula for mgcv::gam. If not, response-predictor relationships are assumed to be linear and behave like a glm rather than a gam. So the user needs to define his smoothing structure using some argument and directly using formula here seems to be the most logic approach.

i know that you did this on purpose, but if we allow this here, we should do that for EVERY learner, right? 
If we want to support this we would need a general mechanism.

@giuseppec was working on this

giuseppec:
@pat-s bernd is basically refering to this incomplete PR https://github.com/mlr-org/mlr/pull/1763 (something similar was also discussed in https://github.com/mlr-org/mlr/pull/1740). 
I think the formula stuff might be solved once we have agreed on a concrete course of action in those other issues/PR.

pat-s:
Ahh I see, I was not aware of that discussions 👍 
sure, then first a general solution should be agreed on. I can use my own implementation for my current work as long as this topic is not solved and subsequently this PR is not mergeable.

berndbischl:
do you want to remove the formula now herE?

pat-s:
> do you want to remove the formula now herE?

remove the current implementation and implement it in some other way (which is not yet agreed on based on what I've read in the other discussions that giuseppe linked). So the state of this PR would go back to "work in progress" - or did I miss a concrete solution for this problem here?

pat-s:
I added the learner to the `mlr-org/mlr-extralearner` repo which holds untested/unstable learners.

PR #1902, opened 2017-07-17
Fda_smilesun for devleopment, will never merge 

smilesun:


ja-thomas:
Can we close this? 

The last commit was in March and we obv. will not merge this.

Issue #1905, opened 2017-07-18
Add test coverage to mlr?


pat-s:
There was already an attempt here: https://github.com/mlr-org/mlr/pull/311

Now we are two years ahead and I think it would be important to have an automatic test coverage.

For my own packages I always use https://codecov.io/ and could implement my working implementation in the travis yaml file. However, @berndbischl or someone with access rights needs to setup the codecov repo. Thoughts?

larskotthoff:
Could you try this first please? The last time we tried this it didn't work because we had too many tests.

mb706:
I guess someone could just unearth the coveralls thing now with #1848 increasing the possible travis time

berndbischl:
every time i checked, covr just took ages. then i did not see any option to proceed further.
happy to include it if it works now

mb706:
Apparently we need a coveralls account for that or something?

pat-s:
> Apparently we need a coveralls account for that or something?

That was also my guess -> https://github.com/mlr-org/mlr/pull/1908#issuecomment-316706011

mb706:
In particular, I get the error ([line 2924ff](https://travis-ci.org/mlr-org/mlr/builds/255862188?utm_source=github_status&utm_medium=notification#L2924))
```
Error in covr::coveralls(quiet = FALSE) : 
  Failed to upload coverage data. Reply by Coveralls: Couldn't find a repository matching this job.
Execution halted
```
so pinging our "core members" (I actually don't know how "core" everybody here is, so sorry if I forgot you) @berndbischl @larskotthoff @mllg (

berndbischl:
so what shall i do now exactly?

pat-s:
go to https://coveralls.io/ and create/re-enable the repo for mlr-org/mlr so that the coverage report can be uploaded there

berndbischl:
just reenabled it on coverall.io

PR #311, opened 2015-05-28
Added coveralls (test coverage report) and lintr (code quality report)

zachmayer:

PR #1908, opened 2017-07-18
add test coverage using coverall

pat-s:
revival of #311, ref to #1905 

now that we have 100 min build time this should work (let's see if build passes because travis updated from precise to trusty today)


.travis.yml
```diff
@@ -72,6 +73,9 @@ after_success:
   - git push
   - "[ $TRAVIS_PULL_REQUEST == \"false\" -a $TRAVIS_BRANCH == \"master\" ] && curl -s -X POST -H \"Content-Type:application/json\" -H \"Accept:application/json\" -H \"Travis-API-Version:3\" -H \"Authorization:token $TT\" -d \"{\\\"request\\\":{\\\"branch\\\":\\\"gh-pages\\\", \\\"message\\\":\\\"commit $TRAVIS_COMMIT $TRAVIS_COMMIT_MSG\\\"}}\" https://api.travis-ci.org/repo/mlr-org%2Fmlr-tutorial/requests"
 
+after_script:
+  - Rscript -e 'covr::coveralls()'
```
mb706: If this fails because of timeout, you could try
```
  - travis_wait 100 Rscript -e 'covr::coveralls()'
```
(if it doesn't fail, just ignore this)

berndbischl:
why is there zachmayer in the badge URL?

pat-s:
> why is there zachmayer in the badge URL?

idk, just used the one from #311

berndbischl:
> idk, just used the one from #311

i already fixed that

mb706:
The reason this fails is [this line](https://github.com/mlr-org/mlr/blob/master/R/listLearners.R#L9):
```R
row = lapply(as.list(functionBody(fun)[[2L]])[slots], eval, envir = ee)
```
`getLearnerTable` tries to get the first function call within the function's body and then parses it to get the arguments. Since all `makeRLearner` call `makeRLearnerXXX` directly, this way the arguments to that function (`cl`, `name`, ...) are retrieved. However, `covr` modifies all the functions a little, as can be seen in the error message itself:
```R
rbindlist(lapply(ids, function(id) {
       {
           covr:::count("listLearners.R:8:5:8:41:5:41:24764:24764")
           fun = getS3method("makeRLearner", id)
       }
```
so this trick doesn't work with covr.

You could try to replace the relevant part in `getLearnerTable` with something like
```R
makefuns = c("makeRLearner", "makeRLearnerClassif", 
  "makeRLearnerCluster", "makeRLearnerCostSens",
  "makeRLearnerMultilabel", "makeRLearnerRegr",
  "makeRLearnerSurv")
fun = getS3method("makeRLearner", id)
environment(fun) = new.env(parent = environment(fun))
for (mf in makefuns) {
  environment(fun)[[mf]] = list
}
row = fun()
```
There are possibly more elegant ways of doing this. If you adapt this, beware that R help tells us not to use `parent.env(env) <- ...` because it can break stuff. If you end up doing this, you should probably run the `test_base_listLearners` test locally to see if it actually works...

pat-s:
```
[0K$ travis_wait 100 Rscript -e 'library(BBmisc);covr::coveralls()'

Still running (42 of 100): Rscript -e library(BBmisc);covr::coveralls()
Still running (43 of 100): Rscript -e library(BBmisc);covr::coveralls()

[31;1mThe command Rscript -e library(BBmisc);covr::coveralls() exited with 1.[0m

[32;1mLog:[0m

Error: Command â€˜'/usr/bin/gcov' '/home/travis/build/mlr-org/mlr/src/init.gcno' '/home/travis/build/mlr-org/mlr/src/smote.gcno' '-o' '/home/travis/build/mlr-org/mlr/src'â€™ failed (1)
Execution halted
/home/travis/.travis/job_stages: line 169: 25014 Terminated              travis_jigger $! $timeout $cmd
```

@berndbischl the coverall repo is active, right? because https://coveralls.io/r/mlr-org/mlr gives me a missing page error

last [build](https://travis-ci.org/mlr-org/mlr/builds/255431289?utm_source=github_status&utm_medium=notification) did not use caching so most of the time was devoted to installing the packages rather than running `covr`

mb706:
Wow, apparently I need to learn how to use github. Sorry for pushing this here, it was by accident.

pat-s:
@mb706 do you know why `GenSA` is not found anymore during testing after including your changes?

mb706:
[I ran into the same issue](https://github.com/mlr-org/mlr/issues/1934#issuecomment-318800507). I assume the problem is somehow caused by travis using an empty cache, that *all* of the things installed in stages 1 + 2 are gone, and that `GenSA` is just the first package that gets used in any test. I don't know travis well enough to solve this, though. Maybe @mllg can help?

pat-s:
I see. Looks weird, I agree. Even though travis states in the session info that packages are installed..mhh

mb706:
Huh, actually hadn't seen the session info part. Now I'm also confused.

pat-s:
for testing purposes I also tried to explicitly install/load the package directly in the test but got the same error message

berndbischl:
do we really want to focus om this, NOW?

it seems to be pretty hard to make this work. and we have 100 more pressing issues.
also this will increase travis runtime by a lot

@mllg also expressed severe doubts that we should do this now

mb706:
>for testing purposes I also tried to explicitly install/load the package directly
If you mean commit `b9a6d20`, I think that one failed because `install.packages(GenSA)` was missing quotes.

pat-s:
> do we really want to focus on this, NOW?

its not that huge of an effort to start the travis builds every once in a while and see what we get.
I would at least go on until we reach a point at which we have no how to proceed.
We know that this PR here has a low priority :)

> If you mean commit `b9a6d20`, I think that one failed because `install.packages(GenSA)` was missing quotes.

ah yes thanks

mb706:
I found the error to be the "too many DLLs" problem; I fixed it in [my own branch](https://github.com/mlr-org/mlr/pull/1919), but it turns out that covr is [hopelessly too slow](https://travis-ci.org/mlr-org/mlr/jobs/261380870) and would probably take around six hours to run. I guess we can let this PR rest until Moore's law catches up with us or R gets a jit compiler.

larskotthoff:
We could revisit this when Jenkins is back up and running.

pat-s:
@mb706 good work. Sad to hear that all efforts can not get it working right now. 

@larskotthoff what is jenkins?

larskotthoff:
Jenkins is another build system that's running on one of my machines. We don't have time limits there and much more flexibility. I had to take it down because I moved.

PR #1912, opened 2017-07-19
use trusty as dist on travis (new default)

pat-s:
`precise` is being decommissioned starting today and completely retired starting in September. See [this blog post](https://blog.travis-ci.com/). If not specified explicitly (which we have not) all linux builds will be running on `trusty` (starting today). So our builds may just break within the next days if we 

1. do not set `dist: precise` explicitly 
2. prepare our travis setup to work on `trusty` (what this PR is currently doing and what we need to do anyways until September)


berndbischl:
can you also please outline potential benefits for these PRs as here?

pat-s:
@berndbischl see updated post above

larskotthoff:
Build is failing.

Stevo15025:
```
Installing estimability
'/home/travis/R-bin/lib/R/bin/R' --no-site-file --no-environ --no-save  \
  --no-restore --quiet CMD INSTALL  \
  '/tmp/Rtmp1PcUA1/devtools45831a3a27eb/estimability'  \
  --library='/home/travis/R/Library' --install-tests 

* installing *source* package â€˜estimabilityâ€™ ...
** package â€˜estimabilityâ€™ successfully unpacked and MD5 sums checked
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded
* DONE (estimability)
trying URL 'https://cloud.r-project.org/src/contrib/lsmeans_2.26-3.tar.gz'
Content type 'application/x-gzip' length 895004 bytes (874 KB)
==================================================
downloaded 874 KB

Installing lsmeans


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

This [comment](https://github.com/travis-ci/travis-ci/issues/8082#issuecomment-315147561) in travis-ci [issue 8082](https://github.com/travis-ci/travis-ci/issues/8082) may be useful

berndbischl:
sure you can set sudo: true, but i remember some other problem with that.....

Stevo15025:
From [stackoverflow](https://stackoverflow.com/questions/26299552/travis-sudo-is-disabled)

> (when you set `sudo: true`) Travis should switch your build to use their "standard infrastructure" (rather than their "container based infrastructure") and then you can use sudo.

So if not having a container based infrastructure is bad then this will cause problems

berndbischl:
> So if not having a container based infrastructure is bad then this will cause problems

ah yeah. thats it. but we are using the contrainer based thing. thats why sudo will not work :(

Stevo15025:
Ugh, Bernd if you can check that our travis instance does not have any of the problems mentioned in that comment I posted above then we should probably report this on that travis issue page. (I would check but tbf I don't know a lot about these kinds of things and am not sure I would be thorough enough)

Stevo15025:
Or I'll review it and you can go review forecasting ;-)

pat-s:
sorry for the noisy commits introduced, accidentally checked out the wrong branch :/

Stevo15025:
> library("devtools"); document()
> Error: â€˜roxygen2â€™ >= 5.0.0 must be installed for this functionality.

Odd, I would try to restart it again. Unless roxygen2 was removed by accident

Stevo15025:
So while this passes, `sudo: false` requires us to install all of the packages each time the tests are run. Can you find an alternative method that allows us to keep the container? If it fails during the package install it may just be breaking the package needs into two or three separate scripts

berndbischl:
> So while this passes, sudo: false requires us to install all of the packages each time the tests are run.

thats infeasible, that just takes too long for us

berndbischl:
@mllg @larskotthoff 
i dont think it makes sense if too many people fiddle around this who dont have too much experience with travis. can you please weigh in?

mllg:
> So while this passes, sudo: false requires us to install all of the packages each time the tests are run.

You can cache the R library (cf. https://github.com/mllg/checkmate/blob/master/.travis.yml). After the first build, it then should be faster than the containerized build.

mllg:
Additionally, we get a more current toolchain with good c++11 support by using trusty.

pat-s:
build passing in 41 mins -> everything's fine?

mllg:
> build passing in 41 mins -> everything's fine?

I might have mixed things up ... We do not need to install a custom toolchain anymore, thus we should be fine with `sudo: false` if the other packages (libgmp-dev and libgsl10-dev) are whitelisted. If we can get along w/o sudo, we should definitely do it.

mllg:
According to https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty, both dependencies are already whitelisted.

larskotthoff:
We switched to no sudo because that makes the builds start quicker and I think also run faster.

pat-s:
https://github.com/mlr-org/mlr/pull/1912/commits/0a68ad0f38c570d7365fa9e6c826ec110cb3f25f was a working setup running 41 mins

FYI: last master build was already automatically redirected to run on `trusty` and failed because of the obsolete `mkdir` call

so we should either merge a working `trusty` solution or explictily add `dist: precise` to `master` builds

larskotthoff:
In the latest build tests are failing?

pat-s:
> In the latest build tests are failing?

yes, I see. @mllg is this related to the removal of  `libgmp-dev` and `libgsl0-dev` or should that have no influence?

mllg:
You need to install both `libgmp-dev` and `libgsl0-dev` so they should stay in the travis.yml. The fact that they are whitelisted just means that we actually can install them via apt without being sudo. 

We could remove the compilers though. But I can also do this in a separate PR.
If @larskotthoff agrees, we can merge and I'll cleanup afterwards.

pat-s:
> You need to install both libgmp-dev and libgsl0-dev so they should stay in the travis.yml. The fact that they are whitelisted just means that we actually can install them via apt without being sudo.

ah yes, make sense

> We could remove the compilers though. But I can also do this in a separate PR.
> If @larskotthoff agrees, we can merge and I'll cleanup afterwards.

@mllg sure no problem, feel free to do so if you have time 👍

larskotthoff:
Sure, looks good to me.

PR #1915, opened 2017-07-19
move travis r_packages and r_github_packages to Remotes field of Description

pat-s:
as suggested in https://github.com/travis-ci/travis-ci/issues/8107#issuecomment-316369135

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,16 +23,6 @@ env:
 warnings_are_errors: true
 r_check_args: "--as-cran --run-donttest"
 
-r_packages:
-  - roxygen2
-
-r_github_packages:
-  - berndbischl/BBmisc
-  - berndbischl/parallelMap
-  - berndbischl/ParamHelpers
-  - RevolutionAnalytics/quickcheck/pkg
-  - jimhester/lintr
-
 before_install:
   # we want to see libpaths and installed packages from cache
   - Rscript -e 'message("LIBPATHS:"); .libPaths()'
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, merging.

berndbischl:
> The remotes field has to be removed prior to CRAN release though: https://github.com/hadley/devtools/blob/master/vignettes/dependencies.Rmd#cran-submission

arg shit. @mllg can we do this in rt ?

mllg:
Do we need the Remotes? Everything except quickcheck is on CRAN. 

We could just skip the respective tests if quickcheck is not installed, and drop the Remotes.

berndbischl:
well the link to the travis thread explains that this will enable the cache. so there is an advantage.
i just dont want to manually remember and remove those lines before a release

regarding quickcheck: we dont use that anyway.

berndbischl:
but we need a decision here: either drop the field now from DESC or rt supports removal with "rbuild"

mb706:
reminder that CRAN `lintr` doesn't work.

pat-s:
also just right now saw the [verification](https://github.com/travis-ci/travis-ci/issues/8107#issuecomment-316480737) that it has to be reverted before a CRAN release

mllg:
> reminder that CRAN lintr doesn't work.

You can still explicitly install it via `r_github_packages`. If it is not compatible with CRAN, we should not use it IMHO.

larskotthoff:
Ok, so revert this?

berndbischl:
> Ok, so revert this?

i think that is the best option

PR #1919, opened 2017-07-20
coveralls experiment II, please disregard

mb706:


berndbischl:
i think we decided not to use coveralls for now. closing

Issue #1933, opened 2017-07-24
fda: extractFDAWavelets: refactor code


berndbischl:
the whole code looks dubious and bad. 

i also had to disable the example under extractFDAFeatures:

```
# FIXME: this does not run because of a problem in extractFDAWavelets
# df = data.frame(x = matrix(rnorm(24), ncol = 8), y = factor(c("a", "a", "b")))
# fdf = makeFunctionalData(df, fd.features = list(x1 = 1:4, x2=5:8), exclude.cols = "y")
# task = makeClassifTask(data = fdf, target = "y")
# extracted = extractFDAFeatures(task,
# feat.methods = list("x1" = extractFDAFourier(), "x2" = extractFDAWavelets()))
# print(extracted$task)
# reextractFDAFeatures(task, extracted$desc)
```

apparently there also is no unit tests for the wavelets, because only the example broke, and no unit test...

pfistfl:
I think this is an issue in the package: 
This works:
```
df = data.frame(x = matrix(rnorm(39), ncol = 13), y = factor(c("a", "a", "b")))
fdf = makeFunctionalData(df, fd.features = list(x1 = 1:4, x2 = 5:13), exclude.cols = "y")
task = makeClassifTask(data = fdf, target = "y")
extracted = extractFDAFeatures(task,
feat.methods = list("x1" = extractFDAFourier(), "x2" = extractFDAWavelets()))
print(extracted$task)
reextractFDAFeatures(task, extracted$desc)
```

and this aswell

```
# FIXME: this does not run because of a problem in extractFDAWavelets
# df = data.frame(x = matrix(rnorm(24), ncol = 8), y = factor(c("a", "a", "b")))
# fdf = makeFunctionalData(df, fd.features = list(x1 = 1:4, x2=5:8), exclude.cols = "y")
# task = makeClassifTask(data = fdf, target = "y")
# extracted = extractFDAFeatures(task,
# feat.methods = list("x1" = extractFDAFourier(), "x2" = extractFDAWavelets(filter  = "haar")))
# print(extracted$task)
# reextractFDAFeatures(task, extracted$desc)
```

The problem is I think when the parameter for the wavelet filter is bigger then the functional length? 
@lbeggel does this make sense?

berndbischl:
can you first clean up the code pls?
then we can discuss this bug

smilesun:
Summary :
1. The default parameters do not match in the unit test
2. Bernd does not like the code

Solution:
[...]

Issue #1930, opened 2017-07-24
extractFDAWavelets: filter arg not doced and bad assert


berndbischl:
it is unclear which values are valid. 
also reiterate the code of  this extractor

pfistfl:
Fixed in #1937

Issue #1928, opened 2017-07-24
FDA: makeExtractFDAFeatsWrapper FIXME


berndbischl:
why is this FIXME still there:

```


  # FIXME:
  # This is stupid, we can not handle multiple tasks for a single wrapper this way.
  # (Impute cant do this neither if using cols = list("X1" = ...)).
  # One solution would be to be able to specify "all" features (regexp would be overkill?).
```

pfistfl:
Fixed in #1935

PR #1935, opened 2017-07-24
Fix various issues related to FDA Project (Wavelts, Wrapper)

pfistfl:
This PR handles
the wrapper in
#1928 
and the wavelets in
#1930 and #1933


berndbischl:
should assertion at least check for a named list?

later on a better check will we done by extractFDAFeatures i guess

pfistfl:
There are 30-40 possible filters.
I am currently doing a sanity check, but the current check will not catch all possibly wrongly specified filters.
Should I do
```
assertChoice(filter, c("la2," "la4", "la8", ......, "c30") or is the regexp sufficient?
```

smilesun:
I have added what Florian did in his branch to fda_lte  #2005.  SO this is not relevant anymore, could be deleted.

PR #1937, opened 2017-07-24
Fix docs and example for extractFDAFeatures

### LINKED ISSUE #1932: extractFDAFeatures: feat.methods docs
the docs said that is was not allowed to work in the the same FDA feature multiple times.
i think this was incorrect

make this clear. 

also reread the whole docs for that function

pfistfl:
Fixes #1932 and re-adds the example


R/extractFDAFeatures.R
```diff
@@ -32,22 +32,21 @@
 #'   functional feature.
 #'   Names of \code{feat.methods} must match column names of functional features.
 #'   Available feature extraction methods are available under family \code{fda_featextractor}.
-#'   Default is \code{list()} which does nothing.
+#'   Default is \code{list()} which does nothing. Specifying a functional feature multiple times
```
berndbischl: the default is always mentioned as the last line in the docs

PR #1938, opened 2017-07-25
set na.rm to TRUE for variance feature filter

### LINKED ISSUE #1906: What is the reason for the "variance" feature filter not handling NAs?
Many other filters seem to handle NAs, but the variance filter explicitly says [`na.rm = FALSE`](https://github.com/mlr-org/mlr/blob/master/R/Filter.R#L574), so I assume there is some reason for that?

mb706:
Fixes #1906


berndbischl:
is a unit test possible here?

giuseppec:
I added a unit test.

larskotthoff:
Thanks, merging.

Issue #1892, opened 2017-07-13
Can printHead be removed?


mb706:
`mlr` defines the function `printHead` in `R/utils.R`, but this is also exported by `BBmisc`. The output of both is subtly different; if this difference is enough to legitimize duplicating this function, I suggest renaming `mlr`'s version, since I get warnings when loading packages that import `mlr` when using `devtools::load_all`.

larskotthoff:
I don't see a problem with renaming it as it's only used internally and not exported.

mb706:
For reference, the function in `BBmisc` is [here](https://github.com/berndbischl/BBmisc/blob/master/R/printHead.R#L15-L23), the one in `mlr` is [this one](https://github.com/mlr-org/mlr/blob/master/R/utils.R#L17-L21).

PR #1939, opened 2017-07-25
Use BBmisc printHead instead of implementing it ourselves.

mb706:
See #1892 
This avoids an annoying namespace warning when `load_all`-ing mlr and a package that imports mlr. Another solution would be to rename `printHead`, but I have the feeling the BBmisc version was specifically written to replace this one?


berndbischl:
you are 100% correct. good change. i am working on travis, will merge anyway now as this is trviial

Issue #1898, opened 2017-07-16
Java based learners fail with parallelMap multicore


mb706:
This is because `fork()`, which multicore is ultimately based on, and the java VM don't play along well if java is started before the forking happens. Loading java based packages, e.g. "RWeka", seems to start the java VM, so if the package gets loaded outside of the `parallelMap` call. it fails.
```R
> library("mlr")
Loading required package: ParamHelpers
> library("parallelMap")
> parallelStartMulticore(2)
Starting parallelization in mode=multicore with cpus=2.
> resample("classif.IBk", pid.task, cv5)  # loads RWeka, then calls parallelMap
Mapping in parallel: mode = multicore; cpus = 2; elements = 5.
[Resample] cross-validation iter 1: [Resample] cross-validation iter 2:
# hang
```
If, on the other hand, the fork is before loading the java vm, it works fine:
```R
> library("mlr")
Loading required package: ParamHelpers
> parallelStartMulticore(2)
Starting parallelization in mode=multicore with cpus=2.
> parallelMap(function(x) resample("classif.IBk", pid.task, cv5), 1:2, simplify=FALSE)
Mapping in parallel: mode = multicore; cpus = 2; elements = 2.
[Resample] cross-validation iter 1: [Resample] cross-validation iter 1: mmce.test.mean=0.275
[Resample] cross-validation iter 2: mmce.test.mean=0.331
[Resample] cross-validation iter 2: mmce.test.mean=0.273
# ...
# no hang
```

I therefore suggest to have a `configureMlr` option to defer loading of packages until a learner's `train` or `predict` function gets called. The user would still need to be careful not to load "RWeka" when he wants to use multicore, but this at least would give him the option. When a learner gets constructed, instead of loading a learner's package, mlr should simply check whether the requested package exists.

mb706:
A current workaround is to load a learner from a savefile. E.g. if a learner is loaded from the `.RData` file at start, resampling with multicore works.
```R
> library("parallelMap")
> parallelStartMulticore(2)
Starting parallelization in mode=multicore with cpus=2.
> library("mlr")
Loading required package: ParamHelpers
> lrn = makeLearner("classif.IBk")
> resample(lrn, pid.task, cv5)
Mapping in parallel: mode = multicore; cpus = 2; elements = 5.
[Resample] cross-validation iter 1: [Resample] cross-validation iter 2: ^C^C^C^C^C
> q("yes")
$ R
[...]

PR #1942, opened 2017-07-25
Make loading of packages when loading a learner optional

mb706:
See issue #1898.
This does not fix that issue, but it may be useful in these kind of cases. This commit can serve as a basis of discussion:
* Do we actually want something like this? If not, feel free to close.
* Should the config option have a different name and default to `TRUE` instead of `FALSE`?
* Possibly more elaborate checks whether the relevant package is at least present?

### DIFF
--- a/R/RLearner.R
+++ b/R/RLearner.R
@@ -64,7 +64,9 @@ makeRLearnerInternal = function(id, type, package, par.set, par.vals, properties
   # must do that before accessing par.set
   # one case where lazy eval is actually helpful...
   assertCharacter(package, any.missing = FALSE)
-  requirePackages(package, why = stri_paste("learner", id, sep = " "), default.method = "load")
+  if (!coalesce(getMlrOption("defer.package.load"), FALSE)) {
+    requirePackages(package, why = stri_paste("learner", id, sep = " "), default.method = "load")
+  }
 
   assertString(id)
   assertChoice(type, choices = c("classif", "regr", "multilabel", "surv", "cluster", "costsens"))
--- a/R/configureMlr.R
+++ b/R/configureMlr.R
@@ -53,12 +53,15 @@
 #'   is \dQuote{warn} or \dQuote{quiet}. If it is \code{TRUE}, the dump can be accessed using
 #'   \code{\link{getFailureModelDump}} on the \code{\link{FailureModel}}, \code{\link{getPredictionDump}} on the failed prediction, and \code{\link{getRRDump}} on resample predictions.
 #'   Default is \code{FALSE}.
+#' @param defer.package.load [\code{logical(1)}]\cr
+#'   When set to \code{TRUE}, packages are not loaded when a learner is constructed.
+#'   Instead it is loaded during train / predict. Default is \code{FALSE}.
 #' @template ret_inv_null
 #' @family configure
 #' @export
 configureMlr = function(show.info, on.learner.error, on.learner.warning,
   on.par.without.desc, on.par.out.of.bounds, on.measure.not.applicable,
-  show.learner.output, on.error.dump) {
+  show.learner.output, on.error.dump, defer.package.load) {
 
   defaults = list(
     show.info = TRUE,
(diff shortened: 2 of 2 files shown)


berndbischl:
can we maybe collect arguments WHY we would ever load the package already during learnner construction?

here is what i know: some hyperpars look (unfornutaley) have values and defaults that are obects from the learner package namespace. so you need that loaded, to contrcuct the paramset of the learner.

we COULD solve that by representing these objects simply as strings, which we map to these objects during training.

but all of these are rather non-trviail changes and I dont know what other problems would additionally appear. hmm maybe not so many, as the learner object is mainly a bunch of strings?

mb706:
The benefits of the current system that I can see is (i) the user gets feedback if a package can not be loaded, right when he tries to construct the learner (ii) the package does not need to be loaded for each thread when using multicore parallelMap. It is possible that some packages need explicit `require()`-ing if they do something important in `.onAttach` (but 1. imho that would be a bug in that package and 2. afaics most learners are only `requireNamespace()`-d anyways).

I was also half wondering if the `requirePackages` calls come from a time when mlr didn't `::`-address all learner functions?

pat-s:
@mb706 What's your opinion on this these days?

> Should the config option have a different name and default to TRUE instead of FALSE?

Imo, change name to be a bit more descriptive. When setting the default to TRUE, how likely is this going to cause problems at some point?

> Possibly more elaborate checks whether the relevant package is at least present?

Would make sense so that we can stop early.

mb706:
I still think this is helpful when trying to dodge #1898. Some more thoughts, in random order:
* Not immediately loading packages would also be helpful when a user is trying to get some info on a learner (e.g. parameter space) without having the relevant packages installed
* While it would be possible to check for package availability without loading we should be aware of possible cases where packages were `load_all`'d, where a package is present but can't be loaded (e.g. compiled for a different R version), or where a user specifically wants the behaviour above
* I'm not 100% sure whether the `helpLearner` and `helpLearnerParam` stuff *still* loads the packages on construction, the relevant code is in `mlr:::makeParamHelpList`. At least if we want to allow constructing learners without the package present there would need to be some tests whether `makeParamHelpList` fails in that case.
* A disadvantage when *not* loading packages on construction is that multicore parallelization would load the package in every thread individually, which will have a performance impact in certain situations. We should therefore probably have default behaviour `FALSE`.
* Another option is to have a flag for individual learners indicating java-dependency. Then an mlr-config option `load.package.on.learner.construction` `"always"`, `"never"`, `"dynamic"`, could have the default "dynamic" which loads java-dependent stuff later, everything else immediately. This would be the most user-friendly option imho, since it behaves the way it does now in most cases except the ones where it causes problems, but still gives the final word to the user.

pat-s:
> A disadvantage when not loading packages on construction is that multicore parallelization would load the package in every thread individually, which will have a performance impact in certain situations. We should therefore probably have default behaviour FALSE.

Yes. Especially because "multicore" is the default for most people on remote machines. 

> Another option is to have a flag for individual learners indicating java-dependency. Then an mlr-config option load.package.on.learner.construction "always", "never", "dynamic", could have the default "dynamic" which loads java-dependent stuff later, everything else immediately. This would be the most user-friendly option imho, since it behaves the way it does now in most cases except the ones where it causes problems, but still gives the final word to the user.

Sounds also good. 

I think these two points are the most important ones. Would you be up implementing these? Including a small section in https://mlr.mlr-org.com/articles/tutorial/configureMlr.html?

I am not sure if we need such detailed checks for the two "edge cases" you outlined ('load all#'ed pkgs and `helpLearnerParam` behavior.) Checking if `makeParamHelpList` fails without the loaded package should be a quick one that should apply to all learners if so?

pat-s:
> We should therefore probably have default behaviour FALSE.

Not changing the current default is a bit in contrast to all the effort and changes of this PR. 
Not many users would use it since they would need to understand the problem in the first place and then take action on their side (find the option, change it).

If we are doing such a fundamental change in terms of how packages are loaded the advantages should be so important that they become the new default, shouldn't they?

To be able to discuss this, we need a clean comparison of the impacts of the change in all possible parts of the package. This sounds like a lot of work.

@mb706 	
Please reply shortly if you have time/motivation to tackle this and continue here.
I suspect that the effort might now be worth the added value in the end?

pat-s:
Closing for now. Please re-open if you are keen on continuing here :)

PR #1943, opened 2017-07-25
Travis cleanup + build stages

mllg:
WIP


mb706:
My PR is the problem. The test doesn't work on my machine so I pushed it expecting travis to find any problems with the tests if there are any. The rest is history I guess. You can rebase [the pr_1943_suggestion branch](https://github.com/mlr-org/mlr/tree/pr_1943_suggestion) (or just apply the diff manually), it should make everything work.

PR #1955, opened 2017-07-31
tuneMBO did return negative performance values when maximizing

jakob-r:
For some optimization startegies we transform the objective function so that the internal optimizer can minimize the function even if the measure is something we want to maximize.
This internal transformation is visible to the user when he sees the result which can be something like
`acc = -0.9`.
This becomes confusing for measures that have values between [-1,1].

See: https://github.com/jakob-r/mlrHyperopt/issues/6


tests/testthat/test_tune_tuneIrace.R
```diff
@@ -11,7 +11,7 @@ test_that("tuneIrace", {
   ctrl = makeTuneControlIrace(maxExperiments = n, nbIterations = 1L, minNbSurvival = 1)
   tr1 = tuneParams(makeLearner("classif.rpart"), multiclass.task, rdesc, par.set = ps1, control = ctrl)
   expect_true(getOptPathLength(tr1$opt.path) >= 10 && getOptPathLength(tr1$opt.path) <= n)
-  expect_true(!is.na(tr1$y))
+  expect_number(tr1$y, lower = 0, upper = 0.2)
```
jakob-r: Well, we want to make sure that the optimizer does not run in the wrong direction. This is a quite save range.

ja-thomas: We had some discussion about this here: https://github.com/mlr-org/mlrMBO/issues/324

jakob-r:
> This will only change the behaviour of optimization that uses smoof, i.e. only MBO based? In that case, maybe a better solution would have been to manipulate the tuning result in the tuneParams function in tuneParams.R, before returning it to the user.

But this way we have correct opt paths

> Otherwise, the PR looks fine to me. I only have a weak opinion that tests should be guaranteed to work, even if a learner's performance were to suddenly deteriorate.

The expectations are really conservative.

berndbischl:
1) can you please post a direct mlr example, without hyperopt, that demos the problem?

2) why are you removing the "force"? that seems to be muddling issues and is uncommented from your side?

3) can you please document the semantics of "always.minimize" at least in one place?

4) i find it confusing why that the value for always.minmize is set to different values in different places.
this is also internally not documented

jakob-r:
> 1) can you please post a direct mlr example, without hyperopt, that demos the problem?

```r
library(mlr)
ps = makeNumericParam
ps = makeNumericParamSet("C", len = 1, lower = 0, upper = 10)
ctrl = makeTuneControlMBO(mbo.control=mlrMBO::makeMBOControl())
tuneParams(learner = "classif.ksvm", task = iris.task, control = ctrl, par.set = ps, resampling = cv2)
# Tune result:
# Op. pars: C=1.37
# acc.test.mean=-0.9400000
```

> 2) why are you removing the "force"? that seems to be muddling issues and is uncommented from your side?

Just some cleanup

> 3) can you please document the semantics of "always.minimize" at least in one place?

I added a documentation

> 4) i find it confusing why that the value for always.minmize is set to different values in different places.
this is also internally not documented

:thinking: Well... when we call `tunerFitnFun` from withing  `tunerSmoofFun` is set it to FALSE because I can set the smoof argument minimize according to the measure.

mb706:
> > This will only change the behaviour of optimization that uses smoof, i.e. only MBO based? In that case, maybe a better solution would have been to manipulate the tuning result in the tuneParams function in tuneParams.R, before returning it to the user.

> But this way we have correct opt paths

Manipulating the tuning result would also entail correcting the `opt.path`.

mllg:
The corrected MRE @jakob-r posted:
```r
set.seed(1)
library(mlr)
ps = makeNumericParamSet("C", len = 1, lower = 0, upper = 10)
ctrl = makeTuneControlMBO(mbo.control=mlrMBO::makeMBOControl())
tuneParams(learner = "classif.ksvm", task = iris.task, control = ctrl, par.set = ps, resampling = cv2, measure=acc)
```

mllg:
I agree with @mb706 that we should not rely on the RNG god being gracious. If we just want to ensure that the measure is not inverted, we could just use seeding here.

jakob-r:
Is it okay to set a seed in the test?

jakob-r:
Anyhow. The seed is already set in `test/testthat/helper_zzz.R`.

larskotthoff:
What's the status here? I'm also a bit concerned about relying on the RNG -- can we instead check somewhere that everything is positive?

jakob-r:
I still prefer the tests how they are atm because we want to make sure, that the tuning walks into the right direction. And as we pointed out the seed is already set.
If they will be errors because these tests are failing we still can discuss about it in the future and correct it.

larskotthoff:
Ok, I'll merge this as it's a pretty bad bug and we can improve later.

jakob-r:
Thanks!

Issue #3, opened 2013-11-13
MultiClass AUC


berndbischl:
See wether we can create an AUC measure for more than 2 classes in mlr.

Here is a hint, sent by Markus by mail.

```r
 learner <- makeLearner('classif.lda', predict.type="prob")
 task <- makeClassifTask(data=iris, target="Species")
 mod <- train(learner=learner, task=task)
 pred.obj <- predict(mod, newdata=iris)
 library(HandTill2001)
 predicted <- as.matrix(pred.obj$data[,paste("prob.", levels(pred.obj$data$response), sep="")])
 colnames(predicted)<-levels(pred.obj$data$response)
 auc(multcap(response=pred.obj$data$response, predicted=predicted))
```

1) Investigate

2) Add Measure, doc. and test it

3) Briefly describe in tutorial / ROC part.

PhilippPro:
I think there is something wrong in the multiclass.auc function. As input there is only the response and the probability, but not the truth. A prediction cannot be evaluated without the true values. Probably the term `response` in the pROC package and the poor documentation is missleading. 
In my opinion `pred$data$response` has to be replaced with `pred$data$truth`.

Also the multiclass.auc and auc should coincide in the two class case. We could transform the probabilities like in #858 or just apply auc. 

Maybe it would also be better to implement the auc stuff by ourselves, maybe I will do it. In this paper there are described several different auc measures for the multiclass case:
https://www.math.ucdavis.edu/~saito/data/roc/ferri-class-perf-metrics.pdf

jakob-r:
I think you are pretty right. We should fix this asap. But the documentation does not enlighten me so much what the correct input should be. But I can see that the current state can not work.

catastrophic-failure:
An example using `caTools`

```splus
library(mlr)

lrn = makeLearner("classif.randomForest", predict.type = "prob")
model = train(lrn, iris.task)
pred = predict(model, iris.task)

library(caTools)
colAUC(getPredictionProbabilities(pred), pred$data$truth)
```

> 
> > colAUC(getPredictionProbabilities(pred), pred$data$truth)
>                                     setosa  versicolor  virginica
> setosa vs. versicolor    1.00     1.0000       0.84
> setosa vs. virginica      1.00      0.7336      1.00
> versicolor vs. virginica  0.54     1.0000       1.00

Each column of probability estimates is a 1-vs-rest binary classifier. Then we can group the outputted AUCs knowing the labels.

[...]

PR #1956, opened 2017-07-31
add auc filter

bommert:


catastrophic-failure:
I don't think you need `ROCR` to do that, you could use the already implemented `auc` functions in `mlr`, and from what I remember these were faster.

larskotthoff:
Looks like the tests are failing.

smilesun:
Hi, We find the following error, not sure yet how this works to constrain to binary class
```
1. Error: filterFeatures (@test_featsel_filters.R#15) --------------------------
  Number of classes is not equal to 2.
  ROCR currently supports only evaluation of binary classification tasks.
  1: filterFeatures(task = multiclass.task, method = filter, perc = 0.5) at testthat/test_featsel_filters.R:15
  2: generateFilterValuesData(task = task, method = method, nselect = nselect, ...)
```

mllg:
You could switch to `pROC` which supports multiclass auc and is already in `Suggests` (albeit we do not use it :-1:).

```r
#' Filter \dQuote{auc} determines for each feature how well the target
#' variable can be predicted only based on this feature. More precisely, the
#' prediction rule is: class 1 if the feature exceeds a threshold and class 0
#' otherwise. The performance of this classification rule is measured by the
#' AUC and the resulting filter score is |0.5 - AUC|.
#'
#' @rdname makeFilter
#' @name makeFilter
makeFilter(
  name = "auc",
  desc = "AUC filter for binary classification tasks",
  pkg  = "pROC",
  supported.tasks = "classif",
  supported.features = "numerics",
  fun = function(task, nselect, ...) {
    data = getTaskData(task, target.extra = TRUE)
    if (length(getTaskClassLevels(task)) == 2L) {
      score = vnapply(data$data, function(x, y) {
        as.numeric(pROC::roc(predictor = x, response = y)$auc)
      }, y = data$target)
    } else {
      score = vnapply(data$data, function(x, y) {
        as.numeric(pROC::multiclass.roc(predictor = x, response = y)$auc)
      }, y = data$target)
    }

    abs(0.5 - score)
  }
)
```

This snippet should already work, but it would be great if @bommert and/or @smilesun could carefully read the manual and confirm that you do not have to set any extra arguments for pROC. Additionally, a short unit test would be great (e.g., check on iris that `Petal.Length` and `Petal.Width` get a high score).

berndbischl:
> You could switch to pROC which supports multiclass auc and is already in Suggests (albeit we do not use it ).

please be careful here. if you search the history, you will notice that we actually USED pROC for multiclass AUC. it was then removed for a reason

berndbischl:
I guess you want to read at least this
https://github.com/mlr-org/mlr/issues/3

berndbischl:
uuuh. just looked at the code. main point is:
we SURELY want to use the *same* calculation here for the AUC that mlr uses for its normal measures?

i DO NOT want be inconsistent. 
so not pROC there, ROCR in another place.

why not simply call into mlr's measureAUC function?

mllg:
Right, multiclass was a bad idea in the first place for the filter. But we are apparently missing a property for the filters to flag them to only work on binary classif tasks.

berndbischl:
> Right, multiclass was a bad idea in the first place for the filter.

why? what would make it more applicable? 

> But we are apparently missing a property for the filters to flag them to only work on binary classif tasks.

separate issue?

mllg:
> why? that would make the filter more applicable? I would like that

`pROC` has a function for that, but they say it is less tested. And the multiclass measures in mlr seem to require probabilities for each class which is something we do not have in the filter.

PR #1961, opened 2017-08-03
Use absolute values in conf matrix for roc measure calculation

### LINKED ISSUE #1958: calcROCMEasures: show abs confusion value counts?
```
lrn = makeLearner("classif.lda", predict.type = "prob")
r = holdout(lrn, sonar.task)
print(calculateROCMeasures(r$pred), abbrev = FALSE)

    predicted
true M         R
   M 0.82      0.18      tpr: 0.82 fnr: 0.18
   R 0.38      0.62      fpr: 0.38 tnr: 0.62
     ppv: 0.66 for: 0.21 lrp: 2.16 acc: 0.71
     fdr: 0.34 npv: 0.79 lrm: 0.29 dor: 7.39
```

the info in the first 4 cells seems very redunant with the other measures.
wouldnt it make more sense to show the normal absolute confmatrix counts here?

would also be nice if i could see the cor and row sums

ja-thomas:
Fixes #1958 

I didn't add the row/col sums, since the ROCmeasures are only defined for two-class problems which makes them still easy to see imo.

If you really want I can add them, but the output is already a bit cluttered.

### DIFF
--- a/R/calculateROCMeasures.R
+++ b/R/calculateROCMeasures.R
@@ -1,7 +1,7 @@
 #' @title Calculate receiver operator measures.
 #'
 #' @description
-#' Calculate the relative number of correct/incorrect classifications and the following evaluation measures:
+#' Calculate the absolute number of correct/incorrect classifications and the following evaluation measures:
 #'
 #'
 #' \itemize{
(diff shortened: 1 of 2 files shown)


mllg:
Thanks!

PR #1966, opened 2017-08-07
Oneclass: Add convert to probability function

MinhAnhL:
e1071::svm and kernlab::ksvm do not return probability for oneclass classification.
Implementation of platt scaling for oneclass according to https://www.cse.buffalo.edu/~jing/doc/ICDM06.pdf
With fix parameters A = 0, B = 1.


lbeggel:
Did you already solve the lintrbot issues?

pat-s:

PR #1967, opened 2017-08-07
Oneclass: Rlearner h2o.autoencoder

MinhAnhL:
Add Autoencoder from the h2o package


pat-s:

PR #1968, opened 2017-08-07
Oneclass: Calculate AMV (area under the mass volume) Anomaly performance measurement

MinhAnhL:
There are no unsupervised performance for anomaly detection in mlr yet.
Paper, Co-winner of Google best paper award:  https://drive.google.com/file/d/0B8Dg3PBX90KNUTg5NGNOVnFPX0hDNmJsSTcybzZMSHNPYkd3/view  introduced the area under the mass volume curve (AMV)
And for high dim data: https://arxiv.org/pdf/1607.01152.pdf (AMVhd)

AMVhd was more complex to implement, as it needs the trained model as input variable.

!Computational expensive


pat-s:

Issue #1852, opened 2017-06-13
Resampling for oneclass-classification/anomaly detection


MinhAnhL:
In oneclass-classification you usually train on data, which only contain one class (the normal class). The trained model should learn the behaviour of the normal class and therefore it should be able to detect anomalies in the test data (if there are any).
To evaluate the performance we would use resampling methods, but mlr does not provide the option yet to only have oneclass in the resampled training data.

Suggestions on how to implement resampling for oneclass:

**1. Hold out** 
- split the data, with only normal data in the train set
- and all anomalies in the test set

**2. k-CV**
- split the data as usual into k-partition
- one partition is the test set as usual
- the rest k-1 partition are the train set as usual, BUT drop anomalies when training the model
- maybe per default stratify == TRUE ? (to make sure we have anomalies in the test data)

**3. Bootstrapping**
- sample with replacement only from the normal data to create train set
- anomaly data are always in the test set.

**4. Subsampling** 
- nearly analog to Bootstrapping
- sample without replacement only from the normal data to create train set (size is $split = a proportion of training cases e.g. 2/3)
- anomaly data are always in the test set.
- MAYBE:  for test set: sample without replacement of anomaly data with size $split (e.g. 2/3) to get some variation in the anomaly data and also to keep the proportion between number of anomalies and number of normal data?

**5. repeat CV**
- stays the same as instantiateResampleInstance.RepCVDesc() in mlr but with one line changed
instead of d = makeResampleDesc("CV", iters = folds) use method 2 from above ("OC-CV")

**6. LOO** 
- not necessary(doesn't make sense to make an own class for oneclass 

**Add-on question:** Is it necessary to have a resampling option that creates test data without anomalies?
I particularly think of the question: Does the Anomaly Detection model learned the normal class well enough to recognize all normal observation as normal in a dataset which only contain normal classes. Or does the model always detect observations with the lowest score as anomalies, although there aren't any/ Test false alarm.
Alternative: Only input normal data, therefore test and train only on normal observation
But, what if: train on data which has normal and anomaly cases but test only on normal data.? Not sure whether this is a common case and worth analysing.

mb706:
Things that I can think of:

> MAYBE: for test set: sample without replacement of anomaly data with size $split (e.g. 2/3)

Depending on the measure, if you always put all the anomalies in the test set there are going to be bias problems with most of these.

> resampling option that creates test data without anomalies

[...]

PR #1969, opened 2017-08-07
Oneclass: Oneclass resampling 

MinhAnhL:
The idea of oneclass classification is basically to train on one-class (the normal class) data. After learning the normal behavior of the data the model should be able to detect anomalies in new data.

In order to do nested resampling it was necessary to add resampling strategies, which only train on normal data. The new strategies have the suffix oc_

See issue #1852


pat-s:

PR #1970, opened 2017-08-07
Oneclass: New measurements

MinhAnhL:
In Anomaly detection research the authors often use supervised performance measurements for evaluation (although they trained on unsupervised).
Besides measurements from the confusion matrix, measurements like weighted balanced accuracy, precision at p, average precision etc. are used. 
These are implemented in this branch.


pat-s:

PR #1971, opened 2017-08-07
Oneclass: Small changes in the variable names in resample.R

MinhAnhL:
resample() and benchmark() have a logical input variable "models", which indicates whether all fitted models should be returned.
It was quite confusing that within the resample function the name of the input variable change from "models" to "model" (without s). 
Additionally in order to make threshold tuning possible for oneclass with the amv measure, we need to pass the actual trained model. The actual trained model is therefore saved in the variable "model" without s (next branch: oneclass_tunethreshold).

To use the variable "model" in the next branch I made the naming consistent in this PR


berndbischl:
discussed with @MinhAnhL 
does seem to make sense

PR #1972, opened 2017-08-07
Oneclass: Make tunethreshold for amv measure

MinhAnhL:
Tuning threshold in anomaly detection is essential, but mlr::tunethreshold() wasn't compatible with the amv measure (unsupervised measurement for anomaly detection), as amv requires the trained model.
I needed to insert some hacks to make threshold tuning possible for amv.


MinhAnhL:
Changes in this PR was made to make threshold tuning for the AMV measure (Area under the mass volume curve) available, but that actually doesn't make sense, as AMV is an unsupervised measure.

Issue #1887, opened 2017-07-10
NA handling of Aggr. result


pat-s:
Given an `resample` output of 

```
resa_ec$measures.test
   iter       auc
1     1 0.9250000
2     2        NA
3     3 0.9737336
4     4 0.9827586
5     5 0.9728261
6     6 0.9919218
7     7 0.9972586
8     8 1.0000000
9     9 0.9852941
10   10 0.9937255
```

Currently, the following is printed to the console:

`[Resample] Aggr. Result: auc.test.mean=  NA`

Wouldn't it be good if the aggregation function would do a `na.rm` for convenience? 

I could not find an option in the docs or in `setAggregation()`.

larskotthoff:
Yeah, having that as an option would be useful. Pull requests welcome :)

pat-s:
I would prefer to hardcode it - I see no advantage in a `NA` result of any aggregation function? 

`NA`s are not lost - they are still stored in the `measures` output.

larskotthoff:
Then the user always has to check very carefully. It's better to keep them and give an option to omit to avoid mistakes.

PR #1980, opened 2017-08-15
Add na.rm flag to remove NAs during resample aggregration

pat-s:
#1887 

When the user sets the `na.rm` flag in `resample()`, all error measures will remove `NA`s encountered during runs and return a numeric value instead of `NA`. 

```r
task = sonar.task
  
rdesc = makeResampleDesc("RepCV", folds = 3, reps = 2)
  
lrn = makeLearner("classif.binomial",
                  link = "logit",
                  predict.type = "prob")
  
resa = resample(lrn, task, rdesc, measures = list(auc, mmce),  na.rm = T)
```

### DIFF
--- a/R/Aggregation.R
+++ b/R/Aggregation.R
@@ -64,7 +64,7 @@ NULL
 #'   properties = "req.test",
 #'   fun = function (task, perf.test, perf.train, measure, group, pred) IQR(perf.test))
 #' @export
-makeAggregation = function(id, name = id, properties, fun, na.rm) {
+makeAggregation = function(id, name = id, properties, fun, na.rm = FALSE) {
   assertString(id)
   assertString(name)
   makeS3Obj("Aggregation", id = id, name = name, fun = fun, properties = properties)
--- a/R/Aggregation.R
+++ b/R/Aggregation.R
@@ -54,6 +54,8 @@ NULL
 #'     \item{\code{pred} [\code{\link{Prediction}}]}{
 #'       Prediction object.}
 #'   }
+#' @param na.rm [\code{logical(1)}]\cr
+#'   Should `NA` values be removed?
 #' @seealso \code{\link{aggregations}}, \code{\link{setAggregation}}
 #' @return [\code{\link{Aggregation}}].
 #' @examples
@@ -62,7 +64,7 @@ NULL
 #'   properties = "req.test",
 #'   fun = function (task, perf.test, perf.train, measure, group, pred) IQR(perf.test))
 #' @export
-makeAggregation = function(id, name = id, properties, fun) {
+makeAggregation = function(id, name = id, properties, fun, na.rm) {
   assertString(id)
   assertString(name)
   makeS3Obj("Aggregation", id = id, name = name, fun = fun, properties = properties)
--- a/R/Aggregation.R
+++ b/R/Aggregation.R
@@ -55,7 +55,7 @@ NULL
 #'       Prediction object.}
 #'   }
 #' @param na.rm [\code{logical(1)}]\cr
-#'   Should `NA` values be removed?
+#'   Should `NA` values be removed? Default `FALSE`.
 #' @seealso \code{\link{aggregations}}, \code{\link{setAggregation}}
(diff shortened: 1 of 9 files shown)


R/Aggregation.R
```diff
@@ -62,7 +64,7 @@ NULL
 #'   properties = "req.test",
 #'   fun = function (task, perf.test, perf.train, measure, group, pred) IQR(perf.test))
 #' @export
-makeAggregation = function(id, name = id, properties, fun) {
+makeAggregation = function(id, name = id, properties, fun, na.rm) {
```
larskotthoff: Please set a default value (`FALSE`).

R/checkMeasures.R
```diff
@@ -3,6 +3,13 @@ checkMeasures = function(measures, obj, aggr = NULL) {
     measures = list(getDefaultMeasure(obj))
   } else {
     measures = ensureVector(measures, n = 1L, cl = "Measure")
+    if (any(names(measures) == "na.rm")) {
```
larskotthoff: Why do you allow `na.rm` in the measure names?

larskotthoff:
I've added some comments; please also write a new unit test for this.

R/checkMeasures.R
pat-s: Because `na.rm` is passed within the measures argument in `resample()`. Hence it will be a subobject within the `measures` object. This call finds it, modifies it as needed (moves it to the according `prop` section of each measure) so that follow up functions will work as before. 

`measures = list(auc, mmce, na.rm = T))`

pat-s:
@larskotthoff 
added unit test and set `na.rm = FALSE` by default

larskotthoff:
Ok, thanks. Then my only question is about the piece of code where you're checking whether any of the measure names are na.rm. What's the purpose of that?

pat-s:
@larskotthoff please see [this comment](https://github.com/mlr-org/mlr/pull/1980#discussion_r133387111)

larskotthoff:
Thanks, I missed that. Why not make `na.rm` a separate argument for `resample`? Having it in the list of measures seems counter-intuitive to me.

pat-s:
@larskotthoff mergeable now?

R/performance.R
```diff
@@ -28,10 +28,10 @@
 #' # Compute multiple performance measures at once
 #' ms = list("mmce" = mmce, "acc" = acc, "timetrain" = timetrain)
 #' performance(pred, measures = ms, task, mod)
-performance = function(pred, measures, task = NULL, model = NULL, feats = NULL) {
+performance = function(pred, measures, task = NULL, model = NULL, feats = NULL, na.rm) {
```
larskotthoff: Document the new flag please.

R/checkMeasures.R
```diff
@@ -1,8 +1,11 @@
-checkMeasures = function(measures, obj, aggr = NULL) {
+checkMeasures = function(measures, obj, aggr = NULL, na.rm) {
   if (missing(measures)) {
     measures = list(getDefaultMeasure(obj))
   } else {
     measures = ensureVector(measures, n = 1L, cl = "Measure")
+    for (i in seq_along(measures)) {
+      measures[[i]]$aggr$na.rm = na.rm
+    }
```
larskotthoff: Document that calling `checkMeasures` sets `na.rm` for all the measures please. And there should be a default value of `FALSE` again.

larskotthoff:
Two very small issues, looks good apart from that!

R/checkMeasures.R
pat-s: I've added a small doc for `checkMeasures` because its internal only and had no doc before. Hope its sufficient?

R/performance.R
pat-s: done

R/checkMeasures.R
larskotthoff: Looks good. An informal comment would have been fine as well :)

larskotthoff:
Thanks, merging.

PR #1985, opened 2017-08-21
update ranger learner definitions

dagola:
Resolve todo's and add standard error prediction.


larskotthoff:
Does this require a particular version of ranger?

dagola:
Yes, the current CRAN version 0.8.0. Do I have to add this somewhere?

larskotthoff:
Thanks, merging.

Issue #1988, opened 2017-08-23
batchmark broken


dagola:
Using `batchmark` in the current version of the master results in an error when estimating the resampling error after tuning. 

The error is:
```r
Error in checkMeasures(measures, pred$task.desc, na.rm = na.rm) : 
  argument "na.rm" is missing, with no default
```

Function `benchmark` works fine.

## Example
```r
library(mlr)
library(batchtools)

# Define task ----
data(BreastCancer, package = "mlbench")
df = BreastCancer
df$Id = NULL
task = makeClassifTask(id = "BreastCancer",
                       data = df,
                       target = "Class")

# Define measures ----
measures = list(mlr::auc)

# Define filter parameter set ----
filter_ps = ParamHelpers::makeParamSet(
  ParamHelpers::makeIntegerParam("fw.abs",
                                 lower = 1,
                                 upper = mlr::getTaskNFeats(task)))

# Define tuning control ----
ctrl = mlr::makeTuneControlRandom(maxit = 10)

# Define resampling strategies ----
inner = mlr::makeResampleDesc("CV", stratify = TRUE, iters = 2)
outer = mlr::makeResampleDesc("CV", stratify = TRUE, iters = 2)

# Define random forest learner ----
rf_lrn = mlr::makeLearner("classif.ranger", predict.type = "prob", 
                          config = list(on.learner.error = "warn"))
rf_lrn = mlr::makeImputeWrapper(rf_lrn, classes = list(factor = mlr::imputeConstant("__miss__")))

rf_ps = ParamHelpers::makeParamSet(
  ParamHelpers::makeIntegerParam("num.trees", 
                                 lower = 100L, 
                                 upper = 5000L, 
                                 trafo = function(x) plyr::round_any(x, 100)),
  ParamHelpers::makeNumericParam("mtry.perc", 
                                 lower = 0.05, 
                                 upper = 0.5),
  ParamHelpers::makeIntegerParam("min.node.size",
                                 lower = 10L,
                                 upper = 100L),
  ParamHelpers::makeLogicalParam("replace"))

rf_lrn = mlr::makeTuneWrapper(rf_lrn,
                              resampling = inner,
                              control = ctrl,
                              measures = measures,
                              par.set = c(rf_ps),
                              show.info = TRUE)

bmr = mlr::benchmark(learners = list(rf_lrn), 
                      tasks = task, 
                      resamplings = outer, 
                      measures = measures, 
                      models = FALSE)

[...]

PR #1989, opened 2017-08-23
add na.rm argument

dagola:
resolves issue #1988.


dagola:
I need help here. Some aggregations of measures seem to have a `na.rm` value, others not.

dagola:
Using commit 31ee122, `devtools::test(filter = "base_resample")` throws 13 errors regarding to the new `na.rm` argument. Was #1980 tested?

ja-thomas:
PR #1980 was reverted. If the `na.rm` flag gets implemented correctly this PR here is not needed

PR #1992, opened 2017-08-23
revert na.rm aggregations (PR #1980)

ja-thomas:
This reverts #1980 as it breaks a lot of stuff and should (hopefully) keep everything we did afterward.

The original PR #1980 should be reopened when this is merged.


larskotthoff:
Ok, thanks. I hate Travis even more now.

PR #2001, opened 2017-08-30
Fix factor encoding for "multi:softmax" in predict.classif.xgboost

### LINKED ISSUE #1999: xgboost predicting NAs
I may be misunderstanding something, but this is looking like a bug to me. The xgboost learner is predicting NA values even on the simplest of tasks. Here is a minimal reproducible example. You can copy and paste that into R and the last line will produce `TRUE` showing that there are NA values. Any idea what's happening here?

```r
library(tidyverse)
library(mlr)
library(xgboost)

# Create data frame
data <- data.frame(
  size = sample(x = c(1:4), size = 100, replace = TRUE),
  vehicles = sample(x = c(0:3), size = 100, replace = TRUE),
  workers = sample(x = c(0:3), size = 100, replace = TRUE)
)

# Convert to factors
data <- data %>%
  mutate_at(
    .vars = vars("vehicles", "workers"),
    .funs = funs(as.factor(.))
  )

# create a training task
trainTask <- makeClassifTask(data = data, target = "size")

# Expand factors to dummy vars
trainTask <- createDummyFeatures(trainTask)

## XGBOOST MODEL

# Create learner
xgb_learner <- makeLearner(
  "classif.xgboost",
  predict.type = "response",
  par.vals = list(
    objective = "multi:softmax",
    nrounds = 200
  )
)

# Create model
xgb_model <- train(xgb_learner, task = trainTask)

# Predict on the same training data
result <- predict(xgb_model, trainTask)

# Are any NA?
any(is.na(result$data$response))

```

ja-thomas:
fixes #1999 

Factor level encoding was wrong, added test.

### DIFF
--- a/R/RLearner_classif_xgboost.R
+++ b/R/RLearner_classif_xgboost.R
@@ -120,7 +120,9 @@ predictLearner.classif.xgboost = function(.learner, .model, .newdata, ...) {
     }
   } else { #multiclass
     if (.learner$par.vals$objective  == "multi:softmax") {
-      return(factor(p, levels = cls)) #special handling for multi:softmax which directly predicts class levels
+      p = as.factor(p) #special handling for multi:softmax which directly predicts class levels
+      levels(p) = cls
+      return(p)
     } else {
       p = matrix(p, nrow = length(p) / nc, ncol = nc, byrow = TRUE)
       colnames(p) = cls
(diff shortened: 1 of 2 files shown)


larskotthoff:
Thanks, merging.

PR #2005, opened 2017-09-06
Functional data analysis

smilesun:
I met a problem saying  "reference not a tree in the old pr" so I created this pr, which is the same with fda2

## Open ToDos
The following issues exist with featextract + tune wrappers. 
Somehow defaults are not set when we use a tune wrapper.
- [x] Rethink lower limit for `shift` parameter in multiRes. 0.1 seems very low.
- [x] `bsignal.df` parameter has a weird maximum. I think we should allow `Inf`.
- [x] Make sure we require `df >= differences ` for `FDboost` 
       I think we can make sure this is the case by using 
       `forbidden = expression(df  < differences)` in the paramset.
- [x]  Dataset Adiac, ranger  and multires:  mtry=1; sample.fraction=0.659; res.level=1; shift=0.868
`
Could not train learner classif.ranger.extracted: Error in extractFDAFeatures.data.frame(obj = structure(list(target = structure(c(1L,  :
  feat.method needs to return a data.frame with one row per observation in the original data and equal nrow per column.
`

- [x] @smilesun What does the `Makefile` do? Can we delete it?

- [x] @smilesun We need to make sure, that during tuning hyperpars are correctly set in the methods, i.e. that the hyperpars end up beeing used when tuning sets them. This is not trivial to check.

- [x] Feature extraction with bsignal seems to be very unstable
```
  methods = list("fd" = extractFDABsignal())

  lrn = makeExtractFDAFeatsWrapper("classif.xgboost", feat.methods = methods)

  ps = makeParamSet(makeNumericParam("eta", lower = 0.0001, upper = 0.3))

  lrn = makeTuneWrapper(learner = lrn, resampling = cv2, measure = acc, par.set = ps, control = makeTuneControlMBO(budget = 10L))

train(lrn, subsetTask(gunpoint.task))
```

- [x] Make Travis pass

- [x] Integrate the classif fdboost and fgam into benchmark.

## Closed ToDos
- [x] Change name of scale param in `tsFeatures` extractor to `tsfeats.scale` or something 
       to reduce further name clashes with learner params.
- [x] The fda.usc "kernel" and "metric" params are not correctly parsed from mlr currently.
We need to give the trainfun functions, not characters (learners: classif.fdausc.knn|np|kernel)
- [x] Test whether params passed to featExtractMethods during tuning actually end up beeing used.
- [x] Add a paramset to `DTWKernel` feature extractor.
  Florian: Using rucrdtw package instead. This seems to work.
- [x] Check if the paramsets for the feature extractors work (i.e. params are actually correctly set)
 Add a unittest for this
- [x] Check if feature extraction parameters work during reextract
- [x] dtw package installation problem => Using rucrdtw
- [x] Fourier Feature Extraction yields a single `NaN` value in the first column on **CBF** dataset when extracting `phase`. Find error and fix eventually.
- [x] `getTaskData(... , functionals.as = "matrix")` is currently not used in the multiClassWrapper. 
      We have to add this in order to be able to use classif.FDboost on multiclass datasets. 
- [x] Fix the following error (job 63:66 on Coffee Dataset)
`
Warning in train(learner, task, subset = train.i, weights = weights[train.i]) :
  Could not train learner classif.xgboost.extracted: Error in (function (data, target, col, res.level, shift, curve.lens)  : 
argument "curve.lens" is missing, with no default
`
- [x] Fix the following error (job 79:82 on Coffee Dataset)
`
  Could not train learner classif.ranger.extracted: Error in checkChoice(x, choices, null.ok, fmatch) : 
  argument "refs" is missing, with no default
`
- [x] Fix the following error (job 75:78 on Coffee Dataset)
`
Warning in train(learner, task, subset = train.i, weights = weights[train.i]) :
  Could not train learner classif.ranger.extracted: Error in refund::fpca.sc(Y = data, pve = pve, npc = npc) : 
  argument "npc" is missing, with no default
`

## Optional:
- [x] Add feature extractor from `tsfeatures`package
      Progress: See `fda_lte_tsfeatures` branch

- [x] There are some more classif learners in the fda.usc package, we have not added yet.


smilesun:
@berndbischl , @lbeggel , @pfistfl , this is currently the major pr for FDA, branch name fda_lte

R/RLearner_classif_fdaFDboost.R
```diff
@@ -0,0 +1,116 @@
+#' @export
+makeRLearner.classif.fdaFDboost = function() {
```
pfistfl: Did we decide to call it fdaFDboost instead of FDboost.

R/RLearner_regr_fdaFDboost.R
```diff
@@ -1,7 +1,7 @@
 #' @export
-makeRLearner.regr.FDboost = function() {
+makeRLearner.regr.fdaFDboost = function() {
```
pfistfl: Same with fdaFdboost instead of FDboost

R/RLearner_classif_fdaFDboost.R
```diff
@@ -0,0 +1,116 @@
...
+  makeRLearnerClassif(
+    cl = "classif.fdaFDboost",
```
pfistfl: see above

R/RLearner_classif_fdaFDboost.R
```diff
@@ -0,0 +1,116 @@
...
+      makeIntegerLearnerParam(id = "knots", default = 10L, lower = 1L),  # determine the number of knots of splines, does not matter once there is sufficient number of knots, 30,40, 50 for example
+      makeIntegerLearnerParam(id = "degree", default = 3L, lower = 1L),  # degree of the b-spline
+      makeIntegerLearnerParam(id = "differences", default = 1L, lower = 1L),  # degree of the penalty
+      makeLogicalLearnerParam(id = "bsignal.check.ident", default = FALSE, tunable = FALSE)  # identifiability check by testing matrix degeneracy
+    ),
+    par.vals = list(family = "Binomial"),
+    properties = c("functionals", "numerics", "twoclass"),
+    name = "Functional linear array classification boosting",
+    short.name = "FDboost",
+    note = "Only allow one base learner for functional covariate and one base learner for scalar covariate, the parameters for these base learners are the same. Also we currently do not support interaction between scalar covariates"
```
pfistfl: Only allows one base learner for each functional covariate and one base learner for each scalar covariate, the parameters for these base learners are the same. Also we currently do not support interaction between scalar covariates.

R/RLearner_classif_fdaFDboost.R
```diff
@@ -0,0 +1,116 @@
...
+#' @export
+trainLearner.classif.fdaFDboost = function(.learner, .task, .subset, .weights = NULL, mstop = 100L,
+  knots = 10L, df = 4L, bsignal.check.ident = FALSE, degree = 3L, differences = 1L, Binomial.link = "logit",
+  nu = 0.1, family = "Gaussian", custom.family.definition = NULL, nuirange = c(0, 100), d = NULL, ...) {
+
+  family = switch(family,
+    Binomial = mboost::Binomial(link = Binomial.link),
+    AdaExp = mboost::AdaExp(),
+    AUC = mboost::AUC(),
+    #PropOdds = mboost::PropOdds(nuirange = nuirange, offrange = offrange),
```
pfistfl: Why is this commented out?

R/RLearner_classif_fdaFDboost.R
```diff
@@ -0,0 +1,116 @@
...
+    fd.grids = lapply(fdns, function(name) seq_len(ncol(m[, name])))
+    names(fd.grids) = fdns
+    fdg = setNames(fd.grids, stri_paste(fdns, ".grid"))
+    # setup mat.list: for each func covar we add its data matrix and its grid. and once the target col
+    # also setup charvec of formula terms for func covars
+    mat.list = namedList(fdns)
+    #formula.terms = setNames(character(length = fdns))
+    formula.terms = namedList(fdns)
+    # for each functional covariate
+    for (fdn in fdns) {
```
pfistfl: Can we write a function which we use for both (regr/classif) learners? Do you think that would make sense?

R/RLearner_classif_fdaFDboost.R
pfistfl: Update note for classif and regr

R/RLearner_classif_fdaFDboost.R
smilesun: because this option is not needed in FDboost

R/RLearner_classif_fdaFDboost.R
smilesun: I think having the "fda" prefix for the FDA learners make it a consistent naming convention. Since we have "fdaknn",  if all learners of FDA have the same prefix, we could easily find out all the fda learners, which is better for maintainence

R/RLearner_regr_fdaFDboost.R
smilesun: see reply above

R/RLearner_classif_fdaFDboost.R
smilesun: same

R/RLearner_classif_fdaFDboost.R
smilesun: because we do not need this in FDBoost

R/RLearner_classif_fdaFDboost.R
pfistfl: ```listLearners(properties = "functionals")```
Does the job aswell

R/makeExtractFDAFeatMethodBsignal.R
```diff
@@ -0,0 +1,35 @@
...
+#' For more details refer to \code{\link[FDboost]{bsignal}}.
+#'
+#' @param bsignal.knots [\code{integer}]\cr
+#' The number of knots for bspline.
+#' @param bsignal.df [\code{numeric}]\cr
+#' The effective degree of freedom of penalized bspline.
+#' @return [\code{data.frame}].
+#' @export
+#' @family fda_featextractor
+extractFDABsignal = function(bsignal.knots = 10L, bsignal.df = 3) {
```
pfistfl: capital B or lowercase b for bsignal?

smilesun: well, bsignal is the orginal version but here use Camel case, so it should be upper case. how do you think?

smilesun: solved

R/extractFDAFeatures.R
```diff
@@ -32,24 +33,23 @@
...
-#'   Default is [list] which does nothing.
+#'   Specifying a functional feature multiple times with different extraction methods allows
+#'   for the extraction of different features from the same functional.
+#'   Default is [list()] which does nothing.
 #' @return ([list])
 #'   \item{data ([data.frame] | [Task])}{Extracted features, returns a data.frame when
-#'   given a [data.frame] and a Task when given a Task.}
-#'   \item{desc (`extractFDAFeatDesc`)}{Description object. See description for details.}
+#'   given a data.frame and a Task when given a Task.}
+#'   \item{desc (`extracFDAFeatDesc`)}{Description object. See description for details.}
```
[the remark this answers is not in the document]
smilesun: yes, back ticks still used

pfistfl:
FIXME: Output slot of extractFDAFeatures "data" is called "task" if the returned data is a task. Make this clear or FIX

pfistfl:
FIXME: We currently do not expose the paramsets via the extractFDAWrappers. We have to add those.

pfistfl:
The fda.usc "kernel" and "metric" params are not correctly parsed from mlr currently.
We need to give the trainfun `functions`, not `characters`

pfistfl:
- [ ] Make sure that `extractFDAFeaturesFPCA()` always returns a matrix for low `pve` values

pfistfl:
Look at https://stats.stackexchange.com/questions/249198/from-a-statistical-perspective-fourier-transform-vs-regression-with-fourier-bas

vrodriguezf:
Hi!

I am new into FDA. Does this branch currently implements the use of unequally spaced data points?

Thanks!

smilesun:
@vrodriguezf  Thanks for the interest!  If I understand your question right, we do have a few learners in mlr-FDA has support for unequally spaced data points where you could provide the sample index or sample time for those data point (fdboost for example) but you have to make each functional instance(curve instance) the same length,  which I myself call it functional imputation(the same length requirement is inevitable due to the matrix calculation). Is that what you are asking?

vrodriguezf:
@smilesun Thanks to you for your great work here!

Yes, that is exactly what I was asking. I have multiple functional instances, each of them with different sample times and different lengths. I understand the need of data imputation due to the matrix calculation, but what I was not sure about, just by taking a look at the function `makeFunctionalData`, is the possibility of adding a list of sample indices for each functional feature.

Thanks again!

pat-s:
@pfistfl 	What is the status here?

pfistfl:
This PR is huge and should be merged at some point. 
Merging it would require quite some review though. 
I would try to merge once we have capacity for the reviews.

pat-s:
@pfistfl I merged most conflicts. Let's see what the tests say.

I can make a review once you say that it's ready.
There is so much in this PR that we should try to get it merged - otherwise it might get lost at some point. 

I do not know much about the practical contents but I'll try to do my best - maybe you can try to give a concise summary about all the changes in here in the first post of this PR?

pat-s:
Closing here for now due to inactivity. Please re-open once you are working on this again.
Happy to review!

PR #2007, opened 2017-09-12
Fix 1957, always set check.packages to FALSE

### LINKED ISSUE #1957: listLearners has different defaults for check.packages depending on S3 generic
This should probably be unified.

ja-thomas:
Sets `check.packages=FALSE` for all Methods of `listLearners` and filterLearnerTable which is only used by `listLearners` internally.

Fixes #1957

### DIFF
--- a/R/listLearners.R
+++ b/R/listLearners.R
@@ -28,7 +28,7 @@ getLearnerTable = function() {
   return(tab)
 }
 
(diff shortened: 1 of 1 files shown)


berndbischl:
can you pls check locally and report back?

berndbischl:
i just checked, looks good i guess

PR #2009, opened 2017-09-12
Add faceting argument to plotROCCurves

### LINKED ISSUE #1959: plotROCCurves: why always facetting
i would like to see the curves in 1 plot? to compare?

also the docs say we return a ggvis object.....

ja-thomas:
fixes #1959 

Add `facet.learner` argument, to choose if faceting or color coding should be used for different learners


R/generateThreshVsPerf.R
```diff
@@ -388,28 +392,29 @@ plotROCCurves = function(obj, measures, diagonal = TRUE, pretty.names = TRUE) {
...
+    aes$color = "learner"
+
+  p = ggplot(obj$data, do.call(aes_string, aes)) + geom_path() + labs(x = mnames[1], y = mnames[2])
 
-  if (length(unique(obj$data$learner)) > 1L)
+  if (mlearn && facet.learner)
     p = p + facet_wrap(~ learner)
 
-  if (diagonal & all(sapply(obj$data[, names(measures)], function(x) max(x, na.rm = TRUE)) <= 1))
+  if (diagonal && all(sapply(obj$data[, names(measures)], function(x) max(x, na.rm = TRUE)) <= 1))
```
larskotthoff: Oh no, there's an `sapply` -- didn't we decide we don't want those anymore? (I know that it was there before, but since you're touching this code anyway, could you change it please?)

larskotthoff:
Tests are failing...

jakob-r:
Okay, I have no clue why `tests/run-regr.R` now fails with a segmentation fault :question:

PR #2012, opened 2017-09-12
Fix Issue 1962 Separate Help Pages

### LINKED ISSUE #1962: Separate help pages for each make*Task
If I I'm right, it seems that if you include the target variable(s) when defining the data parameter in makeCostSensTask, you will have data leakage because there is no target parameter to set. 

I would update the docs with something like this: 

data [data.frame]
A data frame containing the features and target variable(s). _**Except for makeCostSensTask where you should NOT include the target variable(s).**_

QuayAu:
fix #1962


R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
+#' @title Create a classification task
+#'
+#' @description
+#' The task encapsulates the data and specifies - through its subclasses -
```
ja-thomas: This does not make sense anymore

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
...
+#' the type of the task.
+#' It also contains a description object detailing further aspects of the data.
+#'
+#' Useful operators are: \code{\link{getTaskFormula}},
+#' \code{\link{getTaskFeatureNames}},
+#' \code{\link{getTaskData}},
+#' \code{\link{getTaskTargets}}, and
+#' \code{\link{subsetTask}}.
```
ja-thomas: Check for specific operators of every task

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
...
+#'
+#' Object members:
+#' \describe{
+#' \item{env [\code{environment}]}{Environment where data for the task are stored.
+#'   Use \code{\link{getTaskData}} in order to access it.}
+#' \item{weights [\code{numeric}]}{See argument. \code{NULL} if not present.}
+#' \item{blocking [\code{factor}]}{See argument. \code{NULL} if not present.}
+#' \item{task.desc [\code{\link{TaskDesc}}]}{Encapsulates further information about the task.}
```
ja-thomas: check for task specific members

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
...
+#' }
+#'
+#' @template arg_id
+#' @param data [\code{data.frame}]\cr
+#'   A data frame containing the features and target variable(s).
```
ja-thomas: create two templates, one with target one without

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
...
+#' @template arg_weights
+#' @template arg_blocking
+#' @template arg_positive
+#' @template arg_fixup.data
+#' @template arg_check.data
+#' @return [\code{\link{Task}}].
+#' @examples
+#' if (requireNamespace("mlbench")) {
+#'   library(mlbench)
+#'   data(BostonHousing)
```
ja-thomas: we dont need that for this example

R/ClusterTask.R
```diff
@@ -1,4 +1,35 @@
-#' @rdname Task
```
ja-thomas: create a family for task

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
+#' @title Create a classification task
```
ja-thomas: titles should end with a dot

R/Task.R
```diff
@@ -40,39 +49,12 @@
...
-#' @param check.data [\code{logical(1)}]\cr
-#'   Should sanity of data be checked initially at task creation?
-#'   You should have good reasons to turn this off (one might be speed).
-#'   Default is \code{TRUE}.
+#' @template arg_costs
+#' @template arg_weights
+#' @template arg_blocking
+#' @template arg_positive
+#' @template arg_fixup.data
+#' @template arg_check.data
```
ja-thomas: we dont need arguments and return values here anymore

R/ClassifTask.R
```diff
@@ -1,5 +1,49 @@
...
+#'
+#'   makeClassifTask(data = iris, target = "Species")
+#'   # an example of a classification task with more than those standard arguments:
+#'   blocking = factor(c(rep(1, 51), rep(2, 300)))
+#'   makeClassifTask(id = "myIonosphere", data = Ionosphere, target = "Class",
+#'     positive = "good", blocking = blocking)
+#' }
 #' @export
-#' @rdname Task
+#' @family make.task
```
ja-thomas: use family task

R/CostSensTask.R
```diff
@@ -1,6 +1,39 @@
+#' @title Create a cost-sensitive classification task.
+#'
+#' @description
+#' The task encapsulates the data and creates a cost sensitive classification task.
+#' It also contains a description object detailing further aspects of the data.
```
ja-thomas: Add a note about label leakage? That the target should be removed

R/MultilabelTask.R
```diff
@@ -1,5 +1,46 @@
...
+#'
+#' For multilabel classification we assume that the presence of labels is encoded via logical
+#' columns in \code{data}. The name of the column specifies the name of the label. \code{target}
+#' is then a char vector that points to these columns.
+#'
+#' Useful operators are: \code{\link{getTaskFormula}},
+#' \code{\link{getTaskFeatureNames}},
+#' \code{\link{getTaskData}},
+#' \code{\link{getTaskTargets}}, and
+#' \code{\link{subsetTask}}.
```
ja-thomas: check for all available operators

R/SurvTask.R
```diff
@@ -1,5 +1,46 @@
...
+#'  the \dQuote{interval2} format.
+#'  See \code{\link[survival]{Surv}} for details.
+#' @return [\code{\link{Task}}].
+#' @examples
+#' data(lung, package = "survival")
+#' lung$status = (lung$status == 2) # convert to logical
+#' makeSurvTask(data = lung, target = c("time", "status"))
 #' @export
+#' @family make.task
+
```
ja-thomas: No empty line here

R/Task.R
```diff
@@ -5,92 +5,18 @@
...
-#' \item{weights [\code{numeric}]}{See argument. \code{NULL} if not present.}
-#' \item{blocking [\code{factor}]}{See argument. \code{NULL} if not present.}
-#' \item{task.desc [\code{\link{TaskDesc}}]}{Encapsulates further information about the task.}
+#' \itemize{
+#'  \item{For classification see}{ \code{\link{makeClassifTask}}}
+#'  \item{For regression see}{ \code{\link{makeRegrTask}}}
+#'  \item{For survival analysis see}{ \code{\link{makeSurvTask}}}
+#'  \item{For clustering see}{ \code{\link{makeClusterTask}}}
+#'  \item{For cost sensitive classification see}{ \code{\link{makeCostSensTask}}}
+#'  \item{For multilabel classification see}{ \code{\link{makeMultilabelTask}}}
```
ja-thomas: describe instead of itemize

QuayAu: i think "itemize" looks better here

berndbischl:
ok, not so great :)

the trick with these things is to make stuff 
a) maintainable
b) without much copy-paste

usually achieving b) helps with a) ;-)

here are a few options:
a) docu chunks one can define with rox templates
b) some (!) stuff one can document in the parent class
c) we can use doc families to auto-gen links-lists

giuseppec:
@berndbischl as we have many getters that are applicable to all tasktypes I used a description template with template variables instead of using @family to auto-gen links-list.
Or do you prefer to introduce a family for each tasktype?

giuseppec:
`makeSurvTask` does not have a `censoring` arg, however, there is a `@param censoring` in the description. I will remove this.

larskotthoff:
Looks good to me. Any objections to merging?

mb706:
[The roxygen2 vignette](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html) says
> Because it will appear as “Other :”, the @family name should be plural (i.e., “model building helpers” not “model building helper”)

So it would make more sense to have
```R
#' @family task types
```
or something similar.

pat-s:
This was solved in d7546ad296182718571ce29c67cb43844fc154a0.

PR #2013, opened 2017-09-12
fix tunethreshold with MBO, fixes #2002

### LINKED ISSUE #2002: tuneWrapper bug: threshold tuning does not work for binary classification using mbo
when using a tuneWrapper for tuning binary classification with mbo, enabling threshold tuning throws an error. See example below: 

```r
lrn = makeLearner("classif.randomForest", predict.type = "prob")
ps = makeParamSet(
  makeIntegerParam("ntree", lower = 20L, upper = 50L)
)
ctrl = makeTuneControlMBO(budget = 10L, tune.threshold = TRUE)
rdesc = makeResampleDesc("Holdout")
lrn = makeTuneWrapper(lrn, rdesc, mmce, ps, ctrl, show.info = TRUE)
mod = train(lrn, sonar.task)
```

throws the following error:
```r
Error in train(lrn, sonar.task) : 
  Assertion on 'predict.threshold' failed: Must be of type 'numeric', not 'NULL'.
```
However, tuning with ````tune.threshold = FALSE```` or tuning without the wrapper via ````tuneParams```` is working.

this is related to issue #1675

berndbischl:
this fixes the problem reported here: #2002 

what did i do:

- mbo now always pics the best point from the the optstate and ignores option "final.method". this is consistent with other tuners and simplified code, and fixed the issue automatically

- removed option "keep-mbo.result". either we support this for all tuners or for none. 

- cleaned up testcode code which looked ugly when mbo was not on cran

- unit tested original problem


berndbischl:
@jakob-r  @ja-thomas pls review

jakob-r:
@berndbischl please assign reviews on the right hand side -->

jakob-r:
IMHO It doesn't make sense to only allow "best.true.y" for a noisy tuning.
Why do we have to be consistent with other tuners in this extend? To a certain degree each tuner is inconsistent and boiling it down just means cutting features.

ja-thomas:
my 5ct.

-  Consistency between all tuners is not really given, every package in the background offers different things. It might be a good idea at one point to have a generic way on "how to do final evaluation after tuning".

- I think it is ok to remove that option since almost everybody does not really care when using this and if one really cares mlrMBO can be used directly to achieve full flexibility. 


So imo. fix this for now and open an issue for a nicer and general solution for handling final evaluation with tuning, but the bug not being able to tune thresholds seems worse imo.

berndbischl:
> my 5ct.

it's 2cents :) or are your thoughts worth more?

berndbischl:
> @berndbischl please assign reviews on the right hand side -->

will do (in future)

jakob-r:
@ja-thomas Are you confusing `final.method` and `final.evals`?
I could agree to disallow `final.evals` as a part of `mlr` but I really don't see why to dissallow `final.method = "best.predicted". I don't know if it's a better option then `"best.true.y"` but I also don't know if the opposite is true and that's why we should not disallow it. It could happen that the `"best.true.y"` value is just an outlier and the surrogate might save us from selecting this on as a final proposal.

berndbischl:
> I could agree to disallow final.evals as a part of mlr but I really don't see why to dissallow final.method = "best.predicted"

the reason why i did this is clearly stated in the issue.... 
the old code, that allowed all options for final.method, in mbo, throwed always an error when you request tune.threshold = TRUE. thats bad. the whole mbo-code, to extract the final point, after mbo was done, was also not great.

so in conclusion: if you want more options for final.method, tell me how to enable threshold tuning with them. because i checked and didnt easily see how to do that

berndbischl:
PS: 
you kinda need to read evalOptState for that

berndbischl:
@jakob-r i think you were right on insisting here.  i see now how we can keep "final.method" and i can still fix the bug - i hope.
thx

berndbischl:
done. will merge when travis is done. @Coorsaa is waiting for this

ja-thomas:
@jakob-r: yup you're right, confused the two settings

jakob-r:
Merge?

larskotthoff:
Merge indeed. Thanks.

Issue #1986, opened 2017-08-21
Predict probs for surv.randomForestSRC"


edvbb:
In 
`makeLearner("surv.randomForestSRC",  predict.type = "prob")`
`Error in setPredictType.Learner(learner = wl, predict.type = predict.type) : 
  Trying to predict probs, but surv.randomForestSRC does not support that!`
and we get the model with responses

Can we predict risks or survival function with individual prob of survival?
thanks!

berndbischl:
@mllg

PhilippPro:
Probably you can get the probabilities like in the ibrier measure: 
```

targets = getTaskTargets(task)
    tn = getTaskTargetNames(task)
    f = as.formula(sprintf("Surv(%s, %s) ~ 1", tn[1L], tn[2L]))
    newdata = getTaskData(task)[model$subset, ]
    max.time = extra.args$max.time %??% max(newdata[[tn[1L]]])
    grid = seq(0, max.time, length.out = extra.args$resolution)

    probs = predictSurvProb(getLearnerModel(model, more.unwrap = TRUE), newdata = newdata, times = grid)
```
Does it solve your problem?

PR #2018, opened 2017-09-15
Survival integrated brier score

PhilippPro:
Done with @mllg. Currently only works for coxph and randomForestSRC. The function `pec::predictSurvProb` has to be extended to be suitable for more models. (Or enable probability predictions somehow, #1986)

I tested it with our three survival tasks and it works. Maybe put a warning somewhere if another model is used. (Where?)

Probably more testing code needed to fulfill Bernds wishes.


edvbb:
I updated the package 
`devtools::install_github("mlr-org/mlr")`
and tried 
`lrn = makeLearner("surv.coxph", id = "cph", predict.type = "prob")`
but
```
Error in setPredictType.Learner(learner = wl, predict.type = predict.type) : 
  Trying to predict probs, but cph does not support that!

```
Or this works just  with `pec` ?

PhilippPro:
At the moment (in this branch version) it only works for calculating the ibrier score.
You can get the Survival probabilities manually from your model with `model` being your trained mlr model:
`probs = pec::predictSurvProb(model$learner.model, newdata = newdata, times = grid)`

R/measures.R
```diff
@@ -1438,6 +1438,34 @@ ibrier = makeMeasure(id = "ibrier", minimize = TRUE, best = 0, worst = 1,
...
+  fun = function(task, model, pred, feats, extra.args) {
+    requirePackages(c("survival", "pec"))
+    targets = getTaskTargets(task)
+    tn = getTaskTargetNames(task)
+    f = as.formula(sprintf("Surv(%s, %s) ~ 1", tn[1L], tn[2L]))
+    newdata = getTaskData(task)[model$subset, ]
+    max.time = extra.args$max.time %??% max(newdata[[tn[1L]]])
+    grid = seq(0, max.time, length.out = extra.args$resolution)
+    probs = pec::predictSurvProb(model$learner.model, newdata = newdata, times = grid)
+    # this function is only suitable for coxph and randomForestSRC at the moment!
```
mllg: So this only works for CoxPH and rfSRC? If so, this needs to be documented in roxygen.

PhilippPro: It only works for methods, for which pec::predictSurvProb works. Probably we should write it in the note, so the user can see it directly in the tutorial. Otherwise users are confused, when it does not work for certain methods. 

We have currently a master student, who is making a benchmark for several survival methods (also some currently not in mlr) and (TCGA) datasets and he is creating learners as well as predictSurvProb function for each learner, so it will be extended.

PhilippPro: Yes, there is a pull request now (#2410).

PR #2020, opened 2017-09-21
Growing and Fixed Window Cross Validation

SteveBronder:
I'm breaking down the forecasting branch into smaller commits. This is a commit for the growing and fixed window cross validation techniques


man/calculateROCMeasures.Rd
```diff
@@ -80,3 +80,5 @@ Other performance: \code{\link{ConfusionMatrix}},
   \code{\link{performance}}, \code{\link{setAggregation}},
   \code{\link{setMeasurePars}}
 }
+\concept{performance}
```
SteveBronder: For some reason these are being added? Is this because I am on the newest version of roxygen or is this an old thing?

PR #2028, opened 2017-10-09
Add support for missings in h2o glm & deeplearning

ledell:
This PR adds the "missings" property to H2O GLM and Deep Learning classification and regression functions.  Both H2O GLM and Deep Learning support missing values by default using mean imputation.  No action is required by the user, missings are handled automatically.

For the GLM, it was originally added in this [PR](https://github.com/mlr-org/mlr/pull/1710), but there was confusion about whether it was supported, so then it was removed [here](https://github.com/mlr-org/mlr/commit/7a18393cf82615bb6d1f1036400f6056e1cba78d).


berndbischl:
thank you!

PR #2041, opened 2017-10-19
rewrite of generatePartialDependenceData and associated functions

zmjones:
so i've written another package called mmpf (which is on cran and is stable) that should speed up this function by at a minimum about 50%, and potentially much more in other circumstances. this effect is especially consequential if you use any of the derivative estimation. the core of it is now totally working according to the tests, which is why i've pushed it up here. i made as few api changes as i could. it is a bit more general now i think.

i still have to fix-up the main plot function plotPartialDependence.

what are people's feelings about the ggvis functions? ggvis hasn't seen any development from what i can see in a while.

the functional anova stuff i am going to rewrite as well. i'm part of the way done with another package fanova which will provide similar speedups and generality to this.

i'm going to improve the docs and after i finish this i'll rewrite/modify the appropriate tutorial stuff.

and sorry for being absent so long!


larskotthoff:
Great, thanks for doing this! Looks like you need to add mmpf as a dependency.

Good point about ggvis. I would deprecate the ggvis functions and not change them.

zmjones:
I can't leave it in suggests and just check for its installation like with learners? All mmpf imports is data.table and checkmate so it wouldn't be bad to import, but another point of irritation of course.

larskotthoff:
Then you have to prefix the functions you use with `mmpf::`.

zmjones:
sure. you guys are cool with me doing that?

larskotthoff:
If you're planning on using it in other places, it's probably better to put it in imports.

R/generatePartialDependence.R
```diff
@@ -246,78 +190,213 @@ generatePartialDependenceData = function(obj, input, features,
...
-      else
-        doIndividualPartialDependence(out, td, nrow(data), rng, target, x, centerpred)
-    })
-    out = setDF(rbindlist(out, fill = TRUE))
+  # estimation using mmpf::marginalPrediction
+  # derivative estimation using numDeriv::grad or numDeriv::jacobian
+  if (derivative) {
+    ## args = list(model = obj, data = data, uniform = uniform, aggregate.fun = fun,
+    ##   predict.fun = getPrediction, n = n, ...)
+    points = sapply(features, function(x) uniformGrid(data[[x]], n[1]), simplify = FALSE)
```
giuseppec: If `derivative = TRUE` and `uniform = FALSE` we compute the derivatives for the points on a uniform grid, right?

R/generatePartialDependence.R
```diff
@@ -43,57 +44,34 @@
...
+#' @param uniform [\code{logical(1)}]\cr
+#'   Whether or not the prediction grid for the \code{features} is a uniform grid of size \code{n[1]} or sampled with
+#'   replacement from the \code{input}.
+#'   Default is \code{TRUE}.
+#' @param n [\code{integer21)}]\cr
+#'   The first element of \code{n} gives the size of the prediction grid created for each feature.
+#'   The second element of \code{n} gives the size of the sample to be drawn without replacement from the \code{input} data.
+#'   Setting \code{n[2]} less than the number of rows in the \code{input} will decrease computation time.
+#'   The default for \code{n[1]} is 10, and the default for \code{n[2]} is the number of rows in the \code{input}.
+#' @param ... additional arguments to be passed to \code{\link{mmpf}{marginalPrediction}}.
```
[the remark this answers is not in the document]
zmjones: yep

R/generatePartialDependence.R
zmjones: ah yea. i guess that ought to change....

R/generatePartialDependence.R
```diff
@@ -246,78 +190,213 @@ generatePartialDependenceData = function(obj, input, features,
...
-    })
-    out = setDF(rbindlist(out, fill = TRUE))
+  # estimation using mmpf::marginalPrediction
+  # derivative estimation using numDeriv::grad or numDeriv::jacobian
+  if (derivative) {
+    ## args = list(model = obj, data = data, uniform = uniform, aggregate.fun = fun,
+    ##   predict.fun = getPrediction, n = n, ...)
+    points = sapply(features, function(x) uniformGrid(data[[x]], n[1]), simplify = FALSE)
+    if (length(target) == 1L & !multi.fun) {
+      deriv = lapply(seq_along(points), function(i) {
```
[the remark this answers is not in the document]
zmjones: we can but i did a bunch of analysis when i wrote mmpf and found that almost all of the time is spent calling predict. so it is much faster to make fewer calls to predict.

R/generatePartialDependence.R
```diff
@@ -43,57 +44,34 @@
...
-#'   the number of (possibly non-unique) values resampled. If \code{resample = NULL} it defines the
-#'   length of the evenly spaced grid created.
-#' @param range [\code{list}]\cr
-#'   The range of values of the feature you would want the partial plots on - passed as a numeric list
-#' @param ... additional arguments to be passed to \code{\link{predict}}.
+#' @param uniform [\code{logical(1)}]\cr
+#'   Whether or not the prediction grid for the \code{features} is a uniform grid of size \code{n[1]} or sampled with
+#'   replacement from the \code{input}.
+#'   Default is \code{TRUE}.
+#' @param n [\code{integer21)}]\cr
```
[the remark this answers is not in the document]
zmjones: the docs aren't updated so this is just wrong. 

yea n[1] is what gridsize used to be

n[2] is used even when uniform is false. it controls how much of the training data is sampled from. so you can reduce it arbitrarily to speed things up. used to n[2] was fixed to nrow(data)

R/generatePartialDependence.R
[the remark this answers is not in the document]
zmjones: i should obviously fix the docs here

zmjones:
ping @larskotthoff or @berndbischl

zmjones:
or @giuseppec!

DESCRIPTION
```diff
@@ -53,7 +53,8 @@ Imports:
     parallelMap (>= 1.3),
     survival,
     utils,
-    XML
+    XML,
+    mmpf
```
larskotthoff: Packages here should be in alphabetical order.

R/generatePartialDependence.R
```diff
@@ -186,6 +124,9 @@ generatePartialDependenceData = function(obj, input, features,
     assertDataFrame(data, col.names = "unique", min.rows = 1L, min.cols = length(obj$features) + length(td$target))
     assertSetEqual(colnames(data), c(obj$features, td$target), ordered = FALSE)
   }
+  
+  if (is.na(n[2]))
```
larskotthoff: What's n[2] here? Any way to do this without a magic number?

R/generatePartialDependence.R
```diff
@@ -246,378 +172,158 @@ generatePartialDependenceData = function(obj, input, features,
...
+            x = melt(x, id.vars = feature, variable.name = "n", value.name = target)
+            x[, n := stri_replace(n, "", regex = target)]
+            setnames(x, c(feature, if (individual) "n" else "Function", target), names(x))
+          } else {
+            x
+          }
+        })
+    }
+  } else {
+    points = sapply(features, function(x) uniformGrid(data[[x]], n[1]), simplify = FALSE)
```
larskotthoff: We try to not use `sapply` -- could this be `lapply` instead?

tests/testthat/test_base_generatePartialDependence.R
```diff
@@ -1,158 +1,109 @@
...
   nfeat = length(dr$features)
   nfacet = length(unique(regr.df[["chas"]]))
   n = getTaskSize(regr.task)
-  expect_that(max(dr$data$lstat), equals(40.))
-  expect_that(min(dr$data$lstat), equals(1.))
-  expect_that(nrow(dr$data), equals(gridsize * nfeat))
+  expect_that(nrow(dr$data), equals(m[1] * nfeat))
   expect_true(all(dr$data$medv >= min(regr.df$medv) | dr$data$medv <= max(regr.df$medv)))
-
+
```
larskotthoff: Ahhh, single space on line.

DESCRIPTION
zmjones: ok will do

R/generatePartialDependence.R
zmjones: n[2] is the number of points to sample from the training data to average over. that should be in the documentation for n but will double check

R/generatePartialDependence.R
zmjones: the reason i am using it is so that i get a named list. is there an alternative way to do this that isn't terribly verbose?

R/generatePartialDependence.R
larskotthoff: Maybe pull that out into a separate variable to make it more obvious what's going on?

R/generatePartialDependence.R
larskotthoff: There's a few options here: https://stackoverflow.com/questions/9469504/access-and-preserve-list-names-in-lapply-function

R/generatePartialDependence.R
zmjones: i think it makes sense as is as both elements have the same function. it is a resolution parameter

zmjones:
hey @larskotthoff what is up with the lintr? can i not use data.table j expression assignment with := and so on?

larskotthoff:
Ah, the lintr isn't set up for this. @mb706 could you have a look please?

mb706:
Check out #2137 (merge the PR or just merge the branch locally). The trainling whitespace is a true positive and should be removed. I don't get the
```R
​  se = Function = Class = patterns = NULL
```
line, but if it is necessary, just suffix it with a `  # nolint` to ignore the single line.

zmjones:
@mb706 that is because r check views the nse j expressions where i'm assigning to a column as global variables that haven't been defined and throws an error

zmjones:
@mb706 is this still something i need to change? these are column names

mb706:
> suffix it with a ` # nolint` to ignore the single line.

zmjones:
thanks

zmjones:
done @larskotthoff

larskotthoff:
Thanks, merging. Could you update the tutorial as well please?

In the future, please don't commit .Rd files.

zmjones:
ah yea sorry. and yes i will work on it now

PR #2045, opened 2017-10-28
Added @family filter to listFilterMethods (issue #1676)

### LINKED ISSUE #1676: doc of listFilterMethods (and maybe some others) should have '@family filter'
I believe it would be good for `listFilterMethods` to show up in the list of 'Other filter' at the end of help pages of other filter related things.

AlexEngelhardt:
Just a one-liner added to Filter.R (and five respective .Rd files). 
This is my first pull request ever, so I'm happy to receive hints if I did something wrong.
Closes #1676

### DIFF
--- a/R/DownsampleWrapper.R
+++ b/R/DownsampleWrapper.R
@@ -41,8 +41,13 @@ makeDownsampleWrapper = function(learner, dw.perc = 1, dw.stratify = FALSE) {
 #' @export
 trainLearner.DownsampleWrapper = function(.learner, .task, .subset = NULL, .weights = NULL,
   dw.perc = 1, dw.stratify = FALSE, ...) {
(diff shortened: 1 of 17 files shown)


giuseppec:
We never push the. Rd files as they are auto-generated by Travis anyway. But should be OK in this case. Did you find other "filter" functions where adding @family would also make sense?

AlexEngelhardt:
Oh, alright.
There are two other functions with 'filter' in their name and no '@family filter' in them: makeFilter and getBMRFilteredFeatures. Should I add them as well?

AlexEngelhardt:
Three commits: 
- I deleted the .Rd modifications again
- I added "@family filter" to listFilterMethods() and now additionally to makeFilter()
- An update merge from your master

Issue #2047, opened 2017-10-29
makeUndersampleWrapper() does not work properly with weights


banser:
### Problem
makeUndersampleWrapper() does not work properly if there are weights specified. This is independent of the specified learner and data.
Very similar to #838

### Example
```r
library("mlr")

d <- getTaskData(sonar.task)
tsk <- makeClassifTask(data = d, target = "Class", weights = round(d$V1*10))
lrn <- makeUndersampleWrapper("classif.ctree", usw.rate = 0.5)
t <- train(lrn, tsk)
```
### Error:
Error in train(lrn, tsk) : 
  Assertion on 'weights' failed: Must have length 153, but has length 208.

Issue #838, opened 2016-04-18
makeDownsampleWrapper() does not work properly with weights


bkomboz:
**Problem**
`makeDownsampleWrapper()` does not work properly if there are weights specified. This is independent of the specified learner and data. 

**Minimal example**
```
library("mlr")

d <- getTaskData(sonar.task)
tsk <- makeClassifTask(data = d, target = "Class", weights = d$V1)
lrn <- makeDownsampleWrapper("classif.ctree", dw.perc = 0.5, dw.stratify = TRUE)
t <- train(lrn, tsk)

```
**Error**
```
Error in train(lrn, tsk) : 
  Assertion on 'weights' failed: Must have length 103, but has length 208
Timing stopped at: 0.012 0 0.013 
```
**Session Information**
```
R version 3.2.4 Revised (2016-03-16 r70336)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.4 LTS

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=de_DE.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=de_DE.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] mlr_2.8          ParamHelpers_1.7 ggplot2_2.1.0    BBmisc_1.9       fortunes_1.5-2  

loaded via a namespace (and not attached):
 [1] parallelMap_1.3   Rcpp_0.12.4       compiler_3.2.4    plyr_1.8.3        tools_3.2.4       digest_0.6.9     
 [7] lattice_0.20-33   gtable_0.2.0      checkmate_1.7.3   shiny_0.13.2      DBI_0.3.1         parallel_3.2.4   
[13] mvtnorm_1.0-5     coin_1.1-2        stringr_1.0.0     dplyr_0.4.3       ggvis_0.4.2       stats4_3.2.4     
[19] grid_3.2.4        R6_2.1.2          survival_2.38-3   multcomp_1.4-4    TH.data_1.0-7     reshape2_1.4.1   
[...]

PR #2049, opened 2017-11-02
Fix issue #2047 and refix #838

jakob-r:
This fixes the issue in #2047.
Furthermore the fix for #838 was not totally correct in the handling of weights if passed in the `train()` call.

I don't know if it's specified somewhere but if weights are passed alongside with subset I suppose the weights go with the subset vector? At least the test by @berndbischl for issue #838 indicates that.


jakob-r:
thanks

PR #2058, opened 2017-11-09
extend ranger hyperpars

giuseppec:
1) `respect.unordered.factors` has now three options ` c("ignore", "order", "partition")`. It used to be `TRUE` (== `order`) and `FALSE` (== `ignore`). 
2) the `extratrees` splitrule was added and its additional hyperparameter `num.random.splits`.


larskotthoff:
Thanks, merging.

Issue #2057, opened 2017-11-09
Legend issue in plotLearnerPrediction when data points of one class are all misclassified


philipvk:
When using plotLearnerPrediction I noticed some problems with the legend. In the case all data points of one class are misclassied the legend becomes separated. This is caused by that misclassified data points don't have a geom_tile color in the plot. I constructed an extreme example:

```r
data(iris)
iris[2, "Species"] <- "versicolor"
iris[3:150, "Species"] <- "virginica"
task <- makeClassifTask(data = iris, target = "Species")
g = plotLearnerPrediction(makeLearner("classif.rpart"), task)
g
```
In this example the virginica pink color should be on top of the square virginica in the legend. Instead it is now separated with less clear black boxes.

Below is a better example plot. In this plot all data points belonging to group 2 are misclassified. This results a separated and less clear legend. Instead I would like that the legend stays the same, and only the group that is misclassified will be separated from the rest of the legend. 

![models_qda_3](https://user-images.githubusercontent.com/8512559/32605429-a76fe2a6-c551-11e7-9c30-be9eaac32113.png)

> R version 3.4.2 (2017-09-28)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

> Matrix products: default

> locale:
[1] LC_COLLATE=Dutch_Netherlands.1252  LC_CTYPE=Dutch_Netherlands.1252   
[3] LC_MONETARY=Dutch_Netherlands.1252 LC_NUMERIC=C                      
[5] LC_TIME=Dutch_Netherlands.1252    

> attached base packages:
 [1] grid      parallel  stats4    stats     graphics  grDevices utils     datasets  methods  
[10] base     

> other attached packages:
 [1] ggfortify_0.4.1            splatter_1.2.0             scater_1.6.0              
 [4] tidyr_0.7.2                Rtsne_0.13                 RColorBrewer_1.1-2        
 [7] mlr_2.11                   ParamHelpers_1.10          MASS_7.3-47               
[10] gridExtra_2.3              ggplot2_2.2.1              dynamicTreeCut_1.63-1     
[13] dplyr_0.7.4                data.table_1.10.4-3        SingleCellExperiment_1.0.0
[16] SummarizedExperiment_1.8.0 DelayedArray_0.4.1         matrixStats_0.52.2        
[19] Biobase_2.38.0             GenomicRanges_1.30.0       GenomeInfoDb_1.14.0       
[...]

PR #2061, opened 2017-11-14
fix for constant predictions

ja-thomas:
Don't drop missing levels in the geom_tile call.

Fixes #2061


larskotthoff:
Thanks. merging. What issue does this fix though?

larskotthoff:
Ah ok, #2057.

PR #2064, opened 2017-11-20
sort filter values data.table way

jakob-r:
Somehow after updating my packages the tutorial threw an error.

```r
d = generateFilterValuesData(iris.task, method = "anova.test")
plotFilterValues(d)

```

todo: check if this is already covered by the tests


jakob-r:
This error is not cought by any tests. will add test

jakob-r:
I marked this blocking because it is needed in the mlr-tutorial.
Travis fails because of batchmark? Would suggest merge nonetheless.

Issue #1515, opened 2017-02-13
Can't predict on randomForest when test set contains NA's in features


florianfendt:
I don't know if this is a bug in some sort or if I'm overlooking something, but this baffled @ja-thomas and me a bit this morning.
Consider a simple case where you have a missing value somewhere in your test set like in this example:

```r
lrn.rf = makeLearner("classif.randomForest")
mod = train(lrn.rf, iris.task)
test.df = getTaskData(iris.task)
test.df[1L, 1L] = NA
```
mlr then throws an error when you try to predict on this set, randomForest's predict method doesn't though:
```r
# throws error: row names contain missing values
predict(mod, newdata = test.df)
# if I'm directly using the predict method from randomForest it works
predict(mod$learner.model, test.df)
```

I tried printing out `.newdata` in `predictLearner.classif.randomForest` to see if we do sth unwanted with the data.frame before sending it to the learner's predict method but row names / str etc. looks fine.
Any ideas?

larskotthoff:
This sounds like a bug. Could you make a unit test that reproduces it please?

berndbischl:
before you produce a unit test:
please really show the produced output in such cases

berndbischl:
> This sounds like a bug. Could you make a unit test that reproduces it please?

why are you asking him that? he already kinda posted that test....?

the problem is more: what happens here?

berndbischl:
the real problem seems to be:
mlr claims that the rf does not handle missing values. so if you would put the NA into the task data, 
neither training NOR prediction would work, and you would get a meaningful error message.

but here the prediction data frame is directly handled. this also links to the issue 
https://github.com/mlr-org/mlr/issues/1499

we should probably create the task description internally and sanity check it.

QuayAu:
This problem does not only occur with the random forest. I ran every available learner on this problem and here is what i found out:

```
test.df = getTaskData(iris.task)
test.df[1L, 1L] = NA
learners = listLearners(obj = "classif", properties = "multiclass")$class
res = lapply(learners, function(x) {
  mod = train(makeLearner(x), iris.task)
  tryCatch(predict(mod, newdata = test.df), error = function(e) conditionMessage(e))
  }
)
```

So the problem here is, that only the test set contains missing values, and some learners support that, others don't. Here is a small summary of the results:


|learner                        |predict                                            |
|:------------------------------|:--------------------------------------------------|
|classif.bdk                    |setosa                                             |
|classif.boosting               |setosa                                             |
[...]

PR #2068, opened 2017-11-28
add check for missings in newdata

QuayAu:
Fix for #1515

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -45,7 +45,6 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
   model = object
   learner = model$learner
   td = model$task.desc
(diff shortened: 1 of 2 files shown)


mb706:
Usually the prediction made for one row of data should not depend on predictions made for other data rows, or be even dependent on the presence or absence of other data rows. I would suggest to extend this logic here: Don't throw an error if `NA`s are present, and instead predict `NA`s for the rows that contain `NA`s.

In practice, if a `Learner` does not support `NA`s:
1) Remove data rows that contain `NA`s from input data
2) Perform prediction on this reduced data set
3) Insert `NA`s in the predictions returned by the `Learner` at the position of rows that were remove in 1.

pat-s:
Superseeded by #2099

PR #2070, opened 2017-11-28
travis should notice build errors

### LINKED ISSUE #2069: Travis will be green if the build process failed
We had this issue already some time ago. If a PR brakes the building process no unit tests are run (obviously) but **travis will be green**.

see here for example: https://travis-ci.org/mlr-org/mlr/jobs/308349080
The only way to see this is when looking at the time (travis only took ~3 minutes for building stage 3).

mb706:
Fix #2069 (I think)


mb706:
This should do the trick. If anyone wants to trigger the kind of build failure that #2069 talks about for testing, feel free.

larskotthoff:
Could you do this in this branch please so we can see if it works before merging?

larskotthoff:
Great, that fixes it. Merging.

Issue #1700, opened 2017-03-17
Multilabel Learner IDs not unique


QuayAu:
Multilabel Learner IDs are identical for different Wrappers. When benchmarking, we get an error.

```{r}
binary.learner = makeLearner("classif.rpart")
lrnCC = makeMultilabelClassifierChainsWrapper(binary.learner)
lrnBR = makeMultilabelBinaryRelevanceWrapper(binary.learner)
getLearnerId(lrnCC)
"multilabel.classif.rpart"
getLearnerId(lrnBR)
"multilabel.classif.rpart"
```

collinleiber:
We will work on this issue (Kunlan, Collin, Hermann)

PR #2071, opened 2017-11-29
Add different Ids for Multilabel Wrappers

collinleiber:
#1700 

- Changed Ids
- Used getLearnerId(), getLearnerPackages(), getLernerType() instead of learner$...
- Used stri_paste() instead of paste()
- Added test to check Ids


larskotthoff:
Thanks, merging.

Issue #1057, opened 2016-08-03
subsetTask: should also allow to select response labels and classes


berndbischl:
at least for multilabel we need to be able to select response label columns here

christophM:
see file R/Task_operators

likunlan:
working on this issue (Kunlan, Collin, Hermann)

jakob-r:
@christophM Please always use permalinks because now your link highlights an empty line.

PR #2076, opened 2017-12-01
add labels to allow selection of labels in subsetTask

likunlan:
#1057


larskotthoff:
Maybe I'm missing something here, but where does this do the actual subsetting?

likunlan:
@larskotthoff   i create an argument in susetTask()  which named Labels. With the argument,  you are able to get a subtask with any  response label columns.  For example, in unit test , i only choose columns 1-4  as the response Target.

jakob-r:
Ok. So this is about multilabel learning?
I think nonetheless we should stick to the name "targets". From a first sight this only redefines what the target is and the other columns then will become predictors?

mb706:
As far as I understand #1057 it is not supposed to change the *target column names*, but to *subset* the task by removing the columns of the labels not listed in the `labels` argument of `subsetTask`. This would only work for `"multilabel"` type Tasks.

Another option, for `"classif"` tasks, would be to remove all *rows* that have classes not in the requested labels. As I understand the concept of `"multilabel"`, there should be rows allowed that have only `FALSE` target values, so this wouldn't work for `"multilabel"`.

It would look somewhat like (didn't check this):
```R
td = getTaskDesc(task)
if (!is.null(labels)) {
  if (td$type %nin% c("multilabel", "classif"))
    stop("labels only for classif / multilabel")
  assertSubset(labels, td$class.levels)
  if (td$type == "classif") {
    lblselect = target$env$data[[td$target]] %in% labels
    if (is.logical(subset)) {
      subset = subset & lblselect
    } else if (!is.null(subset)) {
      subset = intersect(subset, which(lblselect))
    } else {
      subset = lblselect
    }
  }
  task$task.desc$class.levels = labels
  # +++ plus whatever other changes need to be done in the task.desc
}
newdata = getTaskData(task, subset, features, functionals.as = "matrix")
if (!is.null(labels) && td$type == "multilabel") {
  drop.labels = setdiff(td$class.levels, labels)  # note using non-updated 'td' variable here
  newdata = dropNamed(newdata, drop.labels)
}
changeData(task, newdata, getTaskCosts(task, subset), task$weights)
[......]
```

A different idea is to change the `getTaskData` function to accept a `targets` option. Then the multilabel target column subsetting call would just hand over the `labels` parameter to that one. `subsetTask` would still need to take care of the `task.desc`, and the whole `"classif"` business, however.

pat-s:
This PR is not mergeable in its current state.

- Conditions for other tasks than "multilabel" are needed
- `"labels"` might not the best naming
- improvements of the param desc in the help 

Thanks for contributing @likunlan. You are still welcome to finish this PR here if you want and reopen.

Issue #2066, opened 2017-11-26
Setting hyperparameter "prior" in classif.lda learner


ptl93:
Hello, when I try to set the hyperparamter priori for the linear discrimant analysis which aims to do binary or multiple classification I get the error that the hyperparameter is not included.

I am using the current GitHub Version of mlr (reinstalled today, 26.November 2017)

```r
sessionInfo()
```
R version 3.4.2 (2017-09-28)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252    LC_MONETARY=German_Germany.1252 LC_NUMERIC=C                   
[5] LC_TIME=German_Germany.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] h2o_3.14.0.3      MASS_7.3-47       mlr_2.12          ParamHelpers_1.10

loaded via a namespace (and not attached):
 [1] parallelMap_1.3     Rcpp_0.12.13        compiler_3.4.2      git2r_0.19.0        plyr_1.8.4          bitops_1.0-6       
 [7] tools_3.4.2         digest_0.6.12       jsonlite_1.5        memoise_1.1.0       tibble_1.3.4        gtable_0.2.0       
[13] checkmate_1.8.5     lattice_0.20-35     rlang_0.1.2         Matrix_1.2-11       curl_3.0            parallel_3.4.2     
[19] withr_2.1.0         httr_1.3.1          knitr_1.17          devtools_1.13.3     xgboost_0.6-4       grid_3.4.2         
[25] data.table_1.10.4-3 R6_2.2.2            XML_3.98-1.9        survival_2.41-3     ggplot2_2.2.1       magrittr_1.5       
[31] backports_1.1.1     scales_0.5.0        BBmisc_1.11         splines_3.4.2       colorspace_1.3-2    stringi_1.1.6      
[37] RCurl_1.95-4.8      lazyeval_0.2.0      munsell_0.4.3      

## reproducible example

```r
train = read.csv("train.csv", header = TRUE)
diabetestask = makeClassifTask(id = "diabetesTask", data = train, target = "diabetes", positive = "neg")
##..##
learnerLDA = makeLearner("classif.lda", predict.type = "prob", prior = c(2/3, 1/3) )
```
which returns the error:
[...]

PR #2077, opened 2017-12-01
#2066 add 'prior' hyperparam into classif.lda learner

ptl93:
Fixed issue https://github.com/mlr-org/mlr/issues/2066


larskotthoff:
Thanks, merging.

Issue #2050, opened 2017-11-02
makePreprocWrapperCaret doesn't use methods "corr", "zv", and "nzv"


Prometheus77:
Looking at the code for makePreprocWrapperCaret, it appears that methods "corr" (remove correlated features), "zv" (remove zero-variance features), and "nzv" (remove near zero-variance features) are not supported. ZV can be handled with makeRemoveConstantFeaturesWrapper, but AFAIK there's no way to handle highly correlated and/or duplicate features without making a custom preprocessing wrapper. As SVMs tend to fail when duplicate features exist, this would be a useful enhancement which should be easy to make.

giuseppec:
Thx. We will work on that soon.
Steps that need to be done:
1) look up ?preProcess from caret and compare which methods are not included in makePreprocWrapperCaret https://github.com/mlr-org/mlr/blob/bb32eb8f6e7cbcd3a653440325a28632843de9f6/R/PreprocWrapperCaret.R#L19
2) Extend makePreprocWrapperCaret with the missing methods
3) Write short unit tests for the extended methods

ptl93:
We will work on this : Darya, Michael, Tuan

ptl93:
Created Pull Request : https://github.com/mlr-org/mlr/pull/2078

We found out that ist isnot obvious how to extract the transformed data frame. In the unit test we test that the transformed data frame from caret function is equal to mlr output. As of Clarity we had in mind to add the data after training a model by train(lrn, task), but this might affect other scripts such as: 
makeWrappedModel()

PR #2078, opened 2017-12-01
Fix Issue : Adding 'corr', 'nz' and 'nzv' method from caret::preProcess()

ptl93:
Fixed for https://github.com/mlr-org/mlr/issues/2050


larskotthoff:
Thanks, merging.

Issue #1460, opened 2017-01-27
we should store the class label distribution in the TaskDesc


berndbischl:
that info is useful.
currently we only store the label names, but counts for features types.

we should change to a named int vec for the class labels

Stevo15025:
This would also be useful with #1398

giuseppec:
Also the `mlr:::print.ClassifTask` should be updated so that it does not compute the distribution again. Instead it should print the distribution which is stored in the TaskDesc then.

juliafried:
working on this

PR #2080, opened 2017-12-02
issue1460 - target distribution in taskDesc object

scholbeck:
- For classification tasks, the target distribution was added as an attribute to the `taskDesc` object.
- The according print function was changed, such that the table is not calculated anew for each print, but instead the attribute of `TaskDesc` is printed.
- In the unittest `test_base_createDummyFeatures.R`, the predefined iris task did not contain the added distribution table as an attribute . Therefore the test didn't work.
- The following (classification) tasks were updated: `sonar.task`, `pid.task`, `iris.task`, `bc.task`
-> The data was extracted from the old task and the task was created a new.


scholbeck:
fix for #1460

scholbeck:
The premade tasks in the data folder need to be updated for unittesting.
They currently do not contain the added attribute 'class.distribution'

larskotthoff:
Could you add that please?

R/ClassifTask.R
```diff
@@ -56,10 +57,10 @@ makeClassifTaskDesc = function(id, data, target, weights, blocking, positive, sp
 #' @export
 print.ClassifTask = function(x, ...) {
   # remove 1st newline
-  di = printToChar(table(getTaskTargets(x)), collapse = NULL)[-1L]
+  di = printToChar(x$task.desc$class.distribution)
   m = length(x$task.desc$class.levels)
   print.SupervisedTask(x)
   catf("Classes: %i", m)
-  catf(collapse(di, "\n"))
+  catf(di)
```
mb706: 1) I'm not sure about this, but it looks to me like you're not removing the first newline any more (as the comment in line 59 says). I guess that's ok, but maybe in return you have to add `newline = FALSE` to the `catf("Classes: %i", m)` before this.
2) This was already a bug before: apparently this `catf` will fail if the number of classes is very large:
```R
> makeClassifTask("test", data.frame(x = 1:1000, y = as.factor(1:1000)), "y")
Supervised task: test
Type: classif
Target: y
Observations: 1000
Features:
   numerics     factors     ordered functionals
          1           0           0           0
Missings: FALSE
Has weights: FALSE
Has blocking: FALSE
Classes: 1000
Error in sprintf(...) : 'fmt' length exceeds maximal format length 8192
```
(The `sprintf(...)` referenced is exactly the `catf(di)`.) Can you try to use `cat(di)` instead, and then add a `"\n"` to the next line (`catf("\nPositive class....`)?

mb706: P.S. that `catf` is broken in more ways than just one:
```R
> makeClassifTask("test", data.frame(x = 1:3, y = as.factor(c('a', 'b', '%s'))), "y")
Supervised task: test
[...]
Error in sprintf(...) : too few arguments
```

scholbeck: Thank you for reviewing. I implemented your suggested changes.

`makeClassifTask("test", data.frame(x = 1:1000, y = as.factor(1:1000)), "y")`
and `makeClassifTask("test", data.frame(x = 1:3, y = as.factor(c('a', 'b', '%s'))), "y")`
do not throw an error anymore.

scholbeck:
Yes, I'm going to update the other tasks in the data folder.

scholbeck:
Classification tasks have been updated.

larskotthoff:
Thanks, merging.

Issue #1953, opened 2017-07-28
no getOptPath()?


Prometheus77:
Given that so many mlr classes and objects have getters, it feels inconsistent to have to do as.data.frame(tr$opt.path) instead of just using a getOptPath. Should also include an as.data.frame=TRUE/FALSE option.

larskotthoff:
`getOptPath` is a ParamHelpers function, please open the issue [there](https://github.com/berndbischl/ParamHelpers).

berndbischl:
he has a point here, he means that we usually provide getters for objects.

berndbischl:
also getOptPath just simply does not exist as a function?

giuseppec:
@berndbischl do you want to introduce, e.g., a `getOptPath(obj, as.df = TRUE)` function for `TuneResult` objects?

scholbeck:
Working on it.

PR #2081, opened 2017-12-02
issue 1953 - added getOptPath function to tuneParams.R

scholbeck:
- a getOptPath() function was added to tuneParams.R
- it is a getter method for obtaining the optimization path from a tuning result
- it provides the option to print the optimization path as a data frame
- added a unit test `test_tune_getTuneResultOptPath.R`


scholbeck:
fix for #1953

larskotthoff:
Could you add a unit test for this please?

scholbeck:
Yes. I'm going to add a unit test.

jakob-r:
Following the mlr naming convention this function should be called `getTuneResultOptPath`.
If we want to stick with this naming it has to be S3.

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
 }
 
 
+#' @title Get the optimization path of a tuning result.
+#'
+#' @description
+#' Returns the optPath from a [\code{\link{tuneParams}}] object.
+#' @param tune.result [\code{\link{tuneParams}}] \cr
```
jakob-r: has to be `TuneResult`.

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
...
+#'
+#' @description
+#' Returns the opt.path from a [\code{\link{tuneParams}}] object.
+#' @param TuneResult [\code{\link{tuneParams}}] \cr
+#'   A tuning result of the [\code{\link{tuneParams}}] function.
+#' @param as.df [\code{logical(1)}]\cr
+#'   Should the optimization path be returned as a data frame?
+#'   Default is \code{TRUE}.
+#' @export
+getTuneResultOptPath = function(TuneResult, as.df = TRUE)
```
jakob-r: `tune.result`

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
 }
 
 
+#' @title Get the optimization path of a tuning result.
+#'
+#' @description
+#' Returns the opt.path from a [\code{\link{tuneParams}}] object.
+#' @param TuneResult [\code{\link{tuneParams}}] \cr
```
jakob-r: TuneResult -> `tune.result`
tuneParams -> `TuneResult`

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
 }
 
 
+#' @title Get the optimization path of a tuning result.
+#'
+#' @description
+#' Returns the opt.path from a [\code{\link{tuneParams}}] object.
```
jakob-r: tuneParams -> `TuneResult`

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
...
+#' @param TuneResult [\code{\link{tuneParams}}] \cr
+#'   A tuning result of the [\code{\link{tuneParams}}] function.
```
jakob-r: tuneParams -> `TuneResult`

scholbeck: Are you sure about this one? I just want to show the user in line 135, that the tuning result is created by the tuneParams function.

scholbeck:
I'm sorry about that. I'm new to mlr and couldn't find the naming conventions in the styleguide.

R/tuneParams.R
jakob-r: You are right L:135 it's the result of tuneParams. But in L:134 it has to be TuneResult because that is the class of the object

jakob-r:
Everything looks fine now. Except one last request. I would prefer if we don't have just another tuning call only to test this simple function. Could be incorporated to other tests.

@larskotthoff What do you think?

R/tuneParams.R
```diff
@@ -127,3 +127,21 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
...
+#' @description
+#' Returns the opt.path from a [\code{\link{TuneResult}}] object.
+#' @param tune.result [\code{\link{TuneResult}}] \cr
+#'   A tuning result of the [\code{\link{tuneParams}}] function.
+#' @param as.df [\code{logical(1)}]\cr
+#'   Should the optimization path be returned as a data frame?
+#'   Default is \code{TRUE}.
+#' @export
+getTuneResultOptPath = function(tune.result, as.df = TRUE)
+{
```
larskotthoff: Opening { should be on the same line as the function declaration.

larskotthoff:
Looks good to me apart from the one comment I made. I'm fine with the test code as it is.

scholbeck:
Sorry about that. Fixed it.

larskotthoff:
And a space in between the ( and { :)

jakob-r:
I still think the test is quite "expensive" but ok...

larskotthoff:
It doesn't look like this is increasing build time, so I'll merge it. We can always change it later.

PR #2082, opened 2017-12-04
Tune mbo multi crit

bommert:
Include multi crit tuning with MBO in mlr


R/TuneMultiCritControlMBO.R
```diff
@@ -0,0 +1,67 @@
+#' @export
+#' @param learner [\code{\link{Learner}} | \code{NULL}]\cr
```
jakob-r: These are the same as in TuneControlMBO, right. We should avoid copy & paste. Try to use `@inheritParams` https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html#do-repeat-yourself

R/TuneMultiCritControlMBO.R
```diff
@@ -0,0 +1,67 @@
...
+#'   Number of objectives, i.e. number of \code{\link{Measure}}s to optimize.
+#' @references Bernd Bischl, Jakob Richter, Jakob Bossek, Daniel Horn, Janek Thomas and Michel Lang; mlrMBO: A Modular Framework for Model-Based Optimization of Expensive Black-Box Functions, Preprint: \url{https://arxiv.org/abs/1703.03373} (2017).
+#' @rdname TuneMultiCritControl
+makeTuneMultiCritControlMBO = function(n.objectives = mbo.control$n.objectives,
+  same.resampling.instance = TRUE, impute.val = NULL,
+  learner = NULL, mbo.control = NULL, tune.threshold = FALSE, tune.threshold.args = list(),
+  continue = FALSE, log.fun = "default", final.dw.perc = NULL, budget = NULL, mbo.keep.result = FALSE,
+  mbo.design = NULL) {
+
+  assertIntegerish(n.objectives, lower = 2L)
```
jakob-r: `assertInt()` for a single intergerish value.

R/tuneMultiCritMBO.R
```diff
@@ -0,0 +1,43 @@
+tuneMultiCritMBO = function(learner, task, resampling, measures, par.set, control,
```
jakob-r: For my taste this is too much copy and paste of `tuneMBO.R`. Can you put those changes in `tuneMBO()` with some if else?

R/tuneMBO.R
```diff
@@ -14,6 +14,11 @@ tuneMBO = function(learner, task, resampling, measures, par.set, control,
   force(opt.path)
   force(show.info)
 
+  if (!is.null(mbo.control$n.objectives)) {
```
jakob-r: put that below and just `if (multicrit) {...}` ?

R/tuneMBO.R
```diff
@@ -23,15 +28,33 @@ tuneMBO = function(learner, task, resampling, measures, par.set, control,
     messagef("Resuming previous MBO run using state in '%s'...", state)
     or = mlrMBO::mboContinue(state)
   } else {
-    or = mlrMBO::mbo(tff, design = control$mbo.design, learner = control$learner, control = mbo.control, show.info = FALSE)
+    or = mlrMBO::mbo(tff, design = control$mbo.design, learner = control$learner,
+      control = mbo.control, show.info = FALSE)
+  }
+
+  if (multicrit) {
+    df = convertListOfRowsToDataFrame(or$pareto.set)
```
jakob-r: That looks strange. First you convert a list into a `df` an in the following line you call `as.list` on each row? Can't you just lapply on `or$pareto.set`?

tests/testthat/test_tune_tuneParamsMultiCrit.R
```diff
@@ -158,4 +186,27 @@ test_that("tuneParamsMultiCrit with resample.fun", {
...
+})
+
+test_that("check n.objectives for MBO multi crit", {
+  lrn =  makeLearner("classif.rpart")
+  rdesc = makeResampleDesc("Holdout")
+  ps = makeParamSet(
+    makeIntegerParam("minsplit", lower = 1, upper = 50)
+  )
+
+  expect_error(makeTuneMultiCritControlMBO(1L))
```
jakob-r: I know the checks `expect_error` are all very unspecific here. Could you improve all of them by writing `expect_error(..., "part of expected error message"`)? That would be very nice. This way we can check if the error is actually the error we are expecting.

R/tuneMBO.R
bommert: I will do if (multicrit) but I think that this check should take place before the actual tuning.
Also, I think now that !is.null(...) is not needed at all. mlrMBO makes sure that n.objectives is an integerish number.

jakob-r:
@berndbischl @larskotthoff 
Any objections against merging?

larskotthoff:
Looks good to me.

jakob-r:
Thanks @bommert ! :raised_hands:

PR #2084, opened 2017-12-04
Removing superfluous print statement

mb706:
This print statement looks like it was left over from some debug session. None of the other filters print anything.


jakob-r:
Thanks

Issue #1845, opened 2017-06-12
add adaboostm1 learner from rweka


berndbischl:

berndbischl:
reason: there is no good classic adaboost in R (that i know?).

ada in package ada is EXTREMELY slow

giuseppec:
There is https://cran.r-project.org/web/packages/fastAdaboost/fastAdaboost.pdf which implements Adaboost based on C++ backend code. 
Not sure if this is better than the RWeka learner, at least they have unit tests in their repo https://github.com/souravc83/fastAdaboost

alona-sydorova:
we'll check it out

giuseppec:
Cool. Just do a mini benchmark first and share your results. We can then decide wether you can implement fastAdaboost or the rweka adaboost. (Btw. installing the rweka package can be horrible due to rjava dependency, so if you encounter any problem just ask.)

alona-sydorova:
we have found out that fastAdaboost is implemented only for binary classification while RWeka adaboost can work with multilabel classification tasks. should we still compare their performance or is it more important for you to have the possibility of having more than 2 classes in the task?

giuseppec:
Hm, as the currently used [adaboost learner](https://github.com/mlr-org/mlr/blob/master/R/RLearner_classif_ada.R) is also only able to do binary classification, I would say that a mini benchmark will still help in making a decision.

alona-sydorova:
so, we did some micro benchmarking, with 3 different datasets and 50 to 200 iterations as well as different classifiers for RWeka and found out that RWeka still performs significately better than fastAdaboost

giuseppec:
Do you have the code for this benchmark, so that I could reproduce this on my machine?
But it sounds like you can already implement the RWeka adaboost. Just do a PR for this!

alona-sydorova:
I've sent you the script in Slack

PR #2085, opened 2017-12-05
add adaboostm1 learner from rweka

alona-sydorova:
Solution for this issue: https://github.com/mlr-org/mlr/issues/1845

We have implemented the adaboost M1 learner from RWeka and a test for it


larskotthoff:
Thanks, merging.

Issue #1463, opened 2017-01-30
Getter for learner note


MariaErdmann:
It would be great to have a getter for the learner note. Right now we can only access the note by typing
````
lrn = makeLearner("regr.glmnet")
lrn$note
````
and this looks quite "chaotic" to me:
```
#[1] "Factors automatically get converted to dummy columns, ordered factors to integer.\n      Parameter
#`s` (value of the regularization parameter used for predictions) is set to `0.1` by default,\n      but needs 
#to be tuned by the user."
```
But I think a getter would solve the appearance stuff. 
Although the tutorial lists the learner notes in the list of integrated learners, I couldn't find anything there and neither in the help pages telling the user to look at the learner note and how to do that. Since the note is quite important we should tell the user that he/she really should take a look at the learner's note before getting started with the fitting process.

**Comment by jakob-r** [comment]
The getter would also only return a string that is printed the same way as the string in `lrn$note`. 
You would need a specific class with a own printing method.
A specific class for the note is a bit of an overkill. However I generally favor getters but in this case.

But in this case I think the best idea is to add the note to the printer of the Learner so that when you type
```r
lrn = makeLearner("regr.glmnet")
lrn
```

you will get

```r
Learner regr.glmnet from package glmnet
Type: regr
Name: GLM with Lasso or Elasticnet Regularization; Short name: glmnet
Class: regr.glmnet
Properties: numerics,factors,ordered,weights
Predict-Type: response
Hyperparameters: s=0.01
Note: Factors automatically get converted to dummy columns, ordered factors to integer. Parameter `s` (value of the regularization parameter used for predictions) is set to `0.1` by default, but needs to be tuned by the user.
```

berndbischl:
Lets please not muddle issues so much.

1) One issue is that we simply need a getter for the note.

2) The other is that we have no format definition for the notes. And we are overdoing it at the moment. IMHOi we should completely disallow all of that format stuff that @MariaErdmann posted above. Just plain sentences. If you need more, create a help page, and ref that in the note.

berndbischl:
a note for the format discussion already exists #1462 

please discuss here only the getter, rest there

alona-sydorova:
working on the getter for learner note

PR #2086, opened 2017-12-05
Featuer: add `getLearnerNote()` to access Notes information of learners

alona-sydorova:
fixes https://github.com/mlr-org/mlr/issues/1463


jakob-r:
This will not work for wrapped learners, should we S3 that?

larskotthoff:
Yes. (And please don't regenerate/commit Rd files.)

larskotthoff:
ping @alona-sydorova

pat-s:
> This will not work for wrapped learners, should we S3 that?

Is this really important here? The other operators also do not work in an S3 way.
It's a really small change and just completes the `getLearner*()` family.

I cannot clean up here since its a fork. Should we just add this change in a single commit? @jakob-r 	@larskotthoff

larskotthoff:
Sounds good to me.

jakob-r:
If you want to, you can merge.

PR #2089, opened 2017-12-06
add tests for calculateConfusionMatrix

### LINKED ISSUE #1680: calcConfusionMatrix: more unit tests
from test_base_calculateConfusionMatrix

```
check values itself
 +  task = subsetTask(sonar.task, 1:15)
 +  pred = holdout(makeLearner("classif.rpart"), task)$pred
 +  truth = factor(rep("R", 5), levels = c("M", "R"))
 +  predicted = factor(c("R", "R", "M", "M", "M")) # two correct three wrong
 +  err.abs = 3
 +  err.rel = 3/5
 +  pred$data$truth = truth
 +  pred$data$response = predicted
 +  cm = calculateConfusionMatrix(pred, relative = TRUE)
 +  expect_equal(cm$relative.error, err.rel)
 +  expect_equal(cm$result[3,3], err.abs)
```

this only tests the errrate, but the matrices (abs and rel) contain more entries. add a few lines to the test.

AlexEngelhardt:
- changed the "fake" confusion matrix from `matrix(c(0, 3, 0, 2), nrow=2)` to `matrix(c(1, 5, 3, 7), nrow=2)` so that (a) all cells are filled and (b) quotients are more unlikely to be identical.
- created a second `test_that()` call since that one focuses on the individual elements of the confusion matrix
- added the `split` argument in `holdout()` explicitly
- added one `expect_equivalent()` each for 8 new measures, on the absolute count matrix
- for the relative matrices, I added "transitive" tests that check for equivalence with the absolute count matrix instead of comparing their values to measures.

Closes #1680

### DIFF
--- a/tests/testthat/test_base_calculateConfusionMatrix.R
+++ b/tests/testthat/test_base_calculateConfusionMatrix.R
@@ -56,17 +56,53 @@ test_that("calculateConfusionMatrix", {
   task = makeClassifTask(data = data, target = "Species")
   r = holdout(lrn, task, measures = ber)
   expect_error(calculateConfusionMatrix(r$pred), "FailureModel")
+})
 
+test_that("calculateConfusionMatrix elements are consistent with implemented measures", {
   #check values itself
-  task = subsetTask(sonar.task, 1:15)
-  pred = holdout(makeLearner("classif.rpart"), task)$pred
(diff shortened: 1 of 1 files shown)


larskotthoff:
Great, thanks.

Issue #1416, opened 2017-01-06
default SE estimation in regr.randomForest is set wrong


ja-thomas:
in the function documentation in the parameter definition we state that the default is `jackknife` but we set `se.method = "bootstrap"` in par.vals.

This is quite confusing as everybody (including myself) would assume that when `regr.randomForest` is run in default settings, that jackknife is used...

jakob-r:
should be set to jackknife then - or what is the consensus?

ja-thomas:
imo yes, but we also have to change the default for `keep.inbag`. 

Or we just use `sd` as it is the simplest. I'm not sure, but I would suggest we use something else than bootstrap

berndbischl:
just use sd in mlr, doc it properly and in mbo we have our own defaults anyway

i dont want "keep.inbag" on as a default

riebetob:
working on this

PR #2095, opened 2017-12-08
Issue #1416 - changed default se.method of regr.randomForest to 'sd'

scholbeck:
- fix for #1416 
- changed the default `se.method` in `regr.randomForest` to `sd`
- changed the documentation accordingly


larskotthoff:
Thanks, merging.

PR #2099, opened 2017-12-09
Automatically predict NA for rows w/ NAs and learners that don't support missings

mb706:
A comprehensive fix for the larger issue around #1515, this does what I described in [my comment to #2068](https://github.com/mlr-org/mlr/pull/2068#issuecomment-347551112):
If the learner doesn't support 'missings', the rows containing missings are stripped, and `NA`s are added to the prediction in their place. This has two weaknesses:
1) Apparently there are Learners that don't support "missings" in their training data, but *do* support them in the prediction data. There is no easy fix for this, one might think about adding another Learner-property, or redefining `"missings"` to mean support for missings in the prediction data.
2) If every line of the input contains at least one `NA`, this falls back to the old prediction mode (and possibly creates an error if the Learner doesn't silently ignore `NA`s). A more thorough implementation could create the matrix / vector of `NA`s of appropriate type without calling `predictLearner` at all.

### DIFF
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -362,7 +362,7 @@ plotHyperParsEffect = function(hyperpars.effect.data, x = NULL, y = NULL,
           regr.task = makeRegrTask(id = "interp", data = d.run[, c(x, y, z)],
             target = z)
           mod = train(lrn, regr.task)
-          prediction = predict(mod, newdata = grid)
+          prediction = predict(mod, newdata = grid[c(x, y)])
           grid[, z] = prediction$data[, prediction$predict.type]
           grid$learner_status = "Interpolated Point"
           grid$iteration = NA
--- a/R/predictLearner.R
+++ b/R/predictLearner.R
@@ -62,6 +62,8 @@ predictLearner2 = function(.learner, .model, .newdata, ...) {
     no.na = removeNALines(.newdata)
   else
     no.na = list(newdata = .newdata, inserts = FALSE)
+  if (!nrow(no.na$newdata))
+    no.na = list(newdata = .newdata, inserts = FALSE)  # no choice if all lines contain NA
   p = predictLearner(.learner, .model, no.na$newdata, ...)
   p = checkPredictLearnerOutput(.learner, .model, p)
   return(insertLines(p, no.na$inserts))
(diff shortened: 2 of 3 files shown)


R/generateHyperParsEffect.R
```diff
@@ -362,7 +362,7 @@ plotHyperParsEffect = function(hyperpars.effect.data, x = NULL, y = NULL,
           regr.task = makeRegrTask(id = "interp", data = d.run[, c(x, y, z)],
             target = z)
           mod = train(lrn, regr.task)
-          prediction = predict(mod, newdata = grid)
+          prediction = predict(mod, newdata = grid[c(x, y)])
```
mb706: Bonus bugfix!

tests/testthat/test_base_predict.R
```diff
@@ -144,3 +144,11 @@ test_that("predict works with data.table as newdata", {
   expect_warning(predict(mod, newdata = data.table(iris)), regexp = "Provided data for prediction is not a pure data.frame but from class data.table, hence it will be converted.")
 })
 
+test_that("predict with NA rows for learners that don't support missings automatically returns NA", {
+  mod = train("classif.knn", pid.task)
```
larskotthoff: Could you also add a test for the original random forest problem please?

R/predictLearner.R
```diff
@@ -58,9 +58,41 @@ predictLearner2 = function(.learner, .model, .newdata, ...) {
...
+    no.na = list(newdata = .newdata, inserts = FALSE)  # no choice if all lines contain NA
+  p = predictLearner(.learner, .model, no.na$newdata, ...)
   p = checkPredictLearnerOutput(.learner, .model, p)
-  return(p)
+  return(insertLines(p, no.na$inserts))
+}
+
+removeNALines = function(newdata) {
+  namat = is.na(newdata)
+  if (!any(vlapply(namat, any))) {
```
larskotthoff: Is this check necessary? As far as I can see the code after would do the right thing in this case as well.

mb706: I was wrong about what format the return of `is.na(data.frame)` has, I'll drop this part.

larskotthoff:
Thanks, merging.

PR #2100, opened 2017-12-09
Missing levels suggestion

mb706:
This is my suggestion for the solution of #811 and hence an alternative for the part of #1882 that I was [commenting on](https://github.com/mlr-org/mlr/pull/1882#issuecomment-313875375).

@pat-s maybe you can take this instead of some off the very learner-specific code you have.

This only turns factor levels that were not present during training into `NA`s; it does not help the measures or measure aggregates during resampling or give the user feedback on dropped levels (as #1882 apparently does). I believe handling of `NA`s in predictions during resampling should be independent of `fix.factors.prediction` and in fact happen on the side of the measures (by either dropping or imputing `NA`s) or aggregates (that `na.rm` PR).
I don't think giving messages about dropped factor levels is useful, these measures are not usually given when `fix.factors.prediction` drops levels (when given an entirely novel dataset).

### DIFF
--- a/R/getOOBPreds.R
+++ b/R/getOOBPreds.R
@@ -64,6 +64,7 @@ getOOBPredsLearner.BaseWrapper = function(.learner, .model) {
 # checks if the model was trained on the corresponding task by comparing
 # the descriptions
 checkModelCorrespondsTask = function(model, task) {
-  if (!identical(task$task.desc, model$task.desc))
+  compare = c("id", "type", "target", "n.feats", "has.weights", "has.blocking", "is.spatial", "positive")
+  if (!identical(task$task.desc[compare], model$task.desc[compare]))
     stopf("Description of the model does not correspond to the task")
 }
--- a/R/train.R
+++ b/R/train.R
@@ -33,12 +33,14 @@ train = function(learner, task, subset, weights = NULL) {
   assertClass(task, classes = "Task")
   if (missing(subset))
     subset = NULL
+  if (is.logical(subset))
+    subset = which(subset)  # I believe this is a bug, see #2098
   task = subsetTask(task, subset)
   if (is.null(subset)) {
     subset = seq_len(getTaskSize(task))
   } else {
     if (is.logical(subset))
-      subset = which(subset)
+      subset = which(subset)  # I believe this is a bug, see #2098
     else
       subset = asInteger(subset)
   }
--- a/R/train.R
+++ b/R/train.R
@@ -28,11 +28,9 @@
 #' learner = makeLearner("classif.rpart", minsplit = 7, predict.type = "prob")
 #' mod = train(learner, task, subset = training.set)
 #' print(mod)
-train = function(learner, task, subset, weights = NULL) {
+train = function(learner, task, subset = NULL, weights = NULL) {
   learner = checkLearner(learner)
   assertClass(task, classes = "Task")
-  if (missing(subset))
-    subset = NULL
   if (is.logical(subset))
     subset = which(subset)  # I believe this is a bug, see #2098
   task = subsetTask(task, subset)
@@ -60,7 +58,7 @@ train = function(learner, task, subset, weights = NULL) {
 
   # FIXME: code is bad here, set weights, the simply check it in checktasklearner
   if (!is.null(weights)) {
-    assertNumeric(weights, len = nrow(task$env$data), any.missing = FALSE, lower = 0)
+    assertNumeric(weights, len = getTaskSize(task), any.missing = FALSE, lower = 0)
   } else {
     weights = getTaskWeights(task)
   }
--- a/R/train.R
+++ b/R/train.R
@@ -31,13 +31,21 @@
 train = function(learner, task, subset, weights = NULL) {
   learner = checkLearner(learner)
   assertClass(task, classes = "Task")
-  if (missing(subset) || is.null(subset)) {
+  if (missing(subset))
+    subset = NULL
+  task = subsetTask(task, subset)
+  if (is.null(subset)) {
     subset = seq_len(getTaskSize(task))
   } else {
     if (is.logical(subset))
       subset = which(subset)
     else
       subset = asInteger(subset)
+  }
+  if (learner$fix.factors.prediction) {
+    tdat = getTaskData(task)
+    ttargidx = which(colnames(tdat) %in% getTaskTargetNames(task))
+    task = changeData(task, droplevels(tdat, except = ttargidx))
   }
 
   # make sure that pack for learner is loaded, probably needed when learner is exported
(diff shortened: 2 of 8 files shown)


mb706:
How this works is that in `train`, the task is subsetted first, using the `subset` parameter, and the factor levels saved in the model are thus based only on the training subset of the task. This also has the consequence that the task descriptor of the trained model now saves the target factor level distribution that the learner has *seen*, not the target factor level distribution of the whole resampling task.

A minor consequence of this is that the `task.desc` of a resample prediction is not the `task.desc` of any individual model that was trained on during resampling--it is the `task.desc` of the whole resampling task, while the models have the `task.desc` of the task after subsetting.

This also starts sunsetting the `.subset` parameter of `trainLearner` functions which I've never really seen the point of. The Learners see `.subset` as `NULL`; since all they ever do with it is give it to `subsetTask`, this doesn't have any negative consequences that I can see. I think long-term the `.subset` parameter should be removed.

R/getOOBPreds.R
```diff
@@ -64,6 +64,7 @@ getOOBPredsLearner.BaseWrapper = function(.learner, .model) {
 # checks if the model was trained on the corresponding task by comparing
 # the descriptions
 checkModelCorrespondsTask = function(model, task) {
-  if (!identical(task$task.desc, model$task.desc))
+  compare = c("id", "type", "target", "n.feats", "has.weights", "has.blocking", "is.spatial", "positive")
```
larskotthoff: Is there any way to get this programmatically?

larskotthoff:
I'll let @pat-s comment on the approach first, but deprecating the .subset parameter should be a separate PR.

pat-s:
@mb706 I'm not reluctant dropping my PR and all of its work for a 'better' and cleaner solution. 
Sure, a generic approach is always better. I was just using this as it worked for me and I needed a solution for my work. 

>  I believe handling of NAs in predictions during resampling should be independent of fix.factors.prediction and in fact happen on the side of the measures (by either dropping or imputing NAs) or aggregates (that na.rm PR).

You state that it should be independent but your implementation (as I sse in the tests) still use this flag to correct for it. Or is this just because I have been using this idea in #1882 and you would like to do it in a different way?

> I don't think giving messages about dropped factor levels is useful, these measures are not usually given when fix.factors.prediction drops levels (when given an entirely novel dataset).

Hm, I think this question is two-fold: I see the point in not giving an information as other algs also don't do it (RF, SVM). Actually, I still have no clue how these algorithms handle such issues in detail because it seems to be no problem for them and only applies to parametric ones AFAIK. 
On the other hand, silently changing the data without letting the user know is somehow dirty. However, as said, this is probably also done in ML algs and nobody notices it. 

So this PR is already merge-ready referring to the working tests, isn't it? If yes, sure - it looks cleaner, more generic and simpler than #1882. 
Just happy if we have a working solution in `mlr` finally. 
If my PR at least helped to motivate you doing this PR, it was not completely useless :smile:

mb706:
@pat-s 
> You state that it should be independent but your implementation (as I sse in the tests) still use this flag 

What I mean is that there are two different problems here that should be treated independently:
1) Are there any levels in the test set that were not present in the training set? Some Learners want this to be treated as `NA` values, which they communicate with the `fix.factors.prediction` flag
2) Did any learners predict `NA`s at any points, for reasons that may or may not be related to unseen levels?

As I see it your PR tries ot handle both (1) and part of (2); in particular during resampling it drops predictions that are `NA`, but only when there were unseen levels. IMHO there should be no distinction made between *why* a learner predicted an `NA` (maybe I am wrong?), and the problem of what resampling does with `NA` predictions should  be solved in general, and independently.

This PR doesn't entirely solve the problem you were having, since it only tackles (1), and your problems were (1) *and* (2). But I do think that this is a more elegant solution to (1) since arguably `mlr` was *already* removing unseen factor levels, only with a bug that prevented that from happening sometimes.

All in all I see this PR more as a suggestion (in large part to @pat-s) since I am not so much of a stakeholder when it comes to predictions with unseen levels.

mb706:
@larskotthoff 
> deprecating the .subset parameter should be a separate PR.

In a way this is the PR that depreciates the `.subset` parameter; that it solves a bug related to unseen factor levels is almost a side effect:

The `fix.factors.prediction` machinery was *already* in place, it only gets confused when `train` is called with a `subset` parameter (as it happens in `resample`). The other way of solving this would be to replace the
```R
factor.levels = getTaskFactorLevels(task)
```
line in `train.R` with
```R
factor.levels = getTaskFactorLevels(subsetTask(task, subset))
```
which would work just the same, but perform the subsetting (at least) twice per `train` call.

larskotthoff:
So what's the consensus? I'm happy to merge this, but does it help with the original problem enough to warrant that? @pat-s

mb706:
Also @berndbischl @mllg any opinions on the `.subset` change?

pat-s:
> As I see it your PR tries ot handle both (1) and part of (2); in particular during resampling it drops predictions that are NA, but only when there were unseen levels. IMHO there should be no distinction made between why a learner predicted an NA (maybe I am wrong?), and the problem of what resampling does with NA predictions should be solved in general, and independently.

I agree with you that no distinction should be made why a learner predicated an NA.
My approach was simply a "working" one that handled (1) and (2), well, by accident. 
The main aim for me was that it solved the issues for the models that errored otherwise. 
I'm happy that there is a more general solution that is also way simpler in its implementation. 

I tested it with the tests from #1882 and it works. 
It also works for my dataset at work so I'm happy with the implementation. 

> All in all I see this PR more as a suggestion (in large part to @pat-s) since I am not so much of a stakeholder when it comes to predictions with unseen levels.

Thats humble from you but I'm really fine with merging your PR and eventually building on it for further fixes. Of course I spent dozens of hours for #1882 but I do not insist using this version. 
If (2) is an issue coming up again in the future, we can built upon this PR, possibly with some code from #1882. 

I'm fine putting time into other PRs/issues and marking this as solved :smile:

larskotthoff:
Sounds good -- I'll give @mllg and @berndbischl a chance to comment. Please remind me again in a few days if nothing has happened and I'll merge.

R/getOOBPreds.R
mllg: I'd suggest to write a hash function for task descriptions and save the hashes in the task descriptions. This is cleaner and eventually even cheaper than the bitwise comparison done here. Must not be in this PR though.

R/train.R
```diff
@@ -31,26 +31,36 @@
 train = function(learner, task, subset, weights = NULL) {
   learner = checkLearner(learner)
   assertClass(task, classes = "Task")
-  if (missing(subset) || is.null(subset)) {
+  if (missing(subset))
+    subset = NULL
```
mllg: I assume we break stuff if we let subset default to `NULL`?

R/train.R
```diff
@@ -31,26 +31,36 @@
...
   tn = getTaskTargetNames(task)
 
   # make pars list for train call
-  pars = list(.learner = learner, .task = task, .subset = subset)
+  pars = list(.learner = learner, .task = task, .subset = NULL)
 
   # FIXME: code is bad here, set weights, the simply check it in checktasklearner
   if (!is.null(weights)) {
-    assertNumeric(weights, len = length(subset), any.missing = FALSE, lower = 0)
+    assertNumeric(weights, len = nrow(task$env$data), any.missing = FALSE, lower = 0)
```
mllg: `getTaskSize(task)`?

R/train.R
mb706: I can set the default to `NULL` and see what happens, I was just trying to keep the UI changes as small as possible (since the `missing(subset) || is.null(subset)` bit seemed intentional).

larskotthoff:
Thanks, merging. @pat-s please modify/close your PRs/issues on this as appropriate.

PR #2102, opened 2017-12-10
Fix 2101 surv task type

### LINKED ISSUE #2101: The $type slot of "surv" tasks is "regr"
E.g. `lung.task`:
```R
> getTaskDesc(lung.task)$type
[1] "surv"
> lung.task$type
[1] "regr"
```

mb706:
Fix #2101 
On second thought I'm not 100% sure this is a bug, maybe "surv" is supposed to be a subclass of "regr" in some way? The places I've seen don't use the `task$type` slot at all, so maybe it can also be dropped.

### DIFF
--- a/R/SurvTask.R
+++ b/R/SurvTask.R
@@ -30,7 +30,7 @@ makeSurvTask = function(id = deparse(substitute(data)), data, target, weights = 
     }
   }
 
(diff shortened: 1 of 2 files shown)


mb706:
OK sorry I pushed the wrong thing, will repair in a moment.

larskotthoff:
@mllg ?

mllg:
Survival task are not a subset of regression tasks. Merging. 
Thanks!

PR #2103, opened 2017-12-11
Fix 2092 no check data in cluster task

### LINKED ISSUE #2092: makeClusterTask partially ignores check.data argument
Most `make***Task` functions all call `checkTaskData` (checking for unsupported column types) conditional on `check.data` being `TRUE` ([here](https://github.com/mlr-org/mlr/blob/905421a6ddfdbad399cc0ec5f453579215115736/R/SupervisedTask.R#L4-L11)).  The only exception seems to be `makeClusterTask`, which calls `makeUnsupervisedTask` that then goes on to call `checkTaskData` regardless of the value of `check.data` (see [here](https://github.com/mlr-org/mlr/blob/905421a6ddfdbad399cc0ec5f453579215115736/R/UnsupervisedTask.R#L6)).

mb706:
Fixes #2092, see there why I think this is a bug.


larskotthoff:
Thanks, merging.

PR #2104, opened 2017-12-12
Plot tune multi crit result

bommert:
Enhancement of plotTuneMultiCritResultGGVIS:
So far, only the points in y-space (performance measures) are shown.
Now, information about the x-space (hyper parameters) can be displayed.
This information can be shown on mouse click or on mouse hover (or not at all, if desired).


R/plotTuneMultiCritResult.R
```diff
@@ -68,29 +68,64 @@ plotTuneMultiCritResult = function(res, path = TRUE, col = NULL, shape = NULL, p
...
+    ggvis::prop("y", as.name(colnames(res$y)[2L])),
+    ggvis::prop("key", ~id, scale = FALSE))
+  plt = ggvis::layer_points(plt, ggvis::prop("fill", as.name("location")))
+
+  if (point.info == "hover") {
+    plt = ggvis::add_tooltip(plt, info, "hover")
+  }
+  else if (point.info == "click") {
+    plt = ggvis::add_tooltip(plt, info, "click")
+  }
```
larskotthoff: This block could be just `if (point.info != "none") { plt = ggvis::add_tooltip(plt, info, point.info) }`

larskotthoff:
On a general note, ggvis doesn't seem to be going anywhere, so we'll likely remove the code at some point. It might be worth looking into plotly as an alternative for interactive plots.

jakob-r:
Agreed, it would be preferable to have a ggplot (+plotly) or a pure plotly version.

larskotthoff:
I'll merge this as it's independent of any other functionality and can be deprecated easily.

PR #2106, opened 2017-12-13
calculateConfusionMatrix() gets a new parameter set = "test" | "train" | "both"

### LINKED ISSUE #2046: calculateConfusionMatrix should have a parameter to specify which set to use
E.g. `set = "train" | "test" | "both"`.

AlexEngelhardt:
I did:

- add the argument, defaulting to "both" (for backwards compatibility)
- if it's unequal to "both", first assertClass(pred, classes = "ResamplePrediction"), to make sure the column pred$data$set exists
- then subset the data using this column

See the discussion in #2046 for details.

I also added tests for this argument.

This pull request will now have conflicts with #2089 since both edited `tests/testthat/test_base_calculateConfusionMatrix.R`. I'll update one in case the other gets accepted.

Unrelated: I also changed braces to brackets for one Roxygen parameter.

Fixes #2046


larskotthoff:
Thanks, looks good. Could you add some more checks on the data please? In particular, what happens if a resample prediction is passed, but was computed with `predict = "train"` and `set` is `test` or vice versa?

AlexEngelhardt:
With "checks", did you mean tests in the tests/ directory or in the main function?

Currently, the behavior is to return an all-zero confusion matrix in these cases. I added a test that verifies that for predict = "test" and set = "train". Vice-versa, I can't check it since I get an error:

```r
rdesc_train = makeResampleDesc("CV", iters = 3, predict = "train")
pred_train = resample("classif.rpart", iris.task, rdesc_train)$pred
# Error in FUN(X[[i]], ...) : 
#  Aggregation 'test.mean' not compatible with resampling! You have to set arg 
#  'predict' to 'test' or 'both' in your resample object, instead it is 'train'!
```

larskotthoff:
In the function to calculate the confusion matrix. It would be great if you could make the behaviour consistent, i.e. either always return an error if the required data isn't there, or always return an all-zero confusion matrix. My feeling is that an error is better, as an all-zero confusion matrix is probably never what the user wants and it's easy to miss in large scripts.

AlexEngelhardt:
Okay, I changed the behavior (and tests) so that instead of an all-zero confusion matrix, the function now raises an error. It does it also if the prediction object were empty (which it can't be since that is asserted during the `predict()` call) - I added a test for this too.

This change modifies previous behavior (in the case of empty confusion matrices) - is that a problem?

By the way: Is there any way I can locally limit `test(filter="lint")` to only the few files I modified? The full lintr run takes ca. 8 minutes every time.

R/calculateConfusionMatrix.R
```diff
@@ -54,12 +59,26 @@ calculateConfusionMatrix = function(pred, relative = FALSE, sums = FALSE) {
   n.pred = length(resp)
   truth = getPredictionTruth(pred)
 
+  if (set != "both"){
+      assertClass(pred, classes = "ResamplePrediction")
+      subset.idx = (pred$data$set == set)
+
+      if (!any(subset.idx)){
```
larskotthoff: There should be a space between `)` and `{` (not sure why the lintr isn't catching that).

R/calculateConfusionMatrix.R
```diff
@@ -54,12 +59,26 @@ calculateConfusionMatrix = function(pred, relative = FALSE, sums = FALSE) {
...
+
   cls = union(levels(resp), levels(truth))
   k = length(cls)
   truth = factor(truth, levels = cls)
   resp = factor(resp, levels = cls)
 
   tab = table(truth, resp)
+  if (all(tab == 0)){
+      stop("Confusion matrix is empty!")
+  }
```
larskotthoff: I don't think we need this check here -- the check above catches the error conditions.

tests/testthat/test_base_calculateConfusionMatrix.R
```diff
@@ -118,3 +118,38 @@ test_that("calculateConfusionMatrix with different factor levels (#2030)", {
...
+    # here, you have set=train and set=test in pred3$data:
+    pred2 = resample("classif.rpart", iris.task, rdesc)$pred
+
+    # pred1$data has no column "set" => argument set="train" would *not* make sense
+    expect_error(calculateConfusionMatrix(pred1, set = "train"))
+
+    # pred3 was predicted on both train and test set. Both subsetted matrices should give
+    # a positive total count:
+    expect_gt(sum(calculateConfusionMatrix(pred2, set = "train")$result), 0)
+    expect_gt(sum(calculateConfusionMatrix(pred2, set = "test")$result), 0)
```
larskotthoff: Could you test the actual numbers please, not just greater than 0? If the set argument did nothing and just returned everything all the time this would still pass.

There are some smaller test tasks in tests/testthat/helper_objects.R that you can use to speed things up and make checking the result easier.

tests/testthat/test_base_calculateConfusionMatrix.R
```diff
@@ -118,3 +118,38 @@ test_that("calculateConfusionMatrix with different factor levels (#2030)", {
...
+    pred.test = resample("classif.rpart", iris.task, rdesc.test)$pred
+
+    expect_error(calculateConfusionMatrix(pred.test, set = "train"))
+})
+
+test_that("calculateConfusionMatrix raises error when prediction object is empty", {
+    mod = train("classif.lda", iris.task)
+    pred = predict(mod, iris.task)
+    pred$data = pred$data[FALSE, ]
+    expect_error(calculateConfusionMatrix(pred))
```
larskotthoff: As above, I wouldn't throw an error in this case.

larskotthoff:
Thanks, I've added a few minor change requests -- an all-zero confusion matrix should be fine as long as it's not the result of the subsetting.

There's no easy way to run only a subset of the lint tests unfortunately.

R/calculateConfusionMatrix.R
AlexEngelhardt: Oh. I did that a few other times too, and he never mentioned it. Maybe he doesn't have a rule for it?

R/calculateConfusionMatrix.R
AlexEngelhardt: Oh, okay. I thought that's what you meant with "make the behavior consistent".

It's removed now.

tests/testthat/test_base_calculateConfusionMatrix.R
AlexEngelhardt: The function returns an all-zero matrix now

tests/testthat/test_base_calculateConfusionMatrix.R
AlexEngelhardt: Okay, done. The lines look a bit messy now, I subsetted the confusion matrices to [1:3]. I hope that's still okay.

larskotthoff:
Thanks, merging!

PR #2107, opened 2017-12-13
Fix task data

mb706:
This fixes a few discrepancies I found in the task files in the `data` directory:
* the `"surv"` tasks had type `"regr"` which was fixed in #2102 but had not propagated to the builtin data. Affected `lung.task`, `wpbc.task`.
* The `"surv"` tasks had the `$censoring` slot which was removed from `"surv"` tasks in #1833. Affected `lung.task`, `wpbc.task`.
* Some of the classif and multilabel tasks were missing the `class.distribution` slot. Affected `bc.task.spatial`, `costiris.task`, `gunpoint.task`, `phoneme.task`.
* (minor) `yeast.task`'s `task.desc` had names in a different order than other `task.desc`s

The Assumption I am making here is that `changeData(task, getTaskData(task, functionals.as = "matrix"))` should always be `all.equal` the original task. I added a test that verifies this for the builtin data. Please don't merge this if this assumption is wrong.

The tasks were fixed from the original data using the script
```R
# in the 'data' subdir
for (x in list.files()) {
  load(x)
  xname = gsub("\\.rda$", "", x)
  xdat = get(xname)
  xdat2 = changeData(xdat, getTaskData(xdat, functionals.as = "matrix"))
  if (isTRUE(all.equal(xdat, xdat2))) next
  assign(xname, xdat2)
  save(list=xname, file=x)
}
```


larskotthoff:
Thanks for fixing this, merging.

We should probably make it part of the travis build to regenerate the data files so things like this can't happen.

PR #2110, opened 2017-12-13
Fix costsens tasks

mb706:
So apparently `getTaskCosts` was broken in that its `.CostSensTask` version was not exported. Therefore the 'cost' returned even for CostSensTask was `NULL` when one uses `mlr` from the library (and not with `load_all`).

This also broke the script I used in #2107, so I re-generated `costiris.task` again.


mb706:
(sry I mis-clicked)

larskotthoff:
Thanks, merging.

Issue #1407, opened 2016-12-24
Error in unique.default(x, nmax = nmax) :    unique() applies only to vectors


krishna1072:
Hi,

I am finding error "Error in unique.default(x, nmax = nmax) :    unique() applies only to vectors" when trying to train the model. Below is the complete code and its giving me issue when running the code "xgmodel <- train(learner = lrn_tune, task = traintask)" 

I have also attached the data files. Please guide me with the issue.

[adultdata.txt](https://github.com/mlr-org/mlr/files/671664/adultdata.txt)
[Uploading adulttest.txt…]()

```r
rm(list=ls())

#load libraries and data
library(data.table)
library(mlr)

#set variable names
setcol <- c("age",
            "workclass",
            "fnlwgt",
            "education",
            "education-num",
            "marital-status",
            "occupation",
            "relationship",
            "race",
            "sex",
            "capital-gain",
            "capital-loss",
            "hours-per-week",
            "native-country",
            "target")

#load data
train <- fread("adultdata.txt",header = F,sep = ",",col.names = setcol,na.strings = c(" ?"),stringsAsFactors = T)
test <- fread("adulttest.txt",header = F,sep = ",",col.names = setcol,skip = 1, na.strings = c(" ?"),stringsAsFactors = T)

setDT(train)
setDT(test)

#Data Sanity
dim(train) #32561 X 15
dim(test) #16281 X 15
str(train)
str(test)

#check missing values
table(is.na(train))
sapply(train, function(x) sum(is.na(x))/length(x))*100

table(is.na(test))
sapply(test, function(x) sum(is.na(x))/length(x))*100

#quick data cleaning
#remove extra character from target variable
library(stringr)
test[,target := substr(target,start = 1,stop = nchar(as.character(target))-1)][,target:=as.factor(target)]

#remove leading whitespaces
char_col <- colnames(train)[sapply(train,is.factor)]
for(i in char_col)
  set(train,j=i,value = str_trim(train[[i]],side = "left"))
for(i in char_col)
  set(test,j=i,value = str_trim(test[[i]],side = "left"))

#set all missing value as "Missing"
train[is.na(train)] <- "Missing"
test[is.na(test)] <- "Missing"

#Convert chars to factors
for (col in char_col) {
  cat(paste0(col,'..',"\n"))
  train[,(col):= as.factor(get(col))]
  test[,(col):= as.factor(get(col))]
}
  

#using one hot encoding
labels <- train$target
ts_label <- test$target
[...]

PR #2113, opened 2017-12-19
Fix caret train method name clash

### LINKED ISSUE #1825: While running train() function it is showing "Error: Please use column names for `x`", what does it mean?
I am working on titanic dataset and for initial testing I have ignored variables with NA values and some unimportant values. I am trying to use mlr pacakge and I am not able to understand to find the problem.

`library(mlr)
library(rpart)
library(caret)
library(caTools)
library(data.table)
library(datasets)
library(mlr)``
#reading the dataset
tr<-read.csv("Titanic.csv",header = T,stringsAsFactors = F)
tr$X<-NULL
tr$Name<-NULL
tr$PClass<-as.factor(tr$PClass)
tr$Sex<-NULL
tr$SexCode<-as.factor(tr$SexCode)
tr$Age<-NULL

spl=sample.split(tr$Survived,0.7)

train<-subset(tr,spl==T)
test<-subset(tr,spl==F)

regr<-makeClassifTask(data=train,target ="Survived")

classif.lrn = makeLearner("classif.randomForest", predict.type = "prob", fix.factors.prediction = TRUE)

mod<-train(classif.lrn,regr)

#Output showing  "Error: Please use column names for `x`"

mb706:
## What is this?
The `caret` package exports a `train` function, just like `mlr` does, and masks it when `caret` is loaded after `mlr`. Since apparently many people stumble over this when they load `caret` after `mlr` (see most visited issues in ["Popular Content" in GitHubs traffic analytics](https://github.com/mlr-org/mlr/graphs/traffic)), I think this is something worth fixing.

Closes #1825, also related to #1407 and maybe a few others.

## How it works
`mlr::train` is turned into an `S3` method itself. It would be possible to "depend" on `caret`, import the `train` generic and only define the class methods `train.Learner`, `train.character`. However, trying to not depend on `caret`, the problem is a bit harder: `mlr` must define the `train` generic itself and only call `caret::train` if the caret package is loaded and attached. This is because of how `S3` method lookup works: Apparently it searches the current environment *and* the environment where `UseMethod` is called, so when `mlr` cals `UseMethod("train")`, `caret`'s `train` methods are not found.

This runs into problems if `caret`'s (uncooperative) `train` shadows `mlr`'s `train`: For the reason described above, it will not find `mlr`'s method definitions. Therefore, an additional trick is used in `zzz.R` to detach and re-attach `mlr` whenever `caret` is attached, so that `mlr` is always in the front of `caret`. This should be okay with `caret` users, since `mlr` is nice enough to call their generic.

## How to use
```R
library("mlr")
library("caret")
train(iris[1:4], iris[[5]])  # calls caret
train(Species ~ ., iris)  # calls caret
train("classif.logreg", pid.task)  # calls mlr
```
should all work now.

## Design Decisions
A few design decisions made in this PR might be debatable:
* It is also possible to just "depend" on `caret`. Currently it is in "suggests", and I don't think it would be worth it to depend on a package just because it happens to use the same method names as `mlr` does.
* I renamed the `learner` parameter of `train` to `x`, and turned `mlr::train` into an `S3` method. I think this is the way to be least surprising for people who load `caret` *after* `mlr` and actually intend to use `caret`.
* The `.onAttach` part is possibly a bit tricky; I haven't seen any other way to get what I want here, but if someone has a nicer way, please tell.


larskotthoff:
Build seems to be failing.

mb706:
The error just seems to be some typo in my roxygen doc, i'll fix that later. Whats the general opinion about this PR?

larskotthoff:
Looks fine, but I'm not sure if messing around with attaching/detaching like that might have negative side effects. Somebody who knows more about R internals than me should have a look @mllg @berndbischl

mllg:
I don't think that this is a good idea. The current situation is not perfect, but at least there is a standardized way to deal with name clashes by explicitly loading packages in the desired order. If more packages would follow your approach, chaos ensues.

PR #2121, opened 2017-12-23
merge growing window and fixed window cross validation

SteveBronder:
This is a subset of the forecasting branch that allows mlr to use growing window and fixed window cross validation schemas (like the graph below)

Top is fixed, bottom is growing

Caret has this and so I think it makes sense for us to have it


larskotthoff:
Thanks Steve. If I'm reading your code correctly, `skip` doesn't do what it should according to the documentation -- it's not the number of observations to skip, but the number of observations not to skip in between skipped ones. Is that correct?

SteveBronder:
Thanks for looking this over! 

>  it's not the number of observations to skip, but the number of observations not to skip in between skipped ones. Is that correct?

I'm not sure if I understand. By default, the window's for the test and train indices move one observation ahead in each iteration. A `skip` value of 1 will not cause any observations to be 'skipped'.  (fyi, the `skip` argument is really the `by` argument of `seq()` that operates over the list of indices).

I do think 'the number of observations to skip' is wrong. Maybe it should be, "The number of iterations to skip over. The default of `horizon - 1` creates tests sets which do not overlap."

I think that's more correct, but it does not sound very graceful.

SteveBronder:
The functions is based on Caret's `createTimeSlices()` function. For the `skip` argument they say
```
skip: how many (if any) resamples to skip to thin the total amount
```
Which I think makes the most sense from a user perspective

SteveBronder:
@larskotthoff checks are passing and I changed the skip argument's doc to be in line with caret because I think that makes the most sense. Is that cool with you?

Thanks again for looking this over!

SteveBronder:
ping @larskotthoff

SteveBronder:
ping @larskotthoff

larskotthoff:
Thanks. Is `skip` the standard name for this, or just what caret uses? It sounds like `by` would be a better name.

SteveBronder:
Looking over a few of the other forecast packages I don't see any standard naming convention. Since most people do ML models with growing / fixed window CV through Caret I think `skip` would be the standard name people would expect.

I don't disagree that `by` is more accurate, but I think it is better to stick to the argument name Caret uses since `skip` does the same thing in both packages

R/ResampleDesc.R
```diff
@@ -47,6 +48,13 @@
 #'     Default is 10.}
 #'   \item{folds [\code{integer(1)]}}{Folds in the repeated CV for \code{RepCV}.
 #'     Here \code{iters = folds * reps}. Default is 10.}
+#'   \item{horizon [\code{integer(1)}]}{Number of observations to forecast for \code{GrowthCV}
+#'    and \code{FixedCV}. Default is 1}
```
larskotthoff: Full stop missing at end of sentence.

R/ResampleDesc.R
```diff
@@ -47,6 +48,13 @@
...
+#'   \item{initial.window [\code{numeric(1)}]}{Fraction of observations to start with
+#'    in \code{GrowthCV} and \code{FixedCV}. Default is 0.5}
+#'   \item{skip [\code{integer(1)}]}{ How many resamples to skip to thin the total amount
+#'    for \code{GrowingCV} and \code{FixedCV}. Default is horizon - 1 which gives
+#'    mutually exclusive chunks of test indices.}
```
larskotthoff: Mention relationship to `by` in `seq` here to clarify.

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
   makeResampleInstanceInternal(desc, size, train.inds = train.inds, test.inds = test.inds, group = g)
 }
+
+
+instantiateResampleInstance.FixedCVDesc = function(desc, size, task = NULL, coords) {
+  initial.window.abs = floor(desc$initial.window * size)
+  assertInt(initial.window.abs, lower = 1)
+  if (size - initial.window.abs < desc$horizon)
+    stop(catf("The initial window is %i observations while the data is %i observations. \n
+              There is not enough data left (%i observations) to create a test set for a %i size horizon",
```
larskotthoff: Full stop missing at end of message.

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+              initial.window.abs, size, initial.window.abs - size, desc$horizon))
+  skip = desc$skip
+  stops  = (seq(size))[initial.window.abs:(size - desc$horizon)]
+  starts = stops - initial.window.abs + 1
+  train.inds = mapply(seq, starts, stops, SIMPLIFY = FALSE)
+  test.inds  = mapply(seq, stops + 1, stops + desc$horizon, SIMPLIFY = FALSE)
+
+  thin = function(x, skip = 0) {
```
larskotthoff: Is the default for `skip` needed here? Looks like it will never be called with 0.

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+
+  if (skip > 0) {
+    train.inds = thin(train.inds, skip = skip)
+    test.inds = thin(test.inds, skip = skip)
+  }
+
+  if (length(test.inds) == 0)
+    stop("Skip is too large and has removed all resampling instances. Please lower the value of skip.")
+
+  # If the last value if not included, shift everything over by one
```
larskotthoff: What exactly is this supposed to do? Make sure that the last observation is in there? Wouldn't that require adjusting skip/horizon?

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+              There is not enough data left (%i observations) to create a test set for a %i size horizon",
+              initial.window.abs, size, initial.window.abs - size, desc$horizon))
+  skip = desc$skip
+  stops  = (seq(from = 1, to = size))[initial.window.abs:(size - desc$horizon)]
+  starts = rep(1, length(stops))
+  train.inds = mapply(seq, starts, stops, SIMPLIFY = FALSE)
+  test.inds  = mapply(seq, stops + 1, stops + desc$horizon, SIMPLIFY = FALSE)
+
+
+  thin = function(x, skip = 0) {
```
larskotthoff: Repeated from above -- pull out to be separate function.

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+    n = length(x)
+    x[seq(1, n, by = skip)]
+  }
+
+  if (skip > 0) {
+    train.inds = thin(train.inds, skip = skip + 1)
```
larskotthoff: Why `skip + 1`?

R/ResampleInstances.R
```diff
@@ -67,3 +67,78 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+    test.inds  = thin(test.inds, skip = skip + 1)
+  }
+
+  if (length(test.inds) == 0)
+    stop("Skip is too large and has removed all resampling instances. Please lower the value of skip.")
+
+  # If the last value if not included, shift everything over by one
```
larskotthoff: See above.

R/ResampleInstances.R
SteveBronder: Nope, removed

R/ResampleInstances.R
SteveBronder: It's there to ensure that last observation is included. There are instances where an initial window + horizon + skip can cause the most recent observation to not be included. This was a simple thing to say, "Hey if the most recent observation did not make it into your final iteration then we are going to shift everything up by one." Though this does mean you lose the first observation.

Do you think it would be better to do something that shifts everything forward while keeping that first observation (like below) or should I do nothing at all (maybe a warning?)

```R
train.inds = lapply(train.inds, function(x) c(x, x[length(x)] + 1))
test.inds = lapply(test.inds, function(x) c(x[2:desc$horizon], x[desc$horizon] + 1
```

R/ResampleInstances.R
SteveBronder: This should be `skip` and `horizon - 1` should just be `horizon`

SteveBronder:
Would better names be `GrowingWindowCV` and `FixedWindowCV`?

SteveBronder:
Failing because of the ranger error

R/ResampleInstances.R
larskotthoff: Hmm, I would just output a warning here without adjusting things manually -- maybe this is what the user intended? Certainly changing stuff internally without telling the user isn't good.

Is there some way we could make it easier for the user to specify these values? E.g. allow to specify skip/horizon in terms of the window size (if number >= 1 it's the number of observations, else it's the fraction of the window). Would that make sense?

larskotthoff:
I don't feel strongly about including "Window" in the name. Might be more descriptive to have it in there.

R/ResampleInstances.R
SteveBronder: > Hmm, I would just output a warning here without adjusting things manually -- maybe this is what the user intended? Certainly changing stuff internally without telling the user isn't good.

Yes I agree that a warning would be best here

> Is there some way we could make it easier for the user to specify these values? E.g. allow to specify skip/horizon in terms of the window size (if number >= 1 it's the number of observations, else it's the fraction of the window). Would that make sense?

By these values do you mean skip/horizon? You usually want to forecast for a fixed length of time (say 10 days or so) so I'm not sure how setting these as a fraction of the window would feel. I can play around with setting it as a fraction of the window, but I think it's probably better to leave these as integers

larskotthoff:
Ranger tests are fixed in master.

R/ResampleInstances.R
SteveBronder: I've thought about this more and I think it's best to leave these as integers

SteveBronder:
If you are cool with the changes then I believe this is good to go. I can add these to the tutorial this week or wait until we are using `pkgdown`

R/ResampleInstances.R
```diff
@@ -67,3 +67,63 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+
+  if (skip > 0) {
+    train.inds = thin(train.inds, skip = skip)
+    test.inds = thin(test.inds, skip = skip)
+  }
+
+  if (length(test.inds) == 0)
+    stop("Skip is too large and has removed all resampling instances. Please lower the value of skip.")
+
+  # If the last value is not included, shift everything over by one
```
larskotthoff: This comment is obsolete now, isn't it?

R/ResampleDesc.R
```diff
@@ -47,6 +48,13 @@
 #'     Default is 10.}
 #'   \item{folds [\code{integer(1)]}}{Folds in the repeated CV for \code{RepCV}.
 #'     Here \code{iters = folds * reps}. Default is 10.}
+#'   \item{horizon [\code{integer(1)}]}{Number of observations to forecast for \dQuote{GrowingWindowCV}
+#'    and \dQuote{FixedWindowCV}. Default is 1.}
+#'   \item{initial.window [\code{numeric(1)}]}{Fraction of observations to start with
```
larskotthoff: I would mention that this determines the training set.

R/ResampleInstances.R
```diff
@@ -67,3 +67,63 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+  if (test.inds[[length(test.inds)]][desc$horizon] != size) {
+    warning("The last observation was excluded. To include the final observation please change either the initial.window or horizon.")
```
larskotthoff: So the purpose of this check is to make sure that the user isn't accidentally excluding data, right? Then I would make it more general and check for any number of excluded observations at the end, in particular as excluding e.g. 3 observations at the end is worse than 1.

R/ResampleInstances.R
```diff
@@ -67,3 +67,63 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
   g = as.factor(rep(seq_len(desc$reps), each = folds))
   makeResampleInstanceInternal(desc, size, train.inds = train.inds, test.inds = test.inds, group = g)
 }
+
+
+instantiateResampleInstance.FixedWindowCVDesc = function(desc, size, task = NULL, coords) {
+  initial.window.abs = floor(desc$initial.window * size)
```
larskotthoff: Why `floor` and not `round`?

R/ResampleInstances.R
```diff
@@ -67,3 +67,63 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+  # If the last value is not included, shift everything over by one
+  if (test.inds[[length(test.inds)]][desc$horizon] != size) {
+    warning("The last observation was excluded. To include the final observation please change either the initial.window or horizon.")
+  }
+  desc$iters = length(test.inds)
+  makeResampleInstanceInternal(desc, size, train.inds = train.inds, test.inds = test.inds)
+}
+
+instantiateResampleInstance.GrowingWindowCVDesc = function(desc, size, task = NULL, coords) {
+  initial.window.abs = floor(desc$initial.window * size)
```
larskotthoff: See above.

R/ResampleInstances.R
```diff
@@ -67,3 +67,63 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL, coor
...
+
+  if (skip > 0) {
+    train.inds = thin(train.inds, skip = skip)
+    test.inds  = thin(test.inds, skip = skip)
+  }
+
+  if (length(test.inds) == 0) {
+    stop("Skip is too large and has removed all resampling instances. Please lower the value of skip.")
+  }
+  # If the last value if not included, shift everything over by one
```
larskotthoff: Comment obsolete, check see above.

R/ResampleDesc.R
```diff
@@ -47,6 +48,13 @@
...
+#'    in \dQuote{GrowingWindowCV} and \dQuote{FixedWindowCV}. Default is 0.5.}
+#'   \item{skip [\code{integer(1)}]}{ How many resamples to skip to thin the total amount
+#'    for \dQuote{GrowingWindowCV} and \dQuote{FixedWindowCV}. This is passed through as the \dQuote{by} argument
+#'    in \code{seq()}. Default is horizon which gives mutually exclusive chunks of test indices.}
```
larskotthoff: This seems to be a bit inconsistent to me -- `horizon` and `skip` are numbers of observations, `initial.window` is a fraction. Could we support both in all? I.e. < 1 means fraction, >= 1 number of observations.

larskotthoff:
Thanks, looks good. As always, I have a few comments/requests, but this is very close to ready :)

SteveBronder:
lol of course, they are very appreciated! I'll look these over tmrw

R/ResampleDesc.R
SteveBronder: check

R/ResampleDesc.R
SteveBronder: Yes the more I thought about this the more I liked it! rn I have
If value < 1:
skip is scale by the length the resampling index
initial.window is scaled by size
horizon is scaled by the length of the initial window

Are those sensible?

R/ResampleInstances.R
SteveBronder: yep

R/ResampleInstances.R
SteveBronder: yes, now it checks how much data is missing and includes it in the error

R/ResampleInstances.R
SteveBronder: Yep

R/ResampleInstances.R
SteveBronder: yes

R/ResampleDesc.R
```diff
@@ -178,6 +214,22 @@ print.RepCVDesc = function(x, ...) {
   catf("Stratification: %s", x$stratify)
 }
 
+#' @export
+print.GrowingWindowCVDesc = function(x, ...) {
+  catf("Window description:\n %s: %.2f in initial window, horizon of %i, and skipping %i windows.",
```
larskotthoff: Initial window and skip can be fractions now.

R/ResampleDesc.R
```diff
@@ -178,6 +214,22 @@ print.RepCVDesc = function(x, ...) {
...
+       x$id, x$initial.window, x$horizon, x$skip)
+  catf("Predict: %s", x$predict)
+  catf("Stratification: %s", x$stratify)
+}
+
+#' @export
+print.FixedWindowCVDesc = function(x, ...) {
+  catf("Window description:\n %s: %.2f in initial window, horizon of %i, and skipping %i windows.",
```
larskotthoff: See above.

R/ResampleInstances.R
```diff
@@ -69,3 +69,66 @@ instantiateResampleInstance.SpRepCVDesc = function(desc, size, task = NULL) {
   g = as.factor(rep(seq_len(desc$reps), each = folds))
   makeResampleInstanceInternal(desc, size, train.inds = train.inds, test.inds = test.inds, group = g)
 }
+
+
+instantiateResampleInstance.FixedWindowCVDesc = function(desc, size, task = NULL, coords) {
+  initial.window.abs = scaleWindows(desc$initial.window, size)
```
larskotthoff: Why isn't the scaled initial window stored in the description object like horizon and skip?

SteveBronder: I put the other ones in desc for convenience. I can add initial window to change the value in `desc` as well

larskotthoff: Up to you, but I would make it consistent -- either all or nothing in desc, and if it goes in there I would document that.

SteveBronder: Word, I just set them both to be copied with meaningful names in the func

SteveBronder:
So both growing and fixed window only differ in the 4 lines of code that actually create the indices.

Would it be a good idea to abstract this out and if so where should that function go? helpers?

larskotthoff:
Good idea. Yes, that would go in helpers.

larskotthoff:
Fantastic, thanks for your patience. Merging.

larskotthoff:
Oh and add yourself to the authors list!

SteveBronder:
Awesome! I'll add something to the tutorial this weekend

PR #2122, opened 2017-12-24
Travis: Install R packages as binaries from ppa and reduce build stages

pat-s:
# Advantages

* installation of most R packages as binaries from [c2d4u ppa](https://launchpad.net/~marutter/+archive/ubuntu/c2d4u) 
* solves the problem with rebuilding the cache as most of the installation does not rely on caching anymore (only the packages which are not available as binaries). **IMO thats the major advantage of this PR.**
* single build stage instead of multiple ones as before (warmup1, warmup2, check)
* supports the effort of @marutter (ppa maintainer) building binaries for multiple ubuntu versions

The advantage of using binaries is the highly reduced built time for packages like `Rfast`, `dplyr` etc. that take a lot of minutes when compiled from source. This applies for every new R update when the cache needs to be rebuilt and caused a lot of trouble for @berndbischl and others in the past. 

# Disadvantages

* Apt packages cannot be cached like R packages on Travis and need to be downloaded and installed for every build. This increases build time a bit (current build time 1h 8 min) since downloading a working R package cache is of course faster (current master builds in 1h 1 min).
* New required packages new to be added to `.travis.yml` by hand

# Notes

* @mb noticed that some packages are not at the latest CRAN version (e.g. `data.table` being a minor version behind). I've contacted the maintainer to ask how the building process is done and why this is the case for this (and maybe some other?) packages.
 
Note 09/01/2018: Still no response.


mb706:
Another disadvantage seems to be that all required packages need to be listed explicitly in `.travis.yml`?

> It seems to me that install_github() always reinstalls all dependencies, no matter if deps are already available?

This sounds like a bug in the `devtools` version used on travis? When I `install_github("mlr-org/mlr")` on my own machine, only `mlr` is installed. Do you have a link to a travis build log where this happened?

> Are there any crucial new features in the Github versions of these packages? Build also passes with the CRAN versions.

The question is not whether there are crucial new features *now*, but *ever*. The github versions of `berndbischl` packages sometimes get new features that are used by the github version of `mlr` right away; having to push these changes to CRAN before even being able to test `mlr` with them on travis would probably be an annoyance.

pat-s:
> Another disadvantage seems to be that all required packages need to be listed explicitly in .travis.yml?

Yes, so if we require a new package we need to add it by hand to .travis.yml

> This sounds like a bug in the devtools version used on travis? When I install_github("mlr-org/mlr") on my own machine, only mlr is installed. Do you have a link to a travis build log where this happened?

It seems that the `r-cran-*` packages are not properly detected (maybe because they are installed in a different folder) and `install_github` therefore thinks that the packages are not installed?

> The question is not whether there are crucial new features now, but ever. The github versions of berndbischl packages sometimes get new features that are used by the github version of mlr right away; having to push these changes to CRAN before even being able to test mlr with them on travis would probably be an annoyance.

Right, makes sense. I'll try to tackle the issue with the re-installations of github packages. Maybe it would be sufficient if they could be present in the package cache so they would only be reinstalled for every R version bump :+1:

pat-s:
@mb706 in the current build it seems that `install_github()` does not reinstall as it detects the installation from the package cache. So it should be fine and Github packages will only be installed if the cached is cleared (with a new R version bump).

[This build](https://travis-ci.org/mlr-org/mlr/jobs/320975310) is an example where `devtools::install_github("BBmisc")` re-installs `data.table` although it was previously installed by `r-cran-data.table` via `apt`.

mb706:
Apparently `data.table` gets reinstalled because the installed version is `1.10.4-2` and the CRAN version is `1.10.4-3`. If travis doesn't end up having the most recent CRAN version I would also see that as a pretty grave disadvantage.

pat-s:
@mb706 sure, that would be problematic. Not sure how the building process is done but the repo seems to be pretty active (400 builds in the last month by a single person). Maybe one should ask how the building is done and what can be done to help the project. I think Linux binaries are an important point and if available, should also apply as the default for `install.packages()` in R (but thats a different point).

pat-s:
ping @larskotthoff, @mllg, @berndbischl

larskotthoff:
Since we got a longer time limit for the build, is the package caching still a major issue? Unless it is, I'm not in favour of merging this since it relies on the CRAN and ubuntu versions of packages to be in sync, which will not always be the case.

pat-s:
> Since we got a longer time limit for the build, is the package caching still a major issue? Unless it is, I'm not in favour of merging this since it relies on the CRAN and ubuntu versions of packages to be in sync, which will not always be the case.

Sure, I understand this point. I just thought that it would be cleaner by default and unbind us from potential caching problems in any way. However, of course it makes us depend on the package building of one person. 

Well, at least now we know how to possibly do it in the future :smile:

PR #2129, opened 2018-01-13
Add names for the arguments in the makeCostMeasure function

MinhAnhL:
Just a minor improvement of the makeCostMeasure-function.
When creating a costMeasure `costM = makeCostMeasure (id = "costs", minimize = TRUE, cost = matrix(c(0,9,8,0), 2,2))`, the print function should list all the extra arguments with its name. But it doesn't do it for `makeCostMeasure()`, instead it uses “<unnamed>” as names.
-> Add names in extra.args-lists within the makeMeasure function.


larskotthoff:
Tests seem to be failing.

MinhAnhL:
@larskotthoff tests also fail in the master branch in file: test_regr_range.R (at least when I test it on my laptop)

mllg:
Thanks.

PR #2132, opened 2018-01-16
fix #2125

### LINKED ISSUE #2125: Nested SpCV broken
It works with the [commit](https://github.com/mlr-org/mlr/commit/e865e491dc2062b14ebeb9d5314cec2532c1bc2a) it was introduced but some recent commit broke it.
It seems to only affect nested SpCV and not normal SpCV.

I'll investigate and provide a fix. 

``` r
# install_github("mlr-org/mlr@e865e491dc2062b14ebeb9d5314cec2532c1bc2a")
remotes::install_github("mlr-org/mlr")

library(mlr)

data("bc.task.spatial")

lrn_knn <- makeLearner("classif.kknn",
                       predict.type = "prob")
#> Loading required package: kknn

## Tuning in inner resampling loop
ps <- makeParamSet(makeIntegerParam("k", lower = 10, upper = 400),
                   makeIntegerParam("distance", lower = 1, upper = 100),
                   makeDiscreteParam("kernel",
                                     values = c("triangular", "epanechnikov",
                                                "biweight", "triweight", "cos",
                                                "inv", "gaussian", "optimal")))

ctrl <- makeTuneControlRandom(maxit = 5)
inner <- makeResampleDesc("SpCV", iters = 5)

wrapper_knn <- makeTuneWrapper(lrn_knn, resampling = inner, par.set = ps,
                               control = ctrl, show.info = F, measures = list(auc))

outer <- makeResampleDesc("SpRepCV", folds = 5, reps = 2)

library(parallelMap)
parallelStart(mode = "multicore", level = "mlr.tuneParams", cpus = 1)
#> Starting parallelization in mode=multicore with cpus=1.

set.seed(12345)
resa_knn_spatial <- mlr::resample(wrapper_knn, bc.task.spatial,
                                  resampling = outer, extract = getTuneResult,
                                  show.info = TRUE, measures = list(auc))
#> Resampling: repeated spatial cross-validation
#> Measures:             auc
#> Error in data.frame(task$env$data$x, task$env$data$y): arguments imply differing number of rows: 0, 788
```

Created on 2018-01-07 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

pat-s:
* Removed  [this hacky code part](https://github.com/mlr-org/mlr/blob/7541988ca65c4198907ce63fab7a46d2676449fb/R/Task_operators.R#L344-L349).

* Added a test for nested SpCV.

@mb706 
What's your opinion on the 

```r
task$env$data$x = NULL
task$env$data$y = NULL
```

parts, would it be "better" in terms of mlr coding style (and maybe robustness (?)) to use `getTaskData()` here instead of setting the coordinates `x` and `y` in `$env$data = NULL`?

# Details

When arriving in `train()` at around [l.77](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/train.R#L77) in a `SpCV` call (without tuning), I need to remove coordinates `x` and `y` there because in this setting the models gets trained right [afterwards](https://github.com/mlr-org/mlr/blob/7541988ca65c4198907ce63fab7a46d2676449fb/R/train.R#L96). 

While this works fine in `SpRepCV` because the partitioning is already done at this point, it does not work in a nested setting with `tuneParams()`. 
There, the partitioning is done after I arrive at `train()` [l.77](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/train.R#L77) and if the code removes the coordinates at this point, they are not available when the resampling instance is created later on.

So I somehow have to check whether we are in a `SpRepCV` **with tuning** or `SpRepCV` **without tuning**. 

I now do so by checking in the `pars` object existing at `train` [l.70](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/train.R#L70)) whether the `.learner$id` has "tuned" in its name, e.g. `classif.ksvm.tuned`.
This is only the case if we are in a nested cv setting with tuning.
Subsequently, I make the removal of the coordinates in `train()` conditional on this property.
If we are in a nested setting, I remove the coordinates for the tuning step at `tuneParams()` [l.124](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/tuneParams.R#L124-L127).

We are arriving at this "removal point" (`train()` [l.77](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/train.R#L77)) several times (basically for every model that is fitted, no matter if for tuning or performance estimation) and the `if` part is entered every time and coordinates are removed although they are not existent anymore. 
However, we need to do it at least once for every fold as then the task arrives at `train()` [l.77](https://github.com/mlr-org/mlr/blob/248356ba0412ba5d7d3eb0599604b078ef507fa8/R/train.R#L77) with coordinates.

### DIFF
--- a/R/Task_operators.R
+++ b/R/Task_operators.R
@@ -333,20 +333,6 @@ getTaskData = function(task, subset = NULL, features, target.extra = FALSE, reco
     if (recode.target %nin% c("no", "surv")) {
       res[, tn] = recodeY(res[, tn], type = recode.target, task$task.desc)
     }
-    # first condition checks if 'getTaskData' was called directly, i.e. checks if
-    # the call was from the GlobalEnv
-    # second condition checks which function called 'getTaskData' if call was not
-    # from the GlobalEnv. If cond2 is FALSE, 'getTaskData' was called from
-    # 'subsetTask' in a nested resampling call. In this case we remove x and y
-    # later (later = we arrive in 'getTaskData' twice in a 'resample' call) as
-    # we still need it for partitioning in upcoming function calls and only need
-    # to remove `x` and `y` before we proceed to the training step.
-    if (!identical(parent.frame(n = 1), globalenv()) &&
-        !sys.call(-2) == "subsetTask(.task, .subset)" &&
-        task$task.desc$is.spatial == TRUE) {
-      res$x = NULL
-      res$y = NULL
-    }
   }
   res
 }
--- a/R/predict.R
+++ b/R/predict.R
@@ -118,10 +118,7 @@ predict.WrappedModel = function(object, task, newdata, subset = NULL, ...) {
       on.exit(options(warn = old.warn.opt))
       options(warn = -1L)
     }
-    if (td$is.spatial == TRUE) {
-      pars$.newdata$x = NULL
-      pars$.newdata$y = NULL
-    }
+
     time.predict = measureTime(fun1({p = fun2(fun3(do.call(predictLearner2, pars)))}))
 
     # was there an error during prediction?
--- a/R/train.R
+++ b/R/train.R
@@ -72,6 +72,13 @@ train = function(learner, task, subset = NULL, weights = NULL) {
   vars = getTaskFeatureNames(task)
   # no vars? then use no vars model
 
+  # remove coordinates from data for train call
+  # only do so if we are NOT in a nested cv setting -> there we remove the coordinates in tuneParams()
+  if (isTRUE(task$task.desc$is.spatial) && !grepl("tuned", pars$.learner$id)) {
+    task$env$data$x = NULL
+    task$env$data$y = NULL
+  }
+
   if (length(vars) == 0L) {
     learner.model = makeNoFeaturesModel(targets = task$env$data[, tn], task.desc = getTaskDesc(task))
     time.train = 0
--- a/R/tuneParams.R
+++ b/R/tuneParams.R
@@ -120,6 +120,13 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
     messagef("With control class: %s", cl)
     messagef("Imputation value: %g", control$impute.val)
   }
+
+  # remove coordinates from data for follow-up train() call
+  if (isTRUE(task$task.desc$is.spatial)) {
+    task$env$data$x = NULL
+    task$env$data$y = NULL
+  }
+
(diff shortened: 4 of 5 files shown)


mb706:
I don't like `task$env$data$x = NULL` . You should not write to elements of `task$env`, because it is an `environment`. It so happens that the `subsetTask()` call above makes a deep copy of the task, so in this case this is without consequences, but if that ever changes (if some changes causes the `subsetTask` call to be skipped for optimization, for example) this would be a sneaky bug to catch.

mb706:
Also, you should not check for whether the learner name has "tuned" in its name. The name of a learner can be set upon creation (the `id` parameter).

mb706:
When I was writing my comment on #2125 I didn't understand why you would take the `x` and `y` columns out of the data. If the spatial coordinates are only used for resampling, it makes sense to not have `getTaskData` return them by default. I can think of two possibilities to "solve" this:
1. (Simple) Use a `get.coordinates` (or whatever) flag in `getTaskData`. It should be set in the call in `subsetTask` (because `subsetTask()` of a spatial task should itself be a spatial task), as well as all other calls to `getTaskData` that are used to create copes of a task in some way. Otherwise it should default to `FALSE`. 
2. (Laborious) Save the spatial coordinates in some extra data structure, similar to `blocking`. Make sure that `subsetTask` gets this extra data and handles this correctly (as well as all other places where a task is copied; same as above, basically). `makeXXXTask()` would take the spatial coordinates as a separate dataframe (or matrix), instead of using the hard-coded `x` and `y` columns.

Option 2. is more consistent with the design of `mlr` so far, because the coordinates are similar, in kind, to the `blocking` and possibly `weight` information: They are meta-information and not stored in the same `data.frame` as the data that is given to the learners. Since that would probably mean a lot of rewriting of code you already wrote, however, I think you should probably go for option 1. and earmark this for future refactoring (I think it would be nice to have a unified framework for all the row-wise meta-information inside a task at some point in the future, but for that the `weights` and `blocking` would *also* need refactoring).

larskotthoff:
I think option 2 is better (and I believe we discussed this in the initial PR as well).

larskotthoff:
Unless this absolutely needs to be fixed now, I would prefer to replace with the new approach.

pat-s:
see #2134

PR #2123, opened 2017-12-25
Move from mkdocs to pkgdown and integrate mlr-tutorial into mlr

pat-s:
Hi guys,

sorry for the mass of modified files but I see no option how I could do this in another way?

Looking forward hearing your opinion on this PR :smiley: 

For Travis, please refer to the [recent builds of my fork](https://travis-ci.org/pat-s/mlr) and not the PR 
one.

# Before merging

- [ ]  Update the .travis.yml. Currently its tailored to my fork.

# Features

* Live preview of [mlr-tutorial (devel)](https://pat-s.github.io/mlr/) (will of course be changed from `pat-s` url to `mlr-org` when being merged)
* Integration of [mlr-tutorial (release)](https://pat-s.github.io/mlr/articles/tutorial/release/index.html) with all internal links working. I needed to manually edit most of the `release` files to get the internal links working while maintaining the same url structure as `devel`. However, this should have been a one-time thing and not needed to be done again after the next CRAN release.
* Linking to `pkgdown` implementations of other packages
* Fully R based
* Integrated NEWS menu in menu bar
* Cheatsheet linked on main page
* Links to slack, stackoverflow, github and blog in menu bar
* Better and cleaner look
* Easier build process
* Orcid ID linked next to author name in main page
* Search bar (right now a bit buggy as search need to be entered twice)
* Easier travis build process
* Build of new PDF version based on `bookdown::pdf_document2()` [release](https://github.com/pat-s/mlr/blob/gh-pages/articles/tutorial/release/mlr-tutorial.pdf), [devel](https://github.com/pat-s/mlr/tree/master/vignettes/tutorial/devel/pdf/mlr-tutorial.pdf)
* Uses the new logo that was proposed some months ago in the Slack group

# Miscellaneous
* Cleaned up `mlr-tutorial` sources (new line for every new sentence, consistent new line spacing, consistent space around code-chunks)
* Added [mlr-contribution guide](https://pat-s.github.io/mlr/#mlr-tutorial) to `README.md`
* Older `mlr-tutorial` versions starting from =< 2.10 are placed in `vignettes/tutorial/old/` and are still accessible but only file-by-file, i.e. without working internal linking, e.g. [Bagging section from 2.10](https://pat-s.github.io/mlr/articles/tutorial/old/2.10/html/bagging/index.html)

# Implementation 
* Site is served via `gh-pages` branch and deployed via `ropenscilabs/tic`. 
[...]

PR #2134, opened 2018-01-17
New spatial implementation approach

### LINKED ISSUE #2125: Nested SpCV broken

pat-s:
fixes #2125 
replaces e865e491dc2062b14ebeb9d5314cec2532c1bc2a

This was going faster than I initially thought. Ok, meanwhile I have a better overview of the `mlr` concept than last time.
Thanks to @mb706 and @larskotthoff for your patience and helpful hints :+1: :yellow_heart: 
Let's see what you think about this :smile: 

# Implementation details

* Coordinates are stored in the task during creation as a `data.frame` via argument `coordinates`
* Subsetting for partitioning in a nested setting is done in `instantiateResampleInstance.SpCVDesc` using the respective row indices from the incoming `task$env$data`
* No modification of `task$env` in any place
* No need for an argument `coordinates` in any `instantiateResampleInstance*` because the coordinates are stored within the task
* Cleaned up all the hackish stuff from the previous implementation
* Updated doc
* Added tests


R/SupervisedTask.R
```diff
@@ -1,13 +1,13 @@
-makeSupervisedTask = function(type, data, target, weights, blocking, fixup.data, check.data, spatial) {
-  task = makeTask(type = type, data = data, weights = weights, blocking = blocking, fixup.data = fixup.data, check.data = check.data, spatial = spatial)
+makeSupervisedTask = function(type, data, target, weights, blocking, fixup.data, check.data, coordinates) {
+  task = makeTask(type = type, data = data, weights = weights, blocking = blocking, fixup.data = fixup.data, check.data = check.data, coordinates = coordinates)
```
larskotthoff: Should the `coordinates` argument be checked here? Presumably it needs to have a certain structure.

R/Task.R
```diff
@@ -79,9 +73,9 @@
 #'   Should sanity of data be checked initially at task creation?
 #'   You should have good reasons to turn this off (one might be speed).
 #'   Default is \code{TRUE}.
-#' @param spatial [\code{logical(1)}]\cr
-#'   Does the task contain a spatial reference (coordinates) which should be used for
-#'   spatial partioning of the data? See details.
+#' @param coordinates [\code{data.frame}]\cr
+#'   X and Y coordinates of a spatial data set that will be used for spatial partitioning of the data in a cross-validation resampling setting.
+#'   Coordinates have to be in "Universal Transverse Mercator" (UTM) format, i.e. numeric values.
```
larskotthoff: Document the format -- columns named `x` and `y`, same number of rows as data?

R/Task.R
```diff
@@ -113,13 +107,13 @@ NULL
 #'   weights
 #' @param blocking [\code{numeric}\cr
 #'   task data blocking
-#' @param spatial [\code{logical(1)}]\cr
-#'   whether data is spatial
+#' @param coordinates [\code{logical(1)}]\cr
+#'   whether spatial coordinates have been provided
```
larskotthoff: This isn't a flag now, it's the actual coordinates, isn't it?

tests/testthat/test_base_resample_cv.R
```diff
@@ -38,6 +38,30 @@ test_that("SpCV instance works", {
...
+
+  wrapper.ksvm = makeTuneWrapper(lrn.ksvm, resampling = inner, par.set = ps,
+    control = ctrl, show.info = FALSE, measures = list(auc))
+
+  outer = makeResampleDesc("SpRepCV", folds = 2, reps = 2)
+
+  parallelStart(mode = "multicore", level = "mlr.tuneParams", cpus = 2)
+
+  resa.svm.spatial = resample(wrapper.ksvm, bc.task.spatial,
+    resampling = outer, show.info = TRUE, measures = list(auc))
```
larskotthoff: Could you add some expectations here on data/results please?

tests/testthat/test_base_resample_cv.R
```diff
@@ -38,6 +38,30 @@ test_that("SpCV instance works", {
...
+
+  ctrl = makeTuneControlRandom(maxit = 1)
+  inner = makeResampleDesc("SpCV", iters = 2)
+
+  wrapper.ksvm = makeTuneWrapper(lrn.ksvm, resampling = inner, par.set = ps,
+    control = ctrl, show.info = FALSE, measures = list(auc))
+
+  outer = makeResampleDesc("SpRepCV", folds = 2, reps = 2)
+
+  parallelStart(mode = "multicore", level = "mlr.tuneParams", cpus = 2)
```
larskotthoff: Not a fan of having parallelization in the tests -- is this absolutely necessary?

tests/testthat/test_base_spcv.R
```diff
@@ -0,0 +1,41 @@
...
+    control = ctrl, show.info = FALSE, measures = list(auc))
+
+  outer = makeResampleDesc("SpRepCV", folds = 2, reps = 2)
+
+  library(parallelMap)
+  parallelStart(mode = "multicore", level = "mlr.tuneParams", cpus = 2)
+
+  resa.svm.spatial = mlr::resample(wrapper.ksvm, bc.task.spatial,
+    resampling = outer, show.info = TRUE, measures = list(auc))
+  parallelStop()
```
larskotthoff: See above -- parallelization and checks.

tests/testthat/test_base_spcv.R
```diff
@@ -0,0 +1,41 @@
...
+test_that("SpRepCV works without errors", {
+
+  data(bc.task.spatial, package = "mlr", envir = environment())
+
+  learner = makeLearner("classif.ksvm", predict.type = "prob", kernel = "rbfdot")
+
+  resampling = makeResampleDesc("SpRepCV", fold = 2, reps = 2)
+
+  out = resample(learner = learner, task = bc.task.spatial,
+    resampling = resampling, measures = list(auc))
```
larskotthoff: Checks on data/results.

larskotthoff:
Thanks for your quick work on this! Looks like the tests aren't picking up the changed data (or some of the test data wasn't changed)?

R/SupervisedTask.R
pat-s: This is done [here](https://github.com/mlr-org/mlr/blob/aeda74198da8b51421db795e4666e591dd884b5c/R/Task.R#L148-L154). I think its sufficient?

R/Task.R
pat-s: Updated it in the public part but forgot it in the internal part. Why are there two docs actually?

R/Task.R
pat-s: Updated it in the public part but forgot it in the internal part. Why are there two docs actually?

tests/testthat/test_base_resample_cv.R
pat-s: Ok, removed it.

tests/testthat/test_base_resample_cv.R
pat-s: Did so, see https://github.com/mlr-org/mlr/blob/e4c026700acc840c8d53f8e9cdcd894e49483610/tests/testthat/test_base_spcv.R#L39 and
https://github.com/mlr-org/mlr/blob/e4c026700acc840c8d53f8e9cdcd894e49483610/tests/testthat/test_base_spcv.R#L24

tests/testthat/test_base_spcv.R
pat-s: Did so, see https://github.com/mlr-org/mlr/blob/e4c026700acc840c8d53f8e9cdcd894e49483610/tests/testthat/test_base_spcv.R#L39 and
https://github.com/mlr-org/mlr/blob/e4c026700acc840c8d53f8e9cdcd894e49483610/tests/testthat/test_base_spcv.R#L24

pat-s:
> Thanks for your quick work on this! Looks like the tests aren't picking up the changed data (or some of the test data wasn't changed)?

Yes, task feature `coordinates` was missing for the tasks, I only updated `task$tas.desc` first

I also modifiied `inst/makeData.R`. 
`phoneme` data was broken and I think the `cd..` calls are not needed unless we finally integrate the task creation into Travis. Currently tasks are only generated locally and now the script works as it should.

mb706:
Apparently your PR changes file permissions (adding the `x` bit), which makes viewing the differences quite hard:
view

This also happens in your PR #2123, the source seems to be this [rebase commit](https://github.com/mlr-org/mlr/commit/96f928886bfb34db84bfdae044fc95efd6672814#diff-1e70daafb475c0ce3fef7d2728279182). This has been around for a while and nobody comments on it, is it something with my github settings only I see this?

pat-s:
Thanks @mb706, did not notice that yet! Mysterious, cannot remember I altered any file permissions locally lately. 

I think [this answer](https://stackoverflow.com/a/44866012/4185785) addresses most likely my problem but it didn't help directly. Let's see if it helps setting all permissions back to 644 explicitly.

pat-s:
Furthermore I did not notice that I still had all the `pkgdown` implementation in this branch, obv. because I checked out from the master branch of my fork which holds all that stuff. Should really pay more attentation (or work less but with more focus, mhh ;) )

pat-s:
This [answer](https://stackoverflow.com/a/44866012/4185785) did the trick with a bit of additional modification `chmod +x thirdparty/gen_families.sh` for example.

I think it initially arised because I copied my local git repos to another laptop and git conflicted setting some file permissions on the files because all files were assigned to a different user/laptop.

pkgdown/extra.css
```diff
@@ -0,0 +1,24 @@
+@import url(https://cdn.rawgit.com/tonsky/FiraCode/1.204/distr/fira_code.css);
```
mb706: This file should probably go.

R/datasets.R
```diff
@@ -125,14 +125,32 @@ NULL
...
+#' one from the 'natural' part of the RBSF area and corresponds to landslide
+#' distribution in the year 2000.
+#' @name spatial.task
 #'
-#' @name bc.task.spatial
-#' @references Iturritxa, E., Mesanza, N., & Brenning, A. (2014). Spatial
-#' analysis of the risk of major forest diseases in Monterey pine plantations.
-#' Plant Pathology, 64(4), 880-889. doi:10.1111/ppa.12328
+#' @keywords datasets
+#' @keywords internal
```
mb706: Why internal?

R/SupervisedTask.R
larskotthoff: But the columns need to have specific names, don't they?

pat-s: No, not anymore.

inst/CITATION
```diff
@@ -29,7 +29,7 @@ bref = c(
       person("Helena", "Kotthaus"),
       person("Peter", "Marwedel"),
       person("Claus", "Weihs"),
-      person("Jörg", "Rahnenführer"),
+      person("Joerg", "Rahnenfuehrer"),
```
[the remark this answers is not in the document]
pat-s: This is also a leftover from the `pkgdown` implementation which cannot deal with umlauts in Citation files.

R/ResampleInstances.R
```diff
@@ -1,27 +1,28 @@
...
-  if (is.null(coords)) {
-    coords = data.frame(task$env$data$x, task$env$data$y)
-  } else {
-    coords = coords
-  }
+instantiateResampleInstance.SpCVDesc = function(desc, size, task = NULL) {
+
+  # subset the coordinates with respect to the observations in the data
+  # in a nested cv call we only have a subset of the whole data here (the
+  # indices from the training fold of the outer loop)
```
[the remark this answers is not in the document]
pat-s: Ok, done [here](https://github.com/mlr-org/mlr/blob/1df92b34ae56c77bf8aeee181426636509fd9fe2/R/ResampleInstances.R#L19-L22).

R/Task.R
```diff
@@ -151,12 +148,14 @@ makeTask = function(type, data, weights = NULL, blocking = NULL, fixup.data = "w
...
-
-  if (spatial == TRUE) {
-    # check if coords are named 'x' and 'y'
-    if (!all(c("x", "y") %in% colnames(data))){
-      stop("Please rename coordinates in data to 'x' and 'y'.")
+    if (!is.null(coordinates)) {
+      if (nrow(coordinates) != nrow(data)) {
+        stop("Coordinates have to be of the same length as number of rows in data! Or pass none at all.")
+      }
+      if (class(coordinates)[1] != "data.frame") {
```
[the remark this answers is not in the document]
pat-s: That skips the conversion for tibbles then but I think they cause no troubles anyway.

inst/makeData.R
```diff
@@ -1,95 +1,86 @@
...
-dn = "../data"
-stopifnot(isDirectory(dn))
 DATASEED = 7761  # nolint
 
 # classification
 set.seed(DATASEED)
 data(iris, package = "datasets")
 iris.task = makeClassifTask("iris-example", data = iris, target = "Species")
-save(iris.task, file = file.path(dn, "iris.task.RData"))
+use_data(iris.task, overwrite = TRUE)
```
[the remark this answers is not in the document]
pat-s: Ok, fair enough.

pat-s:
> Don't forget to adjust NEWS.md (the task attribute name is "has.coordinates" instead of "is.spatial" now)

Thanks, did it.

> I think you have to adjust subsetTask to also subset the coordinates.

This is currently done [here](https://github.com/mlr-org/mlr/blob/1df92b34ae56c77bf8aeee181426636509fd9fe2/R/ResampleInstances.R#L26). Should I try to use `subsetTask`for this?

R/ResampleInstances.R
```diff
@@ -1,54 +1,64 @@
...
+  if (is.null(task)) {
+    stopf("Please provide a task.")
+  }
+  if (is.null(task$coordinates)) {
+    stopf("Please provide suitable coordinates for SpCV. See ?Task for help.")
   }
-  # perform kmeans clustering
-  inds = kmeans(coords, centers = desc$iters)
-  inds = factor(inds$cluster)
+  else {
```
mb706: Unnecessary nesting, since `stopf` will end the program flow in any case.
Besides, we do
```R
} else {
```
instead of
```R
}
else {
```

inst/CITATION
```diff
@@ -45,7 +45,7 @@ bref = c(
     key = 'bischl2016class',
     author = c(
       person("Bernd", "Bischl"),
-      person("Tobias", "Kühn"),
+      person("Tobias", "Kuehn"),
```
mb706: Should probably also be undone?

mb706:
> > I think you have to adjust subsetTask to also subset the coordinates.
>
> This is currently done here. Should I try to use subsetTaskfor this?

Is there a reason to carry around all coordinates instead of subsetting right away in `subsetTask`? If not, I would do it just like [with weights and blocking](https://github.com/mlr-org/mlr/blob/7df31a01cff5db5342ce21f9dd923a0ac5ab8283/R/Task_operators.R#L400-L405).

pat-s:
> Is there a reason to carry around all coordinates instead of subsetting right away in subsetTask? If not, I would do it just like with weights and blocking.

Thanks, seems to work and make sense to have it the same way as for `weights` and `blocking`. 
No idea what I did not try that before although you mentioned it, could have been so easy..

FYI: I removed the unnecessary test `binomial_spatial` since it was broken due to the missing coordinates argument. Previously coordinates were stored in the data.
I think the tests in `base_spcv` are sufficient to check if the partitioning works as expected.

R/ResampleInstances.R
pat-s: Ok, removed it.

inst/CITATION
pat-s: Yes, thanks.

pat-s:
@larskotthoff @mb706 everythings fine now? :)

larskotthoff:
I'm still a bit confused about what the requirements on the coordinates are. From the code it looks like there are absolutely none as you're calling `kmeans` on the data frame and clustering whatever is in there. If this is correct, the comment saying that the coordinates need to be UTM should be adjusted accordingly.

pat-s:
Yes, I'm calling `kmeans` on the `task$coordinates` data.frame.
I need numeric vectors that describe coordinate information.
Maybe other projections as UTM might also work but this is untested.
As there is no point in checking a projection, I state that the user should use UTM projection in the help file. By this, it is ensured that numeric values are supplied and `kmeans` can handle them properly. 

The only requirements is that `coordinates` exists, is numeric and has the same number of rows as the data.

larskotthoff:
Then I would document the requirements in your last sentence. There's no reason for it not to work on non-UTM projections. I would also mention that the `coordinates` data frame is clustered by kmeans, so one coordinate would be enough or 3 would also work.

pat-s:
While one coordinate would be enough to get no error message `kmeans`, the partitioning is different than when using 2 coordinates (I just tested it with setting a seed).
This also seems logic to me as we are in a 2-dimensional space and supplying only 1 dimensions should result in a different outcome than using 2-dimensions.

Furthermore, when only one coordinate is supplied, in a nested cv setting the data.frame is somehow converted to a vector in the inner loop when it arrives at `subsetTask()` which then triggers an error.

So for safety and simplicity,  I would just require two coordinates. If coordinates exist, there will always be two anyhow (x and y).

larskotthoff:
Would more than two coordinates be acceptable? E.g. x, y, z.

mb706:
@pat-s are you saying clustering by 1 dimension gives different results from 2 dimensions because information is missing, or because something works fundamentally different in 1D? If data were collected in a 1D world, would the algorithm work?

> when only one coordinate is supplied, in a nested cv setting the data.frame is somehow converted to a vector in the inner loop when it arrives at subsetTask() which then triggers an error.

That sounds like a `drop = FALSE` is missing somewhere

pat-s:
> Would more than two coordinates be acceptable? E.g. x, y, z.

> @pat-s are you saying clustering by 1 dimension gives different results from 2 dimensions because information is missing, or because something works fundamentally different in 1D? If data were collected in a 1D world, would the algorithm work?

I think the algorithm would work in any dimension (1,2,3, X), see also `?kmeans`:

> numeric matrix of data, or an object that can be coerced to such a matrix (such as a numeric vector or a data frame with all numeric columns)

However, when only 1 dimension is supplied there will only be a clustering in this dimension ofc, assuming the second dimension (in a 2-D space) to be constant. This then leads to a different clustering than X-D.

The same for 3-D compared to 2-D and 1-D. 

While I think that 3-D can make sense if data was really collected in a 3-D space, 1-D should not be allowed - because 1D is actually 2D just with one dimension being constant.
If the second dimension is missing one should just supply a constant value for it.

Ofc this can then be said for X-D but I would say 2-D should be the baseline to prevent unwanted outcomes that produce no error.

So coming back to @larskotthoff question: I think yes, we should change it to >= 2D.
TBH I haven't really thought about this before as I only deal with 2-D but I can think of use cases for 3D (as this dimension is also affected by autocorrelation).

> That sounds like a drop = FALSE is missing somewhere

Since we are not allowing it anyways I'm not keen on inspecting this any further. 
If you are curious, the error is thrown by `subsetTask()` in the first nested iteration. The single coordinate arriving there is not a `data.frame` anymore as it was before in the outer resampling iteration but only a vector. This then causes an error in `data.frame[ , ]` obv.

larskotthoff:
Well, 2D data is just 3D data with one dimension constant. I'm not in favour of imposing artificial restrictions. The dimensions don't need to be just space, it could be time. Or something completely different that determines how things belong together. And for time, 1D would make sense.

Regarding the error for 1D, sounds like you just need to add an `as.data.frame()` or something like that somewhere.

pat-s:
> Well, 2D data is just 3D data with one dimension constant. I'm not in favour of imposing artificial restrictions. The dimensions don't need to be just space, it could be time. Or something completely different that determines how things belong together. And for time, 1D would make sense.

Sure, that was the baseline of my last comment.
However, this PR is specifically for spatial data. Idk how temporal correlations are accounted for in detail for performance estimation and if any guys are dealing with 3D data.
Please note that I only restricted on < 2-D.

Imo this PR is dedicated to spatial data, thats why the name of the resampling method is also "SpCV". 
Now we could start to discuss if 1D is already "spatial" or if only > 2-D is "spatial"... :smile: 
If there are similar attempts for temporal and other dimensions, we can think again of aggregating the approaches in a broader approach? 

> Regarding the error for 1D, sounds like you just need to add an as.data.frame() or something like that somewhere.

Sure, but this is not needed currently as a check before errors if `ncol(coordinates) < 2`.

larskotthoff:
Ok, I'll merge this for now to replace the existing broken implementation, but at some point we should revisit and see whether we can make this more general.

PR #2140, opened 2018-01-26
Featsel using mutual information with package praznik and FSelectorRcpp

smilesun:


larskotthoff:
Could you add some tests please?

bommert:
I don't know, if it's really a good idea to put all the praznik functions into one mlr filter.
At least there should be documentation that other praznik methods than _MIM_ can be used.
As far as I can see, you can pass anything for _criteria_, and if what you pass is not the name of a praznik function, you will run into some kind of error.

berndbischl:
+1 to what @bommert said. we should split this up, there can be an internal function makePraznijkFilter to reduce copy pase.

bommert:
I installed the featsel_praznik branch and I ran into problems:

```{r}
test.it = function(data, part) {
  task.data = getTaskData(data, target.extra = TRUE)
  
  # stratified selection of observations
  rdesc = makeResampleDesc("Holdout", split = part, stratify = TRUE)
  rin = makeResampleInstance(rdesc, data)
  ids.obs.new = rin$train.inds[[1]]
  
  data.new = cbind(task.data$data[ids.obs.new, ], target = task.data$target[ids.obs.new])
  new.task = makeClassifTask(data = data.new, target = "target")
  generateFilterValuesData(task = new.task, method = "praznik.MIM")
}

set.seed(2018)
test.it(sonar.task, 7 / 208)
```
Error in (function (X, Y, k = 3)  :  Value of SET_STRING_ELT() must be a 'CHARSXP' not a 'list'

```{r}
Bioresponse = getOMLDataSet(4134)
task = convertOMLDataSetToMlr(Bioresponse)

set.seed(123)
test.it(task, 0.1)
```
Error in cut.default(newX[, i], ...) : 'breaks' are not unique


I also get the second error for other praznik filter methods.

smilesun:
@bommert , D1758 column of this dataset 4134 is constant?

bommert:
D1758 is not constant in the entire dataset. 3704 times it has the value 0 and 47 times the value 1.

smilesun:
@bommert , i mean in the resampling, sometimes it only select 0 or only select 1 although the stratify is set to true.

bommert:
Yes, in the subsampled part of the dataset it is constant. I don't think the user should need to take care of this. It is something that can happen for many datasets with features that have the same value for many instances.

smilesun:
@bommert ,  I found something very interesting.  For the first bug, I post process the split in the resampling by making those constant numerics factors without cutting them using cut.  For the bug 
```
Browse[3]> input$k = 1750
Browse[3]> do.call(what = MIM, args = list(X = input$X, Y = input$Y, k = input$k))
Error in (function (X, Y, k = 3)  : 
  Value of SET_STRING_ELT() must be a 'CHARSXP' not a 'integer'
```
input$k = 1750 is the critical point, setting input$k = 1749 actually work, and  I have now figured out why after the afternoon, since there are 26 constant columns in the resampling.

bommert:
Having had a closer look at the praznik filters, I realized that most of them are greedy iterative procedures. In each iteration they select the feature which maximes a score. But the maximal scores per iteration do not necessarily decrease with the number of iterations. In fact, it' easy to find an example where this is not true:
```
X = iris[, 1:4]
Y = iris[, 5]
praznik::MRMR(X = X, Y = Y, k = 4)

$selection
 Petal.Width Petal.Length  Sepal.Width Sepal.Length 
           4            3            2            1 

$score
 Petal.Width Petal.Length  Sepal.Width Sepal.Length 
  0.90569521  -0.09881598  -0.21764702  -0.16122391 
```
In this example, feature _Sepal.Length_ is chosen last but does not have the smallest score.

If I calculate the filter scores with mlr, the selection order is thrown away and only the scores are kept.
```
generateFilterValuesData(iris.task, "praznik.MRMR")

FilterValues:
Task: iris-example
          name    type praznik.MRMR
1 Sepal.Length numeric  -0.16122391
2  Sepal.Width numeric  -0.21764702
3 Petal.Length numeric  -0.09881598
4  Petal.Width numeric   0.90569521
```

For FilterWrappers this should not be a problem as you set _k_ of MRMR to _nselect_.
Is there anything else where this behavior might cause problems?

smilesun:
I pushed a wrong implementation during the fix the praznik c problem, I will fix this soon. I though you are still using the old version of fd2f28a

smilesun:
After discussion with Andrea, I finally understand the problem: in Praznik package, scores are generated iteratively, if we do not select the full feature sets, unselected features will have the score NA. Instead, in a normal filtering like variance, all feature scores could be generated in parallel which does not depend on each other.  Filtering methods like variance could let it be possible to compare the ranking of features while praznik could not. Is this what you want to say? @bommert . This will be an issue to discuss.

bommert:
Yes. For praznik you cannot generate all scores and get the feature ranking from those scores. For the other filter methods you can do that.

smilesun:
@bommert , I changed some integration code and now it works with highdimensional data for SVM. ksvm still does not work. 
```
library(mlr)
data(chin)
dat = cbind(as.data.frame(chin$x), y = chin$y)
task = makeClassifTask(dat = dat, target = "y")
d = getTaskData(task)
lrn = makeLearner("classif.svm.hd")
mod = train(lrn, task)

```

smilesun:
the praznik problem is fixed:
```
library(OpenML)
Bioresponse = getOMLDataSet(4134)
task = convertOMLDataSetToMlr(Bioresponse)
df = getTaskData(task, target.extra = TRUE)
h = mlr::generateFilterValuesData(task)
```
works correctly now.

bommert:
The *glmnet.lasso* filter is not working properly:

```
generateFilterValuesData(sonar.task, "glmnet.lasso")
```
gives
```
Error in storage.mode(y) <- "double" :  invalid to change the storage mode of a factor
```

smilesun:
lasso bug fixed

bommert:
I ran into a new issue with the praznik filters:

```
lrn = makeLearner("classif.rpart")

lrn1 = makeFilterWrapper(lrn, fw.method = "praznik.CMIM", fw.perc = 0.001)
train(lrn1, sonar.task)
```
throws the error 
```
Error in (function (X, Y, k = 3)  : Parameter k must be positive
```
For other filters, this works fine, e.g.
```
lrn2 = makeFilterWrapper(lrn, fw.method = "anova.test", fw.perc = 0.001)
train(lrn2, sonar.task)
```

The problem is, that 0 features should be selected, but the praznik functions do not accept k=0.
It happend to me, because I tried to tune fw.perc in the range [0, 1] and appearently a very small value got selected by chance.


I think this problem could be fixed by adding
```
k = max(k, 1)
```
after line 659 in helper.cmi.praznik
```
k = min(nselect, length(featnames))
```

In fact, I think _nselect_ can never get larger than _length(featnames)_, if you have a look at how it is defined in filterFeatures.R. So I think that the definition of _k_ could be repaced by
```
k = max(nselect, 1)
```

What do you think?

smilesun:
I agree.

mllg:
What the status here? Do you still have outstanding issues?

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
   }
 )
 
+# Preprocess the data before applying praznik filter. According to Praznik documentation:
+# The best practice is to convert data to factors prior to feeding them in this function. Real attributes are cut into about 10 equally-spaced bins, following the heuristic often used in literature.Precise number of cuts depends on the number of objects; namely, it is n/3, but never less than 2 and never more than 10.
+preprocess.cmi.praznik = function(data) {
+    colns = colnames(data)
+    data = convertDataFrameCols(data, logicals.as.factor = TRUE, chars.as.factor = TRUE)
+    int.yes = vapply(data, is.integer, FUN.VALUE = TRUE)
```
mllg: Use `v[lnc]apply` where appropriate (helper in BBmisc)

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+    if (any(int.yes)) data[int.yes] = lapply(data[int.yes], as.factor)
+    numeric.yes = unlist(lapply(data, function(x) class(x) == "numeric"))
```
mllg: Use `is.numeric()`

mllg: And again, `vlapply()`.

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+    df.num = data.frame(data[, numeric.yes])
+    interval = min(as.integer(nrow(data) / 3.0), 10L)
+    interval = max(interval, 2L)
+    cols.unique.len = unlist(lapply(df.num, function(x) length(unique(x))))
+    const.cols.yes = unlist(lapply(df.num, function(x) length(unique(x)) < 2L))
```
mllg: `viapply()`

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+    colnames(df.num) = colns[numeric.yes]
+    df.nonnum = data.frame(data[, !numeric.yes])
+    colnames(df.nonnum) =  colns[!numeric.yes]
+    data = cbind(df.nonnum, df.num)
+    return(data)
+}
+
+helper.cmi.praznik = function(criteria, preprocess = FALSE) {
+  candiates = c("JMI", "DISR", "JMIM", "MIM", "NJMIM", "MRMR", "CMIM")
+  checkmate::assert_choice(criteria, candiates)
```
mllg: checkmate is imported, no need to explicitly call its functions with the namespace.

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+    data = getTaskData(task)
+    featnames = getTaskFeatureNames(task)
+    targetname = getTaskTargetNames(task)
+    if (preprocess) data = preprocess.cmi.praznik(data)
+    X = data[, featnames]
+    Y = data[, targetname]
+    k = min(nselect, length(featnames))
+    k = max(k, 1)
+    input = list(X = X, Y = Y, k = k)
+    algo = eval(parse(text = criteria))
```
mllg: Never use `eval(parse(text = ...))`. `getFromNamespace()` should work here.

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+  criteria = paste0("praznik::", criteria)
+  function(task, nselect, ...) {
+    org.featnames = getTaskFeatureNames(task)
+    task = removeConstantFeatures(task)  # without removing constant features, praznik will generate Rcpp error
+    data = getTaskData(task)
+    featnames = getTaskFeatureNames(task)
+    targetname = getTaskTargetNames(task)
+    if (preprocess) data = preprocess.cmi.praznik(data)
+    X = data[, featnames]
+    Y = data[, targetname]
```
mllg: `getTaskData(..., target.extra = TRUE)`?

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+#' @rdname makeFilter
+#' @name makeFilter
+#'
+makeFilter(
+  name = "praznik.MIM",
+  desc = "conditional mutual information based feature selection filters",
+  pkg = "praznik",
+  supported.tasks = "classif",  # FIXME: still investigating if regression task could be used
+  supported.features = c("numerics", "factors", "integer", "character", "logical"),
+  fun = helper.cmi.praznik("MIM")
```
mllg: Instead of `getFromNamespace` you could also just pass the function here, right?

smilesun:
from my side, no

R/Filter.R
```diff
@@ -619,6 +619,282 @@ makeFilter(
...
+  fun = function(task, nselect, ...) {
+    mdata = getTaskData(task)
+    fns = getTaskFeatureNames(task)
+    tns = getTaskTargetNames(task)
+    X = mdata[, fns]
+    y = mdata[[tns]]
+    X = as.matrix(X)
+    family = "binomial"
+    if (length(levels(y)) > 2) family = "multinomial"
+    fit = glmnet::glmnet(X, y, family = family)
```
mllg: Not sure if this is the best way. Do we want to provide custom `lambda` values?

R/RLearner_classif_ksvm_hd.R
```diff
@@ -0,0 +1,67 @@
...
+      makeNumericVectorLearnerParam(id = "class.weights", len = NA_integer_, lower = 0),
+      makeLogicalLearnerParam(id = "fit", default = TRUE, tunable = FALSE),
+      makeIntegerLearnerParam(id = "cache", default = 40L, lower = 1L)
+    ),
+    par.vals = list(fit = FALSE),
+    properties = c("twoclass", "multiclass", "numerics", "factors", "prob", "class.weights"),
+    class.weights.param = "class.weights",
+    name = "Support Vector Machines",
+    short.name = "ksvm",
+    note = "Kernel parameters have to be passed directly and not by using the `kpar` list in `ksvm`. Note that `fit` has been set to `FALSE` by default for speed.",
```
mllg: What does the "hd" stand for? How is this learner different from a SVM?

mllg:
@smilesun talk to me about this tomorrow so that we can make progress.

smilesun:
I think I am done,  if you have other comments @mllg

R/Filter.R
```diff
@@ -619,6 +619,254 @@ makeFilter(
...
 )
 
+# Preprocess the data before applying praznik filter. According to Praznik documentation:
+# The best practice is to convert data to factors prior to feeding them in this function.
+# Real attributes are cut into about 10 equally-spaced bins, following the heuristic often used in literature.
+# Precise number of cuts depends on the number of objects; namely, it is n/3, but never less than 2 and never more than 10.
+preprocess.cmi.praznik = function(data) {
+    colns = colnames(data)
+    data = convertDataFrameCols(data, logicals.as.factor = TRUE, chars.as.factor = TRUE)
+    int.yes = vlapply(data, is.integer)
```
mllg: Do we really want this? I think we should treat them like numeric values?

R/Filter.R
```diff
@@ -619,6 +619,254 @@ makeFilter(
...
+    df.num = data.frame(data[, numeric.yes])
+    interval = max(min(as.integer(nrow(data) / 3L), 10L), 2L)
+    const.cols.yes = vlapply(df.num, function(x) length(unique(x)) < 2L)
+    if (any(const.cols.yes)) {
+      df.num.uni =  lapply(df.num[const.cols.yes], as.factor)
+      df.num.cut = df.num[!const.cols.yes]
+      df.num.cut = data.frame(apply(df.num.cut, 2L, cut, interval))
+      df.num = cbind(df.num.cut, df.num.uni)
+    }
+    colnames(df.num) = colns[numeric.yes]
```
mllg: AFAICT the LHS has potentially more columns than the RHS. This would lead to vector recycling. I'm not sure what you are doing here.

R/Filter.R
```diff
@@ -619,6 +619,254 @@ makeFilter(
...
+    df.nonnum = data.frame(data[, !numeric.yes])
+    colnames(df.nonnum) =  colns[!numeric.yes]
```
mllg: Where do we loose the colnames? Do we have to re-set them?

R/Filter.R
```diff
@@ -619,6 +619,254 @@ makeFilter(
...
+    data = cbind(df.nonnum, df.num)
+    return(data)
+}
+
+helper.cmi.praznik = function(fun, preprocess = FALSE) {
+  force(fun)
+  force(preprocess)
+  function(task, nselect, ...) {
+    org.featnames = getTaskFeatureNames(task)
+    task = removeConstantFeatures(task)  # without removing constant features, praznik will generate Rcpp error
```
mllg: This is not feasible, I guess. A filter has to return a score for every feature, not only a subset. You probably just need to return a very bad score for these features.

man/makeFilter.Rd
```diff
@@ -173,6 +183,47 @@ prediction rule is: class 1 if the feature exceeds a threshold and class 0
...
+
+Filters in the praznik package using mutual information criteria greedy search
+Features with higher scores are considered more important features
+Currently only tested with classification task with non-missing values
+candiates = c("JMI", "DISR", "JMIM", "MIM", "NJMIM", "MRMR", "CMIM")
+
+Filters in the praznik package using mutual information criteria greedy search
+Features with higher scores are considered more important features
+Currently only tested with classification task with non-missing values
+candiates = c("JMI", "DISR", "JMIM", "MIM", "NJMIM", "MRMR", "CMIM")
```
mllg: ???

tests/testthat/test_featsel_fselector_rcpp.R
```diff
@@ -0,0 +1,28 @@
...
+    a = c(1, 2, 5.3, 6, -2, 4, 8.3, 9.2, 10.1)  # numeric vector
+    b = c("one", "two", "three")  # character vector
+    c = c(TRUE, TRUE, TRUE, FALSE, TRUE, FALSE)  # logical vector
+    d = c(1L, 3L, 5L, 7L, 9L, 17L)
+    f = rep(c("c1", "c2"), 9)
+    df = data.frame(a = a, b = b, c = c, d = d, f = f)
+    df = convertDataFrameCols(df, logicals.as.factor = TRUE)
+    lapply(df, class)
+    task = makeClassifTask(data = df, target = "f")
+    generateFilterValuesData(task, method = "FSelectorRcpp.infogain", nselect = 2L)
```
mllg: Missing unit tests

tests/testthat/test_featsel_fselector_rcpp.R
```diff
@@ -0,0 +1,28 @@
...
+    lrn = makeFilterWrapper(learner = lrn, fw.method = "FSelectorRcpp.infogain", fw.perc = 0.9)
+    res = resample(learner = lrn, task = binaryclass.task, resampling = cv10, measures = list(mmce, timetrain), extract = getFilteredFeatures, show.info = FALSE)
+    lrn = makeLearner("classif.randomForest")
+    lrn = makeFilterWrapper(learner = lrn, fw.method = "FSelectorRcpp.gainratio", fw.perc = 0.9)
+    res = resample(learner = lrn, task = binaryclass.task, resampling = cv10, measures = list(mmce, timetrain), extract = getFilteredFeatures, show.info = FALSE)
+    lrn = makeLearner("classif.randomForest")
+    lrn = makeFilterWrapper(learner = lrn, fw.method = "FSelectorRcpp.symuncert", fw.perc = 0.9)
+    res = resample(learner = lrn, task = binaryclass.task, resampling = cv10, measures = list(mmce, timetrain), extract = getFilteredFeatures, show.info = FALSE)
+
+    expect_true(TRUE)
```
mllg: This is not sufficient.

R/Filter.R
smilesun: Hi, do you mean whether we should split the data rather than keep the continuous values?  preprocess.cim.praznik is optional,  which we could switch off, but the author said in the docs that it is better to discretize continous values before hand. see praznik doc. what do you think? @mllg

R/Filter.R
smilesun: I just reset them to make them consistent with original data

smilesun: we lose the colnames in line 643

R/Filter.R
smilesun: Could you give more detail? I did not get what you mean here?

mllg:
I don't have a solution for the problem reported by @bommert currently. A simple fix would be to replace the scores with the selection order.

bommert:
How about replacing

```
score = fun(X, Y, k = k)$score

if (length(score) < ncol(X)) {
  unscored = sample(setdiff(names(X), names(score)))
  score = c(score, setNames(rep.int(NA_real_, length(unscored)), unscored))
}

score
```

by

```
sel = fun(X, Y, k)$selection
res = length(sel):1
names(res) = names(sel)
        
unscored = sample(setdiff(names(X), names(sel)))
res = c(res, setNames(rep.int(NA_real_, length(unscored)), unscored))
res
```

in the praznik filter function?

mllg:
A last review would be great @smilesun @bommert ... Did I miss something?

smilesun:
@mllg,  sorry have been sick for a while and the past weeks have been working on correcting exam and making new ones.  I think `score = setNames(rev(seq_along(selected)) / length(selected), selected)` is a good idea since score absolute value is not important.

pat-s:
Hi guys, I would like to push here a bit and get this merged. 

Some structural points:

* Can we deprecate the `FSelector` methods? I don't see why we should keep them on the same level as other filter methods. Might possibly confuse users.
* We should also update the tutorial and not this change (praznik new, tremendous speedup) in the tutorial and NEWS.md

From a technical point of view, the PR seems to be ready? @mllg

pat-s:
We should update the `praznik` filters with more accurate descriptions: https://deploy-preview-2140--nervous-hopper-4136be.netlify.com/articles/tutorial/filter_methods.html

pat-s:
@larskotthoff 	@berndbischl 	@ja-thomas 	
Can we get a review here to finally merge this? Michel is on vacation AFAIK and I would like to get all these filter PRs merged step by step.

pat-s:
ping again @larskotthoff 	@berndbischl 		@mllg 	@ja-thomas

pat-s:
Great. Merging.

berndbischl:
@mllg @pat-s can we please be a bit careful with the API changes? 
first of all, NEWS is not complete. you claim you just changed the implementation of the filters. seems untrue. The new filters are called FSelectorRCPP.*.

I am not saying that is a bad naming scheme. It actually looks good. But NEWS needs to be correct. 
And if we proceed like that:
a) why are the Fselector filters now removed? they could have stayed in?
b) if they are removed why is fselector still in SUGGESTS?

pat-s:
> first of all, NEWS is not complete. you claim you just changed the implementation of the filters. seems untrue. The new filters are called FSelectorRCPP.*.

true. I'll correct this.

> why are the Fselector filters now removed? they could have stayed in?

I found it unnecessary to have multiple implementations of the same filter methods.
But yeah, seeing that they are not completely the same and that we also have multiple versions of other filters (e.g. mrmr from praznik and mrmre) we should probably leave them in. 

> if they are removed why is fselector still in SUGGESTS?

leftover.


Your comments would have been a nice review :) But as I do not know when and if  you can find time for a review I am happy if Lars or Michel finds time and then we merge. More eyes usually see more (I would also appreciate reviews from other people but this never happens).
But luckily we can correct these points in a new PR.

berndbischl:
> Your comments would have been a nice review :)

sorry, wasnt fast enough :(
mistakes happen. normal. and i did check at least now....

do you want to do a correcting, final PR?

pat-s:
> sorry, wasnt fast enough :(
> mistakes happen. normal. and i did check at least now....

no problem, this was just meant as an explanation from my side why we merged it :)

> do you want to do a correcting, final PR?

sure

zzawadz:
@mllg 

> AFAICT the Rcpp implementation does some things differently (e.g., handling of factor variables) which leads to slightly different results (in addition to numerical differences). So this is not completely backward compatible to previous mlr versions. Nevertheless, the NEWS clearly state this change. For me this is acceptable, and getting rid of a Java dependency seems to be worth it.

We are before the next release of the FSelectorRcpp, so we still have some time to make some changes. Right now I think that the biggest difference is that FSelectorRcpp treats `integer` columns like factors and FSelector treats them like numerics. I added a param to control this, so the user can select which type of the behavior he prefers.

I'm more fond to FSelectorRcpp's default, so integers are like factors. However, I can change this to mimic the FSelector. So it will be more backward compatible:)

What do you think?

mllg:
We can overwrite the default in mlr, so basically it's up to you. However, I believe that most R users would expect integers to get treated like doubles. If you only have a few distinct values, you are typically expected to encode as `factor`/`ordered`.

berndbischl:
+1

zzawadz:
I'll change the defaults on my side. We don't want to have a lot of issues like "Why my results are different between mlr and FSRcpp?";)

pat-s:
Great! 
Follow-up question then: Do we still want to keep the `FSelector` filters then if results are the same? Or replace them with the `FSelectorRcpp` ones?

zzawadz:
@pat-s I don't know if they are "always" the same. I need to check how the `NA` are handled in both packages. It's still an untouched issue in FSelectorRcpp (right now we just omit `NA` column wise).

berndbischl:
> Ok, if it is too "unsure" if they are the same always we better leave them in for now (or re-add in this case).

probably better. that would also allow simple technical tests... to check ether they REALLY are the same..

zzawadz:
@berndbischl Yes. They're some tests inside FSelectorRcpp that compare the output with FSelector. In a lot of cases the output should be same, but I don't think that we cover every corner cases. So I would leave this as is.

PR #2143, opened 2018-01-29
Enable roxygen2 markdown support globally

pat-s:
### Reason for the global change
I am about to document a new function (coming soon) using the `roxygen2` markdown syntax.
I think everybody agrees that the markdown syntax makes it easier to document functions than the old .Rd syntax.
Therefore, this PR enables the markdown syntax globally.

(An alternative would be to set a `@md` tag for only those documentation that have a markdown syntax but I think this would result in a confusing mix of markdown and non-markdown documentations).

### Notes

For conflicting files, markdown support was disabled by using the `@noMd` tag in the documentation of the respective file.

The conflicting files can be adjusted in a separate PR.
For conversion, I used https://github.com/r-lib/roxygen2md.
Now if one writes documentation for a function, markdown syntax can (and should) be used.
If still some complex .Rd syntax needs to be used, one can turn off the markdown style by using the mentioned `@noMd` tag in the documentation part.

Conflicting files which are still in the old .Rd syntax:

* generateLearningcurveData.R
* generatePartialDependence.R
* benchmark.R
* batchmark.R
* Measure_custom_resampled.R
* FeatSelWrapper.R
* ModelMultiplexer.R
* Measure.R
* StackedLearner.R
* mergeBenchmarkResults.R
* selectFeatures.R
* TuneParamsMultiCrit.R


larskotthoff:
I like it, but it looks like there's still some old stuff left over -- build is failing.

pat-s:
Yes, seems like there is quite a bunch incorrect.

I just fixed all the errors appearing when running `devtools::document()` after conversion to markdown.
Let's take a look ;)

pat-s:
@larskotthoff All problems resolved.

I decided to go without the `()` in function calls. I thinks its only additional text that is annoying to write and read. However, if your opinion is different we can just use the commit before.

FYI: Rather than using `[numeric]` we now have to use `([numeric])` because everything wrapped in "[ ]" is formatted to be a function link.
Maybe it makes sense to install the branch and check how the help files look now.

larskotthoff:
Thanks, let me know when this is ready.

pat-s:
@larskotthoff ready to merge

larskotthoff:
Thanks, I've had a look at the generated files and at least the ones I checked looked fine. I'm merging this -- could you please update the development instructions in the wiki?

Issue #2144, opened 2018-01-29
can mmpf please be moved to suggests?


berndbischl:
we seem to import it now, but the basic mlr functionality surely does not depend on it?

@zmjones @larskotthoff

zmjones:
yes it can. i had it that way initially but lars suggested we import it. can change it now if you like.

berndbischl:
> i had it that way initially but lars suggested we import it.

@larskotthoff why?

larskotthoff:
To avoid needing `mmpf::` all over the place. Fine with moving it.

PR #2145, opened 2018-01-29
move mmpf to suggests

zmjones:
for issue #2144 

all i do here is assume mmpf is installed and reference it using ::. is there a helper function i should use to throw a informative error message if it isn't installed? i know `requirePackageOrSkip` but that is in the test helper functions. i guess it will just say "mmpf::whateverfunction" isn't found by default.


mb706:
I think what you're looking for is [`requirePackages`](https://www.rdocumentation.org/packages/BBmisc/functions/requirePackages), already imported from `BBmisc`.

zmjones:
@mb706 cool thanks!

larskotthoff:
Thanks, could you also move it from import to suggests please?

zmjones:
yep ready for merge

larskotthoff:
Thanks, merging.

PR #2148, opened 2018-01-30
bugfix and new tests for generatePartialDependenceData

zmjones:
jacobian estimation bug which was hidden in tests because the
multiclass tests where the jacobian was estimated use the iris data
where the number of classes matched the n[1] arg

tests are added that would have caught it and n[1] was changed to
equal 4 as well as the actual bug being fixed


larskotthoff:
Great, thanks.

PR #2153, opened 2018-02-02
partial dependence plot fix and test for binary classification

zmjones:
found the issue while updating the tutorial, added two tests that
would have caught the tutorial bug and another one that could have
occured


zmjones:
ping @larskotthoff

pat-s:
So if @larskotthoff is currently busy and the build passes, lets merge this tomorrow at latest?

pat-s:
merging now so that the fixes for the tutorial can pass again.

PR #2154, opened 2018-02-04
Rewrite measure bac to allow multiclass

### LINKED ISSUE #1834: measures: bac and ber inconsistent?
why does bac not allow multiclass?

AlexEngelhardt:
This PR rewrites the bac measure and the measureBAC function:

- The implementation in bac$fun and measureBAC was redundant. Now bac$fun uses measureBAC
- The balanced accuracy is not anymore computed with the TPR and FPR, but just with the diagonal of the confusion matrix. This makes it able to handle multiclass
- I added two tests: One to make sure this rewrite returns the same result as the old formula, and one to ensure bac works with multiclass now.

Fixes #1834


AlexEngelhardt:
Travis fails with this error:

    object 'mldr_HL' not found
    1: expect_equal(mldr:::mldr_HL(multi.y, multi.p), measureMultilabelHamloss(multi.y, 
           multi.p)) at testthat/test_base_measures.R:657

This is code I haven't touched, and I can't explain why it broke. Does anyone know what happened there? Is it possible that the new version of `mldr` removed a function `mldr_HL` from it?

AlexEngelhardt:
Merged master and especially #2159 so the tests (will hopefully) pass.

AlexEngelhardt:
Hm, I have no idea why this test failed. Can someone help me here?

      ── 1. Error: FDA properties work (@test_base_fda.R#31)  ────────────────────────
      missing value where TRUE/FALSE needed
      1: expect_silent(train(lrn, subsetTask(tsk, features = 2))) at testthat/test_base_fda.R:31

ja-thomas:
yes, that has nothing to do with your PR, you have to wait until #2173 is resolved and then rebase/merge the master in your branch

AlexEngelhardt:
The tests pass. It seems ready to merge - I added the respective label. (Is that okay, workflow-wise, or is that usually done by a reviewer?)

AlexEngelhardt:
oops - misclicked on "close PR". :)

tests/testthat/test_base_measures.R
```diff
@@ -948,3 +948,25 @@ test_that("setMeasurePars", {
...
   expect_equal(mm$extra.args, list(foo = 1))
 })
+
+test_that("bac works as intended with multiclass tasks", {
+  lrn = makeLearner("classif.rpart")
+  mod = train(lrn, task = iris.task)
+  pred = predict(mod, task = iris.task)
+
+  perf = performance(pred, measures = bac)
+  expect_numeric(perf)
```
larskotthoff: Please check whether the actual performance number is correct here.

larskotthoff:
No, please check by hand one particular example and compare to that hand-computed value.

AlexEngelhardt:
I'm a bit confused :/ With "one particular example", you mean one task, not one observation, right?

If yes, do you mean I should rewrite the following line (line 931 in the latest commit) from the test into a different way of computation that yields the same result?

     expected.bac = mean(diag(table(pred$data$truth, pred$data$response) / table(pred$data$truth, pred$data$truth)))

larskotthoff:
See for example https://github.com/mlr-org/mlr/blob/master/tests/testthat/test_base_measures.R#L214 -- the errors there are hand-computed and compared to.

AlexEngelhardt:
Ah, I see. I changed the test case to the small manual example you linked to. Is this what you had in mind?

Would it make sense to incorporate that test case in the existing large `test_that("check measure calculations"` test case, or should I leave it swapped out there?

larskotthoff:
Thanks, that's fine. Merging.

AlexEngelhardt:
Oh okay, I see. 

I thought that hard-coded numbers are more error-prone than formulas, that's why I just discarded that possibility. But ok - now I know for next time.

berndbischl:
> I thought that hard-coded numbers are more error-prone than formulas, that's why I just discarded that possibility. But ok - now I know for next time.

it really is less error prone as it makes you think about the formula you implemented.

PR #2161, opened 2018-02-07
[WIP] Rewrite measure tests for new mldr package functions

### LINKED ISSUE #2159: Suspected: 'mldr' package update broke Travis tests
For the two pull requests #2140 and #2154, Travis fails with the same error:

    ── 1. Error: check measure calculations (@test_base_measures.R#657)  ───────────
    object 'mldr_HL' not found
    1: expect_equal(mldr:::mldr_HL(multi.y, multi.p), measureMultilabelHamloss(multi.y, 
           multi.p)) at testthat/test_base_measures.R:657
    2: quasi_label(enquo(object), label)
    3: eval_bare(get_expr(quo), get_env(quo))
    4: mldr:::mldr_HL
    5: get(name, envir = asNamespace(pkg), inherits = FALSE)

I suspect that the new mldr package version dropped the mldr_HL function.

Can someone confirm? If yes, I can try and fix it.

AlexEngelhardt:
- hamming_loss works
- accuracy works
- subset_accuracy works
- precision works

Problems:

- mldr:::mldr_FMeasure (old) / mldr::fmeasure (new)
  - The mldr issue linked in line 688 seems fixed, so we might be able to remove the `[-3, ]` restriction in line 689
  - I could not reproduce the mlr version of the F1 score using the new mldr functions. Neither mldr::fmeasure, nor micro_fmeasure or macro_fmeasure gave me the same result
- mldr:::mldr_Recall (old) / mldr::recall (new)
  - The new function does not need the `counters` object anymore, instead just `multi.y` and `multi.p`. It disagrees with the old implementation and with the mlr recall - I couldn't find out why
- For these two lines, I left the old code there in a comment for the time being

Fixes #2159


larskotthoff:
Thanks for taking a look at this. If something has changed substantially and you can't make it work, dropping the respective measure may be an option.

jakob-r:
@larskotthoff We just use mldr to check our own measures against something. Do we want to drop them just because we don't get the test working?

Commenting out the test and adding a FIXME would be (although still suboptimal) better solution.

AlexEngelhardt:
It's weird that the definition of recall seems to have changed with the `mldr` update. If I'll be able to understand it, I can rewrite the tests accordingly.

I'll try and reinfuchs myself over the weekend.

AlexEngelhardt:
I asked about the new function implementation [over there](https://github.com/fcharte/mldr/issues/45). I fixed the `recall` test.

With the F-measure, I couldn't bridge the gap from the mlr to the mldr implementation. So I hard-coded the old mldr implementation in the test - is that okay, at least for now?

AlexEngelhardt:
The mldr authors looked into the discrepancy between the two packages' measures [here](https://github.com/fcharte/mldr/issues/45#issuecomment-364716132). The tests pass now, but the two implementations differ when there are observations with no positive labels.

Is this severe enough for a new issue? Would it make sense to re-implement the F-measure by using precision and recall instead? The comment linked above suggests something like this:

```r
mlrAltF1 <- function(truth, response) {
  prec <- mlr::measureMultilabelPPV(truth, response)
  rec <- mlr::measureMultilabelTPR(truth, response)
  2 * prec * rec / (prec + rec)
}
```

larskotthoff:
This looks fine to me. @jakob-r ?

mllg:
Thanks for clearing things up @AlexEngelhardt

jakob-r:
I think we should merge and possibly open an issue about the Fmeasure to change the tests again when `mldr` redefined their fmeasure definition. Furthermore we should fix what happens when _TP + FP + FN = 0_? But this should all be part of another Issue. 

@AlexEngelhardt If you remove _[WIP]_ from this issue I would dare to merge. And it would be awesome if you can briefly sum up the remaining problems in a new issue.

jakob-r:
Ah okay. It's already merged...

PR #2164, opened 2018-02-12
Regenerate example data

mllg:
* Added spam dataset from kernlab
* Fixed `inst/makeData.R`. I don't know how the data which was in `data` before was generated exactly. Last change was in #2134, @pat-s might need to double check that everything is okay with this commit. Additionally, one task could not be constructed.
* Rebuilt all files in `data` with updated `makeData.R`.


inst/makeData.R
```diff
@@ -106,19 +117,21 @@ fs = data.frame("UVVIS" = fuelSubset$UVVIS, "NIR" = fuelSubset$NIR,
...
+save(gunpoint.task, file = file.path(dn, "gunpoint.task.rda"), compress = "xz")
+
 
 # FDA Multiclass Classification
 set.seed(DATASEED)
 data(phoneme, package = "fda.usc")
 ph = as.data.frame(phoneme[["learn"]]$data)
 ph[, "classlearn"] = phoneme[["classlearn"]]
-fdata = makeFunctionalData(ph, fd.features = list(), exclude.cols = "label")
+fdata = makeFunctionalData(ph, fd.features = list(), exclude.cols = "classlearn")
```
mllg: This was broken. @pat-s can you double check this one?

pat-s:
See [here](https://github.com/mlr-org/mlr/blob/9f1814f4ca37f23f2cfc9c1fe2eb6545d0517027/inst/makeData.R) for the approach that generated the data of #2134. I removed it from the PR with the aim to tackle this problem later.

There you also see the `spatial.task` added as the most recent one and what I did with the `phoneme` task.

Feel free to decide if you want to use `devtools::use_data()` or stick with the old approach using `save()`.

larskotthoff:
Thanks, merging.

PR #2165, opened 2018-02-13
add ranger filters

bommert:
Impurity and permutation importance filter based on ranger

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -631,3 +631,46 @@ makeFilter(
     abs(0.5 - score)
   }
 )
+
+
+#' Filter \dQuote{ranger.permutation} trains a ranger learner with
+#' \dQuote{importance = "permutation"} and assesses the variable
+#' importance for each feature.
+#'
+#' @rdname makeFilter
+#' @name makeFilter
+makeFilter(
+  name = "ranger.permutation",
+  desc = "Variable importance based on ranger permutation importance",
+  pkg  = "ranger",
+  supported.tasks = c("classif", "regr", "surv"),
+  supported.features = c("numerics", "factors", "ordered"),
+  fun = function(task, nselect, ...) {
+    lrn.type = paste0(getTaskType(task), ".ranger")
+    lrn = makeLearner(lrn.type, importance = "permutation", ...)
+    mod = train(lrn, task)
+    ranger::importance(mod$learner.model)
+  }
+)
+
+
+#' Filter \dQuote{ranger.impurity} trains a ranger learner with
+#' \dQuote{importance = "impurity"} and assesses the variable
+#' importance for each feature.
+#'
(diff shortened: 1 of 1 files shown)


R/Filter.R
```diff
@@ -631,3 +631,46 @@ makeFilter(
...
+#' @rdname makeFilter
+#' @name makeFilter
+makeFilter(
+  name = "ranger.permutation",
+  desc = "Variable importance based on ranger permutation importance",
+  pkg  = "ranger",
+  supported.tasks = c("classif", "regr", "surv"),
+  supported.features = c("numerics", "factors", "ordered"),
+  fun = function(task, nselect, ...) {
+    lrn.type = paste0(getTaskType(task), ".ranger")
```
jakob-r: Maybe `ranger.permutation` because we have at least two?

R/Filter.R
```diff
@@ -631,3 +631,46 @@ makeFilter(
...
+#' @rdname makeFilter
+#' @name makeFilter
+makeFilter(
+  name = "ranger.impurity",
+  desc = "Variable importance based on ranger impurity importance",
+  pkg  = "ranger",
+  supported.tasks = c("classif", "regr"),
+  supported.features = c("numerics", "factors", "ordered"),
+  fun = function(task, nselect, ...) {
+    lrn.type = paste0(getTaskType(task), ".ranger")
```
jakob-r: see above.

R/Filter.R
bommert: I'm confused. The name of the filter is ranger.permutation already. What exactly do you want me to change/ reconsider?

bommert:
In test_featsel_filters.R all filters are tested automatically.

R/Filter.R
jakob-r: My bad...

jakob-r:
On a third thought... I don't like this so much because it is redundant with `getFeatureImportanceLearner.*` We should have a filter that can use all implemented `getFeatureImportanceLearner`functions. 

But as we already have cforest.importance, randomForestSRC.var.select, randomForestSRC.rfsrc, and randomForest.importance it seems that the horse is already out the barn.

Nonetheless a general solution would be great if you are interested ;)

mllg:
I second what @jakob-r said, a feature importance filter where you pass a learner with the respective property would be useful. Nevertheless, this one here is also fine.

larskotthoff:
This appears to have broken FDA?

mllg:
That's odd. These things have nothing to do with each other. I'll look into it tomorrow.

larskotthoff:
Could be unrelated, but Travis started failing after this was merged.

mllg:
Update of package `fda.usc` to version 1.4.0 caused this, I'm opening a new issue.

PR #2175, opened 2018-02-20
Upsampling and smote now also work for multiclass problems

jakob-r:
- Documentation needs to be improved
- Maybe also handle Undersampling?
- Wrappers could use extra testing for Multiclass


pat-s:
@jakob-r Do you have motivation to finish this up here? I do not know how much work would be left. If not, just close here :)

jakob-r:
If somebody wants this feature he can use this branch or we continue working here but for now let's close it.

PR #2195, opened 2018-03-02
remove Rfast dependency

mllg:


studerus:
Why do you do this?

The FSelector filter is really very slow. It's about 100x slower than Rfast. It really makes a difference if you you do nested repeated cross-validation with a large number of variables, which I guess is not a very uncommon user case. Here's a benchmark:

```
library(microbenchmark)

y <- rnorm(100)
x <- matrix(rnorm(100 * 1000), 100, 1000)
dat <- data.frame(y, as.data.frame(x))

microbenchmark(Rfast::correls(y, x, type = "pearson")[,1],
               FSelector::linear.correlation(y ~ ., dat))
```

```
Unit: milliseconds
                                        expr        min         lq       mean     median         uq        max neval cld
 Rfast::correls(y, x, type = "pearson")[, 1]   1.447481   1.510573   1.642225   1.587398   1.633132   4.826644   100  a 
   FSelector::linear.correlation(y ~ ., dat) 100.393824 102.521574 106.915776 104.146555 105.657127 195.255226   100   b
```

studerus:
Ah sorry. I realized just now, that you use the cor function from base R, which seems to be equally fast as Rfast.

mllg:
https://cran.r-project.org/web/checks/check_results_Rfast.html

Look at the installation times. This was driving us nuts.

PR #2209, opened 2018-03-04
fix a broken unit tests by jakob, so we pass travis

berndbischl:


berndbischl:
i also added a test line NEWS which should be deleted

PR #2210, opened 2018-03-07
put batchmark unit test other testing group so it does not run on winbuilder

berndbischl:
i had to reenable the "parallel" group for travis, too

lets see what happens


berndbischl:
this issue prevents us to go on cran, because of a bug in the datatable package on rdevel / WB
this is caused by batchtools calling into datatable

berndbischl:
i also already uploaded on WB

berndbischl:
WB looks good now (rdevel)

PR #2219, opened 2018-03-12
clean up #2123

pat-s:


.travis.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         - git config push.default matching
         - mv vignettes/tutorial/devel/pdf/_pdf_wrapper.pdf vignettes/tutorial/devel/pdf/mlr-tutorial.pdf
         - git add --force vignettes/tutorial/devel/pdf/mlr-tutorial.pdf
-        - git commit docs -m "update auto-generated tutorial pdf version [ci skip]"
+        - git commit -a -m "update auto-generated tutorial pdf version [ci skip]"
```
jakob-r: if we only push the pdf we don't need `-a` here.

berndbischl:
hi, did this really pass travis? because i really need to fix on cran. 
please do NOT merge if travis is not green.

the master currently seems broken

pat-s:
Hi Bernd,

travis passed for mlr. Only the tutorial (which is now also within the same travis build) did not pass because the ssh keys are not yet properly set up.

So for mlr you only need to check the "check" stage.

berndbischl:
it is still a bit worrysome that we now have a "broken" status on the github badge.
is there an estimate when you guys have this fixed completely?

pat-s:
Jakob and I will have a Skype meeting later on. Stay tuned 🙂 
BTW the behavior of a red badge if either the tutorial or core fails is now the default 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2225, opened 2018-03-17
pkgdown tweaks

pat-s:
### Reasons for changes

New feature of `pkgdown`: Release dates of CRAN versions in NEWS.md.
Currently, this is not yet rendered as my fork does not yet include this commit due to other open problems in the current HEAD of `pkgdown`.

Moving `logo.png` to `man/figures/` as recommended by `?pkgdown::build_site()`


larskotthoff:
So is this something that should only be merged when the problems with pkgdown are fixed? It also looks like you changed version 1.1 to 1.1-18.

pat-s:
> So is this something that should only be merged when the problems with pkgdown are fixed? 

No, otherwise I would not have opened the PR :smile: . Whether logo.png is in root or `man/figures` makes no difference with the current approach. It just cleans up root.

> It also looks like you changed version 1.1 to 1.1-18.

Yes because this is the actual first release version of `mlr` to CRAN. Only if the version is correct, `pkgdown` will correctly attach a date to the release. 

I am currently updating .travis.yml to make all deployments using a single ssh key rather than a bulky git commit chain. This PR will then also include the updated PR changes.

PR #2228, opened 2018-03-21
more pkgdown tweaks

pat-s:
* do not push PDF vignette for cron jobs (maybe we should push to another repo at all to not clutter the main repo with the PDF updates)
* update logo path in README
* incorporate latest pkgdown changes (tested)

We need to use the fork as it holds the custom navbar. I'll check if there is a better way to specify the template than using a fork install. This would release me from having to update the fork all the time which is of course annoying. 

FYI: `pkgdown` is on the way to support a release/dev approach soon.


larskotthoff:
Thanks, merging.

Issue #2240, opened 2018-04-04
Move from party to partykit


HeidiSeibold:
partykit is still being actively moved forwards while party is not. So mlr should switch to using partykit algorithms (see also #1481). For this we need to update:
- [x] classif_ctree
- [x]  regr_ctree 
- [x]  classif_cforest
- [x]  regr_cforest
- [x]  multilabel_cforest
- [ ]  regr_mob
- [ ] tests for classif_ctree 
- [ ] tests for regr_ctree 
- [ ] tests for classif_cforest
- [ ] tests for regr_cforest
- [ ] tests for multilabel_cforest
- [ ] tests for regr_mob

What would be nice would be to add
- [ ] classif_glmtree
- [ ]  regr_glmtree
- [ ]  regr_lmtree

I started working on this in https://github.com/mlr-org/mlr/tree/partykit-transition

HeidiSeibold:
I could use help with:

- [ ] Checking if the parameter sets are correct. In particular I was unsure with perturb in cforest (see https://github.com/mlr-org/mlr/blob/partykit-transition/R/RLearner_classif_cforest.R#L16)
- [ ] How to update the tests. It is not totally clear to me why the tests currently are the way they are.

PR #2241, opened 2018-04-04
Partykit transition

HeidiSeibold:
See #2240 for details. 

This is work in progress.


pat-s:
@HeidiSeibold 	I guess we can close here?

PR #2249, opened 2018-04-16
Add weights for regr.ranger

kysolvik:
- Allows ranger regression to accept weights for case.weights.

- Copied from RLearner_classif_ranger.R, #1409.


larskotthoff:
Thanks, looks like it should be fine apart from the lintr error.

larskotthoff:
Thanks. Could you please add a unit test that checks the learner with weights?

kysolvik:
From the discussion in #1409, it sounded like adding a specific unit test for ranger weights would be redundant because "testThatLearnerRespectsWeights" already checks this.

But I would be happy to work on adding a unit test if it's needed

larskotthoff:
Ok thanks, merging.

PR #2250, opened 2018-04-19
Use rcmdcheck for check stage and deploy via ssh

pat-s:
`tic` now also supports deployment of single files. This means we do not rely anymore on the "git" block for the deployment of any changes in `man/` or `NAMESPACE` but are able to do the deployment through `tic` via `ssh`. We already do so for the tutorial deployment.

Advantages of [rcmdcheck::rcmdcheck()](https://github.com/r-lib/rcmdcheck) over `R CMD Check`:
* Enhanced error and warning print out (no truncation) ([example](https://travis-ci.org/mlr-org/mlr/jobs/368633802))
* Interestingly, quicker build time (Check stage needs only 41 mins instead of 55)

Additionally there is no need anymore for `travis_wait` -> The continuous output of the various sub-test files (introduced by `testthat` v2.0?) does prevent travis from stalling the build due to timeout. This means we have a live log of the check stage.


larskotthoff:
Looks great, thanks!

PR #2252, opened 2018-04-24
Travis adjustments

pat-s:
* Remove grouping of references

Due to the grouping not all references are build by `pkgdown`. Grouping seems not really valuable and the search functionality should then find all available documentation

* Run `devtools::document()` before `rcmdcheck()` so that `man/` checks are successful

Until now, devtools::document()` was only run before deployment but after `rcmdcheck()` which caused some builds to fail (@pfistfl)


larskotthoff:
Thanks, merging.

Issue #2257, opened 2018-04-26
regr.nnet : some output layer params should likely be removed?


sycoforbidden:
When calling getParamSet("regr.nnet"), we get:

                Type  len  Def Constr Req Tunable Trafo
    linout   logical   -  FALSE    -   Y    TRUE     -

However, upon inspection of the mlr:::trainLearner.regr.nnet() function, one gets

> function (.learner, .task, .subset, .weights = NULL, ...) 
> {
>   if (is.null(.weights)) {
>     f = getTaskFormula(.task)
>     nnet::nnet(f, data = getTaskData(.task, .subset), **linout = TRUE**, 
>       ...)
>   }
>   else {
>     f = getTaskFormula(.task)
>     nnet::nnet(f, data = getTaskData(.task, .subset), **linout = TRUE**, 
>       weights = .weights, ...)
>   }
> }
> 

So when making a learner regr.nnet by default, which instance of linout is used?

sycoforbidden:
Alright then, thank you for the quick response.

berndbischl:
@mb706 gblame says you added these some time ago.


```
      makeLogicalLearnerParam(id = "linout", default = FALSE, requires = quote(entropy == FALSE && softmax == FALSE && censored == FALSE)),
      makeLogicalLearnerParam(id = "entropy", default = FALSE, requires = quote(linout == FALSE && softmax == FALSE && censored == FALSE)),
      makeLogicalLearnerParam(id = "softmax", default = FALSE, requires = quote(entropy == FALSE && linout == FALSE && censored == FALSE)),
      makeLogicalLearnerParam(id = "censored", default = FALSE, requires = quote(linout == FALSE && softmax == FALSE && entropy == FALSE)),
```

any reason?

sycoforbidden:
Maybe he added those since they are the default parameters of nnet:::nnet.default() ? Even though what you're applying here should be nnet:::nnet.formula() ?

berndbischl:
the latter 3 only make sense for classification. they should really be removed

mb706:
[...]

PR #2258, opened 2018-04-26
regr.nnet: remove unneeded params

berndbischl:
see issue #2257


larskotthoff:
Could you move the comment to the learner note please?

berndbischl:
> Could you move the comment to the learner note please?

no, we are not adding notes about stuff we didnt do because they dont make sense?

larskotthoff:
That `linout = TRUE` seems to be useful to know for users.

berndbischl:
> That linout = TRUE seems to be useful to know for users.

ah. that i agree with. will do.

berndbischl:
done

larskotthoff:
Thanks, merging.

PR #2259, opened 2018-04-27
Travis: Fix allow_failures

pat-s:
If the tutorial fails, the Travis build should still succeed. 
This did not happen in https://travis-ci.org/mlr-org/mlr/builds/370688009

I did some testing and found out that the cause for this is a slightly wrong specification of the "env" setting in the stages. 

It now works as desired: https://travis-ci.org/mlr-org/mlr/builds/371611368

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,8 +71,7 @@ jobs:
              - $TRAVIS_PULL_REQUEST = false
              - $TRAVIS_EVENT_TYPE != cron
     - stage: Tutorial
-      env:
-        - TUTORIAL=HTML
+      env: TUTORIAL=HTML
       addons:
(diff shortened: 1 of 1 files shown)


larskotthoff:
Hang on, I'm not sure if I agree. If the tutorial fails, something should fail to let us know that something is wrong. And if the tutorial fails because of the change in mlr, this build should fail.

larskotthoff:
Ok, do we still need this PR then?

pat-s:
Yes, I'll change it here 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

larskotthoff:
Thanks, merging.

PR #2043, opened 2017-10-24
Oneclass Local Outlier Factor (LOF) from DMwR

MinhAnhL:
Add Rlearner lofactor (Local Outlier Factor) from DMwR for Anomaly Detection

PR #2264, opened 2018-05-02
Oneclass Local Outlier Factor (LOF) from Package dbscan

MinhAnhL:
Add leaner lof from the dbscan package for Anomaly Detection (one-class classification)


pat-s:

PR #2267, opened 2018-05-08
Enable "blocking" in nested resampling and add getNestedResamplingIndices()

### LINKED ISSUE #2266: Resampling: Add "blocking" resample description to work within nested resampling
Related: https://github.com/mlr-org/mlr/issues/1498

#### Problem

- While the `blocking` argument in `make*Task()` works in a normal resample situation, it does not work in nested situations. 

- This is because the `blocking` information is stored in the task and is not independent from the chosen Resample instances in the outer and inner level. 

- Furthermore it is confusing if one sets `blocking` and then specifies one of the Resampling Descriptions (E.g. CV) although no CV is performed in fact. `blocking` should only be usable in a specific Resample description, e.g. `makeResampleDesc("Blocking")`. 

#### Example
For example: I want to have a "blocking" resampling in the outer level for performance estimation but a CV based hyperparameter tuning in the inner level.
I get an error because the inner resampling instance tries to use the blocked indices but the sample is only a subset of the initial sample at that stage. Subsequently, the function errors.

#### Motivation & Solution

- It should be possible to have two independent resampling strategies. With the current implementation of `blocking` this is not possible.

- I vote for the implementation of a Resample description named "blocking" that accepts a factor variable with length(obs) and returns one repetition of `n` folds `n(= length(levels))`. 

- The `blocking` argument in `make*Task()` can stay and behave similar to `coordinates` which is also only used if either `SpCV` or `SpRepCV` is specified.

Of course this is only an additional resample option and does not solve the possibility of user-defined resampling splits as proposed in #1498

pat-s:
fixes #2266 

### New
* getter for nested resampling indices: `getNestedResamplingIndices()`. Works on a valid `resampleResult` that used `extract = getTuneResult`. Similar to [getNestedTuneResults()](https://github.com/mlr-org/mlr/blob/master/R/getNestedTuneResults.R).

```r
str(getNestedResamplingIndices(p))
List of 5
 $ :List of 2
  ..$ train.inds:List of 4
  .. ..$ : int [1:90] 70 93 44 26 96 115 30 61 9 15 ...
  .. ..$ : int [1:90] 70 93 44 75 115 38 30 61 9 60 ...
  .. ..$ : int [1:90] 44 75 26 96 38 61 9 15 60 36 ...
  .. ..$ : int [1:90] 70 93 75 26 96 115 38 30 15 60 ...
  ..$ test.inds :List of 4
  .. ..$ : int [1:30] 24 89 88 5 86 13 7 48 45 55 ...
  .. ..$ : int [1:30] 102 67 29 71 96 117 77 62 58 53 ...
  .. ..$ : int [1:30] 4 79 73 42 116 97 57 78 25 30 ...
  .. ..$ : int [1:30] 54 28 61 101 32 36 9 41 2 120 ...
 $ :List of 2
  ..$ train.inds:List of 4
  .. ..$ : int [1:90] 15 44 16 35 106 65 1 89 57 119 ...
  .. ..$ : int [1:90] 44 35 106 13 65 1 89 90 57 53 ...
  .. ..$ : int [1:90] 15 44 16 13 89 90 57 53 119 59 ...
  .. ..$ : int [1:90] 15 16 35 106 13 65 1 90 53 119 ...
  ..$ test.inds :List of 4
  .. ..$ : int [1:30] 105 7 77 109 74 90 19 42 76 85 ...
  .. ..$ : int [1:30] 22 25 97 31 58 46 15 69 119 54 ...
  .. ..$ : int [1:30] 106 92 55 87 33 26 65 14 36 79 ...
  .. ..$ : int [1:30] 40 20 4 96 88 57 107 110 59 80 ...
```

#### Implementation

Basically I passed the `resampling` objects upwards through the following functions:

`tuneGrid -> makeTuneResultFromOptPath -> makeTuneResult -> makeOptResult ->  tuneParams -> trainLearner.TuneWrapper#61`. Afterwards it appears within the `extract` object of `getTuneResult()` and can be accessed from there. Of course it still comes with the type of resampling etc. The getter however only gets the raw indices as shown above.

#### Changes to "blocking"

We do not need the "old" "blocking" behaviour anymore as "blocking" is now a valid `ResampleDesc` which it was not before. Before it was hardcoded in `ResampleInstance.R`. 

Not sure how important all the stratification stuff is, I did not deal with it. But as "Blocking" is now a valid `ResampleDesc`, it should be possible to hand it over? Anyhow, combining "blocking" and "stratification" is a very special case and I am not sure if anyone ever used it? Do not want to spend hours on that without any added value. 

#### Tests
* needed some adjustments
* also warn if the user supplies a wrong `iter` count (actually we do not need `iter` at all but during implementation `makeResampleDesc()` did not wanted to work without any argument (https://github.com/mlr-org/mlr/blob/master/R/ResampleDesc.R)

#### Tutorial

I extended the tutorial a bit; maybe we could even think of giving "blocking" a bit more information as it has right now. Looks a bit hidden to me

PS: I have no clue why the diff shows changes for `.travis.yml` and `tic.R`. They are equal with `master` :thinking:


larskotthoff:
Could you add such a getter (not exported) for testing please?

It probably makes sense to output an error message if both blocking and stratification is used -- I want to avoid somebody using it accidentally and then getting weird results. Same thing if the nested resampling would break the blocking -- could you please add a check for that and output an error message?

larskotthoff:
Oh and did you update your master branch? The diff shows differences for the two files for which you see none.

pat-s:
> Could you add such a getter (not exported) for testing please?

Would proably take a bit, but yes.
However, if I would a getter, why not export it? And return the inner indices to the object of `resample()` similar as the done for the outer indices?

> It probably makes sense to output an error message if both blocking and stratification is used -- I want to avoid somebody using it accidentally and then getting weird results.

Or even a stop? I do not see a use case where this combination could be useful. 

> Same thing if the nested resampling would break the blocking -- could you please add a check for that and output an error message?

You mean if the number of blocking levels is below 2 in the inner level?

larskotthoff:
Exporting it is fine as well I guess. And yes, by "error" I mean "stop" :)

What I mean is anything that would break the blocking, e.g. if there's only one blocking level left, or if there's only 9 and the user requested a 10 fold CV.

pat-s:
> Exporting it is fine as well I guess. And yes, by "error" I mean "stop" :)

Idk what I misunderstood here :D 

> What I mean is anything that would break the blocking, e.g. if there's only one blocking level left, or if there's only 9 and the user requested a 10 fold CV.

The latter is already accounted for, no matter what the user specifices, nfold will always be set to nlevels.

I think the other problem only applies if  n=<2 and tuning using blocking is requested, right? Otherwise there will always to 2 levels at least. 

(This seems to become another time-eating implementation :sweat_smile: )

pat-s:
@larskotthoff 

I added a getter for the inner indices and updated the tests.

Questions: 
* I am not really happy with the name "blocking" for the whole approach. I think "grouping" would be more natural. Is that conflicting with another "grouping" option anyhow? 
I know however that chaning the name of the whole concept is a major change. 
* I think we should definitely expand the documentation on this, also give some info on the new getter for the inner indices?

pat-s:
@jakob-r @berndbischl @mllg 
Would be great to have some more thoughts on this PR.

larskotthoff:
Thanks. I think "blocking" is quite descriptive, because it allows to keep observations together in blocks.

Looking at it now, I'm not sure that "blocking" should be another resampling strategy -- I've always seen this as something like stratification that's taken into account by the other strategies. In particular, this doesn't seem to allow to combine blocking with other things, which is one of the things you wanted to make work? Or am I misunderstanding something here?

And yes, more documentation is always good :)

pat-s:
> Looking at it now, I'm not sure that "blocking" should be another resampling strategy -- I've always seen this as something like stratification that's taken into account by the other strategies.

I see this differently. For me it is a valid resampling instance since all we want to do is run one repetition with each factor level specified being once the test set.
The same applies for tuning.

I do NOT want to run a CV with random sampling (i.e. `makeResampleDesc("CV")` and then group SOME observations in that one.

I think these are two different approaches: 

1. Blocking SOME observations in a random sampling CV
2. Running a CV with only "fixed" instances for the whole dataset.

Maybe we should go with both and rename the new approach? Then we can still have the old behaviour of blocking as a helper for any resampling desc and the new approach that does something (completely) different. If so, we would still need a name. "FixedGroupInstances"?

larskotthoff:
Well, blocking doesn't need to cover all the instances, e.g. I could have 1000 observations and only want 10 of them to be blocked.

pat-s:
Yes to use it in CV. But there are not so few cases in which you want to fix all indices upfront with a grouping.
So would do you think about the idea of my last comment?

larskotthoff:
I don't see why we need a different type. This should just come out of blocking + resampling, shouldn't it?

pat-s:
No, this won't work. The "new" blocking does not provide multiple reps, it uses a fixed set of indices for the complete dataset. Same for the inner resampling in which the remaining observations are again grouped by the specified grouping. 

I do not see

1.  How I should get it working to work in both ways (i.e. with partial blocks and full blocks for all obs)
2. Why "blocking" should work for both approaches as it might confuse users

That's why I think it makes most sense to split both approaches. Keep the old blocking as it is and introduce a new one that only does this "1 rep, all indices fixed" approach. 

Currently, `makeFixedHouldoutInstance` does something similar but only tests once (on this one test set). There is no option to test on multiple fixed groupings. I could also extend this resampling method and document it better. For two factor levels it will then behave as before (one training, one test) and for more than 2 it will behave like I want "blocking" to behave.

larskotthoff:
Hmm, any opinions on this @berndbischl @mllg @jakob-r ?

GegznaV:
>  Anyhow, combining "blocking" and "stratification" is a very special case and I am not sure if anyone ever used it?

In my experiments, I combine these two techniques, and it is frustrating to get error messages like this one:

```
Error in makeResampleInstance(resampling, task = task) : 
Blocking can currently not be mixed with stratification in resampling!
```

It the method that combines these two techniques was implemented, I think an error (or warning) message should only appear if stratification contradicts blocking and tries to break it.

pat-s:
@larskotthoff I would like to finish up this PR in the next week.

The main question is still where to implement it.
As it is basically just a single rep CV with predefined indices, we could simply add a factor argument to `"makeResampleDesc("CV", factor = <factor name>")`.
The argument would be called via `...` (further arguments passed to resample strategies).

So instead of performing a single rep CV with random sampling, we use predefined indices from a factor variable stored in the task. What's your opinion on this?

The same would apply for "SpCV" then.

larskotthoff:
That sounds good.

pat-s:
superseeded by #2412

PR #2268, opened 2018-05-15
add colors to plotHyperParsEffect

jakob-r:
How it looked:
image

How it will look like:
image

Notes: 

- It's the same color palette that we use in mlrMBO and that is used in smoof
- I hard coded the color values to avoid RColorBrewer dependencie

### DIFF
--- a/R/generateHyperParsEffect.R
+++ b/R/generateHyperParsEffect.R
@@ -455,6 +455,7 @@ plotHyperParsEffect = function(hyperpars.effect.data, x = NULL, y = NULL,
       }
       if (plot.type == "contour")
         plt = plt + geom_contour()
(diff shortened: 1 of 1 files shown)


mllg:
Much better.

pat-s:
Looks way better. For such visualization tasks I usually use `viridis`. Maybe a more robust approach than hardcoding the values? 

https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html

However, for consistency with other packages it may be preferable to have the same palettes.

jakob-r:
@pat-s They also look nice but I would not include the dependency. After all they do the same and one could also read out the hard coded value generated by `viridisLite::viridis`:
https://github.com/sjmgarnier/viridis/blob/1d6d5084041271e23732f356ce49f290150879f6/R/scales.R#L154

PR #2269, opened 2018-05-15
Cleanup packages

mllg:
* Removed some packages which were not used from DESCRIPTION.
* Added some calls to `requirePackages()` in order to get better error messages.


mllg:
@larskotthoff @pat-s

Could this break the tutorial build? Is this currently tested via travis?

pat-s:
@mllg Tutorial should not be affected by these packages. Also, Travis should reveal if so since the builds are executed but just not deployed. If a package is missing the stage should fail.

jakob-r:
Why are there now two requirePackages calls in the diff? I did not see them in the first draft? @mllg Did you do a squash? I also don't understand why we only load smoof in the multiobjective MBO case?

mllg:
> Why are there now two requirePackages calls in the diff? I did not see them in the first draft? @mllg Did you do a squash? 

AFAICT this was always here. See also my comments in first post of this PR.

> I also don't understand why we only load smoof in the multiobjective MBO case?

It looks like we only directly use smoof in the MO case, right? So we only need to explicitly require the namespace here.

PR #2284, opened 2018-05-29
Deploy release docs again, minor tweaks

### LINKED ISSUE #2254: 'release' documentation is not deployed
Probably due to recent changes in `pkgdown` the `release` folder is not picked up anymore. 

https://github.com/r-lib/pkgdown/issues/644

pat-s:
fixes #2254 

- PDF tutorial for release and dev is now build in parallel stages (see https://travis-ci.org/pat-s/mlr/builds/385260547)

- "Release" tutorial is deployed again. Sadly, there is no other way right now than to also build the release docs on every push again although they are static. No problems with our build time, both versions finish in about 47 mins combined.

- Backup of old release docs to folder /old/2.11

* Change of version indicator in navbar

* update pandoc version

* Update README. Currently, to get to the release docs, one has to manually replace `devel` by `release` in the URL. Loading another site, one lands again in the `devel` version.
There is currently no permanent switch between both version but this is a planned feature of future `pkgdown` versions.

Preview: https://pat-s.github.io/mlr/


pat-s:
Just because the cache needs to be built once. Sharing caches across different envs is not yet supported by travis.

So every env needs to install all deps on its own. After this has happened, simply restart the build (I did) and the second run will be faster.

larskotthoff:
Thanks, merging.

PR #2287, opened 2018-06-05
Test that all learner can handle only one feature!

jakob-r:
Spotted a bug in GPfit. Added general tests for regr and classif


larskotthoff:
Thanks. Looks like we should have tests for the SE -- do you think it would be feasible to add tests for this to this PR?

jakob-r:
How to test it? What are acceptable boundaries for SE values? Or do I calculate the SE for all supported learners on a test dataset and then throw an error for outliers?

larskotthoff:
I would construct a small dataset and make sure that the SE predictions are numbers and not unreasonably large (e.g. larger than the range of values).

jakob-r:
@larskotthoff What do you say to the added test?

larskotthoff:
Looks good, thanks!

PR #2292, opened 2018-06-12
build namespace before html tutorial deployment

pat-s:
Fixes the deployment of the HTML tut. PDF building problems not solved yet.


larskotthoff:
Thanks, merging.

PR #2295, opened 2018-06-14
Re-add missing dependencies

mllg:
Looks like some packages were still in the cached package library, re-adding them here.


jakob-r:
Maybe we can add a not in the commit, so that the next time we know why we need rgenoud and svglite. I usually check the blame before I delete such lines. Then I would see it.

mllg:
R CMD check is not successful, but we get a green checkmark from travis... 

@pat-s @larskotthoff: Can you have a look what is wrong with travis?

pat-s:
This should be related to https://github.com/ropenscilabs/tic/issues/44

@gaborcsardi Any idea when the next CRAN release of `rcmdcheck` will be ready?

Until then we can try installing the dev version manually. As per Gabors comment, it should be fixed in the dev version.. 

You can try replacing the "script" section in `tic.R` with:

```r
get_stage("script") %>%
    add_code_step(devtools::install_github("r-lib/rcmdcheck")) %>%
    add_code_step(devtools::document()) %>%
    add_step(step_rcmdcheck())
```

mllg:
If the last one passes travis, we should merge. `blackboost` can then be re-added in a separate PR.

pat-s:
Grml, the `PDFdev` should not fail. There is no difference to `PDFrelease` which behaves normal. 
Something seems to be corrupted with the cache. I delete the cache and we rebuild it...

pat-s:
Argh, I deleted the cache for the branch but not the PR. Let's try again.

Issue #2236, opened 2018-03-29
Open tasks for pkgdown site


pat-s:
This is a list of open tasks to address in the future.

- [ ] ToC sidebar is missing for all articles because `pkgdown` HEAD only processes h1 headers currently
- [x] Not all functions are processed for "reference", probably due to grouping? -> Yes
- [ ] Write wiki entry explaining the build process and recent travis changes
- [ ] The logo in the navbar always reloads when changing sites. This is not the case for other packages with a logo, e.g. [dplyr](http://dplyr.tidyverse.org/index.html)
- [ ] Some examples in the function reference are not executed properly, e.g. [train()](http://mlr-org.github.io/mlr/reference/train.html)

PR #2329, opened 2018-06-23
Optimize navbar logo rendering

pat-s:
Currently there is a small delay in loading the navbar logo when loading a tutorial page.
The PR now uses the `dplyr` approach (which uses CSS instead of specifying the logo stuff in `navbar.html`) and solves the delay problem.

#2236


larskotthoff:
Thanks, merging.

PR #2332, opened 2018-06-26
fix tuneCMAES and tuneGenSA bug

jakob-r:
They won't work with PH 1.11
This bug was cought by a new ParamHelpers release with more strict assertions.
Actually I don't know what this was supposed to mean or if it simply was accidental nonesense.


mllg:
Closed in favor of #2295

Issue #2335, opened 2018-06-27
Proposed developer change: Commit NS + man files manually


pat-s:
We commit the NS and man/ files via Travis.
This has some advantages but also some practical disadvantages.

#### Disadvantages

- We need to run `devtools::document()` in Travis to build the NS and man/ files on change. This is also needed to get `rcmdcheck()` and `pkgdown()` run properly.
- When developing `mlr`, sometimes the NS needs to be built locally (sometimes `load_all()` is not enough, e.g. when debugging) and then the NS exists in the git log. If one wants to check out to a different branch, it first needs to be discarded. When switching back, it needs to be rebuild. It would be easier if the NS could just be commited to a PR so this is not necessary anymore. 
- If a user accidentally commited man/ and NS to a PR, its not super easy to revert it for a basic git user
- Its common practice for most R packages to commit NS and man/ on your own, so I always find myself out of my habit when doing otherwise in `mlr`. Most people contributing to an R package are used to build the NS and I do not expect many problems.
- We need to bother with the git stuff on Travis to deploy NS + man/ (which is currently a bit problematic due to some `tic` problems)
- If one wants to build the NS manually, the "build" button in the RStudio pane cannot be used as this only triggers a "R CMD INSTALL". This is because `mlr.Rproj` does not contain the setting to use `devtools::document()` when hitting the "build" button.
Currently one always has to go to "build->configure build tools -> create documentation with roxygen". This would be fixed by adding `PackageRoxygenize: rd,collate,namespace` to `mlr.Rproj`. **Edit: This could be changed regardless of the proposed change.**

#### Advantages
- The NS is always correct when building via Travis
- PRs are not cluttered by man/ and NS file changes.

#### Comments
We could still leave the workflow of building the NS on Travis and commit potential changes in case a user forgot to document the latest changes.

Did I oversee something here? Please comment!
@berndbischl @larskotthoff @mllg @jakob-r @giuseppec

larskotthoff:
As a general rule, I try to do things that are done automatically (like generating man pages) automatically. It's very easy to forget to regenerate and commit, and when you're making a release it adds an extra step because you need to regenerate again.

I'm against this change.

pat-s:
Ok, I see the points. 

At least the `.Rproj` file is updated which helps me during development with RStudio :+1: . Did not know that so many develop without it and Michels `rt`, good to know!

PR #2336, opened 2018-06-27
Update mlr.Rproj, update man/

pat-s:
We currently have some problems to build+commit the NS and man/ files on Travis.

This manually fixes man/ and also updates `mlr.Rproj` so that `devtools::document()` will automatically be used when hitting the "build" button in RStudio (instead of triggering a `R CMD Install`). 

See also #2335


larskotthoff:
Thanks, merging.

PR #2339, opened 2018-06-27
Tutorial: Add use case: regression

pat-s:
Cleaned version of https://github.com/mlr-org/mlr-tutorial/pull/100

I removed learner `regr.penalized` as three learners are sufficient for a bmr experiment and the rendering already takes quite some time due to the tuning.
Also penalized regr. might be too "complicated" for a "simple" use case. 

For the moment I would append it to "basics". However, I would want to change the menu ordering in the future completely anyways.


larskotthoff:
Thanks, merging.

PR #2353, opened 2018-07-04
Deactivate all tests for CRAN

jakob-r:
Use the usual if condition to only run on travis and if `R_EXPENSIVE_TEST_OK == 'true'` but added `NOT_CRAN == 'true'` as a possibility to run the base tests as this is used by devtools and testthat to identify not CRAN System.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-![mlr](https://raw.githubusercontent.com/mlr-org/mlr/master/man/figures/logo_navbar.png) Machine Learning in R
-==========================
+# Machine Learning in R <img src="man/figures/logo_navbar.png" align="right" />
+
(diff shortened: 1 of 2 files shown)


pat-s:
I'm trying to understand this since 5 mins. 
Why is the code appended to the old one and not updated? Maybe I need another coffee to get this :duck: 
Could you describe this in more detail Jakob? :)

jakob-r:
Probably because I got distracted :roll_eyes: Sorry.

We should not deactivate anything on Travis. We just skip the base tests like we skip everything else. Just that here there is one additional way to NOT skip this tests by having NOT_CRAN enviroment set (devtools does this automatically)

PR #2354, opened 2018-07-04
link OpenML under "mlr-org packages"

### LINKED ISSUE #2318: Link OpenML in the tutorial
_From @PhilippPro on March 7, 2017 12:47_

Make a own section or link to the OpenML tutorial: [http://openml.github.io/openml-r/vignettes/OpenML.html](http://openml.github.io/openml-r/vignettes/OpenML.html)

Think this would fit best in the Advanced section.

_Copied from original issue: mlr-org/mlr-tutorial#79_

pat-s:
fixes #2318	

Even though its strictly not an "mlr-org" package, I think its still fine. Jakobs package is also listed and not hosted under `mlr-org`.


larskotthoff:
Not sure if we want to have it in that list -- it should definitely be mentioned in the tutorial, but probably in a separate section.

PR #2373, opened 2018-07-08
Base: Add createSpatialResamplingPlots()

pat-s:
This PR enables the option to create multiple `ggplot2` objects at once that visualize spatial resampling structures. 
This si important to get a feeling how the clustering was performed and investigate differences between folds or debug folds in which the model errors. 

Its quite flexible and can deal with multiple `resample` objects to compare spatial and non-spatial resampling. See example 2.

`spatial.task` had unordered rownames that have been fixed. Rownames need to be 1:length(nrow) so that the function can work correctly. This is also checked within the function.

#### Example 1

``` r
library(mlr)
library(cowplot)
#> Loading required package: ggplot2
#> 
#> Attaching package: 'cowplot'
#> The following object is masked from 'package:ggplot2':
#> 
#>     ggsave

rdesc = makeResampleDesc("SpRepCV", folds = 5, reps = 4)
r = resample(makeLearner("classif.qda"), spatial.task, rdesc)

plots = createSpatialResamplingPlots(spatial.task, r, crs = 32717,
                                     repetitions = 2, x.axis.breaks = c(-79.075),
                                     y.axis.breaks = c(-3.970))
plot_grid(plotlist = plots[["Plots"]], ncol = 5, nrow = 2,
          labels = plots[["Labels"]], label_size = 8)
```

Created on 2018-07-08 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

#### Example 2

``` r
library(mlr)
library(cowplot)
#> Loading required package: ggplot2
#> 
#> Attaching package: 'cowplot'
#> The following object is masked from 'package:ggplot2':
#> 
#>     ggsave

rdesc1 = makeResampleDesc("SpRepCV", folds = 5, reps = 4)
r1 = resample(makeLearner("classif.qda"), spatial.task, rdesc1)
rdesc2 = makeResampleDesc("RepCV", folds = 5, reps = 4)
r2 = resample(makeLearner("classif.qda"), spatial.task, rdesc2)

plots = createSpatialResamplingPlots(spatial.task,
                                     list("SpCV" = r1, "CV" = r2), crs = 32717, repetitions = 1,
                                     x.axis.breaks = c(-79.075), y.axis.breaks = c(-3.975))
plot_grid(plotlist = plots[["Plots"]], ncol = 5, nrow = 2,
          labels = plots[["Labels"]], label_size = 8)
```

Created on 2018-07-08 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).


larskotthoff:
Thanks, looks good. I'll get back to this when irace works again.

pat-s:
@larskotthoff Please check again. Travis passes :)

Note: I removed the line `add_code_step(install.packages(old.packages())) %>%` from `tic.R`. Why? Even though it solved the "chicken-and-egg" problem of penalized/survival, it also updates the base R packages (mgcv, MASS etc). However, these will not be cached as they come with the R installation on every travis run. Hence they get updated in every Travis build by this command which is not what we want.

Also, the command even caused build errors recently (https://travis-ci.org/mlr-org/mlr/jobs/403566840). To get the build of this PR passing, I had to include the fix. I did not wanted to create a new PR just for this line. I hope this is ok this time - I keep trying not to mix things in the future ;)

DESCRIPTION
```diff
@@ -48,13 +48,17 @@ Depends:
...
+    hrbrthemes,
     ggplot2,
     stats,
     stringi,
     checkmate (>= 1.8.2),
     data.table,
     methods,
     parallelMap (>= 1.3),
+    purrr,
+    sf,
```
larskotthoff: We try to keep the list of imports as short as possible -- it is absolutely necessary to add all of those packages here?

R/createSpatialResamplingPlots.R
```diff
@@ -0,0 +1,233 @@
...
+#'   plist[["Plots"]][[8]], ncol = 3, nrow = 1, labels = plist[["Labels"]][6:8],
+#'   label_size = 18)
+#' p22 <- plot_grid(plist[["Plots"]][[9]], plist[["Plots"]][[10]], ncol = 2,
+#'   nrow = 1, labels = plist[["Labels"]][9:10], label_size = 18)
+#'
+#' cowplot::plot_grid(p1, p12, p2, p22, ncol = 1)
+#' }
+#' @export
+createSpatialResamplingPlots = function(task = NULL, resample = NULL, crs = NULL,
+  datum = 4326, repetitions = 1, color.train = "#0072B5", color.test = "#E18727",
```
larskotthoff: I'm not a fan of `datum` because datum usually means something completely different in R. How about `epsg.datum`?

DESCRIPTION
pat-s: `purrr` is needed for `imap` (that does not exist in any other package and neither in BBmisc)
`sf` does all the processing in the background of the spatial plotting in `ggplot2`

R/createSpatialResamplingPlots.R
pat-s: This is the exact name of the argument in `coord_sf()` which `datum` is passed on to. Users with a spatial background know what it means in that context.

I see the naming issue but choosing a different name than what users might expect that are used to the `sf` argument naming might introduce new confusion.
I would say its easier staying with `datum` based on the fact that users who will use this function are familiar with its meaning here.

DESCRIPTION
larskotthoff: But does it need to be in imports? Could it be in suggests?

R/createSpatialResamplingPlots.R
larskotthoff: Ok, fine.

DESCRIPTION
pat-s: Both packages are used directly in `createResamplingPlots()`, so we need to have it in imports because I directly import it in the function? (Correct me if I'm wrong)

```
#' @import sf
#' @importFrom purrr map imap is_list flatten map_int
```

But I think I can move `cowplot` and `hrbrthemes` to suggests?

I also made some `importFrom` restrictions for `sf` and `hrbrthemes`.

pat-s:
> Thanks, merging -- what is this appveyor thing?

I enabled `mlr` on appveyor to also test on Windows. Haven't made the PR yet and forgot to restrict it to this branch until then.

Its now restricted and won't appear until the PR is live.

PR #2378, opened 2018-07-11
tests: update learner checks with only one feature

jakob-r:
I was a bit too lazy there. Hope it runs now.


pat-s:
:disappointed: Still errors in the same tests.

pat-s:
:disappointed: New errors in the same tests: https://travis-ci.org/mlr-org/mlr/jobs/402640924

jakob-r:
Yes but this time it actually catches a bug in adabag. They forgot drop = FALSE.
What should we do?
1) Drop adabag to extraLearners
2) Exclude adabag from test
3) Catch this bug in adabag train method.

jakob-r:
I excluded all learners that caused errors not due to mlr from the test. 
I wrote an email to some maintainers (maybe we want more)

What could be done: Use `expect_error()` with concrete error message for the ones that fail so that we catch if we introduce a new error. Flipside: We will fail if they fix or at least change the error msg -- so we probably don't want that.

pat-s:
```
── 1. Error: learners work: regr  (@test_learners_all_regr.R#36)  ──────────────
Argument 'upper' must have length 1, but has length 2
1: lapply(lrns, testBasicLearnerProperties, task = task, hyperpars = hyperpars, pred.type = "se") at testthat/test_learners_all_regr.R:36
2: FUN(X[[i]], ...)
3: expect_numeric(info = info, s, lower = 0, upper = range(y), finite = TRUE, any.missing = FALSE, 
       len = getTaskSize(task)) at /tmp/RtmpHRgahZ/file32127264690b/mlr.Rcheck/tests/testthat/helper_learners_all.R:81

── 1. Error: learners work: classif (@test_learners_all_classif.R#73)  ─────────
x should be a matrix with 2 or more columns
1: lapply(lrns, testBasicLearnerProperties, task = min.task, hyperpars = hyperpars) at testthat/test_learners_all_classif.R:73
```

Thanks for tackling this Jakob :+1:

berndbischl:
i have disabled this for now, as it blocks the release

```
+++ b/tests/testthat/test_learners_all_regr.R
@@ -55,6 +55,6 @@ test_that("learners work: regr ", {
   lapply(lrns, testThatGetOOBPredsWorks, task = task)

   # regr with only one feature
-  min.task = makeRegrTask("oneCol", data.frame(x = 1:10, y = 1:10), target = "y")
-  lapply(lrns, testBasicLearnerProperties, task = min.task)
+  # min.task = makeRegrTask("oneCol", data.frame(x = 1:10, y = 1:10), target = "y")
+  # lapply(lrns, testBasicLearnerProperties, task = min.task)
 })
```

it should be fixed in this PR

berndbischl:
and here:

```
diff --git a/tests/testthat/test_learners_all_classif.R b/tests/testthat/test_learners_all_classif.R
index 19dea492..f51e2190 100644
--- a/tests/testthat/test_learners_all_classif.R
+++ b/tests/testthat/test_learners_all_classif.R
@@ -61,8 +61,8 @@ test_that("learners work: classif", {
   lapply(lrns, testThatLearnerCanCalculateImportance, task = task, hyperpars = hyperpars)

   # classif with only one feature
-  min.task = makeClassifTask("oneCol", data.frame(x = 1:10, y = as.factor(rep(c("a", "b"), each = 5))), target = "y")
-  lapply(lrns, testBasicLearnerProperties, task = min.task)
+  # min.task = makeClassifTask("oneCol", data.frame(x = 1:10, y = as.factor(rep(c("a", "b"), each = 5))), target = "y")
+  # lapply(lrns, testBasicLearnerProperties, task = min.task)
 })
```

pat-s:
@jakob-r Could you describe in the first post what this issue is about? And how close it is to be merge ready?

It is still labeled as "blocking" but I guess this only applied to the release in 2018?

jakob-r:
If I see it correctly we do not test all learners if they only work with one feature. If you enable this test (see [here](https://github.com/mlr-org/mlr/pull/2378/files#diff-b83027ad0658455693111444296c50c8L60) and [here](https://github.com/mlr-org/mlr/pull/2378/files#diff-ef2cc592db5a22d79a152642ab415a79L65)) we will probably get errors in the tests.

In this PR I tried to (re)enable the test and fix all issues on the mlr side. As mentioned above I also tried to reach the maintainers if the problem was in their packages. Those packages were excluded [here](https://github.com/mlr-org/mlr/pull/2378/files#diff-ef2cc592db5a22d79a152642ab415a79R71) and [here](https://github.com/mlr-org/mlr/pull/2378/files#diff-ef2cc592db5a22d79a152642ab415a79R71). Probably some are already updated or new problematic learners were added.

jakob-r:
let me split this PR fast, becasue it has some things that belong into an extra PR

jakob-r:
Should be good to go.

pat-s:
Thanks. To summarize: We now have a list of learners that do not work with only one feature and the test is in general enabled again (excluding all of these learners).

Rather than having this list in the test file, can we have a separate issue for this? 
It will just get forgotten there.
It would be great if you could comment for each learner that does not work what you've tried so far (locally) so that we do not do redundant work (can be really short). 

**General question**
How important is it to test this within _mlr_? This sounds like something the packages should be responsible for. Or are we doing any modifications here so that the problem might be introduced from our side?

jakob-r:
>Rather than having this list in the test file, can we have a separate issue for this?

Problem is: It should be an issue in the respective learners. But we can make one issue sumarizing the state of each failing learner.

> It would be great if you could comment for each learner that does not work what you've tried so far (locally) so that we do not do redundant work (can be really short).

I can't remember what I tried back then. But it is made sure that the problem is was not on the mlr learner implementation side.

> How important is it to test this within mlr? 

It's helpfull if somebody complains that mlr is not working on a specific problem (especially if the one featuer task is generated deeply.

> This sounds like something the packages should be responsible for. 

Yes. As mentioned I sent emails but never got an answer. (unless reported otherwise)

> Or are we doing any modifications here so that the problem might be introduced from our side?

No. The Tests already fail on the learner side.

pat-s:
> Problem is: It should be an issue in the respective learners. But we can make one issue sumarizing the state of each failing learner.

Yes, could you open an issue for this?

> Yes. As mentioned I sent emails but never got an answer. (unless reported otherwise)

I can try to follow up once and then we make a decision what to do in that case.

PR #2385, opened 2018-07-17
Tut: Fix build and always build tutorial

pat-s:
Because the tutorial is only build for `master` commits and nothing else, I did not catch the current Travis error in the last PR.

I modified the setup so that now we always build the tutorial in the "after_script" stage and only the deployment is conditioned on commits to the `master' branch.

With this, we catch potential errors introduced by tutorial PRs.


larskotthoff:
Thanks, merging -- but please make separate PRs for separate changes in the future (the changes to the spatial part of the tutorial).

Issue #2383, opened 2018-07-13
Tutorial: different syntax highlighting for code and comments is needed


GegznaV:
In `mlr` tutorial web pages (e.g., https://mlr-org.github.io/mlr/articles/task.html), it would be less tiring and easier to follow R code in the code chunks, if there was a contrast between the code and the comments/results (e.g., if these elements were in different colors). Is it possible to have this feature?

GegznaV:
It is interesting: lines of code, that start with a single `#` are highlighted differently than lines with several hash symbols (`##`, `###`). There are some examples:

![image](https://user-images.githubusercontent.com/12725868/42722052-fa7d622a-874d-11e8-9695-bd01135f45fe.png)
versus
![image](https://user-images.githubusercontent.com/12725868/42722077-431cf8ce-874e-11e8-8f22-fd566691c2b3.png)

Basing on this, wouldn't it be reasonable to have a guideline to use only one hash symbol for comments in articles of `mlr` tutorial? Different colors really makes it easier to read.

pat-s:
Yes, you're right. We should change this. Most likely this can be found in more tutorial pages.

pat-s:
Fixed in aafc7de61435b111b8d9c6a67ddb70b122e84bff.

GegznaV:
I'm not sure if the issue hat to be closed as it was not solved. 

**The first** thing I suggest is adding more contrast in code highlighting, e.g., this CSS code:

```css
/* Syntax highlighting theme */

.fl      {color: #1514b5;}
.co      {color: #888888;}    /* comments */
.ch, .st {color: green;}      /* strings */
.dv, .ot {color: darkblue;}   /* numbers, logicals */
.sourceCode.r .op {color: #505050;}  /* operators */
.sourceCode.r .dt {color: #303030;}  /* function argument names,  #303030 - darkgray */
.sourceCode.r .kw, .fu, .cf, .sourceCode a:link {color:darkred;} /* function names */
```

in `pkgdown\extra.css` file to get more contrast in code highlighting, e.g.:


![image](https://user-images.githubusercontent.com/12725868/42885081-b5caa47c-8aa8-11e8-989d-0a68834f026e.png)

instead of 

![image](https://user-images.githubusercontent.com/12725868/42885157-ea357516-8aa8-11e8-8bc5-132f0399bb04.png)
 in [this](https://mlr-org.github.io/mlr/articles/functional_data.html) tutorial. And

[...]

PR #2387, opened 2018-07-18
 Tut: syntax highlighting modified (functions are more visible)

GegznaV:
Non-intensive colors added so some R code elements (in code chunks only) to make the elements more distinguishable. Find examples in #2383


larskotthoff:
Thanks, that makes it much more readable!

Issue #2388, opened 2018-07-18
Tut: Why is `knit_hook` used in vignettes?


GegznaV:
Why is **code 1** used instead of just **code 2** in vignettes?

```r
# Code 1
knitr::opts_chunk$set(collapse = FALSE)
knitr::knit_hooks$set(document = function(x){
  gsub("```\n*```r*\n*", "", x)
})
```

```r
# Code 2
knitr::opts_chunk$set(collapse = TRUE)
```

As I noticed, **code 1** makes code less readable as it unnecessarily removes empty lines,
e.g.:

![image](https://user-images.githubusercontent.com/12725868/42903955-0b37f62a-8adc-11e8-9c31-745aa4023630.png)

**_versus_** 

![image](https://user-images.githubusercontent.com/12725868/42903889-df34b2fc-8adb-11e8-8d43-c4b8795160c0.png)

In the second print screen, code highlighting was added, `###` converted into `#`, but the spaces were not added manually, see [here](https://github.com/mlr-org/mlr/blob/master/vignettes/task.Rmd#accessing-a-learning-task)).

I think, if the empty lines are really unnecessary, they can be removed manually.  But I think that it is better to remove `knit_hook` and leave a possibility to insert empty lines between the lines of code. Is there something I am missing about why the `knit_hook` is used? 

Of course, it's a matter of taste and personal preference how the code should be formatted. I have changed the code (removed hooks) on my computer, but I want to consult the authors of `mlr`  (@pat-s , etc.) if it's a good idea to convert the changes into a pull request?

PR #2393, opened 2018-07-19
Tut: Enable `collapse = TRUE`

GegznaV:
In the code of all tutorial `.Rmd` files in the PR, `knit_hook`:

```r
knitr::opts_chunk$set(collapse = FALSE)
knitr::knit_hooks$set(document = function(x){
  gsub("```\n*```r*\n*", "", x)
})
```

 is replaced with `collapse = TRUE` :

```r
knitr::opts_chunk$set(collapse = TRUE)
```
Solves #2388


jakob-r:
Can you preview this pages section (e.g. with a screenshot):

https://mlr-org.github.io/mlr/articles/tune.html#performing-the-tuning

GegznaV:
These print-screens are from the tutorial locally built on my computer:

image

image

image

@jakob-r, are the print-screens what you are interested in?

GegznaV:
If you are searching for code fragmentation as in the example below, `collapse = TRUE` prevents from it.

image

pat-s:
```
knitr::opts_chunk$set(collapse = FALSE)
knitr::knit_hooks$set(document = function(x){
  gsub("```\n*```r*\n*", "", x)
})
```

AFAIK these settings were important in the old `mkdocs` setups that we used before, weren' they @jakob-r?  
One reason was to not have a new output chunk for every code line as it shows up in your last screenshot. 

But apparently we have it there, but only in the "Get started" vignette. That's because we only have `collapse = FALSE` there while in all others we have the code above (probably just a copy/paste miss).

So we basically don't collapse the code by default and then manually remove all the empty lines between the code. Using `collapse = TRUE` does exactly this on a first glance. 

So in summary: This does not change anything to all the pages besides the "Get started" one because there we only used `collapse = FALSE` currently. 

But maybe I am overseeing an important point here from the old days why this was implemented this way. So I would wait for @jakob-r Feedback on this.

jakob-r:
There had been issues with `collapse = TRUE` just working partially which made this workaround necessary at some point. So probably @pat-s is correct. Thanks for providing the sceens @GegznaV. I just wanted to make sure.

GegznaV:
Closes #2388

PR #2394, opened 2018-07-20
integration of regr.gstat learner

pokyah:
Hi mlr's Team,

I'm pleased to submit my first and major contribution to your package for review.

This contribution implements a new learner : `regr.gstat` from the [gstat](https://cran.r-project.org/web/packages/gstat/) package developed by @edzer - http://edzer.github.io/

gstat aims to perform spatial predictions as IDW, various kind of Kriging, trend surfaces, etc.

The learner has been unit tested with the default settings and various parameters. 

An additional `usecase_gstat` vignette has been written. It presents the mlr-gstat functions coding with their raw-gstat counterparts and shows that their outputs are strictly identical.

I need gstat and mlr in the context of my research, this is why I've decided to start this gstat implementation contribution. This is my first experience with package development. I hope I've done things the good way (as explained in your [wiki](https://github.com/mlr-org/mlr/wiki/Developer-Instructions#mlr-coding-guidelines)) but as I'm not a advanced R user, I guess that many parts of my contribution could be improved.

I had a discussion with @pat-s about the kind of learner (`regr.` or `spatial.`). Spatial was retained but as the existing kriging learner `reger.km` is already classified as a regression learner, I've finally chosen `regr.gstat`.

Feel free to contact me for questions, suggestions, etc.

Thanks for your interest.

pokyah


DESCRIPTION
```diff
@@ -159,7 +159,9 @@ Suggests:
     tgp,
     TH.data,
     wavelets,
-    xgboost (>= 0.6-2)
+    xgboost (>= 0.6-2),
+    sp,
+    gridExtra
```
larskotthoff: We try to keep the list of packages sorted alphabetically; please fix that.

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+#' psill, model, range and nugget.
+#' To use auto fitting, simply provide a list containting the types of models to be tested.
+#' Manual fitting example : lrn.man = makeLearner(cl = 'regr.gstat', id = 'manual', psill = 1, model = 'Sph', range = 900, nugget = 1, locations = ~x+y).\n
+# 'Automatic fitting example : lrn.auto = makeLearner(cl = 'regr.gstat', id = 'auto', psill = c('Sph','Exp','Gau', 'Mat'), locations = ~x+y)"
+#'
+#' @references Edzer J Pebesma
+#' Multivariable geostatistics in S: the gstat package
+#' Computers & Geosciences Volume 30, Issue 7, 2004, 683-691.
+#'
+#'
```
larskotthoff: An example would be great.

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+
+#' @export
+makeRLearner.regr.gstat = function() {
+  makeRLearnerRegr(
+    cl = "regr.gstat",
+    package = "gstat",
+    par.set = makeParamSet(
+      # gstat::gstat params
+      makeFunctionLearnerParam(id = "g"),
+      makeUntypedLearnerParam(id = "id"), # FIXME what should be the type ?
```
larskotthoff: What does this parameter do?

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+      makeUntypedLearnerParam(id = "locations", default = ~x+y), # FIXME what should be the type ?
```
larskotthoff: See above. And what are the `x` and `y` here?

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+      makeLogicalLearnerParam(id = "fit.kappa", default = FALSE),
+      # gstat::vgm params
+      makeUntypedLearnerParam(id = "psill", default = NA),
+      makeUntypedLearnerParam(id = "model"), # FIXME what should be the type ?
+      makeUntypedLearnerParam(id = "range", default = NA),
+      makeNumericLearnerParam(id = "kappa", default = 0.5),
+      makeUntypedLearnerParam(id = "nugget"),
+      makeUntypedLearnerParam(id = "add.to"),
+      makeUntypedLearnerParam(id = "covtable"),
+      makeNumericLearnerParam(id = "Err", default = 0)
```
larskotthoff: Can you provide ranges for the parameters where possible please? Do e.g. all the numeric parameters from -inf to inf?

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+    ),
+    par.vals = list(locations = ~x+y, nmax = 0, nmin = 0, omax = 0, maxdist = Inf,
+      dummy = FALSE, fill.all = FALSE, fill.cross = TRUE, variance = "identity", degree = 0, vdist = FALSE,
+      alpha = 0, beta.variogram = 0, cressie = FALSE, dX = 0, boundaries = 0, cloud = FALSE,
+      trend.beta = NULL, debug.level = 1, cross = TRUE, map = FALSE, projected = TRUE, lambda = 1.0,
+      verbose = FALSE, covariogram = FALSE, PR = FALSE, pseudo = -1,
+      fit.sills = TRUE, fit.ranges = TRUE, fit.method = 7, warn.if.neg = FALSE, fit.kappa = FALSE,
+      psill = NA, range = NA, kappa = 0.5, Err = 0),
```
larskotthoff: You don't need to list the defaults again here; only if they are different from the default (with an explanation) and parameters you didn't specify in the list above but that need to be set.

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+    properties = c("numerics", "factors" , "se", "weights", "missings"),
+    name = "Multivariable Geostatistical Prediction And Simulation",
+    short.name = "gstat",
+    note = "The default prediction (when passing no arguments) is IDW (inverse distance weighted).
+    The learner handles gstat variogram autofitting functionnality presented in this post https://www.r-spatial.org/r/2016/02/14/gstat-variogram-fitting.html.\n
+    Manual fitting example : lrn.man = makeLearner(cl = 'regr.gstat', id = 'manual', psill = 1, model = 'Sph', range = 900, nugget = 2, locations = ~x+y).\n
+    Automatic fitting example : lrn.auto = makeLearner(cl = 'regr.gstat', id = 'auto', psill = c('Sph','Exp','Gau', 'Mat'), locations = ~x+y)"
```
larskotthoff: The examples should be in the roxygen doc above.

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+  variogram.names = names(formals(gstat::variogram)) # FIXME cannot retrieve the S3 method for formula arguments
+  variogram.names = c("object", "locations", "beta.variogram")
+  #variogram.names = replace(names(formals(gstat::variogram)), variogram.names == "beta", "beta.variogram")
+  fit.variogram.names = names(formals(gstat::fit.variogram))
+  gstat.names = replace(names(formals(gstat::gstat)), names(formals(gstat::gstat)) == "beta", "beta.gstat")
+
+  d = getTaskData(.task, .subset)
+  f = getTaskFormula(.task, explicit.features = TRUE)
+  # remove location vars as they are handled by gstat - https://stackoverflow.com/questions/40308944/removing-offset-terms-from-a-formula
+  f = update(f, .~.-y-x) # FIXME should be the params entered in locations arg
```
larskotthoff: It sounds like you want to check whether `x` and `y` are present and if not error or at least warn.

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+  if (!is.na(dots$psill)) {
+    # build the samples variogram
+    v = do.call(gstat::variogram, c(list(object = f, data = d), dots[ names(dots) %in% variogram.names] ))
+    # fit the variogram model
+    fit = do.call(gstat::fit.variogram,
+      c(list(object = v,
+        model = gstat::vgm(psill = dots$psill,
+          model = dots$model,
+          range = dots$range,
+          nugget = dots$nugget)),
```
larskotthoff: This is inconsistent with the documentation -- do you want these to be passed as arguments directly, or as a list for the `model` argument?

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+        dots[names(dots) %in% fit.variogram.names[fit.variogram.names != "model"]])
```
larskotthoff: Why is `model` filtered out here?

R/RLearner_regr_gstat.R
```diff
@@ -0,0 +1,164 @@
...
+    )
+    # create the gstat object with a model
+    g = do.call(gstat::gstat,
+      c(list(formula = f,
+        data = d,
+        model = fit
+        ),
+        dots[names(dots) %in% gstat.names[gstat.names != "model"]])
```
larskotthoff: See above.

R/RLearner_regr_gstat.R
pat-s: I assume coordinates? If so, they could be extracted from the task where one can set argument `coordinates` to make it a spatial task. Finally a good use case ;)

pat-s:
Thanks for your work @pokyah! 

Having an own page on just a single learner seems too much weight. I think the comparison on how to use `gstat` traditionally and via `mlr` might better fit into a blog post in r-spatial.org or the mlr-blog. Both parties would certainly be happy about it :)

We can add a little use case section in the `handling_of_spatial_data` vignette but it should be really tidy.

> I had a discussion with @pat-s about the kind of learner (regr. or spatial.). Spatial was retained but as the existing kriging learner reger.km is already classified as a regression learner, I've finally chosen regr.gstat.

Well the spatial thing is kinda new. My thought was just that this learner only makes sense with spatial data. Is this the same for `regr.km`? 
If the learner is named `regr.gstat` and people try to use it for ordinary regr. problems without spatial data, they might be confused that it is not working? So we should tag it before by applying a "spatial" prefix.

DESCRIPTION
```diff
@@ -159,7 +159,9 @@ Suggests:
     tgp,
     TH.data,
     wavelets,
-    xgboost (>= 0.6-2)
+    xgboost (>= 0.6-2),
+    sp,
```
pat-s: Is `sp` still needed for `gstat` We already have `sf`. Not sure if `gstat` has already migrated though.

DESCRIPTION
pat-s: We already use `cowplot` for gridded plot arrangements. Its more powerful in combination with ggplot. Give it a try :)

pat-s:
Ah and please remove all the man/ and NAMESPACE files. We will generate them manually.

You can do so easily via `git checkout master -- NAMESPACE` and `git checkout master -- man/*` if your master branch is up-to-date.

DESCRIPTION
pokyah: OK, I'll follow this rule

pokyah: I'll check that. And as gridExtra was required for the vignette, I'll remove it from the suggest section (see my comment related to sp)

R/RLearner_regr_gstat.R
pokyah: I'll add it

R/RLearner_regr_gstat.R
pokyah: Yes, `x` and `y` are indeed coordinates expressed as a formula. In this case should the I use ` makeFunctionLearnerParam` rather than `makeUntypedLearnerParam` ?

R/RLearner_regr_gstat.R
pokyah: Sure, I'll do this

R/RLearner_regr_gstat.R
pokyah: Ok. Could you than briefly explain me the purpose of par.vals ?

R/RLearner_regr_gstat.R
pokyah: Ok, I'll move it.

R/RLearner_regr_gstat.R
pokyah: I actually need to rewrite the formula as gstat "knows" (i.e. does not required coordinates in its the formula). My problem here is that `y` and `x` are hard coded. So my code assumes  that coordinates will always be referred as `x` and `y`. But it should also work for other names (as e.g. `lat`, `lon`) passed to the parameter with `id = location`. I need to find a way to dynamically retrieve the names of these parameters to update my formula.

R/RLearner_regr_gstat.R
pokyah: from [gstat doc](https://www.rdocumentation.org/packages/gstat/versions/1.1-6/topics/gstat) : *identifier of new variable; if missing, `varn` is used with n the number for this variable*.
But this is still quite cryptic to me... I'll do some more investigations

R/RLearner_regr_gstat.R
pokyah: The `model` argument you must pass to the learner, is the type of model you want `gstat::vgm` to generates. Might be e.g. `"Exp"`, `"Sph"`, `"Gau"`, `"Mat"`. See [vgm doc](https://www.rdocumentation.org/packages/gstat/versions/1.1-6/topics/vgm).
The `gstat::fit.variogram` function also requires a model argument, which is actually the output of the `gstat::vgm` function.

R/RLearner_regr_gstat.R
pokyah: See comment above. `gstat::vgm` and `gstat::fit.variogram` both require a `model` argument. The `model` argument you pass to the learner is the one that goes to `gstat::vgm` which output is the `model` argument required for `gstat::fit.variogram`. So while constructing `fit.variogram`, I must remove the model argument stored in `...` (`dots`) because it only as a meaning in the `gstat::vgm` function.

R/RLearner_regr_gstat.R
pokyah: Same

pokyah:
Thanks for your constructive feedbacks. 

Some additional comments : 

> If the learner is named regr.gstat and people try to use it for ordinary regr. problems without spatial data, they might be confused that it is not working? So we should tag it before by applying a "spatial" prefix.

So, should I use`makeRLearner.spatial.regr.gstat` to construct my learner ? 

> Well the spatial thing is kinda new. My thought was just that this learner only makes sense with spatial data. Is this the same for `regr.km`?

Yes, from what I've read in the [dicekriging doc](https://www.rdocumentation.org/packages/DiceKriging/versions/1.5.5) `regr.km` is also for spatial problems.

pat-s:
@pokyah Now you removed all the `man/` and NS files so that they would be removed in the PR as well. What you need to do is to make them "neutral" so that they are not changed at all. The git command I wrote should do that job.

Also you committed all the `docs/` files. For changes in the tutorial just commit the changed `.Rmd` file. No files in `docs/`.

R/RLearner_regr_gstat.R
larskotthoff: It certainly sounds like you want to set tunable = FALSE for this one.

R/RLearner_regr_gstat.R
larskotthoff: I would be tempted not to make this parameters at all and instead say in the documentation that coordinates always need to be x and y.

R/RLearner_regr_gstat.R
larskotthoff: I would try to make that clearer with different names, otherwise people might get confused.

pokyah:
Hi all ! 
I'll take time to make the proper correction and improvements in the next few days as I'm busy with other urgent tasks at the moment. 
Thanks for your comprehension
All the best !

DESCRIPTION
pokyah: I think we can remove sp from the mlr DESCRIPTION file as gstat also works by passing a dataframe containing longitude and latitude columns (that must be specified). You can check this on the [gstat doc](https://www.rdocumentation.org/packages/gstat/versions/1.1-6/topics/gstat)

R/RLearner_regr_gstat.R
pokyah: Alright, I'll proceed this way

R/RLearner_regr_gstat.R
pokyah: So, what is the exact purpose of par.vals ?

R/RLearner_regr_gstat.R
pokyah: This confusion comes from gstat nomenclature and code itself. What I can do is rename the learner `model` argument by `model.vgm` so that people understand that it is the kind of model expected by the gstat::vgm function

pokyah:
> @pokyah Now you removed all the man/ and NS files so that they would be removed in the PR as well. What you need to do is to make them "neutral" so that they are not changed at all. The git command I wrote should do that job.

What is the exact workflow that I should follow once the code editing has been performed in my gstat branch  ?

I don't understand how to combine the commands you mentioned here with this doc page : https://github.com/mlr-org/mlr/wiki/Developer-Instructions#merging-the-master-in-you-branch

pokyah:
Hi ml-org,

May I ask you to review this new version of my code ?

Thank you.

All the best

pat-s:
> Hi Pat ! I've seen your comment about appVeyor logs. It seems that the warnings and errors are generated by the example I've provided in my RLearner_spatial_regr_gstat.R file. Indeed, the example makes calls to sp (to transform the .sp meuse dataset back to a .df) and gstat libraries. For now I've simply deleted the example to avoid such errors. It seems that other learners don't provide an example. I guess this is not problematic. Let me know if it is ok for you if I simply remove the example

Well, ofc deleting stuff is always the easiest way to overcome problems ;) But that was not my intention here. It is about the correct importing of new packages in a development env in general.
I suggest to become more familiar with the pkg dev workflow: http://r-pkgs.had.co.nz/description.html. It differs a bit from project to project but its good to have be aware of it. 
And it applies here at least for the _dplyr_ and _magrittr_ usage.
I know R pkg dev can be exhausting in the beginning with all these rules but that's how the system works :duck: 	
Once you understood, it all makes sense (most often at least). :sweat_smile: 

Regarding the specific issue here: We usually don't have learner examples, yes. 
So you can remove the example.
However (I think I've said this in the previous review already) we already have a spatial.task for spatial stuff. Is there anything special about the meuse data that cannot be done with spatial.task? Please try to use this task for testing purposes.

pokyah:
> > Hi Pat ! I've seen your comment about appVeyor logs. It seems that the warnings and errors are generated by the example I've provided in my RLearner_spatial_regr_gstat.R file. Indeed, the example makes calls to sp (to transform the .sp meuse dataset back to a .df) and gstat libraries. For now I've simply deleted the example to avoid such errors. It seems that other learners don't provide an example. I guess this is not problematic. Let me know if it is ok for you if I simply remove the example
> 
> Well, ofc deleting stuff is always the easiest way to overcome problems ;) But that was not my intention here. It is about the correct importing of new packages in a development env in general.
> I suggest to become more familiar with the pkg dev workflow: http://r-pkgs.had.co.nz/description.html. It differs a bit from project to project but its good to have be aware of it.
> And it applies here at least for the _dplyr_ and _magrittr_ usage.

dplyr already imports magrittr, so I suppose that I don't need to add magrittr in the imports section of the DESCRIPTION file.

> I know R pkg dev can be exhausting in the beginning with all these rules but that's how the system works 
> Once you understood, it all makes sense (most often at least). 

I've already learnt a lot about package dev with this PR. Thanks for all the constuctive comments, answers. Things become to be less mysterious but I still need to dig further to understand the whole process ;) 

> 
> Regarding the specific issue here: We usually don't have learner examples, yes.
> So you can remove the example.

My choice was to use the meuse dataset because this is the one that comes with the original gstat package. And a lot of online resources about gstat refer to this package. The intention was to make things easy to understand for people already a bit familiar with gstat. For now I'll drop the example as I have a lot of other tasks on my project todo list for the moment. 

> However (I think I've said this in the previous review already) we already have a spatial.task for spatial stuff. Is there anything special about the meuse data that cannot be done with spatial.task? Please try to use this task for testing purposes.

I'll see how I can use spatial.task in my unit testing. This will require some more work but I'm OK to spend some more time on it right now.

pat-s:
Superseeded by #2589

PR #2396, opened 2018-07-23
CI: Add CI for Windows (appveyor)

pat-s:
Should be low maintenance compared to Travis (no deploys, no warmup, no cache problems).

Example: https://ci.appveyor.com/project/pat-s/mlr

## Pro
- CI testing on Windows
- Very fast (with cache: 9 min, without cache: 20 min (!))
- Because we already use `tic`, the main stuff is controlled via `tic.R` 
- We become a bit more independent from Travis 

## Con
- We can only have a summed cache of up to 1 GB in the free plan (one cache needs 5xx GB so we can only have a `master` cache.)
- All PR/branches would need to build without cache (still takes only 20 min and hence is faster than a normal Travis build)

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -27,3 +27,4 @@ logo.png
 ^pkgdown$
 tic.R
 tutorial/*
+appveyor.yml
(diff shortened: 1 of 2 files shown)


larskotthoff:
This is great, thanks for doing it!

PR #2402, opened 2018-07-26
Tut: links fixed

GegznaV:
1. A broken link fixed.
2. A link changed to a more appropriate one which points to the function reference of the  `mlr` tutorial (which is usually more up to date) and no to an external website (www.rdocumentation.org).


larskotthoff:
Thanks, merging.

PR #2404, opened 2018-07-28
remove lqa (fixes #2403)

### LINKED ISSUE #2403: Package lqa was removed from CRAN
On July 20th. That's why the `master` build fails. 

```
── 1. Error: MulticlassWrapper (@test_base_MulticlassWrapper.R#23)  ───────���────
For learner classif.lqa please install the following packages: lqa
1: makeLearner("classif.lqa") at testthat/test_base_MulticlassWrapper.R:23
```

larskotthoff:


berndbischl:
can you put this files alwas in the extra-learners repo in such cases?

berndbischl:
> Well hopefully their suspension from CRAN is only temporarily and then we can easily bring it back from the history. I don't really see the point of keeping things around that can't be used because the underlying package no longer exists.

i kinda find both arguments wrong:
- they are now removed so have serious problems, and this is not the most important learner. so why not stop copying it back and forth, monitoring it, and so on.
- just because the package is not on cran does not mean it does not exist anymore?

also i think we should have a standard procedure for cases like this

pat-s:
> they are now removed so have serious problems, and this is not the most important learner. so why not stop copying it back and forth, monitoring it, and so on

Yeah we have a bunch of learners that are less used and obv not well maintained by the package owners. But I also do not see the point why we should have a learner in the "extra-leaner" repo for which the package is unmaintained. Learners in that repo do not have tests and/or are not fully implemented but at least rely on packages that are maintained on CRAN. 

If we already have tests for a learner and the package is back on CRAN, we can reintegrate it here. Moving it back and forth between the "extra-learners" repo and the main one is kinda confusing. 

The question is: How do we monitor such cases and who does it?

larskotthoff:
And then of course we have todo-files, which also contains a bunch of learners. I just dislike the idea of having a separate place with code which can't quite be used if this is already in the git history.

pat-s:
This seems to be become a general discussion. How do we proceed here? As its prio "blocking", we should agree on one solution soon.

Lars and I are pro "not moving to extra-learners" for different reasons.
@berndbischl, are you still in favor of moving it to the "extra-learners" repo? I guess your vote would override ours ;)

pat-s:
Ping @berndbischl

larskotthoff:
Blocking, so I'll merge this. We can always move stuff to extra learners later.

berndbischl:
merging: well done.

extra-learners:
moving back and forth between extra learners and here:  who says we are doing this? 
what is the downside of having the code there? todo-files is a bad place for such code which MIGHT still be useablke for people. and nobody really add much "maintenance" work to extra-learners. 
the idea is, that in the future learners there get separate tests. but they dont all have to pass.

but can somebody tell me where the code from this PR now went? did you just delete this? this seems bad.

berndbischl:
so now we have 3 places apparently were stuff gets put in these cases. including one which is not easily visible / searchable. can we please agree on a standard here?

berndbischl:
I mean, this is not the most important topic "stuff not that not really works", but it would be great if we agree how this should be handled so we dont have to argue it back and forth. if you dislike moving it to another repo can you at least tell me why?

berndbischl:
> . Learners in that repo do not have tests and/or are not fully implemented but at least rely on packages that are maintained on CRAN.

also this is wrong. the repo was exactly *inventend* for learners NOT on CRAN - as those cannot be put into mlr directly. point in case: mxnet.

pat-s:
todo files becomes bigger and bigger and no one seems to have motivation to clear this up. Putting it there seems to have the effect "out of sight, out if mind". 

* todo files contain learners that are mainly not usable (correct me if I am wrong).
* extra-learners contains learners that are not finished/fully implemented or are not on CRAN and have no tests(?)
* git history can easily be used to revert stuff but we should definitely note it somewhere, otherwise it will be forgotten at some point

So to not forget the lda learner, we should at least place its .R file in todo files and add a note at the top that it can be integrated again by reverting the PR and that it was only removed because the package was removed from CRAN.

> also this is wrong. the repo was exactly inventend for learners NOT on CRAN - as those cannot be put into mlr directly. point in case: mxnet.

Ah ok, thanks for the heads-up!

berndbischl:
i dont get what you are writing. can we move this to slack an clear this uoe?

berndbischl:
i talked with patrick on slack and we agree that in such cases the learner should be removed from mlr and moved to extra-learners. maybe with a note why it was moved.

larskotthoff:
In extralearner repo now.

PR #2407, opened 2018-08-02
Robustify travis

mllg:
* Make sure that lock files will not get cached which can break future builds 
* Make sure that all packages are updated by explicitly calling `update_packages`.   Otherwise only packages in the current depencency graph are updated, neglecting those in the suggests of packages mlr depends on.


larskotthoff:
Great, thanks.

PR #2410, opened 2018-08-14
Change the note of ibrier

PhilippPro:
Add to the note that it currently only works for two methods.


larskotthoff:
No idea why that one check failed, but merging anyway. Thanks!

PR #2412, opened 2018-08-14
Enable using fully predefined indices in resampling

pat-s:
superseeds #2267 

### To do

- [x] Write tutorial section and explain the difference to "blocking"

### Idea

Use fully predefined indices via the `blocking` argument in the task for resampling.

#### Single resampling
4 classes -> 4 folds. 

##### Nested Resampling

Outer: 4 classes -> 4 folds
Inner: 3 classes -> 3 folds 

Only compatible with 1 repetition, hence only with "CV".

To use predefined indices in "RepCV" one should use the already existing "Blocking" implementation. This diiffers in the way what the class number does not also define the number of folds and hence more combinations than just the number of folds can be generated.

### Implementation

The user needs to initiliaze this special resampling by setting `fixed = TRUE` in `makeResampleDesc()`. Otherwise, the "blocking" implementation will be used.

As said, the approach uses the factor variable supplied in the task via the `blocking` argument. I know this can cause confusion between "blocking" and "grouping" but we need to differentiate both approaches somehow.
On the other hand we do not need a new argument taking a factor vector in the task.

In a nested setting, a possible workflow would look as follows:

```r
inner = makeResampleDesc("CV", iters = 4, fixed = TRUE)
outer = makeResampleDesc("CV", iters = 5, fixed = TRUE)
tune_wrapper = makeTuneWrapper(lrn, resampling = inner, par.set = ps,
 control = ctrl, show.info = FALSE)

p = resample(tune_wrapper, ct, outer, show.info = FALSE,
 extract = getTuneResult)
```

So rather than doing a random sampling, we use the predefined indices specified in "blocking".

The function is smart enough to also deal with a little mispecification by issueing a warning:

```r
inner = makeResampleDesc("CV", iters = 5, fixed = TRUE)
outer = makeResampleDesc("CV", iters = 5, fixed = TRUE)

"iters (5) is not equal to length of blocking levels (4)!
```

If inner > outer, an error will be thrown. 

By logic, the inner fold count needs always to be one less the outer count.

Users can also combine using fixed indices in the outer and random sampling in the inner:

```r
inner = makeResampleDesc("CV", iters = 5)
outer = makeResampleDesc("CV", iters = 5, fixed = TRUE)
tune_wrapper = makeTuneWrapper(lrn, resampling = inner, par.set = ps,
  control = ctrl, show.info = FALSE)
expect_success(resample(tune_wrapper, ct, outer, show.info = FALSE,
  extract = getTuneResult))
```

To explicitly avoid clashes between "fixed" and "blocking" when a "blocking" factor was given in the task, I had to add a little helper arg.
To use "blocking" in single "CV", the user now needs to explicitly enable it by using `makeResampleDesc("CV", iters = 5, blocking.cv = TRUE)`.
But I think people would always use "blocking" in "repCV" anyways I guess?

Just for clarification: This PR changes nothing on the existing "blocking" implemen tation besides the need to explicitly trigger it when using "CV".


larskotthoff:
Thanks.

What exactly does `fixed = TRUE` do here? I didn't understand the difference between your first and second example.

Since the number of iterations is fixed by the number of levels, why not make this the automatic choice instead of asking the user to specify it again?

pat-s:
> What exactly does fixed = TRUE do here? I didn't understand the difference between your first and second example.

`fixed = TRUE` triggers the usage of the indices specified in "blocking". Otherwise, a "normal" random sampling CV is applied. 

> Since the number of iterations is fixed by the number of levels, why not make this the automatic choice instead of asking the user to specify it again?

I didn't do this first because I had problems distinguishing between inner and outer and it was easier to get help by the user here (inner should always be one less than outer).
However, because the function can adapt correctly if both levels are set to the maximum number of levels now, I could just hardcode this. Thanks!

(Another reason was that for a long time I did not use "fixed" and so I had no flag telling me whether I am in a "blocking", "fixed" or "normal" setting.)

pat-s:
Vignette update added. Please review using the netlify preview: https://deploy-preview-2412--nervous-hopper-4136be.netlify.com/articles/resample.html

R/ResampleDesc.R
```diff
@@ -64,6 +64,9 @@
 #'     else it will be a fraction of the total training indices. IE for 100 training sets and a value of .2, the increment
 #'     of the resampling indices will be 20. Default is \dQuote{horizon} which gives mutually exclusive chunks
 #'      of test indices.}
+#'   \item{fixed (`logical(1)`)}{Whether indices supplied via argument 'blocking' in the task should be used in resampling. Default is `FALSE`.
```
larskotthoff: This wording suggests that blocking is ignored in resampling, and the documentation for blocking says the opposite.

R/ResampleDesc.R
```diff
@@ -64,6 +64,9 @@
...
+#'     'grouping' only works with 'CV' and the supplied indices must match the number of observations.}
```
larskotthoff: Where does the `grouping` come from?

R/ResampleDesc.R
```diff
@@ -64,6 +64,9 @@
...
+#'   \item{blocking.cv (`logical(1)`)}{Should 'blocking' be used in 'CV'? Default to `FALSE`}
```
larskotthoff: It sounds like this does the same thing as `fixed`.

R/ResampleInstance.R
```diff
@@ -61,7 +61,26 @@ makeResampleInstance = function(desc, task, size, ...) {
   if (length(blocking) && desc$stratify)
     stop("Blocking can currently not be mixed with stratification in resampling!")
 
-  if (length(blocking)) {
+  # 'fixed' only exists by default for 'CV' -> is.null(desc$grouping)
+  # only use this way of blocking of 'fixed = FALSE' -> is.null(desc$grouping)
+  if(is.null(desc$fixed)) {
```
larskotthoff: Space after `if` (isn't this supposed to be checked automatically?).

larskotthoff: Also I would give `fixed` a default value in `makeResampleDesc` so that this can never be `NULL` to remove the additional check.

R/ResampleInstance.R
```diff
@@ -61,7 +61,26 @@ makeResampleInstance = function(desc, task, size, ...) {
...
+      # set to FALSE for all rdesc that do not have the arg
+      if(is.null(desc$blocking.cv)) {
+        blocking.cv = FALSE
+      } else {
+        blocking.cv = desc$blocking.cv
+      }
+    } else {
+      fixed = TRUE
+    }
+  }
```
larskotthoff: With default values for `fixed` and `blocking.cv` this could be simply `fixed = desc$fixed; blocking.cv = desc$blocking.cv`.

R/ResampleInstance.R
```diff
@@ -61,7 +61,26 @@ makeResampleInstance = function(desc, task, size, ...) {
...
+
+  if (length(blocking) && !fixed && blocking.cv) {
```
larskotthoff: I would use an explicit `length(blocking) > 0` here.

R/ResampleInstances.R
```diff
@@ -8,10 +8,52 @@ instantiateResampleInstance.HoldoutDesc = function(desc, size, task = NULL) {
...
+  } else {
+
+    # CV with only predefined indices ("grouping")
+
+    if(is.null(task$blocking)) {
+      stopf("To use blocking in resampling, you need to pass a factor variable when creating the task!")
+    }
+
+    if (desc$iters != length(levels(task$blocking))) {
+      desc$iters = length(levels(task$blocking))
```
larskotthoff: Why is the number of iterations being adjusted here? There should be a warning.

R/ResampleInstances.R
```diff
@@ -8,10 +8,52 @@ instantiateResampleInstance.HoldoutDesc = function(desc, size, task = NULL) {
...
+    }
+    levs = levels(task$blocking)
+    n_levels = length(levels(task$blocking))
```
larskotthoff: Or just `length(levs)`.

R/ResampleInstances.R
```diff
@@ -8,10 +8,52 @@ instantiateResampleInstance.HoldoutDesc = function(desc, size, task = NULL) {
...
+    attr(desc, "class")[1] = "CVHelperDesc"
+    # create fake ResampleInstance
+    inst = instantiateResampleInstance(desc, n_levels, task)
+    attr(desc, "class")[1] = "CVDesc"
+
+    # now exchange block indices with indices of elements of this block and shuffle
+    test.inds = lapply(inst$test.inds, function(i) which(task$blocking %in% levs[i]))
+
+    # Nested resampling: We need to create a list with length(levels) first.
+    # Then one fold will be length(0) because we are missing one factor level because we are in the inner level
```
larskotthoff: What happens if the number of outer folds is less than the number of levels (or a simple train/test split) and more than one factor level is missing?

tests/testthat/test_base_fixed_indices_cv.R
```diff
@@ -0,0 +1,64 @@
...
+  ct = makeClassifTask(target = multiclass.target, data = multiclass.df,
+    blocking = fixed)
+
+  # test blocking in single resample
+  lrn = makeLearner("classif.lda")
+  rdesc = makeResampleDesc("CV", fixed = TRUE)
+  p = resample(lrn, ct, rdesc)$pred
+
+  # check if all test.inds are unique
+  expect_length(unique(unlist(p$instance$test.inds, use.names = FALSE)), 150)
```
larskotthoff: Should also check whether the right observations are together.

tests/testthat/test_base_fixed_indices_cv.R
```diff
@@ -0,0 +1,64 @@
...
+
+  # check that a combination of fixed and normal random sampling works
+  inner = makeResampleDesc("CV", iters = 6)
+  outer = makeResampleDesc("CV", fixed = TRUE)
+  tune_wrapper = makeTuneWrapper(lrn, resampling = inner, par.set = ps,
+                                 control = ctrl, show.info = FALSE)
+  p = resample(tune_wrapper, ct, outer, show.info = FALSE,
+    extract = getTuneResult)
+  expect_length(getResamplingIndices(p, inner = TRUE)[[1]][[1]], 6)
+
```
larskotthoff: I would also check what happens when inner and outer resampling are combined in a way that doesn't split up the levels evenly.

R/ResampleInstance.R
pat-s: Yes this simplifies things. I've set defaults in `makeResampleDesc()`.

R/ResampleInstances.R
pat-s: Two reasons: 

1. If the user inserts a different number than levels used for `fixed` in `makeResampleDesc()`, the function will error. 
2. In the inner call, the function is able to adapt by automatically reducing one level. So having always `length(iters) = length(levels(task$blocking)` is the most safe environment for the function to work. 

I added the explanation as a comment.

R/ResampleInstances.R
pat-s: Yes, thanks.

R/ResampleInstances.R
pat-s: In https://github.com/mlr-org/mlr/blob/192355ca3decbfe8d27d6a658a3072c7e2ac6442/R/ResampleInstances.R#L27-L28 we check that the number of folds is always = number of levels.

tests/testthat/test_base_fixed_indices_cv.R
pat-s: Done.

tests/testthat/test_base_fixed_indices_cv.R
pat-s: I think we cannot check this.

Example: If for 200 obs the user would input 5 levels with 50, 20, 80, 10 and 40 and then tries to do a CV10 in the inner, its a user problem whether it makes sense to have a LOOCV in the inner for one fold (10/10). 

Or do you mean sth different here? An uneven split in the outer should not be a problem. And the inner adapt to the outer anyhow.

R/ResampleInstance.R
pat-s: Set a default value.

R/ResampleDesc.R
pat-s: I added more detail to the doc and a link to the (not yet existing) tutorial page.

R/ResampleDesc.R
pat-s: Leftover, should be `fixed`

R/ResampleDesc.R
pat-s: Made it more clear now. Please take another look :)

pat-s:
@larskotthoff 
Sorry for the delay, I was on vacation and busy with some other stuff.

Hope you still know whats going on here :)

Once the technical part is approved, I'll update the tutorial section.

R/ResampleInstances.R
larskotthoff: Ah ok. Could you add information on what the number of levels was and what it was set to in the warning please?

tests/testthat/test_base_fixed_indices_cv.R
larskotthoff: Ok, as the inner split levels are adjusted automatically this shouldn't be a problem.

vignettes/tutorial/resample.Rmd
```diff
@@ -305,16 +305,84 @@ r = resample("regr.rpart", bh.task, rdesc, show.info = FALSE)
...
+
+To check on the inner resampling indices, you can call `getResamplingIndices(inner = TRUE)`.
+You can see that for every outer fold (List of 5), four inner folds were created that honor the grouping supplied via the `blocking` argument.
+
+Of course you can also use a normal random sampling "CV" description in the inner level by just setting `fixed = FALSE`.
+
+```{r}
+str(getResamplingIndices(p, inner = TRUE))
+```
+
```
larskotthoff: Could you please mention here that the number of inner folds is automatically adjusted based on the available levels?

R/ResampleInstances.R
pat-s: The problem we have here is the following:

* The warning is triggered for both inner and outer level
* Determining the level (inner/outer) is tricky
* The inner level resets to the factor levels of the outer level first and is then further adjusted. So we actually get a false positive for the inner level even if its set correctly (e.g. outer = 5, inner = 4). 

Even if this doesn't sound logical, I would even vote for the complete removal of the warning. Users who set `fixed = T` usually know what they want and what they need to set. 
Its contra-productive if the warning is raised even if the specification is correct (e.g. outer = 5, inner = 4). 

Not sure if this thinking if easy to follow here. Let me know if I should explain it again in more detail.

I would propose to mention the adjustment in the tutorial (and in the help page?).

vignettes/tutorial/resample.Rmd
pat-s: Yes, as said, I will write a new section comparing blocking and grouping and explain what happens.

R/ResampleInstances.R
larskotthoff: But at some point you can figure out whether there are enough levels for the folds, right? So there shouldn't need to be any false positives?

pat-s: Further down in the function I have the final number of levels. But then I have no information if the level count was adjusted or not. And even if I would add a flag, I am missing the original level count since `desc$iters` is reassigned. 

I think the effort of implementing a robust warning for both inner and outer is not worth the effort.
I would prefer to note it in the help page (in details) and in the tutorial.

Something like "Setting iters with fixed = T has no effect. iters will be set to length(blocking.levels) in the outer and length(blocking.levels) - 1 in the inner level".

pat-s: Thanks for discussing this Lars! It's also not a solution I am completely happy with and a problem I thought a lot about. 

I'll do the required changes soon, including the tutorial page. Finally getting this done.

pat-s:
@larskotthoff 	updated help page and tutorial - please take a look.

Remember that you can use the netlify preview of the docs (https://deploy-preview-2412--nervous-hopper-4136be.netlify.com/articles/tutorial/resample.html) once the pkgdown files have been deployed by Travis.

larskotthoff:
Thanks, merging.

PR #2413, opened 2018-08-15
getResamplingIndices(): Translate inner resampling indices to outer indices

### LINKED ISSUE #2409: About inner resampling indices
Maybe the following is already known. Writing it down anyways.

While implementing the feature of using predefined indices in resampling AND nested resampling (see also #2267), I got confused while looking closer at the inner indices. But let's start at the beginning.

With the new function `getResamplingIndices(inner = TRUE)`, its possible to extract resampling indices from `resample` objects.

However, the inner indices do not correspond to indices of the task that was used in the `resample` call. And there is a reason for it. Let's consider the following example:

*Nested CV, 5 folds outer, 5 folds inner, 100 observations*

For Fold 1, 80 observations will be used for training. The alg will be tuned on this 80 obs, hence splitting these again into 5 folds.
Let's assume we used obs 1:70 and 90:100 for Fold 1. 70:90 are the test set.

Since the general idea is based on subsetting the task by indices, we face the following problem:
In the nested setting, we have 80 obs left (for training). Hence, we cannot use the index numbers from the initial task to subset here since these are > 80 and we get the error "subscript out of bounds".

The subsetting is done by taking the indices from the new data.frame that only consists of indices from 1:80.
Now, in the end, these indices will be populated to be the "inner indices".

If one queries the inner indices of a nested resampling call using `getResamplingIndices(inner = TRUE)`, one will think that the indices correspond to the observations in the initial task (which is 100L). 
But that is wrong: They correspond to a (shuffled) subset of the task of Fold 1 which is not populated to the `resample()` result. They will never be > 80 due to the subsetting in the nested setting.

While this does not affect reproducibility because the actual observations used are still the desired ones (even if predefining them via a factor variable through the `blocking` argument), they are basically unusable because they cannot be referred to the anything.
Not only this, they may even misleading if one does that and then gets confused why certain observations got selected.

#### Info

I am currently finishing up #2267 (probably in a fresh branch), implementing the ability to use observations set in "blocking" also in a nested setting. 
While doing so, I stumbled over this issue, questioning myself why the inner indices do not correspond to the grouping variable I used.
While the approach is perfectly fine in general and can be proved by comparing descriptive stats of the initial observations in the task and the inner indices used, it cannot be proved by simply using the indices and relating these to the observations. In facts, its even misleading.

I suggest we issue a warning (at least)?

pat-s:
fixes #2409 

So they match the task indices.

The test I added uses a basic "CV" with predefined indices (`fixed = TRUE`) since this resamp method groups the indices (1:30, 31:60, 61:90, etc) and the translation checking is more easy to understand. 

The implementation is ofc hard to review since it uses three nested `map()` calls. But I think there is no other way to implement this.

The test essentially checks that in the inner inds have values that go up to the max index value of the task (in this case 150). Before, the max of inner inds was "obs - (obs/nfolds)", which evaluates to 120 in this test with 150 obs.

Can only be merged after #2412 since it uses code from this PR.


pat-s:
As a side not: I moved `purrr` from suggest to imports as we use it in various functions now.
Its not about the simple `map()` function but about the enhanced derivates such as `imap()` or `pmap()` that make it worth using `purrr`.

The type safetyness of `map_int()` and others can be seen as additional syntactic sugar.

larskotthoff:
There seem to be a lot of spurious/outdated changes in this branch. Could you please check that?

pat-s:
have to check what's wrong in the example.

tests/testthat/test_base_resample_getResamplingIndices.R
```diff
@@ -45,3 +45,28 @@ test_that("getResamplingIndices works with getFeatSelResult", {
...
+  ctrl = makeTuneControlRandom(maxit = 2)
+  ps = makeParamSet(makeNumericParam("nu", lower = 2, upper = 20))
+  inner = makeResampleDesc("CV", iters = 4, fixed = TRUE)
+  outer = makeResampleDesc("CV", iters = 5, fixed = TRUE)
+  tune_wrapper = makeTuneWrapper(lrn, resampling = inner, par.set = ps,
+    control = ctrl, show.info = FALSE)
+  p = resample(tune_wrapper, ct, outer, show.info = FALSE,
+    extract = getTuneResult)
+
+  inner_inds = getResamplingIndicesNew(p, inner = TRUE)
```
larskotthoff: Typo here? Tests fail.

larskotthoff:
Thanks, merging.

Issue #2405, opened 2018-07-30
Partial Dependence Plot: Flat line despite dependence


leoKampfer:
I try to plot the partial dependence but only receive a flat line - although the data indicates some dependence.
(I am fully aware that the example is not very rational - but it does the trick.)

```r
set.seed(42)
x <- rnorm(50, mean = 2, sd = 0.5)
z <- sample(c(0,1), 50, replace = TRUE)
data <- as.data.frame(cbind(x,z))
data$z = as.factor(data$z)
tsk = makeClassifTask(data = data, target = "z")
lrn = makeLearner("classif.randomForest", predict.type = "prob")
fit = train(lrn, tsk)
pd = generatePartialDependenceData(fit, tsk)
plotPartialDependence(pd)
```
The `plotPartialDependence()` function produces a straight line also `pd$data` indicates dependencies:
```r
#> pd$data
#        0         x
# 1: 0.658 0.6717723
# 2: 0.270 0.9463890
# 3: 0.024 1.2210057
# 4: 0.854 1.4956224
# 5: 0.994 1.7702391
# 6: 0.318 2.0448558
# 7: 0.334 2.3194726
# 8: 0.002 2.5940893
# 9: 0.000 2.8687060
#10: 0.000 3.1433227
```

jakob-r:
I just tracked it down. One could say it's a bug in ggplot. One of your target levels is `0` and defining this column as one axis in ggplot just gets everything to `0`.

jakob-r:
As a workaround you could recode your target.
```r
levels(data$z) = make.names(levels(data$z))
```

jakob-r:
Should work now with PR #2414

PR #2414, opened 2018-08-16
Fix Partial Dependence Plots for Numerical Targets

jakob-r:
What was the problem?

Issue #2405 
`generatePartialDependenceData` created da date.frame with numeric colnames e.g. `0`. This got interpreted by `aes_string` directly as the number and not the column.

Solution

Sanitize names.

What could be another solution?

Avoid numeric colnames in the beginning

Why did I not do this?

Would take some more time to go through the code.

What is missing?

Tests, although I am not sure how to test it most efficiently.


larskotthoff:
Thanks, merging.

Issue #2337, opened 2018-06-27
add case weights for surv.ranger


nikoltoll:
Similarly to classif.ranger and regr.ranger: could case weights be enabled for Rlearner.surv.ranger?

PhilippPro:
Solved in the newest commit (#2418).

Issue #2328, opened 2018-06-23
regr.ranger "minprop" limits are wrong


mb706:
The `minprop` parameter of `regr.ranger` is limited to betwee 0 and 0.5, so the ParamSet upper limit of 1 is wrong.
```r
> train(makeLearner("regr.ranger", splitrule = "maxstat", minprop = 0.6), bh.task)
Error in ranger::ranger(formula = NULL, dependent.variable = tn, data = getTaskData(.task,  : 
  Error: Invalid value for minprop, please give a value between 0 and 0.5.
```

PR #2416, opened 2018-08-17
add case weights and repair minprop in ranger

PhilippPro:
add hyperparameter case.weights (Issue #2337) and set proper limits for minprop (Issue #2328).

### DIFF
--- a/R/RLearner_classif_ranger.R
+++ b/R/RLearner_classif_ranger.R
@@ -11,6 +11,7 @@ makeRLearner.classif.ranger = function() {
       makeIntegerLearnerParam(id = "min.node.size", lower = 1L),
       makeLogicalLearnerParam(id = "replace", default = TRUE),
       makeNumericLearnerParam(id = "sample.fraction", lower = 0L, upper = 1L),
+      makeNumericVectorLearnerParam(id = "case.weights"),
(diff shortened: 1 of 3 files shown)


larskotthoff:
Thanks, merging.

PhilippPro:
There is the problem that case.weights is also used for weighting. Hence, this gives an error:

```
devtools::install_github("mlr-org/mlr", dependencies = FALSE)
library(mlr)
data(iris)
task = makeClassifTask(id = "tutorial", data = iris, target = "Species")
lrn = makeLearner("classif.ranger", par.vals = list(case.weights = c(rep(1,5), rep(0,145))))
train(lrn, task)
```
Is there a proper solution? Users of mlr do not necessarily now the option of using weights in makeClassifTask, at least I was not aware of it and I would rather use the par.vals argument in makeLearner.

berndbischl:
the point is that we have a standard here. caseweights are not exported for any other learner. 
they are mapped via the generic mlr weights argument

@larskotthoff it is particurlarly important that you read this for reviews

berndbischl:
i will revert this now.

PhilippPro:
Maybe you can add the weights to the surv learner if you do that. ;)

larskotthoff:
That standard isn't documented anywhere, is it?

PR #2418, opened 2018-08-20
Ranger case weights

PhilippPro:
Solution to the case.weights problem. Set only weights if hyperparameter was not set.


larskotthoff:
Thanks. Looks like randomForestSRC was updated recently and now tests are failing. Will wait before that passes before merging.

PR #2421, opened 2018-08-23
add some stuff for ranger survival learner

PhilippPro:
add some stuff for the survival learner, that is available for regr and classif


larskotthoff:
Tests are failing.

PR #2425, opened 2018-08-25
fixed ibrier measure

IyarLin:
Reference is a kaplan-meier curve model used to benchmark model performance against (since it uses no covariates). Matrix is the model performance measure. By adding "reference = F" in the pec call, I've removed the reference calculation thus speeding the code and fixing the problem of using reference instead of the fitted model error term.

### DIFF
--- a/R/measures.R
+++ b/R/measures.R
@@ -1429,8 +1429,10 @@ ibrier = makeMeasure(id = "ibrier", minimize = TRUE, best = 0, worst = 1,
 
     probs = predictSurvProb(getLearnerModel(model, more.unwrap = TRUE), newdata = newdata, times = grid)
     perror = pec(probs, f, data = newdata[, tn], times = grid, exact = FALSE, exactness = 99L,
-      maxtime = max.time, verbose = FALSE, reference = F)
+      maxtime = max.time, verbose = FALSE, reference = FALSE)
+    
 
+    # FIXME: this might be the wrong number!
     crps(perror, times = max.time)[1L, ]
   },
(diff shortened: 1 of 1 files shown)


R/measures.R
```diff
@@ -1429,10 +1429,8 @@ ibrier = makeMeasure(id = "ibrier", minimize = TRUE, best = 0, worst = 1,
 
     probs = predictSurvProb(getLearnerModel(model, more.unwrap = TRUE), newdata = newdata, times = grid)
     perror = pec(probs, f, data = newdata[, tn], times = grid, exact = FALSE, exactness = 99L,
-      maxtime = max.time, verbose = FALSE)
+      maxtime = max.time, verbose = FALSE, reference = F)
```
larskotthoff: Please use `FALSE` instead of `F`.

R/measures.R
```diff
@@ -1429,10 +1429,8 @@ ibrier = makeMeasure(id = "ibrier", minimize = TRUE, best = 0, worst = 1,
...
 
-    # FIXME: what is the difference between reference and matrix?
-    # FIXME: this might be the wrong number!
```
larskotthoff: What about the FIXMEs you deleted? I guess that first one is irrelevant now, but what about the second?

IyarLin: Sorry, I thought the second FIXME refers to a possible wrong number caused by the first FIXME. If that's not the case than I'm not sure what is meaning of the second FIXME.  I'll put it back

larskotthoff:
Thanks, merging.

PR #2428, opened 2018-09-06
liquidSVM

PhilippPro:
This is the implementation of liquidSVM as mlr-learner. The probability predictions are still technically flawed and they are not (yet) responding to my emails, so unfortunately probability predictions are not provided at the moment.


larskotthoff:
Could you add some tests please?

PhilippPro:
Why are there some errors on the pr and not on the push? I also do not understand the errors very well...

larskotthoff:
Ok, it looks like the tests are failing with a segmentation fault. I've tried on two different machines and can't reproduce it. Does it work when you run the tests locally?

PhilippPro:
Yes, they were ok. The algorithm only produces warnings in the default version which is not good. Also the probability prediction is not working properly yet. But the package has quite potential regarding performance and runtime and I already contacted the creators.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
+#' @export
+makeRLearner.classif.liquidSVM = function() {
+  makeRLearnerClassif(
+    cl = "classif.liquidSVM",
+    package = "liquidSVM",
+    par.set = makeParamSet(
+      makeIntegerLearnerParam(id = "d", lower = 0L),
+      makeLogicalLearnerParam(id = "scale"),
```
larskotthoff: Default for this is TRUE.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
+#' @export
+makeRLearner.classif.liquidSVM = function() {
+  makeRLearnerClassif(
+    cl = "classif.liquidSVM",
+    package = "liquidSVM",
+    par.set = makeParamSet(
+      makeIntegerLearnerParam(id = "d", lower = 0L),
```
larskotthoff: This should not be tunable (`tunable = FALSE`).

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeLogicalLearnerParam(id = "scale"),
+      makeIntegerLearnerParam(id = "threads", lower = -1L),
```
larskotthoff: Default is 0.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "partition_choice", lower = 0L),
```
larskotthoff: Default is 0.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "grid_choice", lower = -2L, upper = 2L),
```
larskotthoff: How does this interact with `gamma_steps`, `lambda_steps`, etc? Are these options mutually exclusive?

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "adaptivity_control", lower = 0L, upper = 2L),
```
larskotthoff: Default is 0.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "random_seed", default = 1),
```
larskotthoff: I didn't find any documentation on what the default random seed is -- where did you find this? It also shouldn't be tunable.

R/RLearner_classif_liquidSVM.R
larskotthoff: And all of these options should be not tunable.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "fold"),
```
larskotthoff: Should that be `folds`? And there should be a minimum for this (and again not tunable).

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "clipping"),
```
larskotthoff: Seems that the minimum is -1. Also probably shouldn't be tunable.

R/RLearner_classif_liquidSVM.R
```diff
@@ -0,0 +1,49 @@
...
+      makeIntegerLearnerParam(id = "gamma_steps", lower = 0),
+      makeNumericLearnerParam(id = "min_gamma", lower = 0),
+      makeNumericLearnerParam(id = "max_gamma", lower = 0),
+      makeNumericVectorLearnerParam(id = "gammas", lower = 0),
+      makeIntegerLearnerParam(id = "lambda_steps", lower = 0),
+      makeNumericLearnerParam(id = "min_lambda", lower = 0),
+      makeNumericLearnerParam(id = "max_lambda", lower = 0),
+      makeNumericVectorLearnerParam(id = "lambdas", lower = 0),
+      makeNumericVectorLearnerParam(id = "c_values", lower = 0),
+      makeLogicalLearnerParam(id = "useCells")
```
larskotthoff: Seems that the default for this is `FALSE`. Should be not tunable.

R/RLearner_regr_liquidSVM.R
```diff
@@ -0,0 +1,43 @@
+#' @export
+makeRLearner.regr.liquidSVM = function() {
+  makeRLearnerRegr(
+    cl = "regr.liquidSVM",
+    package = "liquidSVM",
+    par.set = makeParamSet(
```
larskotthoff: See comments for `classif` param set.

PhilippPro:
I tried to change all parameters according to your suggestions. I am unsure about clipping, if it is tunable or not, that's why I did not set `tunable = FALSE`.

larskotthoff:
Thanks, looks good -- what about restrictions between parameters that control the tuning though? If you set all the gamma parameters at the same time, I guess that some of them override others? It may make sense to add requirements saying that if you set a certain parameter, others cannot be set to avoid surprising results when users accidentally set conflicting parameters.

PhilippPro:
> Thanks, looks good -- what about restrictions between parameters that control the tuning though? If you set all the gamma parameters at the same time, I guess that some of them override others? It may make sense to add requirements saying that if you set a certain parameter, others cannot be set to avoid surprising results when users accidentally set conflicting parameters.

I changed the parameter settings to take into account requirements. Moreover I looked into https://github.com/cran/liquidSVM/blob/master/R/mlr.R and changed the parameter settings slightly.

larskotthoff:
Thanks for all your work on this. Merging.

PR #2423, opened 2018-08-24
release branch 2.13

berndbischl:
- lqa and elmNN where copied to extralearners, needs to be checked and doced in NEWS

- ranger case weights where handled wrong, reverted. we could update surv ranger.

- travis deploy weird

PR #2429, opened 2018-09-13
Update master to latest CRAN version, travis and tic tweaks

pat-s:
* Merging "release_branch_2_13" #2423 

* Update `.travis.yml` and `tic.R` (using latest `tic`)

* Tidy Description

* To not run into build problems on CRAN during R CMD build, we had to move all the vignettes into a subfolder of `vignettes/` again because R CMD build ignores subfolders of `vignettes/`.


larskotthoff:
Thanks, merging.

PR #2430, opened 2018-09-13
Survival probabilities and ibrier

PhilippPro:
This PR is meant to implement probability predictions and the correct calculation of the ibrier measure based on them.


R/Prediction.R
```diff
@@ -138,7 +138,16 @@ makePrediction.SurvTaskDesc = function(task.desc, row.names, id, truth, predict.
   # FIXME: recode times
   data$truth.time = truth[, 1L]
   data$truth.event = truth[, 2L]
-  data$response = y
+  if (predict.type == "response") {
+    data$response = y
+  } else {
```
larskotthoff: Can you add an explicit check for whether the predict type is probs here please?

R/Prediction.R
```diff
@@ -138,7 +138,16 @@ makePrediction.SurvTaskDesc = function(task.desc, row.names, id, truth, predict.
...
+    data$response = y$preds
+    # Extend probs to the truth time points
+    all.times = sort(unique(c(truth[, 1L], y$train.times)))
+    prob_columns = sapply(all.times, function(t) max(which(y$train.times <= t)))
```
larskotthoff: Please replace `sapply` by the appropriate *apply that doesn't do automatic conversions (lapply?).

R/RLearner_surv_coxph.R
```diff
@@ -26,14 +26,23 @@ makeRLearner.surv.coxph = function() {
...
+    model
   }
 }
 
 #' @export
 predictLearner.surv.coxph = function(.learner, .model, .newdata, ...) {
-  predict(.model$learner.model, newdata = .newdata, type = "lp", ...)
+  if (.learner$predict.type == "response") {
+    predict(.model$learner.model, newdata = .newdata, type = "lp", ...)
+  } else {
```
larskotthoff: See above -- please check explicitly for predict type.

R/measures.R
```diff
@@ -1415,27 +1415,25 @@ iauc.uno = makeMeasure(id = "iauc.uno", minimize = FALSE, best = 1, worst = 0,
...
-    # FIXME: this might be the wrong number!
-    crps(perror, times = max.time)[1L, ]
-  },
-  extra.args = list(max.time = NULL, resolution = 1000L)
+    truth = getPredictionTruth(pred)
+    target = getPredictionTaskDesc(pred)$target
+    newdata = cbind(truth[,1], pred$data$truth.event, feats)
+    colnames(newdata)[1:2] = target
+    probs = getPredictionProbabilities(pred)
+    colnames(probs) = times.train = as.numeric(substr(colnames(probs), 11, 100))
```
larskotthoff: What does this do? It looks like this assumes the column names to be something particular.

R/measures.R
```diff
@@ -1415,27 +1415,25 @@ iauc.uno = makeMeasure(id = "iauc.uno", minimize = FALSE, best = 1, worst = 0,
...
+    times.test = truth[truth[,2] == 1,1]
+    grid = c(sort(unique(c(times.test))))
+    prob_columns = sapply(grid, function(t) max(which(t >= times.train - 10^(-13))))
```
larskotthoff: What's going on here? How does the training time relate to what column names to select?

R/Prediction.R
mllg: +1
You might want to use `viapply()`?

R/Prediction.R
```diff
@@ -138,7 +138,16 @@ makePrediction.SurvTaskDesc = function(task.desc, row.names, id, truth, predict.
...
+    data$prob = y$probs[, prob_columns]
```
mllg: `drop = FALSE`

R/Prediction_operators.R
```diff
@@ -58,6 +60,8 @@ getPredictionProbabilities = function(pred, cl) {
   } else if (ttype == "cluster") {
     y = pred$data[, stri_detect_regex(cns, "prob\\.")]
     colnames(y) = seq_col(y)
+  } else if (ttype == "surv") {
+    y = pred$data[, stri_detect_regex(cns, "prob\\.")]
```
mllg: `stri_startswith_fixed()`? (although two lines above there is the same "mistake")

R/RLearner_surv_coxph.R
```diff
@@ -26,14 +26,23 @@ makeRLearner.surv.coxph = function() {
...
+    preds = predict(.model$learner.model, newdata = .newdata, type = "lp", ...)
+    train.times = c(0, sort(unique(.model$learner.model$times)))
```
mllg: This allows to have a vector with the first two elements being `0`. Is this intended?

R/helpers.R
```diff
@@ -94,6 +94,11 @@ measureTime = function(expr, ee = parent.frame()) {
   proc.time()[[3L]] - before
 }
 
+attachTrainingTime = function(model, task, data) {
+  model$times = data[, getTaskTargetNames(task)[1L]]
```
mllg: Is this the model as returned by the learner, e.g. a coxph object?

Why do you not use `attachTrainingInfo()`?

R/Prediction.R
PhilippPro: I think this is not necessary, as predict.type can only be response or prob. If you try to makeLearner with other character it will give you an error.

R/Prediction.R
PhilippPro: ok

R/Prediction.R
PhilippPro: ok

R/Prediction_operators.R
PhilippPro: ok

R/helpers.R
PhilippPro: Is this an attribute like is used in `attachTrainingInfo()`? It is a vector of times...

R/measures.R
PhilippPro: The training times are saved in the column names character. Maybe a bit weird, but I did not know where to save them elsewhere...

R/measures.R
PhilippPro: This is specific to the pec interface. We need the probabilities to the specific time points given by `times.test`. The probabilities only change from point to point in `times.train` ...

R/measures.R
larskotthoff: Hmm, not sure if I understand. Where exactly are you saving training times, and what is the `substr()` doing?

R/measures.R
larskotthoff: Can you give an example of what this does please?

larskotthoff: Oh wait, you're talking about the survival times, not the time it took to train/test the model, right? Still not sure what's going on though.

R/measures.R
PhilippPro: The training "survival times" are saved as column names of the probability predictions after "prob.time". So that's why "prob.time." is deleted first and then the rest is converted from character to numeric.

R/measures.R
PhilippPro: Yes, I talk about survival times. As the probabilities do not change from "training timepoint to training timepoint" we jus take the probabilities of the specific training time points. For understanding this you have to know some basics of the survival models. The time dependent part is usually not dependent on the variables (see e.g. $\lambda_0(t)$ in the proportional hazards model https://en.wikipedia.org/wiki/Proportional_hazards_model) that's why it can be handled like that.

R/measures.R
mllg: If you have to save numeric values as strings, you are using the wrong data type(s). Such a conversion is not safe.

PhilippPro: Yes, but for example this works: paste0("x", 1:1000 + 10^-10)

It is a bit unlikely to have times smaller than 10^-13 ;). But see also my question above, where should I save the vector instead?

IyarLin:
Tried running the lines below and ran into a few issues:

```
devtools::install_github("mlr-org/mlr", ref = "survival_probabilities_ibrier")
library(mlr); library(survival); library(randomForestSRC); library(dplyr
lung <- lung %>% 
  filter(complete.cases(.)) %>% 
  mutate(status = status == 2)

surv_task <- makeSurvTask(id = "fix ibrier", data = lung, target = c("time", "status"))
holdout_sample <- makeResampleInstance(
  desc = makeResampleDesc(method = "Holdout", stratify = T, split = 0.8),
  task = surv_task
)

cox_lrn <- makeLearner(
  "surv.coxph", 
  predict.type = "prob"
)

# the following doesn't work:
rfsrc_lrn <- makeLearner(
  "surv.randomForestSRC", 
  predict.type = "prob"
)
```

> Error in setPredictType.Learner(learner = wl, predict.type = predict.type) : 
  Trying to predict probs, but surv.randomForestSRC does not support that!

My understanding is randomForestSRC probability should be supported?

Also:

```
res_cox <- resample(learner = cox_lrn, task = surv_task, resampling = cv3, measures = ibrier)
```

> Resampling: cross-validation
Measures:             ibrier    
Error in names(name.vars) <- names(formula)[-1] : 
  'names' attribute [3] must be the same length as the vector [2]

Same happens with 

```
bench_cox <- benchmark(cox_lrn, surv_task, resamplings = cv3, measures = ibrier)
```

> Resampling: cross-validation
Measures:             ibrier    
Error in names(name.vars) <- names(formula)[-1] : 
  'names' attribute [3] must be the same length as the vector [2]

Finally, running

```
cox_model <- train(learner = cox_lrn, task = train_task)
pred <- predict(object = cox_model, surv_task)
```

yields

> Error in names(name.vars) <- names(formula)[-1] : 
  'names' attribute [3] must be the same length as the vector [2]

When I tried debugging the function `resample` I found the error to be on line 56, though I wasn't able to straight away understand what happened.

PhilippPro:
@IyarLin: randomForestSRC is not supported yet, but will be soon, I have already code for that. 

The rest of the code is working for me. Do you have all the required packages in the newest versions?

PhilippPro:
I added probs for rfsrc. there are other learners that will follow soon.

IyarLin:
I have the latest survival package version (2.42.6) and installed mlr as you suggested via `devtools::install_github("mlr-org/mlr", ref = "survival_probabilities_ibrier")` and the above code produces the errors I mentioned.. Hope it's just something weird on my end.

HerrMo:
@IyarLin Which system are you working on? And are you sure the installation was successful? I had some problems installing the branch using devtools on windows. Maybe you encountered the same problem.

PhilippPro:
> @IyarLin Which system are you working on? And are you sure the installation was successful? I had some problems installing the branch using devtools on windows. Maybe you encountered the same problem.

My windows installation and the code was successfully working without any problems. Same on debian.

HerrMo:
> 
> 
> > @IyarLin Which system are you working on? And are you sure the installation was successful? I had some problems installing the branch using devtools on windows. Maybe you encountered the same problem.
> 
> My windows installation and the code was successfully working without any problems. Same on debian.

I see. It also depends on the R version (I use R 3.5.1) and Rtools. I am absolutly no expert, but as far as I understand it is a problem of devtools (they didn't register R 3.5 or something. See devtools:::version_info, I don't get 3.5 listed.). Tried it again and seems to be relatet to that issue. 

When I use 
`devtools::install_github("mlr-org/mlr", ref = "survival_probabilities_ibrier")` 
the installation fails. 

Instead
```
library(devtools)
assignInNamespace("version_info", 
                  c(devtools:::version_info, list("3.5" = list(version_min = "3.3.0", version_max = "99.99.99", path = "bin"))),
                  "devtools")
install_github("mlr-org/mlr", ref = "survival_probabilities_ibrier")
```
works for me.
The issue is known quite a while https://github.com/r-lib/devtools/issues/1772#issue-316998168. No idea why they haven't fixed it yet.

PhilippPro:
That can be a reason, I still use 3.4.4

IyarLin:
I found a fix for the error I got above using cox regression: 

> Error in names(name.vars) <- names(formula)[-1] :
'names' attribute [3] must be the same length as the vector [2]

The code:

```
library(devtools)
install_github("https://github.com/mlr-org/mlr/tree/survival_probabilities_ibrier")
library(mlr); library(survival); library(tidyverse); library(pec)
cox_lrn <- makeLearner(cl = "surv.coxph", predict.type = "prob", x = T)
lung = lung %>% mutate(status = as.logical(status == 2)) %>% filter(complete.cases(.))
task_surv = makeSurvTask(data = lung, target = c("time", "status"))
task = task_surv
model = train(cox_lrn, task, subset = 1:100)
predictSurvProb(model$learner.model, lung[101:120, ], times = 1:100) # this fails
model2 = coxph(Surv(time, status) ~ ., data = lung[1:100, ], x = T)
predictSurvProb(model2, lung[101:120, ], times = 1:100) # this works

# Fix the learner model formula
model$learner.model$formula = as.formula(paste0(gsub(", type = \"right\"", "", model$learner.model$formula[2]), 
                                                   model$learner.model$formula[1], 
                                                   model$learner.model$formula[3]), 
                                         env = attr(model$learner.model$formula, ".Environment"))
predictSurvProb(model$learner.model, lung[101:120, ], times = 1:100) # this works!
```

PhilippPro:
> I found a fix for the error I got above using cox regression:
> ...
Ok, thanks. It looks very complicated, I would prefer simpler solutions. Probably it would be the best to get rid of the `predictSurvProb` function as you suggested in your email by using code similar to the following:

```
pred <- matrix(nrow = length(times), ncol = 20)
for(i in 1:length(times)){
  pred[i, ] = exp(-predict(model$learner.model, newdata =
  task$env$data[101:120, ] %>% mutate(time = times[i]), type = "expected"))
}
```
When I have time, I will try to work on it.;)

pat-s:
@PhilippPro Do you have time in the near future to continue here?

PhilippPro:
It currently works in the branch, but I would need some help and feedback to finally implement it in mlr. I think there is really a demand for benchmarking survival and nobody knows how to exactly do it.

pat-s:
Happy to help if you tell me how :+1: 

Maybe you can finish the review of @larskotthoff first?

PhilippPro:
I think one of the main issues is, that I touched [ResamplePrediction.R](https://github.com/mlr-org/mlr/edit/survival_probabilities_ibrier/R/ResamplePrediction.R?pr=%2Fmlr-org%2Fmlr%2Fpull%2F2430). I think most of the issues of Lars are resolved.

R/measures.R
```diff
@@ -1415,27 +1415,27 @@ iauc.uno = makeMeasure(id = "iauc.uno", minimize = FALSE, best = 1, worst = 0,
...
-  },
-  extra.args = list(max.time = NULL, resolution = 1000L)
+    truth = getPredictionTruth(pred)
+    target = getPredictionTaskDesc(pred)$target
+    newdata = cbind(truth[,1], pred$data$truth.event, feats)
+    colnames(newdata)[1:2] = target
+    probs = getPredictionProbabilities(pred)
+    if (anyMissing(probs))
+      return(NA_real_)
+    colnames(probs) = times.train = as.numeric(substr(colnames(probs), 11, 100))
```
pat-s: Same as above: Is there a need to save the training times at this point? Why not use `time.train` as a normal measure?

E: Ah ok, its about survial times. I have no knowledge in this.

tests/testthat/test_base_performance.R
```diff
@@ -41,7 +41,7 @@ test_that("performance", {
   # custom measure
   mymeasure = makeCustomResampledMeasure(measure.id = "mym", aggregation.id = "train.mean",
     properties = c("classif", "predtype.response"),
-    fun = function(task, group, pred, feats, extra.args) {
+    fun = function(task, model, pred, feats, extra.args) {
```
pat-s: What does this change do? Looks like it will affect other tests as well?

pat-s:
@larskotthoff Maybe you have time to help and resolve your review together with @PhilippPro.

R/ResamplePrediction.R
```diff
@@ -22,8 +22,8 @@ makeResamplePrediction = function(instance, preds.test, preds.train, task.desc)
   if (any(trnull)) pr.tr = preds.train[!trnull] else pr.tr = preds.train
 
   data = setDF(rbind(
-    rbindlist(lapply(seq_along(pr.te), function(X) cbind(pr.te[[X]]$data, iter = X, set = "test"))),
-    rbindlist(lapply(seq_along(pr.tr), function(X) cbind(pr.tr[[X]]$data, iter = X, set = "train")))
+    rbindlist(lapply(seq_along(pr.te), function(X) cbind(pr.te[[X]]$data, iter = X, set = "test")), fill = TRUE),
```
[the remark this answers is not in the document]
IyarLin: you may want to consider using dplyr::bind_rows instead of rbindlist. It's suppose to replace it if I'm not mistaken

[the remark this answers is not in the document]
pat-s: We do not use _dplyr_ in _mlr_ but rather _data.table_. Both should do the same.

R/measures.R
PhilippPro: This is necessary, to get the information about the training times to the model (we predict for each training and prediction time that exist).

R/helpers.R
```diff
@@ -94,6 +94,11 @@ measureTime = function(expr, ee = parent.frame()) {
   proc.time()[[3L]] - before
 }
 
+attachTrainingTime = function(model, task, data) {
```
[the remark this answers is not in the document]
PhilippPro: This is necessary, to get the information about the training times to the model (we predict for each training and prediction time that exist).

tests/testthat/test_base_performance.R
PhilippPro: I think this was just wrong in the former version?

R/ResamplePrediction.R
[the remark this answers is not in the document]
PhilippPro: This is one of the main problems. @HerrMo, do you remember why this was necessary?

larskotthoff:
I think my only concern was about saving the times -- if there's no other way of doing that, then we can take this version.

tests/testthat/test_base_performance.R
pat-s: Maybe its not used in the test and was not catched therefore. If you're sure that its good now than its fine :)

R/ResamplePrediction.R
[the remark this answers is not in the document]
pat-s: Taking the `tests/testthat/test_surv_kaplanmeier.R` test, would I arrive at this point when debugging?

pat-s:
@PhilippPro 	can you resolve the merge conflict? You know best if this change is really needed. Once resolved I'll take a look how we can save the survival times best and if the `rbindlist` change is really needed here (and what the `fill = TRUE`) does here.

In addition, could you summarize in the first post what this PR does? I.e. what features its add in detail, which learners and so on.
This is quite a big one so we need to be clear in the NEWS.md as well later on.

R/ResamplePrediction.R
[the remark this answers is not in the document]
HerrMo: In short, this was implemented to be able to merge the results of successful and unsuccessful CV iterations (i.e., failure models and successful model fits). 
Some details:
Since in `rbindlist(lapply(seq_along(pr.te), function(X) cbind(pr.te[[X]]$data, iter = X, set = "test")))` the data.frames of the single iterations get merged to one single data.frame/data.table rowwise, the data.frames need to have the same number of columns. The problem with survival models is that unsuccessful iterations yield less columns than successful ones. Setting `fill = TRUE` solves this issue, but I'm afraid it's more of a workaround than a satisfactory solution.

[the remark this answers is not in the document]
pat-s: I have the fear that changing this globally could possibly trigger some unwanted side effects even if all tests will pass.

Would it be possible to add the missing columns to survival models manually? This way we would avoid the global change here and encapsulate possible consequences.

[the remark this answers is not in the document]
HerrMo: Probably, but I have no idea how that works and if it's easy. 
I don't know exactly what the cause is, but I think the problem of the different number of columns shouldn't occur in the first place.

pat-s:
- [ ] solve merge conflicts

- [ ] take a look at https://github.com/mlr-org/mlr/pull/2430/files/bdbaa1c017196fec91636e57f0aee723b1deb244#diff-7b657177b2d9da9115e9e031cbe14a52

- [ ] take a look at https://github.com/mlr-org/mlr/pull/2430/files/d2b68f0b9eef7635957d574a769c9b5425a1f4d3#diff-6f0d29667c7c74e4a87ab4c203225f28

@HerrMo Would you have time to do 1)? Afterwards I would take a look at the other two.

PhilippPro:
Merged. The fill = TRUE problem is not treated in the merge and has to be filled in again possibly.

pat-s:
Thanks. I made some minor fixes. I get the following error in `test_surv_measures.R` that needs to be fixed by you guys (please correct me if I'm wrong):


```r
Error in coxModelFrame.coxph(object) : invalid object 
set x=TRUE in the call to coxph 
```

pat-s:
ping @PhilippPro 
Do you think we can continue here or should I close?

pat-s:
I agree that doing it from scratch in mlr3 would be best. I guess the effort needed to get this working does not relate right now. Also you have no time right now for this. Thanks for getting back quickly!

mllg:
If anyone finds this issue and is interested in this, work is continued in https://github.com/mlr-org/mlr3proba.

PR #2440, opened 2018-09-27
Add inline ToC to vignettes

pat-s:
Because the navbar ToC is really messy for long vignettes.

See the netlify preview for a preview (once it has been deployed).

Also template `content-vignette.html` now needs to be `content-article.html`.


larskotthoff:
Thanks, merging.

Issue #2321, opened 2018-06-22
Update feature selection page


pat-s:
_From @schiffner on September 2, 2016 10:12_

We now have `getFeatureImportance` and thus better support for embedded featsel methods.

_Copied from original issue: mlr-org/mlr-tutorial#61_

pat-s:
_From @schiffner on March 30, 2017 13:29_

+ potential changes/improvements in filters

PR #2445, opened 2018-10-05
Tutorial: Update feature selection page

pat-s:
Please use the netlify preview for reviewing.

Tackles #2321 	

- Set toc depth of 3 globally (only works when the next pkgdown version is released to CRAN) https://github.com/r-lib/pkgdown/issues/732
- replace the example "how to fuse a filter with a learner" with an example that includes tuning. Yes, tuning is something different but it is important to see for a user that this is the major advantage here: Having filter and hyperpars both in the param set which enables the tuning of both on the same level.
  - Subsequently, replace `regr.lm` with a learner that can be tuned (`regr.ksvm`)
- remove ggVIS function 
- remove example of multicrit evaluation: Imho it adds no real value to this page and might even be confusing to the reader as it is placed in a sub-sub-sub section at the end of a long section about "fusing a learner with a filter". At least this was the case for me and I thought "does this need to be here?". Maybe we can add it somewhere else or in a different top-level section. 
- add section about `getFeatureImportance()`
- Explain the differences between "filter" and "feature selection"/"wrapper methods"


larskotthoff:
Thanks. Should merging this wait until the next version of pkgdown is on CRAN then?

pat-s:
I am not sure if `pkgdown::build_site()` will complain. Maybe it passes and we can silently wait for the release.

pat-s:
Looks good, feel free to review.

larskotthoff:
Ok, looks like a whole lot of irrelevant stuff got pulled in (random seeds in tutorial), which makes this impossible to review. Could you please revert the non-relevant changes?

Also it looks like Travis doesn't like us anymore.

larskotthoff:
Ok thanks, merging.

PR #2447, opened 2018-10-06
CI: Fix appveyor, XMeans only on Travis, smarten remotes/devtools

pat-s:
* don't install devtools/remotes in every build
* `testthat` needs to be installed explictily since `step_rcmdcheck()` was replaced by a temporal manual `rcmdcheck::rcmdcheck()` call and therefore *tic* does not install it as a dependency anymore.
* Only install Xmeans on Travis
* Clean appveyor.yml


larskotthoff:
Anybody know what's up with Travis?

pat-s:
Both CI's passed, merging.

PR #2449, opened 2018-10-08
Forecasting fix

vrodriguezf:
I just fixed some unit tests on this branch related to the fcregr* tasks and merged the current master.


pat-s:
Looking at the conflicts this PR can be safely closed.

@vrodriguezf 	
Sorry for not reaction on this back then. Do you still know what the purpose was of this PR? I.e. why there were fixed needed for some tests?

vrodriguezf:
I remember I had some troubles when executing `devtools::test()` with some unit tests related to forecasting tasks...they were small fixes, but without them, I could not pass the rest of the tests

pat-s:
Ok, thanks for the quick reply. Right now all tests are fine according to Travis so I do not think that we need to take any action here right now.

PR #2452, opened 2018-10-15
Update Merging surv-prop-FailureModels

HerrMo:
If survival models with predict.type = "prob" lead to an error in a benchmark iteration, merging them with successful iteration raised an error. prob-data.frames differed in dim().


larskotthoff:
Thanks. Could you also add a test that checks for this please?

PhilippPro:
I merged it for now into the branch to make our things work. In general this solution might not be valid, but at the moment I do not know how to make it otherwise.

Issue #1995, opened 2017-08-25
Tuning with filters recalculates filter values each iteration


Prometheus77:
Not a bug, but a recommendation to speed up execution. Currently, when using a filter to tune (e.g. tuning on fw.perc), it recalculates the filter values at each tuning iteration, which in the case of the default filter method (randomForestRFSRC) can be quite time consuming. Would be more efficient to have the tuner check whether there is a filter wrapper in the learner, and if so, to calculate the filter values at the beginning of the tuning process and then simply pass them forward into the filter at each iteration.

A prerequisite for this would be the ability to pass pre-defined filter values into a filter, which I do not believe is currently possible.

mazgit82:
I wonder if it would be possible to define another kind of hyperparameter that must be used at predict time. For instance, this could be useful for methods such as gbm, for which you can do something like preds = predict(model, n.trees = 100) even if the model contains 1000 trees. This way, you do not have to build a 100-tree forest and another 1000-tree forest.

I believe this could also be useful for the filters you mention.

mb706:
(Commenting on the issue as named in the title)

The recalculation of filter values has annoyed me at times as well, and I think it should be accelerated. It would probably be possible to cache the results of `generateFilterValuesData`, globally or with a cache parameter given to that function (and managed by the FilterWrapper / the Filter CPO).

The problem with caching is that some feature filters are nondeterministic and should not be cached.

There are two R libraries that perform caching that I'm aware of: **R.cache** and **memoise**. The latter seems to be more lightweight since `R.cache` apparently doesn't support in-memory caching (?). A potential problem with all of these is that they hash the whole Task data--would probably be fast *relative* to the filter value creation (most filter value functions need to sweep the whole task as well), but might be slow in absolute terms.

An idea that might be useful in other places as well is to have the `Task` store its own hash vale. Since `Task`s already contain an `env`, a `getTaskHash()` function might calculate the hash and store it in this environment to avoid needing to re-calculate it.

pat-s:
I'll tackle this issue in the near future. Together with the `FSelectorRcpp` implementation we should get a significant speed up!

mb706:
Maybe [my implementation of cached preprocessing](https://github.com/mlr-org/mlrCPO/blob/45d54ab096f74ea4378c07fe515f5807c6eb113f/R/CPO_meta.R#L358-L381) can give you some ideas.

pat-s:
[...]

PR #2456, opened 2018-10-19
Filter: Add ensemble filter methods

pat-s:
# Purpose

In the filtering/feature selection field, the idea of ensemble filter methods becomes more and more popular. Ensemble filters aggregate the rankings of multiple single filters and create a new ranking. 

This approach has shown to be superior to single filter methods, e.g. https://ieeexplore.ieee.org/document/8250495.

# Implementation

I decided to establish a new class "FilterEnsemble" which distinguishes the ensemble filter normal the single filters. This decision has (as always) positive and negative side-effects.

Ensemble filters are created in the same way as normal filters in their own file, `R/FilterEnsemble.R`.
They share the same class structure with some minor differences:

- no `pkg`, `supported.tasks`, `supported.features` arguments (all checked by the simple filters)
- new argument `basal.methods` which stands for "single filter methods"

Calculation is done as usual via `generateFilterValuesData()` or `filterFeatures()`:
1. First, a `FilterValues` object is created as usual by calling `generateFilterValuesData()` with the single filters. 
1. Then, the specific ensemble filter calculations are done on the `FilterValues` object (e.g. taking the mean across all voters for each feature).

## Notation

Notation differs a bit among the functions. 
In `generateFilterValuesData()`, an ensemble method is passed in a list together with its required simple methods, .e.g.:

```r
generateFilterValuesData(iris.task, 
  method = list("E-min", c('gain.ratio','information.gain')))
```

To make `makeFilterWrapper()` flexible in the sense that the single methods, which an ensemble method uses, should be tunable, a new argument `base.methods` was introduced. It depends on a ensemble method set either in `filterFeatures(method = "")` or in `makefilterWrapper(fw.method = "")`.

```r
makeFilterWrapper(lrn, fw.method = "E-min", 
  fw.base.methods = c("gain.ratio", "information.gain"),
```

```r
filterFeatures(iris.task, method = "E-min", 
  base.methods = c("gain.ratio", "information.gain"), abs = 2)
```

This gives the user the option to tune

- over multiple filters set in `fw.method`
- over multiple single filters if an ensemble filter is within `fw.method`

Tuning simple filters is not supported due to the lack of sampling without replacement for `DiscreteVectorParams` in ParamHelpers https://github.com/berndbischl/ParamHelpers/pull/206

As multiple rankings are calculated and returned when using an ensemble filter, `filterFeatures()` will always prioritize the ensemble method unless a different method is set via the new `select.method` argument.
This only applies if `filterFeatures()` is called directly as in the wrapper only one filter method is for subsetting anyway (and in the ensemble case, the prioritizing of the ensemble method applies).

# Other changes

- `getFilterValuesData()` now returns a tbl instead of a data.frame (I think there is no reason not to use enhanced data.frame output. I does not harm any internal processes.)
- `plotFilterValues()` got a bit "smarter" and easier now regarding the ordering of multiple facets
- I added multiple examples to the help pages of `filterFeatures()`, `generateFilterValuesData()` and `makeFilterWrapper()`
- Instead of a wide data.frame the values are now returned in a long (tidy) data.frame. This makes it easier to apply post-processing methods (like `group_by()` calls etc)

# To-do

- [x] tests

- [x] Cache filterValues in a tuning process (and don't recalculate them all the time) #1995 

# Examples

``` r
library(mlr)
#> Warning: replacing previous import 'stats::filter' by 'dplyr::filter' when
#> loading 'mlr'
fval = generateFilterValuesData(iris.task, 
  method = list("E-mean", c("gain.ratio", "information.gain")))
fval
#> FilterValues:
#> Task: iris-example
#> # A tibble: 12 x 4
#>    name         type    method           value
#>    <chr>        <chr>   <chr>            <dbl>
#>  1 Petal.Width  numeric E-mean           4    
#>  2 Petal.Length numeric E-mean           3    
#>  3 Sepal.Length numeric E-mean           2    
#>  4 Sepal.Width  numeric E-mean           1    
#>  5 Petal.Width  numeric gain.ratio       0.871
#>  6 Petal.Length numeric gain.ratio       0.858
#>  7 Sepal.Length numeric gain.ratio       0.420
#>  8 Sepal.Width  numeric gain.ratio       0.247
#>  9 Petal.Width  numeric information.gain 0.955
#> 10 Petal.Length numeric information.gain 0.940
#> 11 Sepal.Length numeric information.gain 0.452
#> 12 Sepal.Width  numeric information.gain 0.267

filterFeatures(iris.task, method = "E-min", 
  base.methods = c("gain.ratio", "information.gain"), abs = 2)
#> Supervised task: iris-example
#> Type: classif
#> Target: Species
#> Observations: 150
#> Features:
#>    numerics     factors     ordered functionals 
#>           2           0           0           0 
#> Missings: FALSE
#> Has weights: FALSE
#> Has blocking: FALSE
#> Has coordinates: FALSE
#> Classes: 3
#>     setosa versicolor  virginica 
#>         50         50         50 
#> Positive class: NA


### makeFilterWrapper(), can ofc also be used within tuneParams()
task = makeClassifTask(data = iris, target = "Species")
lrn = makeLearner("classif.lda")
inner = makeResampleDesc("Holdout")
outer = makeResampleDesc("CV", iters = 2)

# usage of an ensemble filter
lrn = makeFilterWrapper(makeLearner("classif.lda"), fw.method = "E-Borda",
  fw.base.methods = c("gain.ratio", "information.gain"),
  fw.perc = 0.5)
r = resample(lrn, task, outer, extract = function(model) {
  getFilteredFeatures(model)
})
#> Measures:             mmce
#> [Resample] iter 1:    0.0533333
#> [Resample] iter 2:    0.0533333
#> 
#> Aggregated Result: mmce.test.mean=0.0533333
#> 
print(r$extract)
#> [[1]]
#> [1] "Petal.Length" "Petal.Width" 
#> 
#> [[2]]
#> [1] "Petal.Length" "Petal.Width"

plotFilterValues(fval)
```

<sup>Created on 2018-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


R/generateFilterValues.R
```diff
@@ -4,12 +4,18 @@
...
+#' @importFrom dplyr group_by arrange mutate summarise rename bind_rows mutate_if
+#' @importFrom tibble as_tibble tibble
 #' @template arg_task
 #' @param method ([character])\cr
 #'   Filter method(s), see above.
 #'   Default is \dQuote{randomForestSRC.rfsrc}.
 #' @param nselect (`integer(1)`)\cr
 #'   Number of scores to request. Scores are getting calculated for all features per default.
+#' @param ensemble.method ([character])\cr
+#'   Ensemble filter method to use. Can only be used with >= 2 filter methods.
```
larskotthoff: Should be consistent with wrapper -- character or logical?

R/generateFilterValues.R
```diff
@@ -77,21 +85,88 @@ generateFilterValuesData = function(task, method = "randomForestSRC.rfsrc", nsel
...
+      map(~ mutate(.x, type = vcapply(getTaskData(task, target.extra = TRUE)$data[fn], getClass1)))
+
+    # rank and merge all information together into one data.frame
+    fval_all_ranked = map(fval_all, ~ mutate(.x, rank = seq(1:length(.x$value)))) %>%
+      imap(~ mutate(.x, method = names(fval_all[.y]))) %>%
+      map_dfr(~ as.data.frame(.x))
+
+    ### ensemble rank aggregation
+
+    if (any(c("E-min", "E-mean", "E-median", "E-max", "E-Borda") %in% ensemble.method)) {
```
larskotthoff: Possible values for ensemble method should be documented.

tests/testthat/test_base_generateFilterValuesData.R
```diff
@@ -143,3 +143,9 @@ test_that("filter method 'variance' works with missing values", {
   fi = generateFilterValuesData(regr.na.num.task, method = "variance")
   expect_false(anyMissing(fi$data$variance))
 })
+
+test_that("ensemble methods work", {
+  fi = generateFilterValuesData(multiclass.task, method = c('gain.ratio','information.gain'),
+                                ensemble.method = c("E-Borda", "E-min"))
```
larskotthoff: What does it mean if multiple ensemble methods are specified?

mb706:
Instead of putting lots of special code into the generateFilterValuesData function the ensemble stuff should probably happen somewhere else. My suggestion is that the filter code should be changed to also accept functions or `Filter` objects (i.e. the objects found in `mlr:::.FilterRegister`). Ensembles (and other interesting things) could then be implemented using functionals that create new filters from existing ones
```r
filterFeatures(pid.task, "univariate.model.score", abs = 3,
  perf.learner = "classif.logreg")
# (there should probably be a better way to access this than by ":::")
filterFeatures(pid.task, mlr:::.FilterRegister$univariate.model.score, abs = 3,
  perf.learner = "classif.logreg")
filterFeatures(pid.task,
  makeFilterEnsemble("median", c("univariate.model.score", "variance", "anova.test")),
  abs = 3,
  univariate.model.score.perf.learn = "classif.logreg")
# alternative:
filterFeatures(pid.task,
  makeFilterEnsemble("median", c("univariate.model.score", "variance", "anova.test")),
    filter.args = list(univariate.model.score = list(perf.learn = "classif.logreg")),
  abs = 3)
```
In these examples, the `makeFilterEnsemble` method would return a `Filter` object (mostly a function with some metadata about allowed task types) that does the ensemble things internally; "generateFilterValuesData" should not be involved in this and call the metafilter just the same way it calls an ordinary filter.

pat-s:
Hi guys, 

This is all WIP here. The main idea is to use them in `makeFilterWrapper()`. I'll come back to your comments later.

R/FilterWrapper.R
```diff
@@ -26,12 +26,18 @@
 #'   Mutually exclusive with arguments `fw.perc` and `fw.abs`.
 #' @param fw.mandatory.feat ([character])\cr
 #'   Mandatory features which are always included regardless of their scores
+#' @param ensemble.method ([character])\cr
+#'   Which ensemble method should be used. Can only be used with >= 2 filter methods.
```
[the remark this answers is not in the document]
pat-s: No, you can use multiple methods.

> Also why is this a character? Comments and code below suggest that this is a logical value.

I'll check again. But as said, you can use multiple ones.

R/generateFilterValues.R
pat-s: Nothing is finished yet :)

R/generateFilterValues.R
pat-s: Yep, I will ofc do this.

tests/testthat/test_base_generateFilterValuesData.R
pat-s: The same as if you would use multiple single ones. You get back a DF with all listed rankings when you use `generateFilterValuesData()`.

pat-s:
@mb706 	Thanks for your input.

Sounds like a good idea. I prefer the following notation:

```r
filterFeatures(pid.task,
  makeFilterEnsemble("median", c("univariate.model.score", "variance", "anova.test")),
  abs = 3,
  univariate.model.score.perf.learn = "classif.logreg")
```

pat-s:
- new class `FilterEnsemble`
- new `listFilterEnsembleMethods()` etc
- wrapper working
- `filterFeatures()` and `generateFilterValuesData()` working

Doc, tests and more concrete examples in the next days.

pat-s:
@larskotthoff 	@mb706 	
Looking forward to your comments now - see first post.

pat-s:
@larskotthoff 	@mb706 	looking forward to your comments.

larskotthoff:
Looks like builds are failing...

larskotthoff:
I'm not sure about `basal.methods` -- I find this name really unintuitive (although I see why you chose it). Why not `base.methods` instead? This would also be consistent with e.g. `makeStackedLearner()`.

pat-s:
> I'm not sure about basal.methods -- I find this name really unintuitive (although I see why you chose it). Why not base.methods instead? This would also be consistent with e.g. makeStackedLearner().

I also do not like the name that much but that how it is called in the literature: https://ieeexplore.ieee.org/document/8250495

`base.methods` sounds a bit better yes and is consistent with our naming scheme. 
I guess the term `basal.methods` is not so widely known that we could not deviate from it. I'll change it.

pat-s:
@larskotthoff 	If you approve I would like to get a second review on this.

@mllg 	Would you have time to take a look?

R/FilterEnsemble.R
```diff
@@ -0,0 +1,296 @@
+.FilterEnsembleRegister = new.env() # nolint
```
larskotthoff: Why do we need a registry for filter ensembles?

R/FilterEnsemble.R
```diff
@@ -0,0 +1,296 @@
...
+    fval = generateFilterValuesData(task, method = base.methods,
+      nselect = nselect, more.args = ...)
+
+    # rank base filters by method
+    fval_all_ranked_simple = transform(fval$data,
+      rank = ave(1:nrow(fval$data), method,
+        FUN = function(x) order(fval$data$value[x])))
+
+    fval_all_ranked_simple = fval_all_ranked_simple[with(fval_all_ranked_simple,
+      order(value, rank)), ]
```
larskotthoff: It looks like there's a lot of repeated code in these calls (for example the line above). Any way of moving this into the `makeFilterEnsemble` implementation?

tests/testthat/test_base_benchmark.R
```diff
@@ -271,6 +271,41 @@ test_that("drop option works for BenchmarkResults_operators", {
     wrapper.class = "cl")
 })
 
+test_that("benchmark works with ensemble filters", {
+  lrn = makeLearner("classif.ksvm")
+  lrn = makeFilterWrapper(lrn, fw.method = "E-Borda")
+
+  par.set <- makeParamSet(
```
larskotthoff: style: <-

tests/testthat/test_base_benchmark.R
```diff
@@ -271,6 +271,41 @@ test_that("drop option works for BenchmarkResults_operators", {
...
+  tune_ctrl = makeTuneControlRandom(maxit = 3)
+
+  tune_wrapper_svm = makeTuneWrapper(lrn, resampling = rin, par.set = par.set,
+    control = tune_ctrl, show.info = FALSE,
+    measures = list(acc))
+
+  expect_class(benchmark(learners = tune_wrapper_svm, task = tasks,
+    resampling = rin, measures = list(acc)), "BenchmarkResult"
+  )
+
```
larskotthoff: Could you add a test that runs each of the available methods at least once please?

R/FilterEnsemble.R
pat-s: We cannot move it into `makeFilterEnsemble()` directly since S3 kicks in and causes troubles. But I moved the shared code into two helper funs. Good point.

tests/testthat/test_base_benchmark.R
pat-s: I don't think that this is needed here. The purpose is not to test the simple filter execution but the aggregation when being run in combination with an ensemble filter. 
As long as the dedicated simple filter tests are fine there is no need to test over all here again. For simplicity we could even select only two distinct ones to make sure the `makeTuneWrapper()` implementation is able to deal fine with a FilterWrapper containing ensemble filters.

I did not change anything how simple filters are calculated per se. After they have been, the follow up processing is the same for all of them.

R/FilterEnsemble.R
pat-s: We need to distinguish the ensemble filters from the normals ones to treat them more easily within the package, e.g. here

https://github.com/mlr-org/mlr/blob/e64fa48bb548bab6ec2e9eec611fd5ac5c020a2a/R/filterFeatures.R#L88

I just sticked to the "Register" approach to keep it consistent with the current implementations of simple filters.

pat-s:
The biggest problem right now is that we cannot tune the simple methods because of https://github.com/berndbischl/ParamHelpers/pull/206.

Tl,DR: We need distinct methods in the VectorParam to tune over and it doesn't make sense to have certain simple methods in twice. These would just bias the ensemble filter towards a certain simple filter (basically an unwanted weighting).
Unfortunately the first case (sampling without replacement) is not supported by ParamHelpers for `DiscreteVectorParams`.

So instead of putting simple filters into the ParamSet, one should set it as fixed params in the FilterWrapper only.
I don't have much hope that this case will be ever supported for _mlr_ anymore.

I will just make a note in the tutorial and NEWS that tuning of simple filters is not supported. 

(Another problem in tuning character vector params would be the choice of the optim algorithm. 
MBO cannot handle character vectors probably as a surrogate model would not produce meaningful outcomes.)

R/FilterEnsemble.R
larskotthoff: Hmm, a solution that uses properties like for the learners is probably better, but probably also too much work to implement.

tests/testthat/test_base_benchmark.R
larskotthoff: Ok, I see your point.

R/filterFeatures.R
```diff
@@ -41,14 +46,55 @@
...
+    method = method[[2]]
+    assertSubset(ens.method, choices = ls(.FilterEnsembleRegister), empty.ok = FALSE)
+    if (length(method) == 1) {
+      warningf("You only passed one base filter method to an ensemble filter. Please use at least two base filter methods to have a voting effect.")
+    }
+  } else {
+    assertChoice(method, choices = append(ls(.FilterRegister), ls(.FilterEnsembleRegister)))
+  }
+
+  method %in% ls(.FilterEnsembleRegister)
```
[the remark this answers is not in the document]
pat-s: Yes!

R/filterFeatures.R
```diff
@@ -41,14 +46,55 @@
...
+  # base.methods arrive here in a list when called from 'tuneParams'.
+  # we need them as a chr vec for further proc, so transforming
+  if (is.list(base.methods)) {
+    base.methods = as.character(base.methods)
+  }
+
+  # if a list is passed to arg 'method' (ensemble)
+  if (class(method) == "list") {
+    ens.method = method[[1]]
+    method = method[[2]]
```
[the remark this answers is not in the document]
pat-s: Nope. And I removed it now. Might probably cause confusion having multiple ways to specify an ensemble filter.
Using `method` in combination with `base.methods` is enough and keeps the code tidy.

R/generateFilterValues.R
```diff
@@ -28,16 +28,44 @@
...
+#'   method = list("E-mean", c("FSelectorRcpp_gain.ratio", "FSelectorRcpp_information.gain")))
 #' @export
-generateFilterValuesData = function(task, method = "randomForestSRC_importance",
-  nselect = getTaskNFeats(task), ..., more.args = list()) {
+generateFilterValuesData = function(task, method = "randomForestSRC_importance", nselect = getTaskNFeats(task), ..., more.args = list()) {
+
+  # ensemble
+  if (class(method) == "list") {
+    ens.method = method[[1]]
+    method = method[[2]]
```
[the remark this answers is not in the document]
pat-s: Ah, looks like we indeed need the list specification in `generateFilterValuesData()` if we do not want to complicate the function more.
But no, it was not documented until now. 

Still I would like to stay with only specification for `filterFeatures()`.
The rest is handle internally (passing the list within `filterFeatures()` to `generateFilterValuesData()`)

R/generateFilterValues.R
```diff
@@ -76,32 +104,49 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
-    x[match(fn, names(x))]
-  })
-
-  fval = do.call(cbind, fval)
-  colnames(fval) = method
-  types = vcapply(getTaskData(task, target.extra = TRUE)$data[fn], getClass1)
-  out = data.frame(name = row.names(fval),
-    type = types,
-    fval, row.names = NULL, stringsAsFactors = FALSE)
+  if (exists("ens.method")) {
```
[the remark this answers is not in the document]
pat-s: Since I do not use an explicit argument in the function for an ensemble filter method but do the checking and assignment internally, I am not aware of a more elegant way.
Probably there is and I do not know it.

Is there anything particular bad about `exists()`? Works and backed up by the `assertSubset()` check for ensemble filters in l.57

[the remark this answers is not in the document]
larskotthoff: This probably comes with a performance hit -- may not matter much as it's only in one place. You could simply set `ens.method = NA` in the main body of the function and then check whether it's `NA`.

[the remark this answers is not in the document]
pat-s: > You could simply set ens.method = NA in the main body of the function and then check whether it's NA

That does not sound very elegant. Do you have any reference about the performance drop of `exists`? At least style wise I would like it more than the proposed workaround.

R/FilterEnsemble.R
pat-s: Yes, there's not really much time left in the box for this implementation here :sweat_smile:

R/generateFilterValues.R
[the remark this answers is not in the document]
larskotthoff: I don't, but in general reflection tends to be expensive.

What happens if an object called `ens.method` is defined somewhere else, outside of this method?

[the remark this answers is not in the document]
larskotthoff: I don't, but in general reflection tends to be expensive.

What happens if an object called `ens.method` is defined somewhere else, outside of this method?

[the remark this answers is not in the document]
larskotthoff: I don't, but in general reflection tends to be expensive.

What happens if an object called `ens.method` is defined somewhere else, outside of this method?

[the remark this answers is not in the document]
larskotthoff: Ok, I didn't mean to post this three times, apparently github is having issues.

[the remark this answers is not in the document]
pat-s: > What happens if an object called ens.method is defined somewhere else, outside of this method?

This is a valid concern, yes. In this case indeed this variable from the global env is used and we error with an unspecific message. I've added another assert check in that condition to see if the defined filter makes sense.

But that does not prevent that the variable from the global env is not chosen in the first place. Guess we really need to define it in the function body. Done in https://github.com/mlr-org/mlr/pull/2456/commits/55c67c181087dc3b9b073cbd00a0e1cf734d0594.

pat-s:
`plotFilterValues()` needs adjustment: https://deploy-preview-2456--mlr-preview.netlify.com/reference/plotfiltervalues

larskotthoff:
Thanks, looks good.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # mlr 2.14.0.9000
 
+## Breaking
+
+- Instead of a wide `data.frame` filter values are now returned in a long (tidy) `tibble`. This makes it easier to apply post-processing methods (like `group_by()`, etc) (@pat-s, #2456)
```
jakob-r: I could imagine that this is actually something more people programmed against. So a breaking change here might irritate some people. Also I don't see the need to use a tibble here. If someone wants to do anything to the values one is free to transform and convert them as one pleases.
Regarding internal calculations: As we have data.table as a dependency - why don't we just use that?

pat-s: > I could imagine that this is actually something more people programmed against.

This refers to returning a long DF?
Changing this again would involve several hours since I have to re-arrange all outputs..

> Regarding internal calculations: As we have data.table as a dependency - why don't we just use that?

Besides one expection for which I failed with base R, we are not using any dplyr stuff internally.

> Also I don't see the need to use a tibble here. If someone wants to do anything to the values one is free to transform and convert them as one pleases.

Yes, sure. For now it is only used for printing, not internally. (i.e. the DF is coerced right before its returned). Which ofc makes no difference for the Import of tibble. I just hate it to print a DF that fills my console to Inf...

jakob-r: > This refers to returning a long DF?

Yes

> Changing this again would involve several hours since I have to re-arrange all outputs..

Just saying that it could break someones code. It is something that could make a reverse dependency check worth it before going on cran.

> Besides one expection for which I failed with base R, we are not using any dplyr stuff internally.

Can you point me to it?

>  I just hate it to print a DF that fills my console to Inf...

Then just add the following to your `.Rprofile` No need to add a whole package to the dependencies.

```r
.First = function() {
  ee = new.env()
  if (interactive()) {
    if ("tibble" %in% rownames(installed.packages())) {
      ee$print.data.frame = function(x, ...) {
        tibble:::print.tbl(tibble::as_tibble(x), ...)
      }
    }
  }
  attach(ee, warn.conflicts = FALSE)
}
```

pat-s: > Just saying that it could break someones code. It is something that could make a reverse dependency check worth it before going on cran.

It will break code since the structure of the returned filter value is different, yes. 

> It is something that could make a reverse dependency check worth it before going on cran.

Yes, I always do that.

> Then just add the following to your .Rprofile No need to add a whole package to the dependencies.

Nice hack. I'll use it :) - and get rid of using _tibble_ then in the package.

jakob-r: I updated it using melt from data.table

pat-s: :rocket: 	
so easy, damn...

pat-s:
I assume that the ensemble filters are not using caching since I see long runtimes for them in my study. The aggregation step cannot cause this so I assume the simple filters are not being taken from the cache. Have to inspect.

pat-s:
I made a mistake during merging (most likely) - caching was not used so far in this PR because the memoized function was not used. See https://github.com/mlr-org/mlr/pull/2456/commits/049969f1cb4baa2d35c0065db50070a7a4d9eb64. Fixed it now.

I was wondering heavily why everything took so long in my project.. :roll_eyes: :man_facepalming:

pat-s:
@larskotthoff 	@jakob-r 	
I guess we're good for now here. If I encounter more issues along the way I'll fix them separately. 

@jakob-r 	If you approve your review, feel free to merge.

DESCRIPTION
```diff
@@ -207,6 +208,7 @@ Suggests:
     LiblineaR,
     lintr (>= 1.0.0.9001),
     MASS,
+    magrittr,
```
jakob-r: Not needed.

DESCRIPTION
```diff
@@ -147,6 +147,7 @@ Imports:
     ggplot2,
     methods,
     parallelMap (>= 1.3),
+    rlang,
```
jakob-r: where is `.data` used?

pat-s: Leftover :+1: 	See commit.

pat-s:
@jakob-r 	Mergeable now?

pat-s:
Merging now.

PR #2463, opened 2018-10-23
Enable caching of filter values during tuning

### LINKED ISSUE #1995: Tuning with filters recalculates filter values each iteration

pat-s:
fixes #1995 	

# Implementation

- caching via `memoise::memoise()`
- caching is done on the filesystem because memory caching is not supported in parallel processes (https://github.com/r-lib/memoise/issues/77)
- argument `cache` can be passed to `makeFilterWrapper()`. It works with `resample()`, `tuneParams()`, `filterFeatures()`. 
- `cache` accepts a logical vector (using default cache dirs determined by the [rappdirs](https://github.com/r-lib/rappdirs) pkg) or a chr vector specifying a custom directory
- new function: `delete_cache()`: Deletes ONLY the default cache dirs. 
- new function: `get_cache_dir()`: Returns the default cache dirs. 
- removal of argument `nselect` in `generateFilterValuesData()` and all filters because its not used
- finally remove deprecated `getFilterValues()`

# Benchmark example

#### WITH Cache

``` r
library(mlr)
library(microbenchmark)

lrn = makeFilterWrapper(learner = "regr.ksvm", fw.method = "chi.squared", 
                        cache = TRUE)
ps = makeParamSet(makeNumericParam("fw.perc", lower = 0, upper = 1),
                  makeNumericParam("C", lower = -10, upper = 10, 
                                   trafo = function(x) 2^x),
                  makeNumericParam("sigma", lower = -10, upper = 10,
                                   trafo = function(x) 2^x)
)
rdesc = makeResampleDesc("CV", iters = 3)

y <- rnorm(100)
x <- matrix(rnorm(100 * 2000), 100, 2000)
dat <- data.frame(data.frame(y, as.data.frame(x)))
task = makeRegrTask(target = "y", data = dat)


set.seed(123)
microbenchmark(tuneParams(lrn, task = task, resampling = rdesc, par.set = ps,
                          control = makeTuneControlRandom(maxit = 10), show.info = T),
               times = 1)
#> [Tune] Started tuning learner regr.ksvm.filtered for parameter set:
#>            Type len Def    Constr Req Tunable Trafo
#> fw.perc numeric   -   -    0 to 1   -    TRUE     -
#> C       numeric   -   - -10 to 10   -    TRUE     Y
#> sigma   numeric   -   - -10 to 10   -    TRUE     Y
#> With control class: TuneControlRandom
#> Imputation value: Inf
#> [Tune-x] 1: fw.perc=0.962; C=4.08; sigma=1.23
#> [Tune-y] 1: mse.test.mean=1.0250471; time: 0.5 min
#> [Tune-x] 2: fw.perc=0.403; C=195; sigma=0.152
#> [Tune-y] 2: mse.test.mean=1.0250471; time: 0.0 min
#> [Tune-x] 3: fw.perc=0.288; C=0.0104; sigma=0.0106
#> [Tune-y] 3: mse.test.mean=1.0222349; time: 0.0 min
#> [Tune-x] 4: fw.perc=0.482; C=0.0326; sigma=0.0196
#> [Tune-y] 4: mse.test.mean=1.0226891; time: 0.0 min
#> [Tune-x] 5: fw.perc=0.674; C=0.00189; sigma=16.2
#> [Tune-y] 5: mse.test.mean=1.0226888; time: 0.0 min
#> [Tune-x] 6: fw.perc=0.352; C=0.283; sigma=85.6
#> [Tune-y] 6: mse.test.mean=1.0273247; time: 0.0 min
#> [Tune-x] 7: fw.perc=0.919; C=0.0491; sigma=597
#> [Tune-y] 7: mse.test.mean=1.0242091; time: 0.0 min
#> [Tune-x] 8: fw.perc=0.728; C=13.2; sigma=0.00203
#> [Tune-y] 8: mse.test.mean=1.0249831; time: 0.0 min
#> [Tune-x] 9: fw.perc=0.395; C=0.736; sigma=2.31
#> [Tune-y] 9: mse.test.mean=1.0307186; time: 0.0 min
#> [Tune-x] 10: fw.perc=0.698; C=318; sigma=5.16
#> [Tune-y] 10: mse.test.mean=1.0250471; time: 0.0 min
#> [Tune] Result: fw.perc=0.288; C=0.0104; sigma=0.0106 : mse.test.mean=1.0222349
#> Unit: seconds
#>                                                                                                                             expr
#>  tuneParams(lrn, task = task, resampling = rdesc, par.set = ps,      control = makeTuneControlRandom(maxit = 10), show.info = T)
#>       min       lq     mean   median       uq      max neval
#>  42.67322 42.67322 42.67322 42.67322 42.67322 42.67322     1
```

<sup>Created on 2018-11-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

#### WITHOUT caching

``` r
library(mlr)
library(microbenchmark)

lrn = makeFilterWrapper(learner = "regr.ksvm", fw.method = "chi.squared", 
                        cache = FALSE)
ps = makeParamSet(makeNumericParam("fw.perc", lower = 0, upper = 1),
                  makeNumericParam("C", lower = -10, upper = 10, 
                                   trafo = function(x) 2^x),
                  makeNumericParam("sigma", lower = -10, upper = 10,
                                   trafo = function(x) 2^x)
)
rdesc = makeResampleDesc("CV", iters = 3)

y <- rnorm(100)
x <- matrix(rnorm(100 * 2000), 100, 2000)
dat <- data.frame(data.frame(y, as.data.frame(x)))
task = makeRegrTask(target = "y", data = dat)


set.seed(123)
microbenchmark(tuneParams(lrn, task = task, resampling = rdesc, par.set = ps,
                          control = makeTuneControlRandom(maxit = 10), show.info = T),
               times = 1)
#> [Tune] Started tuning learner regr.ksvm.filtered for parameter set:
#>            Type len Def    Constr Req Tunable Trafo
#> fw.perc numeric   -   -    0 to 1   -    TRUE     -
#> C       numeric   -   - -10 to 10   -    TRUE     Y
#> sigma   numeric   -   - -10 to 10   -    TRUE     Y
#> With control class: TuneControlRandom
#> Imputation value: Inf
#> [Tune-x] 1: fw.perc=0.962; C=4.08; sigma=1.23
#> [Tune-y] 1: mse.test.mean=0.7753929; time: 0.4 min
#> [Tune-x] 2: fw.perc=0.403; C=195; sigma=0.152
#> [Tune-y] 2: mse.test.mean=0.7753929; time: 0.3 min
#> [Tune-x] 3: fw.perc=0.288; C=0.0104; sigma=0.0106
#> [Tune-y] 3: mse.test.mean=0.7776375; time: 0.4 min
#> [Tune-x] 4: fw.perc=0.482; C=0.0326; sigma=0.0196
#> [Tune-y] 4: mse.test.mean=0.7772678; time: 0.4 min
#> [Tune-x] 5: fw.perc=0.674; C=0.00189; sigma=16.2
#> [Tune-y] 5: mse.test.mean=0.7778738; time: 0.4 min
#> [Tune-x] 6: fw.perc=0.352; C=0.283; sigma=85.6
#> [Tune-y] 6: mse.test.mean=0.7774229; time: 0.4 min
#> [Tune-x] 7: fw.perc=0.919; C=0.0491; sigma=597
#> [Tune-y] 7: mse.test.mean=0.7766936; time: 0.4 min
#> [Tune-x] 8: fw.perc=0.728; C=13.2; sigma=0.00203
#> [Tune-y] 8: mse.test.mean=0.7762520; time: 0.4 min
#> [Tune-x] 9: fw.perc=0.395; C=0.736; sigma=2.31
#> [Tune-y] 9: mse.test.mean=0.7763041; time: 0.4 min
#> [Tune-x] 10: fw.perc=0.698; C=318; sigma=5.16
#> [Tune-y] 10: mse.test.mean=0.7753929; time: 0.3 min
#> [Tune] Result: fw.perc=0.403; C=195; sigma=0.152 : mse.test.mean=0.7753929
#> Unit: seconds
#>                                                                                                                             expr
#>  tuneParams(lrn, task = task, resampling = rdesc, par.set = ps,      control = makeTuneControlRandom(maxit = 10), show.info = T)
#>       min       lq     mean   median       uq      max neval
#>  224.1479 224.1479 224.1479 224.1479 224.1479 224.1479     1
```

<sup>Created on 2018-11-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


larskotthoff:
Could you add a test that checks that results are the same with and without caching please?

R/filterFeatures.R
```diff
@@ -42,7 +46,21 @@ filterFeatures = function(task, method = "randomForestSRC.rfsrc", fval = NULL, p
...
-    fval = generateFilterValuesData(task = task, method = method, nselect = nselect, ...)$data
+    if (isTRUE(cache)) {
+
+      if (!requireNamespace("memoise", quietly = TRUE)) {
+        stop("Package \"memoise\" needed for this function to work. Please install it.",
+             call. = FALSE)
+      }
+
+      # caching calls to `generateFilterValuesData()` with the same arguments
+      cache = memoise::cache_filesystem("~/.cache")
```
larskotthoff: Does this work on Windows? What if the directory doesn't exist? I would make this an option, i.e. instead of a Boolean make `cache` a character with `NULL` meaning no caching, otherwise use that directory. And if it's empty, use the system temp dir `tempdir()`.

R/resample.R
```diff
@@ -73,10 +75,11 @@
 #'   measures = list(mmce, setAggregation(mmce, train.mean)))
 #' print(r$aggr)
 resample = function(learner, task, resampling, measures, weights = NULL, models = FALSE,
-  extract, keep.pred = TRUE, ..., show.info = getMlrOption("show.info")) {
+  extract, keep.pred = TRUE, cache = FALSE, ..., show.info = getMlrOption("show.info")) {
```
larskotthoff: I would make `cache` an argument to the functions that create the filter wrappers, not to `resample`. I think this is misleading in that in most cases this won't do anything because nothing can be cached.

R/resample.R
```diff
@@ -103,7 +106,8 @@ resample = function(learner, task, resampling, measures, weights = NULL, models
 
   rin = resampling
   more.args = list(learner = learner, task = task, rin = rin, weights = NULL,
-    measures = measures, model = models, extract = extract, show.info = show.info)
+    measures = measures, model = models, extract = extract, show.info = show.info,
+    cache = cache)
```
larskotthoff: Probably not necessary to pass the `cache` argument down as it's in `learner`.

R/train.R
```diff
@@ -28,7 +31,7 @@
 #' learner = makeLearner("classif.rpart", minsplit = 7, predict.type = "prob")
 #' mod = train(learner, task, subset = training.set)
 #' print(mod)
-train = function(learner, task, subset = NULL, weights = NULL) {
+train = function(learner, task, subset = NULL, weights = NULL, cache = FALSE) {
```
larskotthoff: See above -- I would not add `cache` as an argument here.

tic.R
```diff
@@ -56,6 +56,7 @@ if (Sys.getenv("TUTORIAL") == "HTML") {
       add_step(step_setup_ssh())
 
     get_stage("deploy") %>%
+      add_code_step(devtools::document()) %>%
```
larskotthoff: This seems unrelated -- is it meant to be here?

R/filterFeatures.R
larskotthoff: Also if the cache was saved as part of the learner (instead of just the Boolean) repeated uses of the learner would benefit from the caching.

pat-s: > Does this work on Windows? What if the directory doesn't exist?

Good point. Didn't check on this. Will do it now.

>  instead of a Boolean make cache a character with NULL meaning no caching, otherwise use that directory.

Sounds good. 

> And if it's empty, use the system temp dir tempdir().

Tried tempdir() first and faced some weird errors, that's why I went with `~/.cache`. 

> Also if the cache was saved as part of the learner (instead of just the Boolean) repeated uses of the learner would benefit from the caching.

There are more opportunities for which we could use caching. But I think we should do this in a new PR.

R/resample.R
pat-s: I tried this at first but faced errors when calling `resample()` then stating that `cache` was not found etc pp.

R/resample.R
pat-s: Same as above - if I remove it I face errors that cache is not populate through the function calls.

R/train.R
pat-s: Same as above. I initially tried without adding `cache` as an arg but it seems we cannot go without.

tic.R
pat-s: Yes, its indeed needed. Even though `pkdown::build_site()` writes NAMESPACE before the call, we get NS errors. So somehow the NS change is not picked up.

It only works if we explicitly create the NS using `devtools::document()`.

In this case this happened because I removed `getFilterValues()` entirely in this PR. Then pkgdown complained about the functions still being listed as "exported" in the NS and errored. Adding this step before solved the problem.

R/resample.R
larskotthoff: Hmm, I would investigate that further. If the cache isn't found in some cases it may indicate that not as much caching is done as you may think.

R/filterFeatures.R
```diff
@@ -41,8 +53,36 @@ filterFeatures = function(task, method = "randomForestSRC.rfsrc", fval = NULL, p
...
+      if (class(cache) == "character") {
+        if(!fs::dir_exists(cache)) {
+          fs::dir_create(cache)
+        }
+        cache_dir = cache
+      } else {
+        if(!fs::dir_exists(rappdirs::user_cache_dir("mlr", "mlr-org"))) {
+          fs::dir_create(rappdirs::user_cache_dir("mlr", "mlr-org"))
+        }
+        cache_dir = rappdirs::user_cache_dir("mlr", "mlr-org")
```
larskotthoff: This should use `get_cache_dir()`. I would also check whether it exists and if it doesn't create it in there.

R/resample.R
```diff
@@ -77,6 +77,9 @@ resample = function(learner, task, resampling, measures, weights = NULL, models
 
   learner = checkLearner(learner)
   learner = setHyperPars(learner, ...)
+  # if (is.null(learner$cache)) {
+  #   learner$cache = cache
+  # }
```
larskotthoff: Remove commented code if no longer relevant please.

R/tuneParams.R
```diff
@@ -99,6 +101,7 @@ tuneParams = function(learner, task, resampling, measures, par.set, control, sho
   assertFlag(show.info)
   checkTunerParset(learner, par.set, measures, control)
   control = setDefaultImputeVal(control, measures)
+  #learner$cache = cache
```
larskotthoff: See above.

larskotthoff:
Thanks, I've made some minor comments. It also looks like the build is failing.

R/filterFeatures.R
pat-s: > This should use get_cache_dir().

Completely agree.

> I would also check whether it exists and if it doesn't create it in there.You 

You mean whether the cache is actually written to this directory?

pat-s:
> It also looks like the build is failing.

The "funny" thing is that master is also failing after the last "green" PR got merged.
It looks like some pkg now depends on a Bioconductor pkg and we need to install this first, otherwise `devtools::update_packages(T)` fails.

R/filterFeatures.R
mllg: Do we really want the extra `fs` dependency just for this?

mllg:
Some comments: 

* I don't see that we need the extra dependency `fs` for`mlr`. 
* Although `nselect` is not used by any `mlr` filter currently, I had some custom filters which used it. I would prefer to keep the few extra lines.

pat-s:
> I don't see that we need the extra dependency fs formlr.

I tried with base `dir.create()` and it did silently fail to create the directory (if a custom cache directory was given). Instead, using `fs::dir_create()` worked.

That's why I used the _fs_ pkg.
I think that's a valid reason. Also it is in SUGGESTS and should not have that much impact then.

> Although nselect is not used by any mlr filter currently, I had some custom filters which used it. I would prefer to keep the few extra lines.

Ok, fair enough.

berndbischl:
@pat-s you might get angry at me saying this now:
but i am REALLY unsure whether we should "shoehorn" such a mechnism into mlr now. 
caching just for filter values. that seems not that reasonable?
what are other thoughts here? i wnat this at least discussed before we do this, this is a major change to the base system
@mllg @larskotthoff

pat-s:
> but i am REALLY unsure whether we should "shoehorn" such a mechnism into mlr now.
> caching just for filter values. that seems not that reasonable?

I understand your point. 
However, filtering is the code part that will profit from caching most (since it is the only part that is recalled with the same settings every time). 
Enabling caching for all _mlr_ parts would involve a lot of more work with not so much impact as for the filter stuff.

Imo it is sufficient to only have caching for filtering in _mlr_ and implement it properly (pkg wide) in _mlr3_ right from the start.

I do not really a big disadvantage of this PR. Yes it is not pkg wide but does this point make it not mergeable?

But if you do not want to have this in _mlr_ because of this point I'll just leave it in the branch.

mllg:
> I tried with base `dir.create()` and it did silently fail to create the directory (if a custom cache directory was given). Instead, using `fs::dir_create()` worked.

If `dir.create()` fails, the alarm bells should start to ring. I assume this is because of race conditions during parallelization. `libuv` used by `fs` might have some fallbacks (e.g., timeouts, retries, ...) to solve this. Nevertheless, you still will encounter race conditions for the files. Memorization will not work reliably in parallel.


> That's why I used the _fs_ pkg.
> I think that's a valid reason. Also it is in SUGGESTS and should not have that much impact then.

I disagree. Most of our problems regarding maintainability is because of the long list of packages in Suggests, not the few packages in Imports.

pat-s:
> If dir.create() fails, the alarm bells should start to ring. I assume this is because of race conditions during parallelization. libuv used by fs might have some fallbacks (e.g., timeouts, retries, ...) to solve this. Nevertheless, you still will encounter race conditions for the files. Memorization will not work reliably in parallel.

`dir.create()` is just doing a one-time call creating the directory for caching. Even when executing it "manually" it failed to create the dir. I assume some permission problems here? But I did not debug further.
To be more explicit, `dir.create(rappdirs::use_cache_dir())` failed while `fs::dir_create(rappdirs::use_cache_dir())` worked on my machine. 

> Memorization will not work reliably in parallel.

As far as I've read the docs of `memoise`, parallelization is not a problem as long as caching is done **on the filesystem**. Parallelization and caching do not work when caching in memory as the processes do not share the same memory.

> I disagree. Most of our problems regarding maintainability is because of the long list of packages in Suggests, not the few packages in Imports.

I think we can debate about a lot of packages in SUGGESTS (to possibly clean up) but isn't `fs` one of those that really do a better (safer) job than the base R implementation?
If you insist on not using `fs`, it will take quite some time to figure out why `dir.create()` failed silently (!) :/.

mllg:
> > If dir.create() fails, the alarm bells should start to ring. I assume this is because of race conditions during parallelization. libuv used by fs might have some fallbacks (e.g., timeouts, retries, ...) to solve this. Nevertheless, you still will encounter race conditions for the files. Memorization will not work reliably in parallel.
> 
> `dir.create()` is just doing a one-time call creating the directory for caching. Even when executing it "manually" it failed to create the dir. I assume some permission problems here? But I did not debug further.
> To be more explicit, `dir.create(rappdirs::use_cache_dir())` failed while `fs::dir_create(rappdirs::use_cache_dir())` worked on my machine.

Try
```r
if (!dir.exists(path)) dir.create(path)
```
If this does not work, the permissions are indeed set strangely on your machine.


> > Memorization will not work reliably in parallel.
> 
> As far as I've read the docs of `memoise`, parallelization is not a problem as long as caching is done **on the filesystem**. Parallelization and caching do not work when caching in memory as the processes do not share the same memory.

Then the docs are wrong. You spawn multiple threads / processes which do the following concurrently:

1. Check if cache dir exists and create it, if necessary
2. Hash the inputs to the function call
3. Run the filtering
4. Store result as `cache_dir/[hash].rds`
5. Next computation will load the stored file

So lets assume that the cache dir already existed. Then (1) is no problem. The calculated hash will be the same on all / many workers. As soon as we try to store the results in (4), all workers with the same hash will try to write to the exactly same file concurrently. Depending on the file system different things will happen now. 

* On a local file system with a working file locking, the workers will write the file, one after another. Other workers in the meantime might try to load the file for (5), and now have to wait (which can lead to timeouts). 
* On a network file system without proper locking, you will corrupt your files, and system admins will get very angry because you overburden the file system and make the system unresponsive for everyone.

You either need a thread-safe storage system (like a data base), or pre-compute the results before the parallelization.

mllg:
Caching would still be nice to have for sequential execution though.

Issue #2468, opened 2018-11-05
Request for example on usage of `bit.names` and `bit.to.feature` arguments to `selectFeatures`


missuse:
Hi,

First I'd like to thank you for all the effort you put into mlr.
Second I'd like to apologize for my question since it is of the "How can this be done?" type, I decided to post it here contrary to the guidelines since I haven't received any answers or suggestions on [Stack Overflow](https://stackoverflow.com/questions/52986493/using-bit-names-and-bits-to-features-arguments-to-makefeatselwrapper-mlr-to-pe) in 10 days (even the bounty did not attract answers).

The question:
I would like to perform feature selection by a wrapper method on the iris data set using mlr package, however I would like to look only at groups of features associated with Petal and/or Sepal. So instead of looking at 4 features in different combinations the wrapper routine would look at two groups of features in different combinations.
If I understand correctly  this can be performed using two arguments `bit.names` and `bit.to.feature` to `selectFeatures` however I have not found usage examples and my own attempts were not fruitful. Could you provide an example of usage?

How would `bit.names` and `bit.to.feature` arguments be incorporated in the following code to accommodate search over the mentioned groups of features.
```r
library(mlr)
rdesc <- makeResampleDesc("Holdout")
ctrl <- makeFeatSelControlSequential(method = "sfs",
                                    maxit = NA)
res <- selectFeatures("classif.rpart",
                     iris.task,
                     rdesc,
                     control = ctrl)
analyzeFeatSelResult(res) 
```

Thank you,

Milan

berndbischl:
> Second I'd like to apologize for my question since it is of the "How can this be done?" type,

totally fine.


> . So instead of looking at 4 features in different combinations the wrapper routine would look at two groups of features in different combinations.

thats what bits.to.features is for

missuse:
I'd like to thank you both for the rapid response.

I added a comment on SO asking you to elaborate on the btf function:

```r
bitnames = c("Sepal", "Petal")
btf = function(x, task) {
  sets = list(
    c("Sepal.Length", "Sepal.Width"), 
    c("Petal.Length", "Petal.Width")
  )
  res = unlist(sets[as.logical(x)])
  if (is.null(res)) {
    return(character(0L))
  } else {
    return(res)  
  }
}
```

1. why does it have a task argument which is not used? Just because of the function check?
[...]

PR #2469, opened 2018-11-05
fix binaryToFeatures (Part of #2468)

jakob-r:
Clean up the code as most parts arent used anymore. Add a check to avoid unwanted vector recycling.

### DIFF
--- a/R/selectFeatures.R
+++ b/R/selectFeatures.R
@@ -21,9 +21,11 @@
 #' @param bit.names [character]\cr
 #'   Names of bits encoding the solutions. Also defines the total number of bits in the encoding.
 #'   Per default these are the feature names of the task.
+#'   Has to be used together with `bits.to.features`.
 #' @param bits.to.features [function(x, task)]\cr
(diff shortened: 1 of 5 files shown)


larskotthoff:
Can you add a test please?

jakob-r:
Hmm. It's an internal function which is already called numerous times by other tests. And it seems that it is never triggered (fortunately). But I probably could write a test with a wrong bits.to.features function.

larskotthoff:
That would be great, in particular for unwanted vector recycling.

berndbischl:
how do you know that this is correct? you just removed the matrix case.
sure that this not used on parallelization?

berndbischl:
can we try to fix the problem in issue #2468 here too?

jakob-r:
@berndbischl We could. But I would still like to seperate the two things. First lets do the basic stuff a bit better and afterwards we have to fix the FeatSelResult.
I would do a second PR after this here is accepted.

berndbischl:
Totally good and agreed

larskotthoff:
Thanks, merging.

PR #2474, opened 2018-11-06
Re-add FSelector filters

pat-s:
finishing up #2140 (probably should have used a new branch)

adding the comments in `Filter.R` helps navigating in RStudio:

selection_003


R/Filter.R
```diff
@@ -344,6 +360,65 @@ makeFilter(
...
+#' @rdname makeFilter
+#' @name makeFilter
+makeFilter(
+  name = "information.gain",
+  desc = "Entropy-based information gain between feature and target",
+  pkg  = "FSelector",
+  supported.tasks = c("classif", "regr"),
+  supported.features = c("numerics", "factors"),
+  fun = function(task, nselect, ...) {
+    y = FSelector::information.gain(getTaskFormula(task), data = getTaskData(task))
```
larskotthoff: Wait, either I'm missing something here or you're using the wrong functions -- shouldn't this be `FSelectorRcpp::` instead of `FSelector::`?

pat-s: The purpose is to re-add the removed FSelector filters because of [bernds comment](https://github.com/mlr-org/mlr/pull/2140#issuecomment-435698491). 

Because FSelector != FSelectorRcpp.

The FSelectorRcpp funs are already merged.

berndbischl:
should we rename the FS stuff to
FSelector.gainratio
FSelector.infogain
FSelector.symuncert

i know this can break code, but i dislike the  non-unified naming scheme

pat-s:
Sounds reasonable. Even though the naming scheme is still different to all other filters then (without the pkg prefix). And the praznik filters are again different..

R/Filter.R
larskotthoff: Ah yes, sorry!

NEWS.md
```diff
@@ -7,18 +7,31 @@
...
+* praznik_JMIM
+* praznik_MIM
+* praznik_NJMIM
+* praznik_MRMR
+* praznik_CMIM
+* FSelectorRcpp_gain.ratio
+* FSelectorRcpp_information.gain
+* FSelectorRcpp_symuncert
+
+## filter - renamed
```
mb706: Why was this heading added when all the actual renamings are documented further down under "filter - general"?

pat-s: Thanks for spotting, I corrected it :+1:

Issue #2472, opened 2018-11-06
TuneWrapper with clssif.glmnet does not seem to apply tuned hyperparameter in outer cross-validation


MartinKlm:
I came across this bug when comparing different predictors with "benchmark" and found Lasso to perform surprisingly poor.

So, what I think happens, is that the tuned "s" hyperparameters of "classif.glmnet" are not applied within the outer cross-validation when used in the TuneWrapper setting. The results marelely depend on the initial value of "s"; the extracted tuning values for each cross validation loop appear correct.

To illustrate this, I added three examples on the standard iris data set:
1) Not initializing "s" when creating the learner and subsequently tuning "s"
2) Initializing "s" with the a reasonable value (0.01; might be the default according to my simulation, but 0.1 is stated in the appendix) when creating the learner and subsequently tuning "s"
3)  Initializing "s" with the a unreasonable value (100) when creating the learner and subsequently tuning "s"

In general, all these exaples should produceto identical results, as the initial value should not matter, but only the tuned value. This was the case when executing the examples with the ksvm classifier, but does not hold for glmnet).

Here is the code:

```r
##
##
### optimize hyper parameter without initializing it

set.seed(123)
ps = makeParamSet(
  makeDiscreteParam("s", values = 2^(-10:10))
)
ctrl = makeTuneControlGrid()
inner = makeResampleDesc("Subsample", iters = 2)
lasso_lrn <- makeLearner("classif.glmnet", alpha=1)
lrn = makeTuneWrapper(lasso_lrn, resampling = inner, par.set = ps, control = ctrl, show.info = FALSE, measure=list(mlr::acc))

# Outer resampling loop
outer = makeResampleDesc("CV", iters = 3)
r1 = resample(lrn, iris.task, resampling = outer, extract = getTuneResult, show.info = FALSE, measure=list(mlr::acc))


###
###
### optimize hyper parameter and initializing it with default(?)

set.seed(123)
ps = makeParamSet(
  makeDiscreteParam("s", values = 2^(-10:10))
)
ctrl = makeTuneControlGrid()
inner = makeResampleDesc("Subsample", iters = 2)
lasso_lrn <- makeLearner("classif.glmnet", alpha=1, s=0.01)
lrn = makeTuneWrapper(lasso_lrn, resampling = inner, par.set = ps, control = ctrl, show.info = FALSE, measure=list(mlr::acc))

# Outer resampling loop
outer = makeResampleDesc("CV", iters = 3)
[...]

PR #2479, opened 2018-11-09
fix bug in TuneWrapper for param in used in predict function, issue 2472

berndbischl:
fixes issue #2472 

unit tests are missing. please help and merge


R/TuneWrapper.R
```diff
@@ -75,7 +75,10 @@ trainLearner.TuneWrapper = function(.learner, .task, .subset = NULL,  ...) {
 #' @export
 predictLearner.TuneWrapper = function(.learner, .model, .newdata, ...) {
   lrn = setHyperPars(.learner$next.learner, par.vals = .model$learner.model$opt.result$x)
-  predictLearner(lrn, .model$learner.model$next.model, .newdata, ...)
+  arglist = list(.learner = lrn, .model = .model$learner.model$next.model, .newdata = .newdata)
+  arglist = insert(arglist, list(...))
+  arglist = insert(arglist, .model$learner.model$opt.result$x)
```
mb706: Should probably only insert the parameters that have `when %in% c("predict", "both")`.

Maybe one could just copy [the lines from predict.R](https://github.com/mlr-org/mlr/blob/e547bae5e776c9378a764e33962f82cb44f95ecd/R/predict.R#L101-L106), if we are sure there are no other wrappers that introduce a discrepancy between the `par.vals` of a learner and the `...` parameters on purpose.

jakob-r:
Some notes regarding my observation:

- `setHyperPars` within `predictLearner` is without effect because they will never be requested. The hyperparams got extracted in `predict.(WrappedModel)` and are put into `...`. 
- We don't know if the `par.vals` got changed after the training of the TuneWrapped Learner. 
- We will always overwrite the existing par.vals with the ones from the learner.
- We assume that the user does not change par.vals and between training and predict. This holds especially as in most cases both is done internally in one step.

mb706:
> setHyperPars within predictLearner is without effect because they will never be requested. The hyperparams got extracted in predict.(WrappedModel) and are put into ....

we should probably not rely on this; right now the parameter values are given both inside the `...` and as part of the `.learner` object; its probably good to make sure they agree.

> We don't know if the par.vals got changed after the training of the TuneWrapped Learner.

How would that happen (assuming the user does not directly modify the `model$learner` object, in which case, too bad)?

> We will always overwrite the existing par.vals with the ones from the learner.

You mean the ones being tuned over? It might be worth considering to *remove* the parameters found in the tuning param set from the wrapped learner's param set, so the user gets an error message when he tries to set a param that will be overwritten later.

>We assume that the user does not change par.vals and between training and predict. This holds especially as in most cases both is done internally in one step.

I'm confused again, how would that happen? Does any part of `mlr` support that?

jakob-r:
> > setHyperPars within predictLearner is without effect because they will never be requested. The hyperparams got extracted in predict.(WrappedModel) and are put into ....
> 
> we should probably not rely on this; right now the parameter values are given both inside the `...` and as part of the `.learner` object; its probably good to make sure they agree.

That is the architecture of mlr at the moment and we are relying on it everywhere. In `train` and `predict` the parameters are read and put into the arguments.
This is probably not the best design but keeping the arguments and the par.vals "in sync" would only be a lot of effort.

> > We don't know if the par.vals got changed after the training of the TuneWrapped Learner.
> 
> How would that happen (assuming the user does not directly modify the `model$learner` object, in which case, too bad)?

Like that, yes., ? `mod = train(...); setHyperPars(mod$learner, ...); predict(mod, ...)`

> > We will always overwrite the existing par.vals with the ones from the learner.
> 
> You mean the ones being tuned over? It might be worth considering to _remove_ the parameters found in the tuning param set from the wrapped learner's param set, so the user gets an error message when he tries to set a param that will be overwritten later.

A warning somewhere might be useful. But I am afraid we would generate warnings for cases where we have set the par.vals in the definition of the learner. This can entail a lot of work.

> > We assume that the user does not change par.vals and between training and predict. This holds especially as in most cases both is done internally in one step.
> 
> I'm confused again, how would that happen? Does any part of `mlr` support that?
See above.

My suggestion: We should merge this PR asap and can add convenience later.

mb706:
> That is the architecture of mlr at the moment and we are relying on it everywhere. In train and predict the parameters are read and put into the arguments.

Do you consider [this](https://github.com/mlr-org/mlr/blob/f2817efb5d033adf2d090cd435852c974f1722d1/R/RLearner_classif_gamboost.R#L65) to be a bug?

>`mod = train(...); setHyperPars(mod$learner, ...); predict(mod, ...)`

has "`model$learner <- [something]`" ever been documented as supported behaviour? If the user modifies internals of objects then of course he can break things without bounds, I thought of `mod$learner` as such an internal.

berndbischl:
i do not understand the unit test. can you please add at least some comments? 
how the test is setup and why you expect certain things?

berndbischl:
also please note: a PR for fixing a certain bug is probably not the best place to dicuss general architecture questions. please move this to a clean issue if you want to discuss this...

jakob-r:
> Do you consider [this](https://github.com/mlr-org/mlr/blob/f2817efb5d033adf2d090cd435852c974f1722d1/R/RLearner_classif_gamboost.R#L65) to be a bug?

I would not call it a bug as it's just affecting some error message. But the whole code is strange. How do we know that it's really the wrong family that leads to NAs? But this is another issue and should not be discussed here.

> i do not understand the unit test.

Better now?

jakob-r:
@berndbischl *ping*

R/TuneWrapper.R
```diff
@@ -74,8 +74,16 @@ trainLearner.TuneWrapper = function(.learner, .task, .subset = NULL,  ...) {
 
 #' @export
 predictLearner.TuneWrapper = function(.learner, .model, .newdata, ...) {
-  lrn = setHyperPars(.learner$next.learner, par.vals = .model$learner.model$opt.result$x)
```
mb706: FYI I am relatively sure that removing this line (and not putting `.learner = lrn` further down etc) will break things.

jakob-r: you mean `.learner$next.learner`? That moved to another line. But I added setHyperPars now again.

pat-s:
@jakob-r If the build passes, can this be merged? I would add an entry to NEWS.md

PR #2480, opened 2018-11-09
Travis: Re-enable auto-deployment of docs

pat-s:
I finally managed to change the vignettes and function examples so that we have a _deterministic_ travis run that only auto-deploys if real changes are made.

Furthermore the **deployment will only be done in the master branch**. Why? This prevents the PRs being cluttered with `man/` and `docs` files. However, the tutorial will always be build. 

#### What about the branch preview of the tutorial then?
It still works if changes in `docs/` (i.e. in the tutorial) are manually committed to the PR. This is anyway the way how it should be done: We want to see changes in the tutorial slides and not really changes in the function reference (which can still be added). With this approach we really only have the important files that changed in the PR listed in the "files changed" tab. 

Deterministic travis run: https://travis-ci.org/mlr-org/mlr/builds/452883066 (hopefully ;) )

## Other changes

- Deployment is again done via _tic_ and restricted to man/*, DESCRIPTION, NAMESPACE and docs/* files.
- removed the slack build notifications as the channel does not even exist in the slack group anymore (or its hidden) and probably nobody uses it?
- removed bernds mail from the build notifications to prevent another full inbox ;)
- update to pandoc 2.3.1


pat-s:
@jakob-r 	@larskotthoff 	would be cool if you could have a look before we merge this.

larskotthoff:
Thanks, looks good.

PR #2483, opened 2018-11-13
Filter: Improve method of loading required filter pkgs

pat-s:
As @mb706 	pointed out, this should be handled (for filters) in 

https://github.com/mlr-org/mlr/blob/f2817efb5d033adf2d090cd435852c974f1722d1/R/generateFilterValues.R#L41-L42

However,issue https://github.com/mlr-org/mlr/issues/2484 applies.

This PR solves the issue by

- setting the `pkg` slot of a filter that requires no pkg to be loaded to `NULL`.
- changing the if statement from above so that it actually works


mb706:
Doesn't the `pkg  = "FSelectorRcpp",` argument in the `makeFilter` call already cause the package to be loaded when the filter is being used?

berndbischl:
I don't like this custom change. We should load packages always in the same way

pat-s:
@mb706 I need to check this.

@berndbischl we already use this way of loading the NS in the pkg. On the phone atm so can't refer to an example.

Just discovered today that all these msg are really annoying during usage.

berndbischl:
Ok that's new to me? Please link to code

berndbischl:
Actually please explain the problem first. That should always be done for a PR...

mb706:
See #2484, apparently `requirePackages` *should* be called, but *isn't*.

pat-s:
Take a look at https://github.com/mlr-org/mlr/issues/2484#issuecomment-438668903.

pat-s:
Not yet finished/working. Will continue after the meeting.

pat-s:
>  Also, since a user can define custom filters relying on other packages, the set of possible packages should not be enforced.

That's a good point. I changed it back to `assert_character(null.ok = TRUE)`.

berndbischl:
I don't get the change. See Martins last comment. Why should packages be Null? And why should Nas be allowed? Packages should be a char vec, without Nas. If empty no packages are required. If not empty all packs are required

pat-s:
> Why should packages be Null? And why should Nas be allowed?

Packages are still a chr vec. The only thing that changed is that filters that require NO pkg to be loaded (i.e. these pkgs are in base-R and loaded by default) now do not have a slot `pkg = character(0)` but `pkg = NULL` instead.

That's why I changed the assert function to accept `NULL`. `NULL` is imo more clean than using `character(0)` to me.
If we do only want to accept chrs, we should explicitly load the respective pkgs even if they are already loaded, e.g. for filter `variance` the `stats` pkg. 


The param decription of `pkg` in `makeFilter()` is: "Source package where the filter is implemented."
To me it does not make sense why this should evaluate to `character(0)` in any case?

R/Filter.R
```diff
@@ -525,26 +525,26 @@ univariate = makeFilter(
...
-    fns = getTaskFeatureNames(task)
-    res = double(length(fns))
-    for (i in seq_along(fns)) {
-      subtask = subsetTask(task, features = fns[i])
-      res[i] = resample(learner = perf.learner, task = subtask, resampling = perf.resampling, measures = perf.measure, keep.pred = FALSE, show.info = FALSE)$aggr
-    }
-    if (perf.measure[[1L]]$minimize)
-      res = -1.0 * res
-    setNames(res, fns)
+      if (is.null(perf.measure))
```
mb706: Changing indentation here is probably going to be misleading.

berndbischl:
> Packages are still a chr vec. The only thing that changed is that filters that require NO pkg to be loaded (i.e. these pkgs are in base-R and loaded by default) now do not have a slot `pkg = character(0)` but `pkg = NULL` instead.

i still dont get why we do that? as this is not consistent with what we do in learners? 
character(0) should be used if no packages are supposed to be loaded?

pat-s:
> i still dont get why we do that?

I found it more logic to set a not-existend value to `NULL` rather than `character(0)`. But if we introduce inconsistency with that, fair enough :)

@berndbischl 	tidied up everything, ready-to-go now?

pat-s:
ping @berndbischl

Issue #2208, opened 2018-03-02
Memory and speed issues related to h2o model training and predicting via mlr


kdpsingh:
I love utilizing h2o through mlr's interface. This issue relates to all h2o learners.

# Memory issues

When data is passed from mlr to h2o, the following code is used in the training...

```
 h2of = h2o::as.h2o(d)
```

... and similar code is used in generating predictions...

```
h2of = h2o::as.h2o(.newdata)
```

Each time the `as.h2o()` function is run, a **new copy** of the data is saved in h2o. This is not a major problem in routine use, but after prolonged use or when using mlr's `generateFeatureImportanceData()` function to compute permutation importance, this can effectively crash h2o.

I would suggest 2 changes to fix this:

1. I would suggest adding the argument `destination_frame="mlr_data_frame"` to the `as.h2o()` function. This will make sure that every time data is passed, it is given the same name as opposed to being assigned a new name and generating a copy.

2. I would suggest adding `h2o.rm("mlr_data_frame")` at the end of mlr's training and prediction functions for all of the h2o learners. This will ensure that even this single copy of the data that has been passed to h2o will be removed.

# Speed issues

When data is converted to h2o using the `as.h2o()` function, the default behavior is to write the data.frame to file and then read this into h2o. This can be slow for large files. h2o now supports the use of data.table's `fwrite()` function to make the file writing much faster. This is not enabled by default because h2o does not have the data.table package as a dependency. However, mlr does have the data.table package as a dependency, so there is no reason (in my mind) to not turn on this faster data conversion.

All it takes is the following code:

```
options("h2o.use.data.table"=TRUE)
```

I would consider adding this to the makeRLearner functions for all of the h2o learners.

@ledell, thoughts?

ledell:
Hi @kdpsingh, these are all really good ideas!  

- There is no issue with removing the H2OFrames from the H2O cluster once they've been used for training or prediction, so that would be a good thing to do.
- Setting `options("h2o.use.data.table"=TRUE)` will speed things up so that's a good optimization as well.

Thank you for considering these improvements.

j-hartshorn:
I've had a go at adding this into my PR #2508, should fix the problem.

pat-s:
Fixed in #2508. Thanks @j-hartshorn.

ledell:
Awesome, thank you @j-hartshorn!

PR #2508, opened 2018-12-17
Expanded functionality for regr.h2o.gbm

j-hartshorn:
This pull request implements the suggestions in #2208 for `regr.h2o.gbm` and elaborates on the available parameters.

In particular, it removes the decision to hard code `distribution = "gaussian"` from the learner and allows the use of others.


larskotthoff:
Thanks, looks good. Could you add a few of those new parameters in the test please so that they're checked automatically?

pat-s:
Will take a closer look in the next days, please ping if not ;)

PR #2538, opened 2019-02-02
Run checks on Appveyor

pat-s:
The current appveyor setup does not run tests and is somewhat unstable.

The current setup runs the tests and is (hopefully) a bit more robust.

As installing and caching both `pandoc` and `LateX` is tedious and error prone (segfaults in some runs), we run RCMD check with the following options:

```r
c("--as-cran", "--no-manual", "--no-vignettes",
        "--no-build-vignettes", "--no-multiarch")
```

We still have travis which does "all" checks. Most important is that we install and run the tests on Windows.

Latest appveyor builds can be found [here](https://ci.appveyor.com/project/mlr-org/mlr/history).


pat-s:
Sometimes it hangs when initializing tests. Let's see how it behaves over time. Apart from this, it looks ok.

PR #2543, opened 2019-02-08
replace geom_tile with geom_raster

fabian-s:
no ugly grid-artefacts (tile would also need a colour-aes for the tile border, not just fill...) & slightly faster rendering


larskotthoff:
Thanks, merging.

PR #2579, opened 2019-04-30
Bugfix: Aggregation of probabilities

### LINKED ISSUE #2578: Bug in predictions due to recent change in data.table
`data.table` changed its behaviour wrt. rbind by name/position. Recent version issue a message about this, e.g. if you run this code:
```r
set.seed(123)
lrn = makeLearner("classif.svm", predict.type = "prob")
task = iris.task
ctrl = makeTuneControlRandom(maxit = 2)
ps = makeParamSet(makeNumericParam("cost", lower = 0.01, upper = 0.1))
lrn = makeTuneWrapper(lrn, cv3, par.set = ps, control = ctrl)
resample(lrn, task, cv3)
```

This leads to errors down the analysis, where probabilites are not joint corretly together in prediction objects. I.e., probabiltites for the first class where rbinded to the column for the 2nd class, and vice versa. See https://github.com/mlr-org/mlr/blob/3b52b1213ee26fbe53ce9a6461510bf32f25b807/R/ResamplePrediction.R#L26.

To turn the message into an error, you can set the option `options(datatable.rbindlist.check="error")`.

mllg:
1st attempt to fix #2578


DESCRIPTION
```diff
@@ -261,9 +261,6 @@ Suggests:
 VignetteBuilder:
     knitr
 Remotes:
-    berndbischl/BBmisc,
```
pat-s: Was this change on purpose or bc the Travis build failed back then?

mllg: Is there a reason why why install the devel version? If yes, we need to upload BBmisc before the next CRAN release. Otherwise it would be best to use the exact same version CRAN uses.

pat-s: AFAIK there was but probably all problems have been fixed? I released v2.14.0 ofc without the Remotes Field and everything seems to be fine. So yes, I guess we can drop these requirements.

pat-s:
@mllg I added NEWS, please merge if you approve.

mllg:
Perfect, thanks!

PR #2589, opened 2019-05-23
Add regr.gstat learner

pat-s:
Clean version of #2394

@pokyah 

If you want, you can start from here and finish it up :) 

I've changed the data to use `spatial.task` from our package. There is no need for a second spatial task in the package for testing purposes only.


pat-s:
@pokyah 	If you want to continue here, let me know. Even better would be to implement this learner in https://github.com/mlr-org/mlr3, more specifically in https://github.com/mlr-org/mlr3spatiotemporal.

pokyah:
Hi @pat-s, 

Sorry for my late answer. 

Will try to find some time in the coming weeks to finish this implementation in the mlr3 version

PR #2591, opened 2019-05-27
Tests: Various Learner fixes

jakob-r:
Please check if we still need them. Originally they were in PR #2378.


jakob-r:
Notes:
* RLearner_classif_cforest.R  
  We do not handle defaults like that.
* RLearner_regr_crs.R  
  I cannot remember but I probably checked the se 
* helper_learners_all.R  
  Wanted to check upper values for se estimation. Strangely `regr.gausspr` gives realy high values althogh they should be se estimates (and not var or etc)

pat-s:
@jakob-r 	
Removing the lines from cforest seems to break the tests.
Do you know OOB how the defaults of the learner should be set correctly?
Haven't yet taken a closer look since I suspect you might know it.

The other two fixes do not seem to cause any issues.

jakob-r:
Ok. This is more complex then I expected.
`cforest()` has the following default `controls = cforest_unbiased(...)` which is just a wrapper for `cforest_control` with some defaults. Unfortunately these defaults can not be overwritten, so we have to do manual default handling somehow and try to mimic the behavior of `cforest()`. So we can just leave it as it is, because other solutions are equally unelegant.

pat-s:
Thanks for looking into this! Hope it did not take too much time. Merging master should resolve the Travis error you got.

pat-s:
Thanks again @jakob-r!
The one CI build that didn't pass had a problem pushing the cache.
Merging.

PR #2599, opened 2019-06-06
Change argument default for `benchmark(models = )` and add `keep.extract` argument

pat-s:
# Background

Until now it was not possible to get rid of the `$extract` slot when calling `benchmark()`.

This slot stores several artifacts created during the execution of `benchmark()`, e.g. the tuning results.
These can increase the resulting object size tremendously (from MB to GB) even for normal datasets (~ 1k obs). 
The problem are the extensive tuning results that are stored (e.g. when running a 100 reps nested CV with MBO).

Due to this behavior we also agreed internally to throw away all of these artifacts by default in _mlr3_ right from the start.
Usually tuning results are not explicitly needed when just doing a benchmark comparison.
In comparison storing the predictions with `keep.pred` does almost make no different in terms of R object size.

No action is needed for `resample()` since tuning results are not stored by default but only if argument `extract` is specified in the first place.
For `benchmark()`, this happens automatically which is why so far the tuning results were always stored.

Changing the argument default here is a major change but is an important change in the long run.
It prevents users from running into memory problems without knowing what the reason might be.
If users enable this switch on their own they have a higher change of detecting this change as the possible reason for subsequent errors.
Also it enables consistency with _mlr3_ which will also act like this by default.

# Changes

- adds the `keep.extract = FALSE` argument to `benchmark()`.
- changes the default for argument `models` from `TRUE` to `FALSE`


larskotthoff:
Looks good, but this should also be mentioned/explained in the tutorial (and check whether everything in the tutorial still works with this change).

pat-s:
> and check whether everything in the tutorial still works with this change).

~~The tutorial is checked during CI and should fail if something does not work.~~

It does not fail, it just returns `NULL` for the empty slots.
Will update the tutorial!

> this should also be mentioned/explained in the tutorial

indeed, adjustments are needed here. Thanks!

pat-s:
- We actually did not change anything for `resample()` -> removed the NEWS entry

- updated the benchmark vignette to reflect the new changes

larskotthoff:
Yes, in an ideal world we would notice problems in the tutorial automatically... ;)

R/batchmark.R
```diff
@@ -126,7 +127,13 @@ reduceBatchmarkResults = function(ids = NULL, keep.pred = TRUE, show.info = getM
       res = batchtools::reduceResultsList(tab[problem == prob & algorithm == algo], reg = reg)
       models = !is.null(res[[1L]]$model)
       lrn = data$learner[[algo]]
-      extract.this = getExtractor(lrn)
+      if (isTRUE(keep.extract)) {
```
jakob-r: Why is this done here at the reducing stage and not in `getAlgoFun`. Now it just saves all the stuff to throw it away at the reduce stage, right?

pat-s: Good catch! :+1:

R/batchmark.R
```diff
@@ -73,7 +73,13 @@ getAlgoFun = function(lrn, measures, models) {
   force(measures)
   force(models)
   function(job, data, instance) {
-    extract.this = getExtractor(lrn)
+    if (isTRUE(keep.extract)) {
```
jakob-r: This value is not available here. Did the test catch that?

pat-s:
@jakob-r 	Putting the if clause into `getAlgoFun()` breaks a lot of tests. I do not see the problem right away. Seems `getAlgoFun()` is called from a variety of places and some other funs might not be working as expected if the extract result is missing.

Should we just leave it as in https://github.com/mlr-org/mlr/pull/2599/commits/be55c8d595850702abcb40f13ef84e9396289a84?

jakob-r:
> Seems getAlgoFun() is called from a variety of places

No. It's just called from within `batchmark()` which is quite easy to find out. But I fixed it now.

Besides: We should have a test that checks that setting keep.extract even has an effect.

R/batchmark.R
```diff
@@ -73,14 +73,14 @@ getAlgoFun = function(lrn, measures, models, keep.extract) {
...
-        NULL
-      }
+  if (isTRUE(keep.extract)) {
+    extract.this = getExtractor(lrn)
+  } else {
+    extract.this = function(model) {
+      NULL
     }
+  }
+  function(job, data, instance) {
```
pat-s: Shouldn't we pass `extract.this` along here?

pat-s:
> Besides: We should have a test that checks that setting keep.extract even has an effect.

Already exists.
https://github.com/mlr-org/mlr/blob/a4bcd2ab7f23f5f170cb07740dfaf5444bf83477/tests/testthat/test_basenocran_batchmark.R#L141-L151

Travis is green btw and just not updated here.

R/batchmark.R
jakob-r: No, it lives now in the Environment of the function we are returning. Together with `lrn`, `measures` and `models`.

jakob-r:
> Already exists.

This test is a pain to read. All I am seeing is that with `keep.extract = TRUE` we see that I can extract the tuneResult. But I am missing a test that checks wheter it is not there if `keep.extract = FALSE`.

pat-s:
> But I am missing a test that checks whether it is not there if keep.extract = FALSE.

This is indeed missing, yes. 
The test errored before I added `keep.extract` explicitly after the change but yes, there is none right now that checks for the missing extract slot.

pat-s:
@jakob-r I now added a test to check if the `extract` slot is really empty when setting `keep.extract = FALSE`.

If you're ok with it, feel free to merge.

PR #2622, opened 2019-07-18
add argument `progress` to `generateFeatureImportanceData()`

pat-s:
With lots of features this function can take very long.

This change adds a `progress` argument which prints the following information is set to `TRUE`:

- the current feature name being processed
- the index of the feature in the list of all features
- the elapsed time since the last feature finished (hardcoded to minutes). Imo this should be sufficient even for long running processes

Small example: 

``` r
library("mlr")

fit = train(lrn, iris.task)
imp = generateFeatureImportanceData(iris.task, "permutation.importance",
  lrn, nmc = 10L, progress = TRUE)
#> Feature: 'Sepal.Length' [1/4, 0 min]
#> Feature: 'Sepal.Width' [2/4, 0 min]
#> Feature: 'Petal.Length' [3/4, 0 min]
#> Feature: 'Petal.Width' [4/4, 0 min]
```

<sup>Created on 2019-07-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


larskotthoff:
This looks great and is probably something we should have in mlr3 as well. Ideally as a generic thing that allows to show progress for all operations.

berndbischl:
But all other functions call this arg show.info? Pls always strive for consistency in the api

pat-s:
> But all other functions call this arg show.info? Pls always strive for consistency in the

Yes, thanks! Changed it.

PR #2646, opened 2019-10-03
Bug fix for Issue 2639

annette987:
When using the RF min depth filter with a method other than "md" (i.e.
"vh" or "vh.vimp"), an error is generated at line 298:
	Error in -im[, 1L]: invalid argument to unary operator.
This happens because im = NULL.
From code: im = (result of RF var.select)$md.obj$order
From randomForestSRC documentation: md.obj - Minimal depth object. NULL
unless method = "md".
Therefore the object md.obj cannot be used for methods "vh" and "vh.vimp".
The solution is to use object varselect instead, which is a matrix
containing the tree depth and relative frequency for each variable. Note
also that object topvars contains the names of the top variables.

P.S. This is my first PR and I don;t have access to the coding guidelines or other docs. Could I be added please?

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -276,8 +276,8 @@ rf.min.depth = makeFilter(
   supported.features = c("numerics", "factors", "ordered"),
   fun = function(task, nselect, method = "md", ...) {
     im = randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),
-      method = method, verbose = FALSE, ...)$md.obj$order
-    setNames(-im[, 1L], rownames(im))
+		method = method, verbose = FALSE, ...)$varselect
+    setNames(im[, 1L], rownames(im))
   })
 .FilterRegister[["rf.min.depth"]] = rf.min.depth
 .FilterRegister[["rf.min.depth"]]$desc = "(DEPRECATED)"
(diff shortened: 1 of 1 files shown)


annette987:
Sorry, I have never done this before and I don't really know what to do next. Could I have some guidance please?

R/Filter.R
```diff
@@ -276,8 +276,8 @@ rf.min.depth = makeFilter(
   supported.features = c("numerics", "factors", "ordered"),
   fun = function(task, nselect, method = "md", ...) {
     im = randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),
-      method = method, verbose = FALSE, ...)$md.obj$order
-    setNames(-im[, 1L], rownames(im))
+		method = method, verbose = FALSE, ...)$varselect
```
larskotthoff: Please don't change the indentation.

annette987: OK. But is that the reason it failed?

larskotthoff:
Thanks for the PR, looks good apart from some minor issues I raised in the review. You can find our contributor guidelines in the wiki at https://github.com/mlr-org/mlr3/wiki/pr-guidelines

For bug fixes, we ask that you write a unit test that checks whether the bug is fixed in addition to the fix itself. This is to make sure that we don't accidentally reintroduce the same bug in future changes. Could you do that please? Thanks!

pat-s:
> The solution is to use object varselect instead, which is a matrix

Was there an upstream change in _randomForestSRC_?
Is the change robust and applies to all methods, i.e. will everything else work as before with added benefits?

Reading your initial post is not easy, please take a few more minutes next time to format it nicely. Thanks.

PR #2649, opened 2019-10-07
Support multiple choices in argument `method` for `randomForest…

### LINKED ISSUE #2639: Parameter naming conflict with RF Min Depth filter
I am aware of issue https://github.com/mlr-org/mlr/issues/1066 about a conflict with the parameter name method when using the RF Min Depth filter. Although this issue was resolved I am still experiencing the same problem:

``` r
library(survival)
#> Warning: package 'survival' was built under R version 3.5.3
library(mlr)

data(veteran)
set.seed(24601)
task_id = "VET"
vet.task <- makeSurvTask(id = task_id, data = veteran, target = c("time", "status"))
vet.task <- createDummyFeatures(vet.task)
tuning = makeResampleDesc("CV", iters=2, stratify=TRUE)                             # Tuning: 5-fold CV, no repeats
outer = makeResampleDesc("CV", iters=2, stratify=TRUE)                              # Tuning: 5-fold CV, no repeats

filt = makeFilterWrapper(
    makeLearner(cl="surv.coxph", id = "cox.filt.rfsrc", predict.type="response"), 
    fw.method="randomForestSRC_var.select",
    cache=TRUE,
    ntree=500,
    method="vh"
)
bmr = benchmark(filt, vet.task, outer, list(cindex), show.info = TRUE, models=TRUE, keep.extract=FALSE)
#> Task: VET, Learner: cox.filt.rfsrc.filtered
#> Error in (function (task, method = "randomForestSRC_importance", fval = NULL, : formal argument "method" matched by multiple actual arguments
```

<sup>Created on 2019-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I installed the latest development version of mlr. Am I doing something wrong or is this bug still in play?

pat-s:
closes #2639 

superseeds #2646 

@annette987 Could you try if this PR solves the issue you faced in #2639?

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -266,8 +266,6 @@ randomForestSRC.rfsrc = makeFilter(
 #' @name makeFilter
 NULL
 
-# for some reason we cannot call 'randomForestSRC_var.select' directly as we then face
-# nested recursion problems when using other methods than "md".
 rf.min.depth = makeFilter(
   name = "randomForestSRC_var.select",
   desc = "Minimal depth of / variable hunting via method var.select on random forests fitted in package 'randomForestSRC'.",
@@ -275,9 +273,7 @@ rf.min.depth = makeFilter(
   supported.tasks = c("classif", "regr", "surv"),
   supported.features = c("numerics", "factors", "ordered"),
   fun = function(task, nselect, method = "md", ...) {
-    im = randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),
-      method = method, verbose = FALSE, ...)$md.obj$order
-    setNames(-im[, 1L], rownames(im))
+    # redirected to randomForestSRC.var.select()
   })
 .FilterRegister[["rf.min.depth"]] = rf.min.depth
 .FilterRegister[["rf.min.depth"]]$desc = "(DEPRECATED)"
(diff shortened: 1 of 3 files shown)


R/Filter.R
```diff
@@ -266,18 +266,14 @@ randomForestSRC.rfsrc = makeFilter(
...
   name = "randomForestSRC_var.select",
   desc = "Minimal depth of / variable hunting via method var.select on random forests fitted in package 'randomForestSRC'.",
   pkg = "randomForestSRC",
   supported.tasks = c("classif", "regr", "surv"),
   supported.features = c("numerics", "factors", "ordered"),
   fun = function(task, nselect, method = "md", ...) {
-    im = randomForestSRC::var.select(getTaskFormula(task), getTaskData(task),
-      method = method, verbose = FALSE, ...)$md.obj$order
-    setNames(-im[, 1L], rownames(im))
+    # redirected to randomForestSRC.var.select()
```
larskotthoff: Wait, I'm missing something here. How does this redirect work?

pat-s: Because down the line, `randomForestSRC.var.select` does another `.FilterRegister[["randomForestSRC.var.select"]] = randomForestSRC.var.select` call.

There can only one function be registered. I found this when debugging; we always jump to the body of the function who last called `.FilterRegister` for a unique name.

So in detail this is what happens

1. `rf.min.depth` registers `rf.min.depth`, calls it "deprecated" and then registers the filter under the name `randomForestSRC_var.select`.

2. `randomForestSRC.var.select` does the same as 1) and hence the body of this filter will finally be used when someone calls `randomForestSRC_var.select` as it overwrites 1)

Yes, this is "deprecation hell" but I am not eager on touching more in here.
So currently we have 2 deprecated methods (`randomForestSRC.var.select` and `rf.min.depth`) and the current one `randomForestSRC_var.select` which all call the same function body

larskotthoff: Ok, thanks.

annette987:
Yes that is working for me. Thank you.

PR #2653, opened 2019-10-09
Use `data.table::melt()` instead of `reshape2::melt()`

### LINKED ISSUE #2652: tuneParams() triggers data.table reshape2::melt warning via melt() in R/generateFilterValues.R
I hope this can be investigated without a minimal working example. I'm getting warnings while invoking tuneParams():

> In melt(out, value.name = "value", measure.vars = method,  ... :
>
> The melt generic in data.table has been passed a data.frame and will attem
> pt to redirect to the relevant reshape2 method; please note that reshape2 is
> deprecated, and this redirection is now deprecated as well. To continue usi
> ng melt methods from reshape2 while both libraries are attached, e.g. melt.l
> ist, you can prepend the namespace like reshape2::melt(out). In the next ver
> sion, this warning will become an error.

The only line of code matching the pattern `melt(out, value.name = "value", measure.vars = method,` is in `R/generateFilterValues.R`. The warning itself comes from data.table. Data.table has been deprecating reshape2::melt since version v1.12.4 -- see NEWS.md, note 8 for that version.

My package versions: mlr 2.15.0, data.table 1.12.4, reshape2 1.4.3, R 3.6.1.

pat-s:
fixes #2652 

When passing a `data.frame` to `melt()`, _data.table_ redirects to _reshape2_. This redirection is removed with the next version.
To use the S3 generic from _data.table_, we simply need to pass a `data.table` instead of a `data.frame`.


larskotthoff:
Can this be tested with a unit test?

larskotthoff:
Ok, sounds good.

Issue #2645, opened 2019-10-02
Inspect changes in praznik v7.0 release


pat-s:
which broke CI: https://travis-ci.org/mlr-org/mlr/jobs/592255714#L1659

pat-s:
See https://gitlab.com/mbq/praznik/issues/19

PR #2654, opened 2019-10-11
Temporary fix for praznik v7.0

pat-s:
#2645 

## Why temporary?

Currently I think that there is an [bug](https://gitlab.com/mbq/praznik/issues/19) upstream in the new release. 
There is a new engine for the "mutual condition" engine in v7.0 and it seems that is does only return NA's for factors right now.
I fixed the tests so that they pass for now and we have a green build again.

## Cleanup praznik test

I cleaned the praznik test:

- We do not need a logical since `makeClassifTask()` does not accept logicals and we convert it to a factor anyways using `convertDataFrameCols()`
- We do not need multiple factor variables to test on


larskotthoff:
Looks like a bunch of other changes slipped into this PR -- could you make a PR with only the relevant changes please?

pat-s:
These are the deployed pkgdown files by Travis.

`master` hasn't run successfully in a while and didn't deploy the pkgdown site. 
I've now done it manually so that there won't be deployed pkgdown changes in this PR.

PR #2668, opened 2019-10-27
Update `RLearner_classif_h2odeeplearning.R`

### LINKED ISSUE #2658: h2o.deeplearning early stopping: stopping_rounds is missing?
Doesn't seems to be `stopping_rounds` in `getParamSet("classif.h2o.deeplearning")`.
I'm aware of `classification_stop` but it's not the same.
Am i missing something here?

albersonmiranda:
Added params.

fixes #2658


larskotthoff:
Thanks, looks good. Could you also please add at least some of the new parameters to the tests for this learner?

Are these parameters relevant for the regression learner as well? If so, it would be great if you could add them to that as well please.

jakob-r:
We should check, because some of these "new" ones were already in the regr learner.

albersonmiranda:
@jakob-r i just verified and all of them are already set in regr learner, except for `stopping_rounds` which is se as integer in regr and i've set as numeric with another default. It worked locally but i'll make it match regr just in case.

jakob-r:
> Also, there's a few more `stopping_metric` values in documentation (which i've added in classif.) that aren't defined for regr.

I would assume that `"deviance", "logloss", "MSE", "RMSE", "MAE", "RMSLE"` are stopping metrics for `regr` and not `classif`. Unfortunately the h2o documentation does not give hints here.

> `stopping_rounds` which is set as integer in regr and i've set as numeric with another default. It worked locally but i'll make it match regr just in case.

It might work because h2o converts it but certainly e.g. 43.4 would not be a plausible value.

berndbischl:
Log loss is for classify... Come on... Deviance probably too

jakob-r:
My bad. But deviance seems to be for regression according to [this page](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/deep-learning.html) Although I do not understand the "formula".

berndbischl:
i was wrong. deviance is for regression. i think it should be possible to use it also for classif, but maybe not supported, maybe simply test it for 1 min.

also i had to stare 5 min at the formula to understand it

Loss = Quadratic -> MSE==Deviance For Absolute/Laplace or Huber -> MSE != Deviance

you have to read it like this

if loss=l2, then mse == dev
if loss=l1 or huber, then mse!=dev

pat-s:
Whats the status here?

@albersonmiranda Will you incorporate the changes from Jakob?

albersonmiranda:
@pat-s sorry for the delay. Updated  `stopping_metric` as Jakob pointed out and added `distribution` with values `c("bernoulli", "multinomial")`. I'm not sure if others available make sense for classification problems.

Also, `c("bernoulli", "multinomial")` distributions and `"logloss",  "AUC", "misclassification"` `stopping_metric`s are defined for regression. Do you agree to open a PR to exclude them?

jakob-r:
From my side it is related enough to do it in one PR

albersonmiranda:
Done!

pat-s:
Thanks for contributing!

PR #2674, opened 2019-11-05
Robustify 1-row matrix returns from prediction

### LINKED ISSUE #2673: gbm multiclass classification predict throws an error with only one observation
Hi,

I would like to report a bug, that I have met today.
Execution of the following example 
```
library(DALEX)
library(mlr)
task <- makeClassifTask(data = HR, target = "status")
learner_gbm <- makeLearner("classif.gbm", predict.type = "prob", par.vals = list(n.trees=280, n.minobsinnode=10, shrinkage=0.00736))
model <- train(learner_gbm, task)
predict(model, newdata = HR[1,])
```
throws a following error
```
Error in checkPredictLearnerOutput(.learner, .model, p) : 
  predictLearner for classif.gbm has returned a class numeric instead of a matrix!
```
Have to say that it is quite problematic for me. Also I assume it is a bug since for randomForest predict with only one observation works without any problem
```
library(DALEX)
library(mlr)
task <- makeClassifTask(data = HR, target = "status")
learner_rf <- makeLearner("classif.randomForest", predict.type = "prob")
model <- train(learner_rf, task)
predict(model, newdata = HR[1,])
```

Best,
Szymon

pat-s:
fixes #2673 

Previously we had `p[, , 1]` in the predict functions of some learners.

If `newdata` has only one row, the prediction result will be a named numeric instead of a matrix. This causes an assertion error later on as outlined in the linked issue.

The following replacement ensures that also prediction objects with 1 row will be of class `matrix`.


larskotthoff:
Could you add a test for this as well please (e.g. the original bug)?

larskotthoff:
Thanks, but some unrelated changes seem to have slipped in. And *cough* style guide *cough* (<-).

larskotthoff:
Thank you!

PR #2675, opened 2019-11-05
Add pkgdown reference index

pat-s:
We now have groupings in our reference index on the pkgdown site. This is really helpful for getting a grouped overview of the functions.

Also, three learners had their own help file. To stay consistent, I removed these help pages.


larskotthoff:
Hmm, so the documentation for those learners is just completely gone now? I'd rather be inconsistent and have more documentation :)

pat-s:
> Hmm, so the documentation for those learners is just completely gone now? I'd rather be inconsistent and have more documentation :)

We have somewhat about 90 learners, 3 of them had a doc (for no reasons). I cannot see why this would help a user in particular.
Also I doubt that the doc was up-to-date.

This was most likely a leftover from early times when there was motivation to document every learner with a help page.
It only became visible because of the help page grouping in this task. 
I would really vote for their removal.

@berndbischl @mllg third opinion?

berndbischl:
> This was most likely a leftover from early times when there was motivation to document every learner with a help page.

no not really. it was like this:
- we never planned for learners to have help pages (due to maintenance, maybe)
- than people requested extra info an help on learners. it is obviously useful right?
- we decided (unfortunately?) to out this info into learner s3 slots. the idea was to have this on the console and in a programmable fashion
- we then figured out that this probably a very bad approach. you cannot really have "structuing" elements, lists. quotes, links, in these simple string fields
- so then at the end some learners, for which more complex help pages where necessary, we created individual help pages.

---> conclusion:
the bst apprpoach might have been a single, manually crafted page with commets on some learners

does this clear things up?

pat-s:
> does this clear things up?

a bit yes, but this is not a statement on what to do with the existing pages now.

For me, I've not even been aware of those three until now.
Should be maybe just put them into a "section" in `?makeLearner`?
The issue I see is that if there are only pages for 3/80 people won't look them up because they are not aware of them. However, chances are higher that people read `?makeLearner` in more detail.

berndbischl:
> Should be maybe just put them into a "section" in `?makeLearner`?

totally agree, also with your argument below

pat-s:
Put the learner help pages as separate sections in `makeLearner()`.

pat-s:
Merging now, the R 3.5.3 build failure relates to a bug in glmnet (confirmed via mail) and the rest seems to be fine.

PR #2684, opened 2019-11-25
Fix for issue #2685 - Ensemble filters

annette987:
## The Problem: 
Ensemble filters do not select the top-ranked features - see Issue #2685

## The Reasons

1. The ranking calculation in calcBaseFilters  (FilterEnsemble.R )  is incorrect. See evidence in issue #2685.

2. In filterFeatures.R, the method to use for the final ranking must be selected. This is done by the line below, but this has no effect:
`fval = fval[fval$method == fval$method, ]`
As a result fval still contains the data for all methods, not just the ensemble method and the final ranking is done using one of the base methods.

3. When the top features are selected by setting a threshold, this must be done after the subsetting referred to in point 2. If not nselect can be greater than the total number of features.

```
  if (select == "threshold") {
    nselect = sum(fval[["value"]] >= threshold, na.rm = TRUE)
  }
```

If fval contains data for all methods, there will be multiple rows for each feature - one per method. So the calculation above can include the same feature multiple times and thus nselect could be greater than the number of features.

### DIFF
--- a/R/FilterEnsemble.R
+++ b/R/FilterEnsemble.R
@@ -272,9 +272,9 @@ calcBaseFilters = function(task, method = method,
     nselect = nselect, more.args = more.args)
 
   # rank base filters by method
-  fval.all.calced.simple = transform(fval.calc$data,
-    rank = ave(1:nrow(fval.calc$data), method,
-      FUN = function(x) order(fval.calc$data$value[x])))
+  fval.all.calced.simple = transform(fval.calc$data, 
+                  rank = ave(value, method, 
+                            FUN = function(x) rank(x, ties.method = "first")))
 
   fval.all.calced.simple = fval.all.calced.simple[with(fval.all.calced.simple,
     order(value, rank)), ]
--- a/R/filterFeatures.R
+++ b/R/filterFeatures.R
@@ -150,9 +150,7 @@ filterFeatures = function(task, method = "randomForestSRC_importance", fval = NU
     # Set the the filter values of the mandatory features to infinity to always select them
     fval[fval$name %in% mandatory.feat, "value"] = Inf
   }
-  if (select == "threshold") {
-    nselect = sum(fval[["value"]] >= threshold, na.rm = TRUE)
-  }
+
   # in case multiple filters have been calculated, choose which ranking is used
   # for the final subsetting
   if (length(levels(as.factor(fval$method))) >= 2) {
@@ -163,12 +161,17 @@ filterFeatures = function(task, method = "randomForestSRC_importance", fval = NU
       stopf("You supplied multiple filters. Please choose which should be used for the final subsetting of the features.")
     }
     if (is.null(select.method)) {
-      fval = fval[fval$method == fval$method, ]
+      fval = fval[fval$method == method[[1]], ]
     } else {
       assertSubset(select.method, choices = unique(fval$method))
       fval = fval[fval$method == select.method, ]
     }
   }
+  
+  if (select == "threshold") {
+    nselect = sum(fval[["value"]] >= threshold, na.rm = TRUE)
+  }
+
   if (nselect > 0L) {
 
     # order by method and (desc(value))
--- a/tests/testthat/test_base_generateFilterValuesData.R
+++ b/tests/testthat/test_base_generateFilterValuesData.R
@@ -150,3 +150,31 @@ test_that("ensemble methods work", {
   fi = generateFilterValuesData(multiclass.task, method = list("E-min", c("FSelectorRcpp_gain.ratio", "FSelectorRcpp_information.gain")))
   expect_true(all(!is.na(fi$data$value) == TRUE))
 })
+
+test_that("Ensemble filters select the top features", {
+library(survival)
+
+data(veteran)
+vet.task <- makeSurvTask(id = "VET", data = veteran, target = c("time", "status"))
+vet.task <- createDummyFeatures(vet.task)
+
+set.seed(24601)
+filt = mlr:::.FilterEnsembleRegister[["E-mean"]]
+cox.lrn <- makeLearner(cl="surv.coxph", id = "coxph", predict.type="response")
+dat = filt$fun(vet.task, 
+               base.methods = c("univariate.model.score", "randomForestSRC_importance"),
+               nselect = 5,
+               more.args = list("univariate.model.score"=list(perf.learner=cox.lrn))
+)
+ordered_dat = dat[with(dat, order(method, -value))]
+top_features1 = ordered_dat[ordered_dat$method == "E-mean"]$name[1:5]
+
+set.seed(24601)
+new.task = filterFeatures(vet.task,
+                          method="E-mean",
+                          abs = 5,
+                          base.methods = c("univariate.model.score", "randomForestSRC_importance"))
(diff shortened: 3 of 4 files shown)


pat-s:
@annette987 Thanks for checking.
Could you please post the problem as an issue first? Everything in a PR makes things messy.

I am not sure about what you are doing here

```r
set.seed(24601)
filt = mlr:::.FilterEnsembleRegister[["E-mean"]]
cox.lrn <- makeLearner(cl="surv.coxph", id = "coxph", predict.type="response")
dat = filt$fun(vet.task, 
               base.methods = c("univariate.model.score", "randomForestSRC_importance"),
               nselect = 5,
               more.args = list("univariate.model.score"=list(perf.learner=cox.lrn))
)
```

So I would wait until the issue is verified before coming up with a PR unless you have strong evidence that there is really a bug.
I wanted to release v2.16.0 today but will postpone it to inspect this one first.

NB: If you use the dev version you can get rid of the warning messages in your reprex.
Also for contributions to packages it often is helpful to be on the latest R version (just to avoid ugly conflicts) - because I see some R 3.5.3 statements.

annette987:
I have created Issue #2685, updated my description above, updated my version of R and installed the development version. I hope this satisfies all requirements and makes more sense now.

In the code above that you said you did not understand, I am simply calling the function for ensemble filter E-mean directly, to see what it should produce and later comparing that to the results produced by filterFeatures.

For future reference, if I do a PR for an enhancement, not a bug fix, should I create an issue first?

pat-s:
Thanks, I will take a detailed look in the next days. Not sure when I will find time though, my list is quite packed right now.

Actually we should have tests for this :thinking: 

> For future reference, if I do a PR for an enhancement, not a bug fix, should I create an issue first?

If it is a major one like here, yes please.
If it is only a minor one that does not need much discussion/explanation than a direct PR is fine. Otherwise the PR will not be about the proposed changes but discussing the issue itself.

Thanks for all the effort. If there is really a bug this is quite important. I use it myself in practice and like seeing other people also trying it out. :+1:

R/filterFeatures.R
```diff
@@ -150,9 +150,7 @@ filterFeatures = function(task, method = "randomForestSRC_importance", fval = NU
     # Set the the filter values of the mandatory features to infinity to always select them
     fval[fval$name %in% mandatory.feat, "value"] = Inf
   }
-  if (select == "threshold") {
-    nselect = sum(fval[["value"]] >= threshold, na.rm = TRUE)
-  }
```
pat-s: This is a separate issue, isn't it?

R/filterFeatures.R
```diff
@@ -163,12 +161,17 @@ filterFeatures = function(task, method = "randomForestSRC_importance", fval = NU
...
     } else {
       assertSubset(select.method, choices = unique(fval$method))
       fval = fval[fval$method == select.method, ]
     }
   }
+  
+  if (select == "threshold") {
+    nselect = sum(fval[["value"]] >= threshold, na.rm = TRUE)
+  }
+
```
pat-s: This is a separate issue, isn't it?

R/FilterEnsemble.R
```diff
@@ -272,9 +272,9 @@ calcBaseFilters = function(task, method = method,
     nselect = nselect, more.args = more.args)
 
   # rank base filters by method
-  fval.all.calced.simple = transform(fval.calc$data,
-    rank = ave(1:nrow(fval.calc$data), method,
-      FUN = function(x) order(fval.calc$data$value[x])))
+  fval.all.calced.simple = transform(fval.calc$data, 
+                  rank = ave(value, method, 
+                            FUN = function(x) rank(x, ties.method = "first")))
```
pat-s: I think I'll solve this using data.table as shown in https://github.com/mlr-org/mlr/issues/2685#issuecomment-561333330.

R/filterFeatures.R
```diff
@@ -163,12 +161,17 @@ filterFeatures = function(task, method = "randomForestSRC_importance", fval = NU
       stopf("You supplied multiple filters. Please choose which should be used for the final subsetting of the features.")
     }
     if (is.null(select.method)) {
-      fval = fval[fval$method == fval$method, ]
+      fval = fval[fval$method == method[[1]], ]
```
pat-s: This looks dangerous?

tests/testthat/test_base_generateFilterValuesData.R
```diff
@@ -150,3 +150,31 @@ test_that("ensemble methods work", {
...
+set.seed(24601)
+filt = mlr:::.FilterEnsembleRegister[["E-mean"]]
+cox.lrn <- makeLearner(cl="surv.coxph", id = "coxph", predict.type="response")
+dat = filt$fun(vet.task, 
+               base.methods = c("univariate.model.score", "randomForestSRC_importance"),
+               nselect = 5,
+               more.args = list("univariate.model.score"=list(perf.learner=cox.lrn))
+)
+ordered_dat = dat[with(dat, order(method, -value))]
+top_features1 = ordered_dat[ordered_dat$method == "E-mean"]$name[1:5]
```
pat-s: I am still not fully understanding what's happening here. Based on your comparison later, this will be the "true" reference. How can we be sure that the internal functions do a correct ranking only based on this code?

This code is also using 

https://github.com/mlr-org/mlr/blob/27f0ba48a11ca979e9a20dfec9b1df769e83f827/R/FilterEnsemble.R#L275-L277

at a certain stage, isn't it?

I think the only robust check we can do here is to check whether the code we are using for ranking/ordering the `rank` and `value` columns (see above) is working correctly by checking that these columns are in a asc/desc order afterwards?

R/filterFeatures.R
annette987: Perhaps. But do you agree that 
`fval = fval[fval$method == fval$method, ]`
has no effect?

R/filterFeatures.R
annette987: It could be considered a separate issue, but it is part of the problem with ensemble filters not working correctly, which is why I included it. Happy to do a separate PR if you like.

R/filterFeatures.R
annette987: As above (it is the same issue) - it could be considered a separate issue, but it is part of the problem with ensemble filters not working correctly, which is why I included it. Happy to do a separate PR if you like.

R/filterFeatures.R
annette987: Please see my comment in Issue #2685 regarding this.

R/filterFeatures.R
annette987: Please see #2685 (comment) regarding this.

tests/testthat/test_base_generateFilterValuesData.R
annette987: Yes I think you are right that this is not an adequate test. Do we need to check the output from calcBaseFilters, after ranking?

pat-s:
@annette987 

The threshold thing is a separate issue that we should tackle after the issue with the ranking/sorting.

I'll want to go step by step here to not loose overview.

Happy to take a review from your side then.
This is not an action against your approach but it is just easier for me to understand what is going wrong if I'll redo it on my own :) 

I'll close here and will focus on #2698

Issue #2688, opened 2019-11-26
Enhancement: Allow ensemble filters to include multiple base filters of the same type


annette987:
Currently, when creating an ensemble filter, the base filters must be unique. This is limiting for two reasons:
1. I cannot use both random forest minimal depth and random forest variable hunting in the same ensemble filter as they both have the same base filter -randomForestSRC_var.select
2. In this paper: http://www.cs.cornell.edu/~alexn/papers/shotgun.icml04.revised.rev2.pdf Caruana et al propose creating ensemble models from a large library of models, where there may be many models of the same base type, but with different parameters. Models are added to the ensemble one at a time using a greedy, forward step-wise selection process to add the model that improves the ensemble performance the most at each step. The same process could be applied to filter ensembles.

I propose this be achieved in the following way:

**Existing Method:**
```
ensemble = 
  makeFilterWrapper(
    makeLearner(cl="surv.coxph", id = "coxph", predict.type="response"), 
    fw.method = "E-mean",
    fw.base.methods=c("univariate.model.score", "mrmr"),
    fw.perc=0.3,
    more.args=list("univariate.model.score"=list(perf.learner=cox.lrn))
  )
```

**New Method:**
```
ensemble1 = 
  makeFilterWrapper(
    makeLearner(cl="surv.coxph", id = "coxph", predict.type="response"), 
    fw.method = "E-mean",
    fw.base.methods = list("filter1" = "univariate.model.score", "filter2"= "univariate.model.score"),
    fw.perc = 0.3,
    more.args=list("filter1"=list(perf.learner=cox.lrn), "filter2"=list(perf.learner=cox.lrn))
  )
```

1. The base methods are passed as a named list, where the names are identifiers and the values are the filter methods.
2. The arguments to the filters are passed in more.args as a named list, the same as in the current implementation, but in this case the names are the identifiers, not the filter names.
3. In generateFilterValues.R, we can no longer distinguish an ensemble method from a base method by assuming that only ensemble methods are passed as lists. So instead we check if the name of the filter exists in the FilterEnsembleRegister.

I have coded these and other changes neede and can do a PR if this idea is acceptable.

larskotthoff:
That sounds good -- please go ahead and make the PR and we'll review it.

pat-s:
fixed in #2693

PR #2693, opened 2019-11-29
Allow ensemble filters to include multiple base filters of the…

annette987:
PR for Issue #2688 

The base methods of ensemble filters currently have to be unique, which limits the types of ensembles that can be created. This PR removes that restriction by:

- Passing the base methods  as a named list, where the names are identifiers and the values are the filter methods.
- Passing the arguments to the filters in more.args as a named list, the same as in the current implementation, but in this case the names are the same identifiers, not the filter names.
- Distinguishing an ensemble method from a base method by checking if the name of the filter exists in the FilterEnsembleRegister.


R/generateFilterValues.R
```diff
@@ -55,15 +55,16 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
-    ens.method = method[[1]]
-    method = method[[2]]
-    assertSubset(ens.method, choices = ls(.FilterEnsembleRegister), empty.ok = FALSE)
-    if (length(method) == 1) {
-      warningf("You only passed one base filter method to an ensemble filter. Please use at least two base filter methods to have a voting effect.")
+    if (method[[1]] %in% ls(.FilterEnsembleRegister)) {
+	  ens.method = method[[1]]
+	  method = method[[2]]
+	  if (length(method) == 1) {
+		warningf("You only passed one base filter method to an ensemble filter. Please use at least two base filter methods to have a voting effect.")
```
larskotthoff: Level of indentation missing here.

R/generateFilterValues.R
```diff
@@ -125,21 +125,26 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
-    })
+    index_names = names(method)
+	if (is.null(index_names))
+		index_names = method
+	fval = mapply(function(x, name) {
+			x = do.call(x$fun, c(list(task = task, nselect = nselect), more.args[[name]]))
+			missing.score = setdiff(fn, names(x))
+			x[missing.score] = NA_real_
+			x[match(fn, names(x))]
+    }, filter, index_names, SIMPLIFY=FALSE)
```
larskotthoff: Space before and after =.

tests/testthat/test_featsel_filters.R
```diff
@@ -72,3 +72,16 @@ test_that("randomForestSRC_var.select filter handles user choices correctly", {
...
+
+test_that("Ensemble filters can accept duplicate base methods", {
+  lda = makeLearner(cl = "classif.lda", id = "lda_class", predict.type = "response")
+  ff = filterFeatures(multiclass.task, 
+                      method = "E-mean", 
+                      base.methods = list("filter1" = "univariate.model.score", "filter2" = "univariate.model.score"),
+                      abs = 2,
+                      more.args = list("filter1" = list(perf.learner = lda), "filter2" = list(perf.learner = lda))
+       )
+  expect_equal(getTaskFeatureNames(ff), c("Petal.Length", "Petal.Width"))
```
larskotthoff: Could you also please check that the returned data frame has the right column names, i.e. the filter names are reflected correctly in the output? Ideally with different methods that return different features so it's obvious which one is which.

annette987: Just to clarify, when you say "the returned data frame", which data frame are you referring to and which function is returning it? Thanks.

R/generateFilterValues.R
annette987: Yes I will fix that.

R/generateFilterValues.R
annette987: Yes I will fix that.

tests/testthat/test_featsel_filters.R
larskotthoff: Sorry, ignore that -- I thought it returned something different.

pat-s:
I'll come to this after #2698 is done.

NEWS.md
```diff
@@ -19,6 +19,7 @@
 ## misc
 
 - Add `configureMlr()` to `.onLoad()`, possibly fixing some edge cases (#2585) (@pat-s, #2637)
+- Allow ensemble filters to include multiple base filters of the same type (#2688)
```
pat-s: Please merge the latest master and put the item below `## filters - general`.

R/generateFilterValues.R
```diff
@@ -55,15 +55,16 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
 
   # ensemble
   if (class(method) == "list") {
-    ens.method = method[[1]]
-    method = method[[2]]
-    assertSubset(ens.method, choices = ls(.FilterEnsembleRegister), empty.ok = FALSE)
```
pat-s: Isn't this assertion still valid? We only want to have valid Ensemble filters as `ens.method`?

R/generateFilterValues.R
```diff
@@ -55,15 +55,16 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
+			ens.method = method[[1]]
+			method = method[[2]]
+			if (length(method) == 1) {
+				warningf("You only passed one base filter method to an ensemble filter. Please use at least two base filter methods to have a voting effect.")
+			}
     }
   }
 
-  assertSubset(method, choices = append(ls(.FilterRegister), ls(.FilterEnsembleRegister)), empty.ok = FALSE)
+  lapply(method, assertSubset, choices = append(ls(.FilterRegister), ls(.FilterEnsembleRegister)), empty.ok = FALSE)
```
pat-s: would it be better to unlist and then take the old approach rather than using `lapply()` here?

Also doesn't `lapply()` trigger console output here since the result is not assigned to an object?

Also note the superfluous white-spaces at the end.

R/generateFilterValues.R
```diff
@@ -125,21 +125,26 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
   } else {
-    fval = lapply(filter, function(x) {
-      x = do.call(x$fun, c(list(task = task, nselect = nselect), more.args[[x$name]]))
-      missing.score = setdiff(fn, names(x))
-      x[missing.score] = NA_real_
-      x[match(fn, names(x))]
-    })
+    index_names = names(method)
+	if (is.null(index_names))
+		index_names = method
```
pat-s: Do we want to allow this in the first place? Or assert this and stop early?

R/generateFilterValues.R
```diff
@@ -125,21 +125,26 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
+			x = do.call(x$fun, c(list(task = task, nselect = nselect), more.args[[name]]))
+			missing.score = setdiff(fn, names(x))
+			x[missing.score] = NA_real_
+			x[match(fn, names(x))]
+    }, filter, index_names, SIMPLIFY = FALSE)    
     fval = do.call(cbind, fval)
     colnames(fval) = method
-    types = vcapply(getTaskData(task, target.extra = TRUE)$data[fn], getClass1)
+	if (!is.null(names(method)))
+		colnames(fval) = names(method)
```
pat-s: Would this be obsolete if we would require named inputs right from the start?

R/generateFilterValues.R
```diff
@@ -125,21 +125,26 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
+	types = vcapply(getTaskData(task, target.extra = TRUE)$data[fn], getClass1)
 
     out = data.table(name = row.names(fval),
       type = types, fval, row.names = NULL, stringsAsFactors = FALSE)
 
     # variable.factor = FALSE has no effect
-    out = melt(out, value.name = "value", measure.vars = method,
+    out = melt(out, value.name = "value", measure.vars = colnames(fval),
       variable.name = "method")
```
pat-s: FYI: This will trigger a merge conflict after #2698 has been merged

tests/testthat/test_featsel_filters.R
```diff
@@ -72,3 +72,16 @@ test_that("randomForestSRC_var.select filter handles user choices correctly", {
     more.args = list("randomForestSRC_var.select" = c(method = "vh.imp"))))
   )
 })
+
+
+test_that("Ensemble filters can accept duplicate base methods", {
```
pat-s:
```suggestion
test_that("Ensemble filters can deal with non-unique base methods ", {
```

tests/testthat/test_featsel_filters.R
```diff
@@ -72,3 +72,16 @@ test_that("randomForestSRC_var.select filter handles user choices correctly", {
...
+  lda = makeLearner(cl = "classif.lda", id = "lda_class", predict.type = "response")
+  ff = filterFeatures(multiclass.task, 
+                      method = "E-mean", 
+                      base.methods = list("filter1" = "univariate.model.score", "filter2" = "univariate.model.score"),
+                      abs = 2,
+                      more.args = list("filter1" = list(perf.learner = lda), "filter2" = list(perf.learner = lda))
```
pat-s: Should we also assert that `more.args` is a named list?

NEWS.md
annette987: OK.

R/generateFilterValues.R
annette987: The assertion is still valid but no longer needed because, in the code below,  we test to see if method[[1]] is in the FilterEnsembleRegister, to determine if this is an ensemble filter, then set ens.method to method[[1]]. Therefore ens.method is guaranteed to be in the FilterEnsembleRegister. 
```
  if (class(method) == "list") {
    if (method[[1]] %in% ls(.FilterEnsembleRegister)) {
	ens.method = method[[1]]
	method = method[[2]]
	if (length(method) == 1) {
	  warningf("You only passed one base filter method to an ensemble filter. Please use at least two base filter methods to have a voting effect.")
       }
    }
  }

```

R/generateFilterValues.R
annette987: This is for backwards compatibility. You can still pass a vector of filters in argument method: 
```
#' @examples
#' # two simple filter methods
#' fval = generateFilterValuesData(iris.task,
#'   method = c("FSelectorRcpp_gain.ratio", "FSelectorRcpp_information.gain"))
#' # using ensemble method "E-mean"
#' fval = generateFilterValuesData(iris.task,
#'   method = list("E-mean", c("FSelectorRcpp_gain.ratio", "FSelectorRcpp_information.gain")))
```
In this case names(method) will be NULL and so we want the column names of fval (prior to melting) to be the names of the filters that were passed in.

R/generateFilterValues.R
annette987: Yes, but that would break a lot of code. I am trying to make it backwards compatible.

R/generateFilterValues.R
annette987: Yes it does.

annette987: Yes it does. What can I do about this?

R/generateFilterValues.R
annette987: OK. 
Note the call to lapply a few lines below that, which is not my code. Won't that trigger console output as well?
```
  if (!(any(sapply(filter, function(x) !isScalarNA(filter$pkg))))) {
    lapply(filter, function(x) requirePackages(x$pkg, why = "generateFilterValuesData", default.method = "load"))
```

tests/testthat/test_featsel_filters.R
annette987: If you like, but I think "accept duplicate" is clearer and more succinct.

tests/testthat/test_featsel_filters.R
annette987: There is already such an assertion in generateFilterValues.R, line 94:
`  assertList(more.args, names = "unique", max.len = length(method))
`

R/generateFilterValues.R
pat-s: Ok. Can we replace the if clause with the proper assertion? 

https://github.com/mlr-org/mlr/blob/2a10e5a9a4a8a1ad3bff3501b05e07d423c76c21/R/generateFilterValues.R#L58

NEWS.md
pat-s: This should go below `## filters - general` not below `##misc`.

R/generateFilterValues.R
pat-s: Looks like it, yes. Feel free to also fix it, maybe even just by assigning the outcome to an object. We have no side-effect helper funs such as `walk` in _mlr_ unfortunately.

R/generateFilterValues.R
pat-s: Ah I see, and because we are now allowing user-defined names via named lists, we need to check if `index_names` are NULL in fact.

(PS: Please try to format your code blocks as R code and also use inline-code highlighting. This makes comments easier to read. Thanks.)

R/generateFilterValues.R
pat-s: Ok fine, same as the comment before then.

R/generateFilterValues.R
pat-s: Nothing, you cannot prevent a merge conflict ;) I just stated this because I was changing some code in this part in another PR, so I knew the conflict would appear.

tests/testthat/test_featsel_filters.R
pat-s: "duplicate" indicates that is indeed a full duplicate - which it is not, because the params are different. 

"non-unique base methods" indicates that some names/methods might not be unique in the selection but not necessarily a duplicate (due to different params).

tests/testthat/test_featsel_filters.R
pat-s: Just FYI: You can cross-link code by clicking on the line number of a file on Github and then select "copy permalink". This way the code line are embedded and you do not need to c/p :)

NEWS.md
annette987: That's weird. It is under filters - general in my version.

tests/testthat/test_featsel_filters.R
annette987: That's a good point. I will change it.

tests/testthat/test_featsel_filters.R
annette987: Thanks for the tip.

R/generateFilterValues.R
annette987: I don't understand. Do you want:
 if (assertSubset(ens.method, choices = ls(.FilterEnsembleRegister), empty.ok = FALSE)) {

R/generateFilterValues.R
annette987: OK done.

R/generateFilterValues.R
pat-s: Replacing

https://github.com/annette987/mlr/blob/541300f7b086a21d4a3787429f4a0359f8a98c34/R/generateFilterValues.R#L58

by 

`assertSubset(method[[1]], choices = ls(.FilterEnsembleRegister), empty.ok = FALSE)`

Then we simply throw an assertion error now if method[[1]] is not an ensemble method. We could be more specific with the error message here but this should suffice for now.

annette987: No, that line needs to be an if statement. Previously, the base methods had to be a character vector of unique filter names. Now, we are allowing the base methods to be a named list. That meant having to delete this block of code:  
https://github.com/annette987/mlr/blob/e7ec04d99a857bf77b555ab03d2dcecadfc9ccc0/R/filterFeatures.R#L69-L73 
Note that it says "base.methods arrive here in a list when called from 'tuneParams'".
Therefore, we can no longer assume that the only methods passed as a list are ensemble methods. So we first check to see if method is of class list, and if it is we still need to check whether or not method[[1]] is an ensemble method. It may not be. If it is we do what we need to do for ensemble methods and if it is not we don't need to do anything. We can't simply make an assertion there.

pat-s: > we still need to check whether or not method[[1]] is an ensemble method. It may not be. If it is we do what we need to do for ensemble methods and if it is not we don't need to do anything. 

Now I see. With an assertion we would error which is not what we want in this case.
Thanks for resolving this in such detail!

tests/testthat/test_featsel_filters.R
```diff
@@ -72,3 +72,16 @@ test_that("randomForestSRC_var.select filter handles user choices correctly", {
     more.args = list("randomForestSRC_var.select" = c(method = "vh.imp"))))
   )
 })
+
+
```
pat-s: this empty line is not needed

tests/testthat/test_featsel_filters.R
```diff
@@ -72,3 +72,16 @@ test_that("randomForestSRC_var.select filter handles user choices correctly", {
...
+test_that("Ensemble filters can deal with non-unique base methods", {
+  lda = makeLearner(cl = "classif.lda", id = "lda_class", predict.type = "response")
+  ff = filterFeatures(multiclass.task, 
+                      method = "E-mean", 
+                      base.methods = list("filter1" = "univariate.model.score", "filter2" = "univariate.model.score"),
+                      abs = 2,
+                      more.args = list("filter1" = list(perf.learner = lda), "filter2" = list(perf.learner = lda))
```
pat-s: Here the "mlr-style" is still missing

tests/testthat/test_featsel_filters.R
annette987: Wow! Thank you so much. I have added my name to the list of contributors.

tests/testthat/test_featsel_filters.R
annette987: OK done.

annette987:
Sorry, I am really not sure what is going on here. I resolved the conflicts, did and git pull and a git push and the tests failed. I tried to fix things by doing a reset and starting over, but I may have made things worse. I am not sure what to do now. Can you advise please. Thanks.

DESCRIPTION
```diff
@@ -121,7 +121,11 @@ Authors@R:
       person(given = "Christoph",
              family = "Molnar",
              role = "ctb",
-             email = "christoph.molnar@stat.uni-muenchen.de"))
+             email = "christoph.molnar@stat.uni-muenchen.de"),
+			person(given = "Annette",
```
pat-s: Just a general comment: It is preferred to use whitespace instead of tabs in most coding areas. You're still often using tabs. I'll clean this myself here. 

But please really make sure to stick with the alignment rules and "whitespace only" in mlr in all cases. This prevents me from cleaning up and commenting on it each time :)

R/generateFilterValues.R
```diff
@@ -125,22 +125,29 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
     types = vcapply(getTaskData(task, target.extra = TRUE)$data[fn], getClass1)
 
     out = data.table(name = row.names(fval),
       type = types, fval, row.names = NULL, stringsAsFactors = FALSE)
 
     # variable.factor = FALSE has no effect
-    out = melt(out, value.name = "value", measure.vars = method,
-      variable.name = "filter")
+    out = melt(out, value.name = "value", measure.vars = colnames(fval),
+      variable.name = "method")
```
pat-s: @annette987 This change caused the test failures. It probably slipped while solving the merge conflicts in master.

DESCRIPTION
annette987: Sorry! Thanks.

R/generateFilterValues.R
annette987: Actually that change was deliberate and was there before the merge. The reason for it is that method can now have duplicate filter names, which will cause the melt to fail. So I used colnames(fval) because that is set to names(method) if not null and method otherwise. In fact I could have used index_names, which is set to teh same thing.

pat-s: It could be that this change is still needed then. However, in one of the latest PRs there was the change to `variable.name = "filter"` which definitely needs to come into this PR here.

(The column name listing all the methods used is no longer called "method" but "filter" now to avoid name clashes in other places.)

If tests fail you should then know why and be able to solve the issue quickly :)

annette987: All OK now. It was just the change to variable.name = "filter" as you suggested. I think the test I already added should be sufficient: https://github.com/annette987/mlr/blob/21c3b8b768baa0b1d7fcad5faf1bb38d157d7fbc/tests/testthat/test_featsel_filters.R#L149
Thanks for your patience with a Github newbie! :-)

pat-s: Thanks for your patience - we know how hard it is to make PRs to packages to comply with everything and especially in mlr 😄 

It seems that you need to update `man/` again, just run `devtools::document()´ or press CTRL+Shift+D in case you are using RStudio.

annette987: :relaxed:
I ran devtools::document() and it has changed a huge number of files, not just the two with conflicts. Is that OK to push them all?

pat-s: I'd assume that you are not running the latest version of _roxygen_?
When developing pkgs, it is important to always have the most recent versions of pkgs - we usually update every few days :)

pat-s:
Thanks @annette987 for all your contributions!

R/generateFilterValues.R
annette987: Oh OK. I will do that from now on. Thanks.

annette987:
My pleasure. I have learnt a lot along the way. Thanks for all your help.
Do I need to click the 'Update branch' button, or should I leave that to you? How do we finalise this?

PR #2696, opened 2019-12-03
Make `n.show` arg work in `plotFilterValues()`

### LINKED ISSUE #2689: n.show argument  in `plotFilterValues` ignored
I'm working on a data set with several features and just noticed `plotFilterValues` is ignoring `n.show` argument. Code inspected, couldn't find how it should work.

I ran something like
`importance = generateFilterValuesData(train.task, method = "FSelectorRcpp_information.gain")
plotFilterValues(importance, n.show = 10)`
and got all features, instead of 10.

Changed line 188 on the file for
`  data = fvalues$data %>%
    arrange(desc(value)) %>%
    slice(1:n.show)`
and it seems to do the trick.

Could you guys check it?

Edit by @pat-s: Reprex added

``` r
library(mlr)

importance = generateFilterValuesData(spam.task, method = "FSelectorRcpp_information.gain")
plotFilterValues(importance, n.show = 10)
```

<sup>Created on 2019-12-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #2697: Test "plotFilterValues" should be optimized
https://github.com/mlr-org/mlr/blob/eb14d977f6e0e88dbdc14dcd09e0c5df614736b6/tests/testthat/test_base_generateFilterValuesData.R#L62-L89

- No need to use "permutation importance" as the filter
- "randomForestSRC_importance" also is quite slow and causes problems if called iteratively
- should use the _vdiffr_ approach for comparing visual outputs

pat-s:
fixes #2689 including some style adjustments
fixes #2697 

``` r
library(mlr)

fv = generateFilterValuesData(spam.task, 
                              method = "FSelectorRcpp_information.gain")
plotFilterValues(fv, n.show = 10, sort = "dec")
```

``` r

plotFilterValues(fv, n.show = 10, sort = "inc")
```

<sup>Created on 2019-12-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/generateFilterValues.R
```diff
@@ -177,15 +180,17 @@ print.FilterValues = function(x, ...) {
...
 
   assertClass(fvalues, classes = "FilterValues")
   assertChoice(sort, choices = c("dec", "inc", "none"))
+
   if (!(is.null(fvalues$method))) {
     stop("fvalues must be generated by generateFilterValuesData, not getFilterValues, which is deprecated.")
   }
 
-  data = fvalues$data
+  data = fvalues$data[1:n.show, ]
```
larskotthoff: Shouldn't this happen after sorting?

larskotthoff: Sorting and heading could also happen in one step here.

pat-s:
Ok, diving into `plotFilterValues()` really brought up a lot of issues:

- the "facet" arguments inside the function were not exposed in the constructor
- `n.show` was not used within the function

Facetting is complex on its own but what is really tricky is to get an orderding within facets.
After realizing there is no canonical way but just a lot of hacky workarounds, I decided to clean up and drop this one completely. 
I was not even exposed yet to the user so this shouldn't make any problems.

Also, since different filters might come up with different features, features across facets will be empty and the resulting plots will look really chaotic.

Users who want this functionality should just create single plots of each method and then combine those in a grid manually.

## Additional changes

- Currently we've been running `generateFilterValuesData()` for all methods in the test for `plotFilterValues()`. This is redundant and there is no need to check this.
We now check for the correct function execution including some optional arguments using the canonical way to check plotting via the _vdiffr_ package.

- The title of the plots now contains a "X out of Y" depending on the `n.show` argument. If one supplies `n.show` > nfeat, it is automatically recycled.

- I moved the tests for `plotFilterValues()` to its own file

- tests for generateFilterValuesData() now live in the _featsel_ group rather than in the _base_ group

larskotthoff:
Thanks, this looks good, but I wonder whether a better default would be to show all features. This might lead to some head-scratching otherwise.

pat-s:
I agree, let's default to all.

R/generateFilterValues.R
```diff
@@ -161,61 +161,72 @@ print.FilterValues = function(x, ...) {
...
+#'   Whether to color different feature types (e.g. numeric | factor).
+#'   Default is to use no colors (`feat.type.cols = FALSE`).
 #' @template arg_facet_nrow_ncol
 #' @template ret_gg2
 #' @export
 #' @examples
 #' fv = generateFilterValuesData(iris.task, method = "variance")
 #' plotFilterValues(fv)
-plotFilterValues = function(fvalues, sort = "dec", n.show = 20L, feat.type.cols = FALSE, facet.wrap.nrow = NULL, facet.wrap.ncol = NULL) {
+plotFilterValues = function(fvalues, sort = "dec", n.show = nrow(fv$data),
```
larskotthoff: Don't think that this will work -- I would set a dummy value (`NULL`) as default and then check that inside the function to get the number of features in the data.

pat-s: It works, I just had a typo in the arg. There is no problem with setting a default arg dependent on another arg of a function.

I don't think setting the default to `NULL` has any advantage here: We are not doing any if-branching on this arg down in the function and just use it as an integer value for subsetting.
In general setting args to `NULL` by default is the preferred way in many cases but there are cases like this where deviating from this default makes sense (imo).

larskotthoff: Ok thanks.

PR #2701, opened 2019-12-06
Update tree_method values for XGBoost learners

albersonmiranda:
Added 'auto', 'approx' and 'gpu_hist' to `tree_method` param for XGBoost learners.

According to https://xgboost.readthedocs.io/en/latest/parameter.html

Is that ok, @pat-s ?


larskotthoff:
Does this require a specific version of xgboost? Would be great if you could update that in DESCRIPTION.

albersonmiranda:
@larskotthoff sorry about that. Indeed it was added in XGBoost v0.7 back in december/2017. Updated.

pat-s:
Thanks!

PR #2702, opened 2019-12-07
Move Paramhelpers from Depends to Imports

pat-s:
To avoid the autoloading of ParamHelpers when calling `library(mlr)`.


jakob-r:
Okay. But then users have to load ParamHelpers manually as soon as they want to define ParamSets etc. So this would break a lot of old code once people get this update. Also this would have to be made clear in the tutorial etc.
I don't think this is a good idea. Why is it even necessary?

jakob-r:
@berndbischl @mllg Thoughts?

berndbischl:
i dont know. i have the feeling we are changing way too much for "maintenance mode"?
@mllg ?

pat-s:
My thoughts to this:

- One can do a lot in _mlr_ without _ParamHelpers_
- In mlr3 we also have to load _paradox_ externally now
- We get rid of the "loading required package" message
- People would become more aware for what ParamHelpers is actually there for and would load it on purpose (I've come across a bunch of cases where people asked me what this message actually says or what this package does)

Downside: This can be seen as a breaking change to older scripts since one would need to add `library(ParamHelpers)` now in scripts before constructing a ParamSet

jakob-r:
All the arguments seem to be minor inconveniences in comparison to the annoyance it will create for many users. I think about batch(tool) scripts that suddenly fail, stackoverflow answers that have to be edited, tutorial pages that have to be edited. Of course it is an easy fix if the people are aware of ParamHelpes. But you specifically say that this PR is targeted towards beginners or people that are confused by loading messages. Imagine what those kind of people will think if they suddenly see `ParamSet function not found`
So I very much vote for _no merge_.

pat-s:
OK I agree, this will cause more confusion than it helps.
Thanks @jakob-r for bringing this up.

berndbischl:
Agreeing with Jakob

jakob-r:
:raised_hands:

PR #2712, opened 2020-01-09
Enable passing filter params via more.args w/o filter name

### LINKED ISSUE #2709: Can't pass additional arguments via more.args to simple filter
After recent changes to filters (#2693) it is no longer possible to pass additional arguments to a simple filter via more.args:
``` r
library(survival)
library(mlr)
#> 'mlr' is in maintenance mode since July 2019. Future development
#> efforts will go into its successor 'mlr3' (<https://mlr3.mlr-org.com>).

set.seed(24601)
data(veteran)
mas.task <- makeSurvTask(id = "VET", data = veteran, target = c("time", "status"))
outer = makeResampleDesc("CV", iters = 2, stratify = TRUE)

cox.filt.mindepth.lrn = makeFilterWrapper(
    makeLearner(cl = "surv.coxph", id = "cox.filt.mindepth", predict.type = "response"), 
    fw.method = "randomForestSRC_var.select",
    fw.abs = 5,
    more.args = list("metho" = "md", ntree=100, nsplit = 10, nodesize = 3)
  )

learners = list(cox.filt.mindepth.lrn)   
bmr = benchmark(learners, mas.task, outer, show.info = TRUE)
#> Task: VET, Learner: cox.filt.mindepth.filtered
#> Error in generate.fv.data(task = task, method = method, nselect = getTaskNFeats(task), : Assertion on 'more.args' failed: Must have length <= 1, but has length 4.
```

<sup>Created on 2020-01-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I can submit a PR to fix this.

annette987:
Fixes #2709 

When using a filter via makeFilterWrapper() or generateFilterValuesData(), additional parameters can be passed down to the filter using the argument more.args. If more than one filter is passed to these functions, or an ensemble filter is used,  more.args must be a named list, using the filter name to determine which filter the parameters belong to. For example:
```
generateFilterValuesData(task = iris.task,
                         method = c("randomForestSRC_var.select", "FSelectorRcpp_information.gain"),
                         more.args = list("randomForestSRC_var.select" = list("metho" = "md", ntree=100, nsplit = 10, nodesize = 3) ,
                                          "FSelectorRcpp_information.gain" = list(equal = TRUE)))
```
But if only one simple filter is passed, the filter name need not be included in more.args. For example:
```
generateFilterValuesData(task = iris.task,
       method = "randomForestSRC_var.select",
       more.args = list("metho" = "md", ntree=100, nsplit = 10, nodesize = 3) )
```
However, the latter method, without the filter name included, currently does not work. To overcome this issue, more.args is converted to a named list of the correct format for this particular case. This will maintain backwards compatibility.

Rather than adding an additional test, the test for this PR uses an exiting test but removes the filter name from the parameter list, as it was not necessary and this case had not been tested.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 - Fixed an issue which caused the random forest minimal depth filter to only return NA values when using thresholding. 
   NAs should only be returned for features below the given threshold. (@annette987, #2710)
-- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
+- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annette987, #2709)
   
 # mlr 2.17.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,7 @@ PR: #2638 (@pfistl)
 
 - `filterFeatures()`: Arg `thresh` was not working correctly when applied to ensemble filters. (@annette987, #2699)
 - Fixed incorrect ranking of ensemble filters. Thanks @annette987 (#2698)
+- Fixed problem of not being able to pass arguments using more.args to a simple filter (@annete987, #2709)
 
 # mlr 2.16.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 
 - Fixed an issue which caused the random forest minimal depth filter to only return NA values when using thresholding. 
   NAs should only be returned for features below the given threshold. (@annette987, #2710)
+- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
   
 # mlr 2.17.0
 
@@ -46,7 +47,6 @@ PR: #2638 (@pfistl)
 
 - `filterFeatures()`: Arg `thresh` was not working correctly when applied to ensemble filters. (@annette987, #2699)
 - Fixed incorrect ranking of ensemble filters. Thanks @annette987 (#2698)
-- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
 
 # mlr 2.16.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,7 +39,7 @@ PR: #2638 (@pfistl)
 
 - `filterFeatures()`: Arg `thresh` was not working correctly when applied to ensemble filters. (@annette987, #2699)
 - Fixed incorrect ranking of ensemble filters. Thanks @annette987 (#2698)
-- Fixed problem of not being able to pass arguments using more.args to a simple filter (@annete987, #2709)
+- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
 
 # mlr 2.16.0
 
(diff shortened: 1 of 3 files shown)


R/generateFilterValues.R
```diff
@@ -110,7 +109,11 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
     } else {
       stopf("You use more than 1 filter method. Please pass extra arguments via 'more.args' and not '...' to filter methods!")
     }
+  } else if (length(method) == 1L && length(more.args) > 1L) { # simple filter with 1 method and more.args
```
larskotthoff: What if I want to set only parameters that are not `method`?

annette987: That won't be a problem. "method" here is the argument to generateFilterValuesData, specifying the type of filter to use, not the method argument to a specific filter like randomForestSRC_var.select. So this test is just checking to see if a simple filter has been used (as opposed to an ensemble filter), with arguments passed inline, not as part of a named list, for example:
`fv = generateFilterValuesData(iris.task, method = "randomForestSRC_var.select", nselect = 2, metho="md", nrep=5)`
If that is the case then the arguments are formatted as a named list, named by the filter method.
I hope that makes sense. Am I understanding your concerns correctly?

NEWS.md
```diff
@@ -39,6 +39,7 @@ PR: #2638 (@pfistl)
 
 - `filterFeatures()`: Arg `thresh` was not working correctly when applied to ensemble filters. (@annette987, #2699)
 - Fixed incorrect ranking of ensemble filters. Thanks @annette987 (#2698)
+- Fixed problem of not being able to pass arguments using more.args to a simple filter (@annete987, #2709)
```
pat-s:
```suggestion
- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
```

Please move the entry to the section for the dev version :)

R/generateFilterValues.R
```diff
@@ -110,7 +109,11 @@ generateFilterValuesData = function(task, method = "randomForestSRC_importance",
...
+    more.args = namedList(method, more.args)
   }
+  assertList(more.args, names = "unique", max.len = length(method))
+
```
pat-s: Please remove the unnecessary empty line.

R/generateFilterValues.R
larskotthoff: Ok thanks.

R/generateFilterValues.R
annette987: OK.

pat-s:
@annette987 Please update NEWS.md and resolve the conflicts. Thanks.

NEWS.md
```diff
@@ -4,6 +4,7 @@
 
 - Fixed an issue which caused the random forest minimal depth filter to only return NA values when using thresholding. 
   NAs should only be returned for features below the given threshold. (@annette987, #2710)
+- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annete987, #2709)
```
pat-s:
```suggestion
- Fixed problem which prevented passing filter options via argument `more.args` for simple filters (@annette987, #2709)
```

PR #2713, opened 2020-01-09
RF md filter: Do not return only NA values when thresholding

### LINKED ISSUE #2710: Random Forest minimal depth filter returns all NA's
The random forest minimal depth filter is returning all NA values. This is because the vector of features returned by the filter is not a named vector and generateFilterValuesData expects a named vector. I can submit a PR to fix this and it will explain the solution.

``` r
library(survival)
library(mlr)
#> 'mlr' is in maintenance mode since July 2019. Future development
#> efforts will go into its successor 'mlr3' (<https://mlr3.mlr-org.com>).

set.seed(24601)
data(veteran)
mas.task <- makeSurvTask(id = "VET", data = veteran, target = c("time", "status"))
outer = makeResampleDesc("CV", iters = 2, stratify = TRUE)

cox.filt.mindepth.lrn = makeFilterWrapper(
    makeLearner(cl = "surv.coxph", id = "cox.filt.mindepth", predict.type = "response"), 
    fw.method = "randomForestSRC_var.select",
    fw.abs = 5,
    more.args = list("randomForestSRC_var.select" = list("metho" = "md", ntree=100, nsplit = 10, nodesize = 3))
  )

learners = list(cox.filt.mindepth.lrn)   
bmr = benchmark(learners, mas.task, outer, show.info = TRUE)
#> Task: VET, Learner: cox.filt.mindepth.filtered
#> Error in (function (task, method = "randomForestSRC_importance", fval = NULL, : Filter method returned all NA values!
```

<sup>Created on 2020-01-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

annette987:
Fixes #2710 

In #2690 the random forest minimal depth filter was modified so that it returns a value of NA for any features below the threshold calculated by the filter. It was actually returning NA for all features, but the tests did not pick that up. This PR fixes that.

The reason for this bug is that the minimal depth filter currently does not return a named vector. Then in generateFilterValuesData, in this block of code: https://github.com/mlr-org/mlr/blob/e12475364a8412e94b19b99d329ed98b5d174c8f/R/generateFilterValues.R#L137-L142
in line 139, names(x) is NULL, so missing.score is set to the entire feature set and then the value of each feature in the feature set is set to NA in line 140.

Simply returning a named vector from the filter fixes this problem. A new test is added to check that all values returned are not NA.


tests/testthat/test_featsel_filters.R
```diff
@@ -109,7 +109,8 @@ test_that("randomForestSRC_var.select minimal depth filter returns NA for featur
     method = "randomForestSRC_var.select",
     nselect = 5,
     more.args = list("randomForestSRC_var.select" = list(method = "md", nrep = 5)))
-  expect_equal(is.na(dat$data$value[dat$data$name %in% c("Sepal.Length", "Sepal.width")]), TRUE)
+  expect_equal(all(is.na(dat$data$value[dat$data$name %in% c("Sepal.Length", "Sepal.Width")])), TRUE)
+  expect_equal(all(is.na(dat$data$value)), FALSE)
```
larskotthoff: Could you check explicitly for the ones that should not be NA please?

annette987: OK. Done.

NEWS.md
```diff
@@ -39,6 +39,7 @@ PR: #2638 (@pfistl)
 
 - `filterFeatures()`: Arg `thresh` was not working correctly when applied to ensemble filters. (@annette987, #2699)
 - Fixed incorrect ranking of ensemble filters. Thanks @annette987 (#2698)
+- Fixed problem of random forest minimal depth filter returning all NA values. Should only return NAs for features below the threshold. (@annette987, #2710)
```
pat-s:
```suggestion
- Fixed an issue which caused the random forest minimal depth filter to only return NA values when using thresholding. 
  NAs should only be returned for features below the given threshold. (@annette987, #2710)
```

Please move this point to the section for the current dev version.

annette987:
Hahaha. Sorry! I'll get it right next time! :-)

Issue #2716, opened 2020-01-12
Problem with adding fregre.glm learner from fda. usc package


NvNorsinski:
Hello,
I am trying to add the learner fregre.glm from package fda.usc. For the integration I followed the code for the already implemented and very similar integration of classif.glm from the same package in RLearner_classif_fdausc.glm.R. But I am having troubles when it comes to using the implementation, because I get an error in the for the trainLearner function, which seems to have an error. In the following you can see the code for the integration. Maybe I am doing a beginner's :) mistake.
``` r
library(mlr)
library(fda.usc)
library(parallelMap)

makeRLearner.regr.fregre.glm = function() {
  makeRLearnerRegr(
    cl = "regr.fregre.glm",
    package = "fda.usc",
    par.set = makeParamSet(
      makeDiscreteLearnerParam(id = "family", default = "binomial()", values = list("binomial()", "gaussian()", "Gamma()", "inverse.gaussian()", "poisson()")),
      makeUntypedLearnerParam(id = "basis.x"),
      makeUntypedLearnerParam(id = "basis.b"),
      makeLogicalLearnerParam(id = "CV", default = FALSE)
    ),
    properties = c("functionals"),
    name = "Generalized Linear Models regression on FDA",
    short.name = "fregre.glm",
    note = "model$C[[1]] is set to quote(fregre.glm)"
  )
}

trainLearner.regr.fregre.glm = function(.learner, .task, .subset, .weights = NULL, ...) {

  # Get and transform functional data
  d = getTaskData(.task, subset = .subset, target.extra = TRUE, functionals.as = "matrix")
  fd = getFunctionalFeatures(d$data)
  # transform the data into fda.usc:fdata class type.
  data.fdclass = fda.usc::fdata(mdata = as.matrix(fd))
  # transform the data into fda.usc:fdata class type and save in a list
  dat = list(df = data.frame(d$target), x = data.fdclass)

  model = fda.usc::fregre.glm(d.target ~ x, data = dat)

  # Fix bug in package. The changed slot looks different when called with
  # `fda.usc::lassif.glm()` than just `classif.glm()`
  model$call[[1]] = quote(fregre.glm)

  return(model)
}

predictLearner.regr.fregre.glm = function(.learner, .model, .newdata, ...) {
  # transform the data into fda.usc:fdata class type.
  fd = getFunctionalFeatures(.newdata)
  nd = list(x = fda.usc::fdata(mdata = fd))

[...]

PR #2718, opened 2020-01-13
Add fregre learner and test file

NvNorsinski:
Here I added the fregre.glm learner from package fda.usc and the corresponding test script.


pat-s:
#2716 

@pfistfl Appreciated if you could take a quite look if you find some time.

pfistfl:
Hi, 

some thoughts on your problem: 
Do you aim to add this is to **mlr** or just as a script that you source before running your experiments?
In the former case, please move the tests to the `tests/testthat` folder.

1) I tested your code, and it seems to work fine if you do not parallelize.
    I would assume that it works fine if you add your new learner to mlr, install that package and source 
    it.
    Alternatively, try to `parallelSource` your learner.

2) You might want to add #' @export tags

If there is anything else, feel free to @ me again.

NvNorsinski:
It should be a test. I rewrote it. The Problem is that the test fails. It looks to me like a numerical issue, because results of the original learner und mlr output differ at 6th place after decimal point.

pat-s:
@NvNorsinski Do you want to finish this up here and maybe also add the classification pendant?

NvNorsinski:
@pat-s I would like to finish the started work, but at the moment I have to organize my time. I think at the latest after finishing my thesis in march, I will try to finish this.

pat-s:
@NvNorsinski Any updates?

PR #2727, opened 2020-02-07
Remove threshold slot from `ResamplePrediction` container

### LINKED ISSUE #2289: Problems with feature selection and threshold tuning: optimal threshold of 0.5 all the time
Hi all,
I am trying to solve a binary classifaction problem with logistic regression (for example, spam detection). Via nested resampling with cross-validation, I select proper features and (want to) tune the threshold in the inner loop, and evaluate the performance in the outer.

Performing the resampling with ```classif.logreg```, the optimal threshold is exactly 0.5 every time, which seems to be odd. Furthermore, the plot ```ThreholdVsPerformance``` suggest a optimale threhold which is above 0.5.

Can anyone help me to understand why the value is 0.5, please?

#### Minimal reproducible example

(reprex added by @pat-s)

``` r
library(mlr)
data(spam, package = "kernlab")
task = makeClassifTask(data = spam, target = "type")
# inner loop (tune.threshold = TRUE to allow threshold tuning)
ctrl = makeFeatSelControlRandom(tune.threshold = TRUE, maxit = 5L)
inner = makeResampleDesc("CV", iters = 3)
base = makeLearner("classif.logreg", predict.type = "prob")
learner = makeFeatSelWrapper(base, resampling = inner, control = ctrl,
  show.info = FALSE, measure = mmce)
# outer loop
outer = makeResampleDesc("CV", iters = 3)
res = suppressWarnings(resample(learner, task = task, resampling = outer,
  show.info = FALSE, measures = list(auc, mmce))
)
# results
res$aggr
#>  auc.test.mean mmce.test.mean 
#>     0.96414281     0.09084793
res$pred$threshold
#> nonspam    spam 
#>     0.5     0.5
```

<sup>Created on 2019-12-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Many thanks.

pat-s:
fixes #2289 

The `threshold` slot in a `ResamplePrediction` container has no relevance.
Thresholds are optimized within the folds and the optimized threshold is stored within the respective models.

In fact, the slot in the `ResamplePrediction` object might even confuse users because they assume its the average or best threshold of the `resample()` call.

Instead, users are advised to call `tuneThreshold()` manually on a `Prediction()` object.
It is questionable if this makes sense for the `ResamplePrediction()` object because this contains the predictions of all folds of the CV (i.e. predictions from multiple datasets) and the result can be seen questionable.


R/resample.R
```diff
@@ -131,6 +131,11 @@ resample = function(learner, task, resampling, measures, weights = NULL, models
     printResampleFormatLine("Measures:", measure.lognames)
   }
 
+  # FIXME
+  # iter.results contains only thresholds of 0.5
+
+  # The learner going in here should contain a $predict.threshold slot if tune.threshold = TRUE during tuning
```
larskotthoff: Should this be a FIXME or WONTFIX? Not a fan of adding FIXMEs in PRs.

larskotthoff: This behaviour should also be documented.

mllg:
I'm not fond of this change. We unnecessarily break the API here for a package which is in maintenance. I'd suggest to just update the documentation with some useful hints/comments, and leave the object untouched.

pat-s:
> We unnecessarily break the API here for a package which is in maintenance. I'd suggest to just update the documentation with some useful hints/comments, and leave the object untouched.

Is it likely that this slot is used in any workflow? Imo its really **wrong** within the RR container.
So if we document it, we would do it like "this should be ignored/is wrong, please do not rely on it/interprete it". Why then not remove it completely?

Note that I do only want to remove it from the RR container - not from normal `pred` slots of fitted models.

PS: How do we handle this in mlr3?

R/resample.R
pat-s: oh yeah, this was just some internal leftover of myself. Cleared. Thanks!

mllg:
Ah, now I got it. Yes, can probably be removed.

PR #2730, opened 2020-02-08
Move pkg function calls within ParamSets to helper file

### LINKED ISSUE #2729: `listLearners()` fails if certain packages are not installed
Because of `package::` calls in the respective learner. Those calls should be outsourced to a helper file, just like for the "fda" learners.

pat-s:
fixes #2729 

We cannot have `pkg::<fun>` call within the ParamSets of learners as they will cause `listLearners()` to fail if the learner package is not installed.


jakob-r:
It looks a bit cumbersome. 
I would avoid extra helper files for each learner
Can you just write `function(...) mda::polyreg(...)` directly inside the paramset?

pat-s:
Thanks. Seems to work.

jakob-r:
Sorry. This was a stupid suggestion. Now we will run into problems if we want to set e.g. `method` to `mda::mars` because the check will say that `mda::mars` is not a valid value.
I think here it is done correctly:
https://github.com/mlr-org/mlr/blob/7ebc36456815719d02c41dd1acdad93e0ad12e8e/R/RLearner_classif_fdausc.np.R#L39-L40

Otherwise just use `makeUntypedLearnerParam` I guess. Does not really hurt and allows the user to supply custom functions.

pat-s:
> I think here it is done correctly:

This is within `train()` which is easier than in the ParamSet.

I will use the `makeUntypedLearnerParam` then.

R/RLearner_classif_mda.R
```diff
@@ -21,7 +21,7 @@ makeRLearner.classif.mda = function() {
       makeLogicalLearnerParam(id = "trace", default = FALSE, tunable = FALSE),
       makeDiscreteLearnerParam(id = "start.method", default = "kmeans", values = c("kmeans", "lvq")),
       makeIntegerLearnerParam(id = "tries", default = 5L, lower = 1L),
-      makeDiscreteLearnerParam(id = "criterion", default = "misclassification", values = c("misclassification", "deviance"))
+      makeDiscreteLearnerParam(id = "criterion", default = "misclassification", special.vals = c("misclassification", "deviance"))
```
jakob-r:
```suggestion
      makeDiscreteLearnerParam(id = "criterion", default = "misclassification", values = c("misclassification", "deviance"))
```
`special.vals` are meant for values that do not match the param type. It also does not make much sense to define a DiscreteParam without discrete values.

R/RLearner_classif_mda.R
```diff
@@ -11,13 +11,17 @@ makeRLearner.classif.mda = function() {
       makeIntegerLearnerParam(id = "dimension", lower = 1L),
       makeNumericLearnerParam(id = "eps", default = .Machine$double.eps, lower = 0),
       makeIntegerLearnerParam(id = "iter", default = 5L, lower = 1L),
-      makeDiscreteLearnerParam(id = "method", default = mda::polyreg,
-        values = list(polyreg = mda::polyreg, mars = mda::mars, bruto = mda::bruto, gen.ridge = mda::gen.ridge)),
+      # see helpers_mda.R for objects
+      makeUntypedLearnerParam(id = "method", default = function(...) mda::polyreg(...),
+        values = list(polyreg = function(...) mda::polyreg(...),
```
jakob-r: for an untyped Param you cannot set `values`.

jakob-r:
Now everyhting looks mixed up...

pat-s:
@jakob-r approved?

jakob-r:
Ok - i just did a commit how I would do it. If you dislike it feel free to revert to another solution

pat-s:
Thanks. I am not an expert in advanced Param specifications so I'll take whatever works :)

PR #2732, opened 2020-02-10
Enable tuneThreshold to also maximize measures

### LINKED ISSUE #2731: Really surprising result in tuneThreshold with three classes
I have a classification task with three classes (`Higher`, `Lower`, `Stagnation`).
Roughly 2/3 of the data is in one class `Higher`. I initially wanted to set the thresholds to get the optimal balanced accuracy. The results where really bad (worse than without threshold tuning) so I tried to tune the threshold to get the best accuracy to have a look on the results. They were even more surprising and bad:

```
> Predictions.rf.test <- predict(Model.rf, 
+                                task = Classification.Task,
+                                subset = test.id)
> 
> mlr::performance(Predictions.rf.test, measure = list(bac,
+                                                      acc))
      bac       acc 
0.3338171 0.6508621 
> # acc is not a big surprise because 2/3 of the data is in one class (Higher)
> 
> # "Best" threshold for acc:
> Threshold.rf <- tuneThreshold(Predictions.rf, measure = list(acc))$th
> Threshold.rf
  Higher         Lower            Stagnation 
0.442058919 0.004732797 0.553208283 
> 
> Predictions.rf.test <- 
+   setThreshold(Predictions.rf.test, threshold = Threshold.rf)
> 
> mlr::performance(Predictions.rf.test, measure = list(bac,
+                                                      acc))
       bac        acc 
0.33542977 0.09482759 
``` 

I am quite confused. If I set the threshold of `Higher` very low I always get an accuracy of 2/3 because 2/3 of the data is with class `Higher`. Of course this is trivial but how can the optimal threshold for `Higher` that high (0.442)? With this threshold the performance of accuracy gets really, really bad as you can see.
Am I doing something wrong probably?
Thank you very much!

jakob-r:
Fixes #2731.
`tuneThreshold` always minimized the performance measure and did not check the "direction" of the measure.

### DIFF
--- a/R/tuneThreshold.R
+++ b/R/tuneThreshold.R
@@ -53,7 +53,7 @@ tuneThreshold = function(pred, measure, task, model, nsub = 20L, control = list(
     if (ttype == "multilabel" || k > 2) {
       names(x) = cls
     }
-    performance(setThreshold(pred, x), measure, task, model, simpleaggr = TRUE)
+    ifelse(measure$minimize, 1, -1) * performance(setThreshold(pred, x), measure, task, model, simpleaggr = TRUE)
   }
 
   if (ttype == "multilabel" || k > 2L) {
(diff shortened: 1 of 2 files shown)


pat-s:
Thanks. Blocked by #2733

PR #2773, opened 2020-07-13
Enable FDboost tests again

### LINKED ISSUE #2758: FDBoost tests need to be fixed
Currently skipped because they are constantly failing and I do not see why / do not have time to debug further.

- test_regr_FDboost.R
- test_base_fda.R#386
- vignettes/functional_data.Rmd#191

pat-s:
After the package is back on CRAN.

@pfistfl 
It looks like we have some true positive test failures for the FDboost tests now: https://github.com/mlr-org/mlr/runs/832076608?check_suite_focus=true#step:17:305

Would be great if you could have a look.

fixes #2758

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     -   id: no-browser-statement
     -   id: readme-rmd-rendered
     #  R package development
-    -   id: roxygenize
+    #-   id: roxygenize
     #-   id: use-tidy-description
     #-   id: deps-in-desc
         args: [--allow_private_imports]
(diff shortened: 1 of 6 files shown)


pat-s:
@pfistfl Do you have time for this in the near future? If not, I might need to remove the FDboost learner again if we can't test it.

pat-s:
Starting to remove the FBboost learner then in the near future, I do not have capacities/knowledge to maintain it.

pfistfl:
Sorry for the long silence, will let you know by tonight what to do with this

pfistfl:
Seems to be an issue in FDboost, see https://github.com/boost-R/FDboost/issues/17

pat-s:
Thanks Flo. Let's see if someone replies.

PR #2790, opened 2020-09-25
praznik filters applicability (#2786)

### LINKED ISSUE #2786: praznik filters applicability
Why are only the praznik filters praznik_CMIM and praznik_MRMR applicable for regression data sets and the rest of the praznik filters can only be applied to classification data?
In the praznik package, it says for all of the filter functions that the target can be real-valued (but will be made categorical).

Also, I think that praznik_MRMR should not be allowed for survival data sets. I believe that it is conceptually not applicable and this error occurs when I apply the filter to a survival data set.
```
generateFilterValuesData(wpbc.task, "praznik_MRMR")
Error in .subset2(x, i, exact = exact) : subscript out of bounds
```

bommert:
fixes #2786 

As discussed in issue #2786, I added "regr" and removed "surv" from the praznik filters' supported tasks.

### DIFF
--- a/R/Filter.R
+++ b/R/Filter.R
@@ -951,7 +951,7 @@ makeFilter(
   name = "praznik_JMI",
   desc = "Joint mutual information filter",
   pkg = "praznik",
-  supported.tasks = "classif",
+  supported.tasks = c("classif", "regr"),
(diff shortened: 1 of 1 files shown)


pat-s (approved):
Thanks!

Issue #2771, opened 2020-07-04
ksvm fails for if task has (too) many factor variables.


bommert:
When I use the example code from makeModelMultiplexer and use my task instead of the iris task, tuning does not work. With the iris task, it does work. With my task without a model multiplexer (single learners), tuning works, too.
Can you please help me figure out what the problem is?

```
# download data into working directory
# http://www.salvirt.com/?smd_process_download=1&download_id=4128
# description of data set: http://www.salvirt.com/b2bdataset/

data = read.csv2("Salvirt_B2B_ML_dataset.csv", header = TRUE)
task = makeClassifTask(data = data, target = "Status")

# this code is taken from ?makeModelMultiplexer
# the only change is using "task" instead of "iris.task" in tuneParams
set.seed(123)
library(BBmisc)
bls = list(
  makeLearner("classif.ksvm"),
  makeLearner("classif.randomForest")
)
lrn = makeModelMultiplexer(bls)
ps = makeModelMultiplexerParamSet(lrn,
  makeNumericParam("sigma", lower = -10, upper = 10, trafo = function(x) 2^x),
  makeIntegerParam("ntree", lower = 1L, upper = 500L)
)
rdesc = makeResampleDesc("CV", iters = 2L)
ctrl = makeTuneControlRandom(maxit = 10L)
res = tuneParams(lrn, task, rdesc, par.set = ps, control = ctrl)
```
[Tune] Started tuning learner ModelMultiplexer for parameter set:
                               Type len Def                            Constr Req Tunable Trafo
selected.learner           discrete   -   - classif.ksvm,classif.randomForest   -    TRUE     -
classif.ksvm.sigma          numeric   -   -                         -10 to 10   Y    TRUE     Y
classif.randomForest.ntree  integer   -   -                          1 to 500   Y    TRUE     -
With control class: TuneControlRandom
Imputation value: 1
[Tune-x] 1: selected.learner=classif.rand...; ntree=152
[Tune-y] 1: mmce.test.mean=      NA; time: 0.0 min
[Tune-x] 2: selected.learner=classif.rand...; ntree=420
[Tune-y] 2: mmce.test.mean=      NA; time: 0.0 min
[Tune-x] 3: selected.learner=classif.rand...; ntree=244
[Tune-y] 3: mmce.test.mean=      NA; time: 0.0 min
[Tune-x] 4: selected.learner=classif.rand...; ntree=365
[...]

Issue #2611, opened 2019-06-26
"Error: Missing data in columns" due to `fix.factors.prediction=TRUE`


feinmann:
```r
library(mlr)

train_data <- data.frame(
  A = runif(100), B = factor(sample(c("A", "B"), 100, replace = T)))

test_data <- data.frame(
  A = runif(100), B = factor(sample(c("A", "B", "C"), 100, replace = T)))

lrn <- makeLearner("regr.ranger", fix.factors.prediction = TRUE)

train_task <- makeRegrTask(
  data = train_data,
  target = "A"
)

model <- train(lrn, train_task)

predictions <- predict(model, newdata = test_data)
```

Gives `Error: Missing data in columns: B.`, although there is no missing data. Same for classification.

Kind regards

pat-s:
This issue sounds familiar to me - I think it hit me some time in the past as well.
I cannot tell you when I will have time to look into this.

Thanks a lot for the reprex in the first place!

jakob-r:
The reason is that the new factor `C` is converted to an `NA` because of `fix.factors.prediction = TRUE`.
As you can see in the documentation this is feature was intended for cases where the test data has less factors than the training set. However, it has the side effect that it reduces the levels to the one seen in the training and R then just sets new factor levels to NA. Maybe some learners can deal better with an `NA` then with an unseen factor? However this is not really intended and definitely has to go into the documentation of the `fix.factors.prediction` argument. 

Also we might want  to deal with it better. 
#2771 is kind of related.

pat-s:
The PR from Jakob provides a good approach to the problem. Most likely there is not much else we can do in such situations to account for all possible issues with missing data in prediction scenarios.

Issue #2602, opened 2019-06-13
`make_learner(fix.factors.prediction = TRUE)` does not have an effect


pat-s:
See this SO question: https://stackoverflow.com/questions/56553479/r-predicting-with-new-factor-levels-in-mlr-with-regr-svm-task/56587590#56587590

I did a very dirty fix in the `fix-factors` branch that works for this example but probably not if more factor levels in `newdata` have missing values.
In fact, ,`fix.factors.prediction` sets the missing level to `NA` rather than to a level existing in the training data.

PR #2794, opened 2020-10-26
add error and warning on new factor levels for prediction

jakob-r:
General problem: New factor levels appeared during prediction.

Behaviour: `fix.factors.prediction = TRUE` will convert new factors to NAs

Concrete problem: There is no warning that this happens.

More concrete problem: Some learners (eg. ksvm) will just drop rows with NAs in prediction, leading to fewer predictions than in newdata. (#2771)

Less heavy problem: Some learners (eg. ranger) will complain about NAs, but the user does not now where these are comming from (#2611)

Solution: Create a warning, when NAs are created, throw an error if prediction has not the right number of items. Both in combination will hopefully give the user a clue of what is happening.

Also related: #2602

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # mlr 2.18.0.9000
 
-- Internal changes only.
+- Warning if `fix.factors.prediction = TRUE` causes the generation of NAs for new factor levels in prediction.
+- Clear error message if prediction of wrapped learner has not the same length as `newdata`.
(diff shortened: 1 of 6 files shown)


pat-s:
Thanks @jakob-r! Could you also add a descriptive NEWS entry for this?

jakob-r:
done.

pat-s (approved):
Thanks. LGTM.

PR #2799, opened 2020-11-25
impute.val mention configureMlr

jakob-r:
Previously the information on how to actually use the imputation was very hidden.


jakob-r:
The improved documentation might have helped here: https://github.com/mlr-org/mlrMBO/issues/501

jakob-r:
All the failing CIs can be ignored, right?

pat-s:
> All the failing CIs can be ignored, right?

Yes, for now. Parallel tests are not fully stable yet. 
Even though the message that no tests file were found in these builds is strange.

PR #2825, opened 2021-12-21
fix pre-commit CI

lorenzwalthert:
@pat-s by excluding `todo-files/` for {styler}. Maybe we can re-activate the roxygenize hook and only skip it on CI.


pat-s:
Excluding `todo-files` is fine, they should not block anything :) I was not aware that `exclude` can be used for any hook - I am now 👍 

Are the multiple `styler` entries in the YAML intended?

> Maybe we can re-activate the roxygenize hook and only skip it on CI.

Jup, sounds like a good idea.

lorenzwalthert:
> Are the multiple styler entries in the YAML intended? 

Yeah, I tried to see if timeouts are applicable only to individual hooks or to all hooks. Unfortunately the latter.

lorenzwalthert:
Ok, now we seem to stay under the timeout (2min 36s), compromising on the files styled though. Seems we can't get tests styled as well in that time window. -.- Given that mlr3 repos are much smaller and this repo is in maintenance mode only, I think it's acceptable. Note that this also applies to local execution.

lorenzwalthert:
Just to be on the save side, let's wait until you can make `main` pass, then we rebase these change (and I push some styling of the tests directory that are not catched with pre-commit).

pat-s:
> Given that mlr3 repos are much smaller and this repo is in maintenance mode only, I think it's acceptable. Note that this also applies to local execution.

Fully agree.

> Just to be on the save side, let's wait until you can make main pass, then we rebase these change (and I push some styling of the tests directory that are not catched with pre-commit).

Main is ok for now, it's unstable anyhow since some tests fail in a non-deterministic way.
E.g. the nightly CRON finished just fine - however it might be that the next commit to `main` will fail again.
Things will sort itself out in the long run but there is unfortunately some constant variation among the GHA runs.
This is not a new thing and is happening since more than a year already and meanwhile I am suspecting that GHA is the main problem.

But anyhow! LGTM and thanks a lot for the help 🚀

pat-s:
I've tidied the precommit config, re-added the `roxygenize` hook but skipping it during CI.

PR #2829, opened 2022-02-28
Remove randomForestSRC learner

pat-s:
Since v3 of {randomForestSRC} we see non-deterministic test failures, especially for the multilabel learner. The learner has seen many changes and it would take too much time to implement these into {mlr}, especially given that it is not the most used RF learner.
Please use {mlr3} and {mlr3extralearners} if you want to use {randomForestSRC}.

- Changed default filter for `makeFilterWrapper()` from `randomForestSRC_importance` to `FSelectorRcpp_information.gain`.

### Removed learners

- `classif.randomForestSRC`
- `regr.randomForestSRC`
- `multilabel.randomForestSRC`
- `surv.randomForestSRC`

### Removed filters

- `randomForestSRC_importance`
- `randomForestSRC_var.select`


pat-s:
Removing learner due to non-deterministic test failures since v3. Please use {mlr3} and {mlr3extralearners} or the {ranger} implementation of RF.

